PR #13, opened 2019-09-04
Feature/9 minimal layout

anxolin:
- Creates some minimal markup with styles
- Defines a new component Layout with a header, a footer and the content
- Define some placeholders for adding later other functionalities: i.e. connect wallet

For now it looks like:
image

Test: run and see in the browser


src/components/layout/Header.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
+import styled from 'styled-components'
+
+const Wrapper = styled.header`
```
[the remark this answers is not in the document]
anxolin: I know, blame your former company for that. Now everything is JS, they killed the separation
You can still do it separately, but this is the new trend

src/components/layout/Header.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    &:hover {
+      color: white;
+      border-color: white;
+      cursor: pointer;
+    }
+  }
+
+  .logo {
+    font-size: 1.2em;
```
fleupold: Why are we sometimes using rem and other times em?

anxolin: The logo is a placeholder. So it will replace by an image as soon as we have a logo.

rem and em are using depending on "in relation to what" you want to define the size. A global reference, or a relative (current paragraph)

They both have it's use

PR #16, opened 2019-09-05
Feature/9 linter

anxolin:
- Add husky githook to lint ant test automatically before commiting and pushing.
- Fix lint errors/warnings
- Changes the rules for prettifier/linter to use 2 spaces instead of 4 for indentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ cache: 'yarn'
 script:
   - yarn test
   - yarn build
+before_install:
+  # Needed to deploy pull request and releases
+  - sudo apt-get -y install python-pip python-dev
+  - pip install awscli --upgrade --user
 after_success:
   # Pull Request - Deploy it to a review environment
   # Travis doesn't do deploy step with pull requests builds
--- a/package.json
+++ b/package.json
@@ -6,10 +6,17 @@
   "scripts": {
     "start": "NODE_ENV=development webpack-dashboard -- webpack-dev-server --mode development --hot",
     "build": "rimraf ./dist && NODE_ENV=production webpack --mode production",
-    "eslint": "eslint . --ext .ts,.tsx,.js",
-    "eslint-fix": "eslint --fix . --ext .ts,.tsx,.js",
+    "eslint": "eslint src --ext .ts,.tsx,.js",
+    "eslint:fix": "eslint --fix . --ext .ts,.tsx,.js",
     "stats": "NODE_ENV=production webpack --mode production --env.stats --profile --json > stats.json && webpack-bundle-analyzer ./stats.json ./dist",
-    "test": "echo Testing... && echo OK"
+    "test": "echo Testing... && echo OK",
+    "verify": "npm run eslint && npm run test"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run verify",
+      "pre-push": "npm run verify"
+    }
   },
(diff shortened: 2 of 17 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr16--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr16--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr16--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr16--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr16--dexreact.review.gnosisdev.com

package.json
```diff
@@ -6,10 +6,17 @@
...
+    "lint": "eslint src --ext .ts,.tsx,.js",
+    "lint:fix": "eslint --fix . --ext .ts,.tsx,.js",
     "stats": "NODE_ENV=production webpack --mode production --env.stats --profile --json > stats.json && webpack-bundle-analyzer ./stats.json ./dist",
-    "test": "echo Testing... && echo OK"
+    "test": "echo Testing... && echo OK",
+    "verify": "npm run lint && npm run test"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run verify",
```
fleupold: Do you need the same hook for commit and push?

.travis.yml
```diff
@@ -6,6 +6,10 @@ cache: 'yarn'
 script:
   - yarn test
   - yarn build
+before_install:
```
fleupold: Is this PR introducing linting or fixing the travis script?

anxolin: Yep, you are right, I was fixing the travis script in the wrong branch

package.json
anxolin: I'll leave only the pre-commit for now

fleupold:
@anxolin the gnosis info messages are awesome, but can we adjust the script to only post once per  PR (it should still build on every push, just not comment)? At the moment this guy is extremely spammy.

Worst case, the URL can be derived manually from the PR number.

Or maybe this guy can be a non-blocking github check that you usually find at the end of a PR.

anxolin:
I'll bring in to conversation to Raul (@luarx ), father of the amazing PRaul functionality.

What would you say is the best way to do this part only one time?
image

luarx:
Maybe that could be achieved if we check that there is already a comment which contains the PR preview URL before posting a new one. We can try it!

fleupold:
>We can try it!

That would be cool. I was hoping travis would set a flag when it's the first time a PR is built, but I couldn't find anything.

anxolin:
Yes I also looked for a env var, but didn't find anything relevant.

Also other approach is to create a `dummy file` in node_modules, that it's cached between compilations. Although I'm not 100% sure how the caching works, and if it's just caching between compilations of the same branch

```bash
IF dummy_file doesnt exist THEN
   touch dummy_file
   publish_comment()
ENDIF 
```

anxolin:
I'm thinking, even if the `node_modules` would be reused between different branches, we can still use  the `PR number` in the name of the *dummy file*

luarx:
We can try the @anxolin option also if the first one doesn't work as expected!

PR #18, opened 2019-09-06
Feature/9 show version

anxolin:
Shows the current version in the footer:

image

### DIFF
--- a/src/components/layout/Footer.tsx
+++ b/src/components/layout/Footer.tsx
@@ -4,6 +4,8 @@ import { Link } from 'react-router-dom'
 
 const Wrapper = styled.footer`
   text-align: center;
+  position: relative;
+  color: var(--color-text-secondary);
 
   ul {
     margin: 0;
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr18--dexreact.review.gnosisdev.com

src/components/layout/Footer.tsx
```diff
@@ -35,6 +45,7 @@ const Footer: React.FC = () => (
         <Link to="/source-code">Source code</Link>
       </li>
     </ul>
+    <div className="version">dFusion PoC v{VERSION}</div>
```
alfetopito: Newbie question: I see that you imported the file containing the `VERSION` variable on `App.tsx`, but not here. I assume it works. The question is: why it works?

anxolin: Hehe, good question :) 

If I'm honest, this was also new to me, I haven't used too much typescript in projects before.
What I think is, by using `declare` it becomes global. I just imported it in the App cause is the entry point (could have been imported in `index.ts` too).

PR #26, opened 2019-09-10
21/setup testing

alfetopito:
- [x] Enabling test command using `jest`
- [x] Adding test dependencies such `jest`, `enzyme` and related deps
- [x] Adding a simple test of a function on `DepositWidget`

Fixes https://github.com/gnosis/dex-react/issues/21

Open to suggestions regarding test structure/what to test.

```
» yarn test                                                                                                                               alfeto@amebadell
yarn run v1.17.3
warning package.json: No license field
$ jest
 PASS  src/test/components/DepositWidget.test.ts
  ✓ Deposit button text (3ms)
  ✓ Withdraw button text
  ✓ Deposit button text (1ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        1.212s
Ran all test suites.
Done in 2.23s.

```


src/test/components/DepositWidget.test.ts
```diff
@@ -0,0 +1,11 @@
+import { getButtonText, ButtonTypes } from '../../components/DepositWidget/Row'
```
anxolin: instead of `src/test`, I think it would be better to use `test` for the test.
This way the `src` is just for source code

src/test/components/DepositWidget.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+test('Deposit button text', () => {
+  expect(getButtonText(ButtonTypes.DEPOSIT, '')).toMatch(/Deposit/)
```
anxolin: Why do you use regex?
it's not enough to use a string `toBe('Deposit')`

alfetopito:
Should we add `yarn build` as a pre-commit hook as well? If I were aware of this, the build would have failed before I pushed it.

gnosis-info:
Travis automatic deployment:
 https://pr26--dexreact.review.gnosisdev.com

src/test/components/DepositWidget.test.ts
```diff
@@ -0,0 +1,11 @@
+import { getButtonText, ButtonTypes } from '../../components/DepositWidget/Row'
+
+test('Deposit button text', () => {
```
anxolin: All this one's test the same function, consider using scoping for grouping together:

For example:
```js
describe('Get text for buttons', () => {
  test('Deposit', () => {
    expect(getButtonText(ButtonTypes.DEPOSIT, '')).toMatch(/Deposit/)
  })
  test('Withdraw', () => {
    expect(getButtonText(ButtonTypes.WITHDRAW, '')).toMatch(/Withdraw/)
  })
  test('Enable', () => {
    expect(getButtonText(ButtonTypes.ENABLE, 'TTT')).toMatch(/Enable TTT/)
  })
})
```

image

package.json
```diff
@@ -44,8 +45,11 @@
     "@types/web3": "^1.0.19",
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
+    "babel-jest": "^24.9.0",
     "babel-loader": "^8.0.6",
     "babel-plugin-styled-components": "^1.10.6",
+    "enzyme": "^3.10.0",
+    "enzyme-adapter-react-16": "^1.14.0",
```
anxolin: Where did you use enzyme?

anxolin:
> Should we add `yarn build` as a pre-commit hook as well? If I were aware of this, the build would have failed before I pushed it.

Not sure, it can be too much. Right now the build is more or less fast, but if it delays all our commits some seconds it can be an issue.

Most stuff gets caught with the lint and just running the app. It's fine that we break travis, that's what it there for :)

We could add it before pushing, but not 100% sure if it's good or bad (i have missed fillings about that)

package.json
alfetopito: Nowhere yet. It's supposed to be used to test components. In case we don't, this might be unnecessary. Would you like me to remove that?

src/test/components/DepositWidget.test.ts
alfetopito: Mainly because right now there are `-` and `+` signs that, from the mocks, might be an img or another element for styling later on. Otherwise I can use `toBe`

src/test/components/DepositWidget.test.ts
alfetopito: Fair enough, will move it.

src/test/components/DepositWidget.test.ts
alfetopito: Thanks for the suggestion, will update.

alfetopito:
Understood, let not add `yarn build` then.

package.json
anxolin: I think it's better to include it in a separate PR, just with a basic render of a component

gnosis-info:
Travis automatic deployment:
 https://pr26--dexreact.review.gnosisdev.com

src/test/components/DepositWidget.test.ts
anxolin: Okay :)

gnosis-info:
Travis automatic deployment:
 https://pr26--dexreact.review.gnosisdev.com

package.json
alfetopito: Removed in latest commit.

PR #31, opened 2019-09-12
30/stable coin api: Part 1, Deposit API

anxolin:
Defines the DepositAPI, an implementation of it, and unit testing.

**Part of issue**: #30 

**Context**
This is the first of two PRs that will add an interface for interacting with the contract.
The idea is to: 
1. Create an interface in Typescript defining the method signatures
2. Provide a simple, "in memory", pure JS implementation. This is for making development simpler. We will be able to start the app using this mock implementation.
3. Add tests for this implementation

**Scope of the PR**
This first one defines the API for the deposit methods (`DepositApi`). 
It's provides a similar interface as EpochTokenLocker.

It also includes the tests, and some utils/constants/helpers that were required in order to do these tests.

**Main components**
* `DepositApi`: The typescript interface. Also as part of this interface it was required to define: `BalanceState` and `PendingFlux`
* `DepositApiMock`: Basic implementation of this API
* Tests:
The following test are being done:
image

**How to test?**
Just execute:
```bash
yarn test
```

Issue #30, opened 2019-09-12
Implement basic stable coin service API: Interface + Mock implementation


anxolin:
**Includes**:
- [ ] Define interface for the stable coin converter
- [ ] Provide a mock implementation

**Related PRs**:
- [ ] #31 Deposit API
- [ ] #36 Exchange API

PR #36, opened 2019-09-13
30/stable coin api  part2

anxolin:
*  [X] Addresses the EchangeAPI for #30 

**Context**
This is the second of two PRs that will add an interface for interacting with the contract.
The idea is to: 
1. Create an interface in Typescript defining the method signatures
2. Provide a simple, "in memory", pure JS implementation. This is for making development simpler. We will be able to start the app using this mock implementation.
3. Add tests for this implementation

> See first PR: https://github.com/gnosis/dex-react/pull/31

**Scope of the PR**
This second one defines the API for the deposit methods (`DepositApi`). 
It's provides a similar interface as `StablecoinConverter`.

It also includes the tests, and some utils/constants/helpers that were required in order to do these tests.

**Main components**
* `ExchangeApi`: The typescript interface. Also as part of this interface it was required to define: `Order` and `AddOrderParams`
* `ExchangeApiMock`: Basic implementation of this API
*Tests*:
The following test are being done:
TODO: Add picture once all test are defined

**How to test?**
Just execute:
```bash
yarn test
```


gnosis-info:
Travis automatic deployment:
 https://pr36--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApiMock.ts
```diff
@@ -0,0 +1,108 @@
...
+    assert(typeof this.registeredTokens[tokenId] === 'string', 'Must have ID to get Address')
+    return this.registeredTokens[tokenId]
+  }
+
+  public async getTokenIdByAddress(tokenAddress: string): Promise<number> {
+    assert(typeof this.tokenAddressToId[tokenAddress] === 'number', 'Must have Address to get ID')
+    return this.tokenAddressToId[tokenAddress]
+  }
+
+  public async getCurrentPrice(_tokenId: number): Promise<BN> {
```
anxolin: Maybe we should remove this?
It's there an equivalent in the contract?

alfetopito: I left it there in case we need it, because I couldn't find an easy way to implement this in the mock.
The contract has a mapping of tokens and prices which is updated when the solution is submitted:

```
/** @dev mapping of type tokenId -> curentPrice of tokenId */
    mapping (uint16 => uint128) public currentPrices;
```

I'll remove it for now. We can add it back if required.

gnosis-info:
Travis automatic deployment:
 https://pr36--dexreact.review.gnosisdev.com

PR #47, opened 2019-09-16
Remove table component

anxolin:
I deleted the table component and move it to the main component because I think for now, so it's better not to split so much the components while they are small.


gnosis-info:
Travis automatic deployment:
 https://pr47--dexreact.review.gnosisdev.com

PR #49, opened 2019-09-16
Remove widget component

anxolin:
Removed the Widget component cause it only added style.

For now it would be just a css class for simplicity


gnosis-info:
Travis automatic deployment:
 https://pr49--dexreact.review.gnosisdev.com

Issue #43, opened 2019-09-16
Load balances for each token


alfetopito:
**Description:** Wallet, exchange wallet, pending deposits, pending withdraws

**Estimation:** 1

**Depends on:** https://github.com/gnosis/dex-react/issues/41

PR #51, opened 2019-09-17
43/create mock deposit api

anxolin:
Includes:
- [X] Instanciates a DepositApi using the mock implementation
- [X] Includes a MOCK env var to control whether we want to run it in mock mode or not. 
- [X] Throws an exception if we don't run it in MOCK mode, since the real implementation is not ready

Part of #43


gnosis-info:
Travis automatic deployment:
 https://pr51--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr51--dexreact.review.gnosisdev.com

src/api/index.ts
```diff
@@ -1,5 +1,22 @@
...
+
+function createTokenApi(): TokenApi {
+  return new TokenApiImpl([Network.Mainnet, Network.Rinkeby])
+}
+
+function createDepositApi(): DepositApi {
+  if (true || isMock) {
+    return new DepositApiMock({ balanceStates })
+  } else {
+    throw new Error('Only has been implemented')
```
alfetopito: In this case, I think you could either have the `true` in the if condition or throw the error. Because the `else` will never be reached, thus the error is not necessary.

Also, the error message is probably missing a word.

anxolin: Yep, this `true` was a test that I forgot from to remove...

I realised only when I did this other PR: https://github.com/gnosis/dex-react/pull/57
Since they go in serie (every PR uses the previous one as it's base),I fixed it in #57

fleupold:
For testing error behaviour it might be in the longterm cool to also have a mock that always fails?

anxolin:
@fleupold It's not done yet, but I want to add some initial test data to the mocks, also I'll take your advice and make that the implementation throw an error if you enter a magic number

fleupold:
> I'll take your advice and make that the implementation throw an error if you enter a magic number

Maybe even just a flag that you can set via the console that will switch the Mock into one that always fails. This way you should be able to test all kinds of errors pretty dynamically.

anxolin:
> Maybe even just a flag that you can set via the console that will switch the Mock into one that always fails. This way you should be able to test all kinds of errors pretty dynamically.

It can be done, but I think for  a demo or for testing quick is easier if we do the magic number approach, cause you don't need to go open the console, change a prop, then do the use case, then unset the prop. My idea is just, if you add 100 as the amount, for example, it fails

PR #52, opened 2019-09-17
Use BN for balances and rename constants

anxolin:
**Includes**
* Replace Number for BN in the balance amouts. This is required to match the API (and ofcourse the future web3 implementation)
* Renames the constants file because of a naming collision with another file with the same path

Part of #43


gnosis-info:
Travis automatic deployment:
 https://pr52--dexreact.review.gnosisdev.com

anxolin:
@alfetopito it's explained in the PR description: "Renames the constants file because of a naming collision with another file with the same path"

This is why I created a folder and a index file and I was requiring it with `constants/` to differentiate from `constants`, but I run into issues and solved it with this rename

gnosis-info:
Travis automatic deployment:
 https://pr52--dexreact.review.gnosisdev.com

Issue #42, opened 2019-09-16
Enable tokens button and action


anxolin:
Part of the deposit widget.

**Includes**
- [*] Detect if the user have set the allowance
- [*] Show the button
- [*] Implement the approve

**Depends on**
- #41

PR #61, opened 2019-09-19
42/enable token

anxolin:
Part of #42 

Allows to enable a token. It handle the call to the API, displaying a loading spinner, and updating the interface once it's approved.

> Please accept my apologize because this PR contains some extras that it was required, but that it's not directly related to "Enabling a token". 

The extras are details in the below. 

**Included**: 
- [X] Adds a `getContractAddress` function for the deposit api, required for the approve call
- [X] Wait utility: Used to add some delays to the mock functions. This way, the mock will wait for 2.5 seconds before success. This is used in the deposit to make it look more real, and to test the loading.
- [X] Created a new hook for enabling a token. This way the logic is extracted from the visual component
- [X] Show loading spinner while the token is being approved. See image below
image

**Extras**: 
- [X] Adds mock mode for the test, otherwise they will fail once a component depend on an API
- [X] Create a log util function. It just uses `console.log`, but for tests uses a `noop` function. This was added because it's useful to log the mock functions of the API, but it's not good that it's showing that on the tests too.
- [X] Created an utility to format amounts. It accepts a BN, and it formats the wei. Delegates on web utils.
- [X] WalletAPIMock now fails if you try to do something before connecting
- [X] Added a disconnect function to the wallet


gnosis-info:
Travis automatic deployment:
 https://pr61--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr61--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr61--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr61--dexreact.review.gnosisdev.com

PR #63, opened 2019-09-19
42/hightlight enabled token

anxolin:
Part of #42 

Highlights the row when it's loading (doing the approval for example), this way, it's clear that it's doing something :) 

Screenshot at Sep 19 18-22-56

Then, when the approve is done, the row is highlighted for 5 seconds:
Screenshot at Sep 19 18-23-18

Part of #42


gnosis-info:
Travis automatic deployment:
 https://pr63--dexreact.review.gnosisdev.com

anxolin:
@alfetopito this branch has some placeholders for adding the messages using the library you are integrating:

image

You can use it as a base for adding it

gnosis-info:
Travis automatic deployment:
 https://pr63--dexreact.review.gnosisdev.com

src/utils/format.ts
```diff
@@ -0,0 +1,10 @@
+import BN from 'bn.js'
+import { fromWei } from 'web3-utils'
+
+export function formatAmount(amount?: BN): string | null {
+  if (!amount) {
+    return null
```
alfetopito: Since the format function return a string, why not return an empty string here?

anxolin: It might be more comfortable what you suggest. 
I just thought that empty string doesn't need to be the representation of an undefined value. I return null in this case so we define what to do.

I actually don't mind to change it to empty string since it also evaluates to a falsy value

alfetopito: I don't have a strong opinion either. Just thinking in terms of semantics and to be more strict to default to an empty string when the function is supposed to format something for output.

anxolin: I went for your approach, we can also change it one way or the other in the future. The idea is to use this function always to format amounts so we change it just in a single place. Actually this function doesn't deal yet with decimals, but that we'll come in a future PR.

https://github.com/gnosis/dex-react/commit/f9e6792d6c68508ae4972fc10b50c1ee943ddb0c

PR #69, opened 2019-09-23
Load contract address for etherscan link

anxolin:
Tiny PR to load the URL from the contract to be used in the Etherscan link


gnosis-info:
Travis automatic deployment:
 https://pr69--dexreact.review.gnosisdev.com

anxolin:
> thought the contract we would link to is referring to the Dfusion exchange contract. Not the case?

Yes, but this is mock data. I'm using fake tx and fake contract. The relevant part is that it's using the API to get the URL, then when we implement the real API, we will return the correct contract

PR #71, opened 2019-09-23
66/format amounts

### LINKED ISSUE #66: Format amounts
There's currently a format amount function that is used all over the app:
```js
export function formatAmount(amount?: BN): string | null {
  if (!amount) {
    return null
  }

  return fromWei(amount).toString()
}
```

This function should also:
- [ ] Have a decimal strategy (how many decimals we use in the different cases). It makes sense to use the same as the Safe: https://github.com/gnosis/safe/blob/master/specs/common/format_amounts.rst
- [ ] Take into account that the amounts are for tokens with different number of decimals

anxolin:
Fixes #66 

Adds formatting for the amounts. Takes into consideration the precision and the decimals we want to show.

This PR was harder than I thought initially, I assumed there was nice formatting tools for BN.js, but I couldn't find any. Also it was difficult to handle the big number formatting in JS, this is why the followed strategy is to format the `string` instead of trying to convert to a JS number.  For that, we need to split the number in the integer part and the decimal. For the decimal, we discard the numbers we are not interested as a substring (for the same reason)

If someone provides a nicer way to format, we can incorporate at anytime, we only need to respect the interface:

```js
function formatAmount(
  amount?: BN,
  amountPrecision = 18,
  decimals = 4,
): string {  ... }
```

Includes:
- [X] Adds the format amount util (18 amount precision, 4 decimals by default, but configurable)
- [X] Adds the format full amount util (all decimals)
- [X] Format numbers in Deposit widget
- [X] Adds unit testing for representative cases

image

**Out of the scope:**
* Format big amount abbreviation like 23K, 1M, or 1.5B
* Format nicely arbitrary big amounts, like `2**256 -1`
* Use the browser locale for the symbols --> Addressed in https://github.com/gnosis/dex-react/pull/75

**Test**
* Check web link
* `yarn test`


gnosis-info:
Travis automatic deployment:
 https://pr71--dexreact.review.gnosisdev.com

test/utils/format.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+describe('Tokens with precission 6', () => {
+  test('1 unit', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6)).toEqual('1')
+  })
+
+  test('12,345 units', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6)).toEqual('1')
```
alfetopito: Test case doesn't match the test name

test/utils/format.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  })
+
+  test('4,567,890.123444 units, round down', async () => {
+    const amount = new BN('4567890123444')
+    expect(formatAmount(amount, 6)).toEqual('4,567,890.1234')
+  })
+
+  test('4,567,890.123555 units, round up', async () => {
+    const amount = new BN('4567890123455')
+    expect(formatAmount(amount, 6)).toEqual('4,567,890.1235')
```
alfetopito: Number in the test name doesn't match value used (`.123555` vs `123455`)

test/utils/format.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+describe('2 decimals', () => {
+  test('1 unit', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6, 2)).toEqual('1')
+  })
+
+  test('12,345 units', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6, 2)).toEqual('1')
```
alfetopito: Test case doesn't match the test name

test/utils/format.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  })
+
+  test('4,567,890.12333 units, round down', async () => {
+    const amount = new BN('4567890123333')
+    expect(formatAmount(amount, 6, 2)).toEqual('4,567,890.12')
+  })
+
+  test('4,567,890.123555 units, round up', async () => {
+    const amount = new BN('4567890125555')
+    expect(formatAmount(amount, 6, 2)).toEqual('4,567,890.13')
```
alfetopito: Number in the test name doesn't match value used (`123555` vs `125555`)

gnosis-info:
Travis automatic deployment:
 https://pr71--dexreact.review.gnosisdev.com

test/utils/format.spec.ts
anxolin: That's a good eye, it was bad in the test case name

anxolin:
A applied the suggested fixes. 

Also found one issue when displaying lot's of decimals. Additionally added a test for the `formatAmountFull` which is used to format amounts with all their decimals

src/utils/format.ts
```diff
@@ -1,10 +1,61 @@
...
+
+function _decomposeBn(
+  amount?: BN,
+  amountPrecision = DEFAULT_PRECISION,
+  decimals = DEFAULT_DECIMALS,
+): { integerPart: BN; decimalPart: BN } {
+  // Discard the decimals we don't need
+  //  i.e. for WETH (precision=18, decimals=4) --> amount / 1e14
+  //        16.5*1e18 ---> 165000
+  const amountRaw = amount.divRound(TEN.pow(new BN(amountPrecision - decimals)))
```
alfetopito: Just occurred to me. `amount` is an optional input (line 12), correct? What happens here if it's `undefined`?

I guess it shouldn't be optional here.

test/utils/formatAmount.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+describe('2 decimals', () => {
+  test('1 unit', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6, 2)).toEqual('1')
+  })
+
+  test('12,345 units', async () => {
+    const amount = new BN('1000000')
+    expect(formatAmount(amount, 6, 2)).toEqual('1')
```
alfetopito: The amount doesn't match the test name: `12,345` vs `1000000`

test/utils/formatAmount.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  })
+
+  test('4,567,890.123333 units, round down', async () => {
+    const amount = new BN('4567890123333')
+    expect(formatAmount(amount, 6, 2)).toEqual('4,567,890.12')
+  })
+
+  test('4,567,890.125555 units, round up', async () => {
+    const amount = new BN('4567890125555')
+    expect(formatAmount(amount, 6, 2)).toEqual('4,567,890.13')
```
alfetopito: Question: Should 0.5 round up to 1? 
My concept is that, from 0.0 to 0.5, round up == 0, while 0.6 to 0.9, round up == 1
Not the case?

src/utils/format.ts
anxolin: Good point. We ensure that is not undefined before, but it should be part of the signature as you suggest

test/utils/formatAmount.spec.ts
anxolin: fixed

test/utils/formatAmount.spec.ts
anxolin: It's pretty common to round the `5` up
image

Issue #46, opened 2019-09-16
Deposit Widget: Detect withdrawability and withdraw 


anxolin:
**Includes**:
- [ ] Detect withdrawability
- [ ] Show withdraw button
- [ ] Withdraw

**Depends on**:
- #30 
- #38

PR #77, opened 2019-09-25
46/detect withdrawability

alfetopito:
- [x] Detect withdrawability
- [x] Show withdraw button
- [x] Show pending withdraw icon
- [x] Withdraw

withdraw

### Additionally:

- Added a eslint plugin to check react hooks.
- Added the utils function `abbreviateString` to, well, abbreviate strings.
- Using a [ref](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) to track hook mounted state as opposed to a regular variable.
- Did a bit refactoring to reuse the same hook (`useTokenBalances`) for optionaly updating a single token balance as opposed to all balances at once.
  Added a `update` function to the hook, to manually update the values.
  Not sure this is the best approach though. Open to suggestions.

--- 

Closes https://github.com/gnosis/dex-react/issues/46

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,32 +1,38 @@
 module.exports = {
-    parser: '@typescript-eslint/parser',  // Specifies the ESLint parser
-    extends: [
-        'plugin:react/recommended',  // Uses the recommended rules from @eslint-plugin-react
-        'plugin:@typescript-eslint/recommended',  // Uses the recommended rules from @typescript-eslint/eslint-plugin
-        'prettier/@typescript-eslint',  // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
-        'plugin:prettier/recommended',  // Enables eslint-plugin-prettier and displays prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
+  parser: '@typescript-eslint/parser', // Specifies the ESLint parser
+  extends: [
+    'plugin:react/recommended', // Uses the recommended rules from @eslint-plugin-react
+    'plugin:@typescript-eslint/recommended', // Uses the recommended rules from @typescript-eslint/eslint-plugin
+    'prettier/@typescript-eslint', // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
+    'plugin:prettier/recommended', // Enables eslint-plugin-prettier and displays prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
+  ],
+  parserOptions: {
+    ecmaVersion: 2018, // Allows for the parsing of modern ECMAScript features
+    sourceType: 'module', // Allows for the use of imports
+    ecmaFeatures: {
+      jsx: true, // Allows for the parsing of JSX
+    },
+  },
+  rules: {
+    // Place to specify ESLint rules. Can be used to overwrite rules specified from the extended configs
+    // e.g. "@typescript-eslint/explicit-function-return-type": "off",
+    '@typescript-eslint/explicit-function-return-type': [
+      'error',
+      {
+        allowTypedFunctionExpressions: true,
+      },
     ],
-    parserOptions: {
-        ecmaVersion: 2018,  // Allows for the parsing of modern ECMAScript features
-        sourceType: 'module',  // Allows for the use of imports
-        ecmaFeatures: {
-            jsx: true,  // Allows for the parsing of JSX
-        },
+    'react/prop-types': [1, { skipUndeclared: true }],
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
+  },
+  plugins: ['react-hooks'],
+  settings: {
+    react: {
+      version: 'detect', // Tells eslint-plugin-react to automatically detect the version of React to use
     },
-    rules: {
-        // Place to specify ESLint rules. Can be used to overwrite rules specified from the extended configs
-        // e.g. "@typescript-eslint/explicit-function-return-type": "off",
-        "@typescript-eslint/explicit-function-return-type": ["error", {
-            "allowTypedFunctionExpressions": true
-          }],
-          "react/prop-types": [1, { "skipUndeclared": true }]
-    },
-    settings: {
-        react: {
-            version: 'detect',  // Tells eslint-plugin-react to automatically detect the version of React to use
-        },
-    },
-    globals: {
-      VERSION: true
-    }
-};
+  },
+  globals: {
+    VERSION: true,
+  },
+}
--- a/src/api/exchange/DepositApiMock.ts
+++ b/src/api/exchange/DepositApiMock.ts
@@ -5,7 +5,7 @@ import { getEpoch, formatAmount, log } from 'utils'
 import { ZERO, BATCH_TIME } from 'const'
 import { CONTRACT, RECEIPT } from '../../../test/data'
 
-import { DepositApi, BalanceState, TxResult, TxOptionalParams } from 'types'
+import { DepositApi, BalanceState, TxResult, TxOptionalParams, Erc20Api } from 'types'
 import { waitAndSendReceipt } from 'utils/mock'
 
 export interface BalancesByUserAndToken {
@@ -14,9 +14,11 @@ export interface BalancesByUserAndToken {
 
 export class DepositApiMock implements DepositApi {
   private _balanceStates: BalancesByUserAndToken
+  private _erc20Api: Erc20Api
 
-  public constructor(balanceStates: BalancesByUserAndToken) {
+  public constructor(balanceStates: BalancesByUserAndToken, erc20Api: Erc20Api) {
     this._balanceStates = balanceStates
+    this._erc20Api = erc20Api
   }
 
   public getContractAddress(): string {
--- a/src/api/exchange/DepositApiMock.ts
+++ b/src/api/exchange/DepositApiMock.ts
@@ -148,6 +148,9 @@ export class DepositApiMock implements DepositApi {
     pendingWithdraws.amount = ZERO
     balanceState.balance = balanceState.balance.sub(amount)
 
+    // mock transfer tokens to user's mock `wallet`
+    await this._erc20Api.transfer(tokenAddress, this.getContractAddress(), userAddress, amount)
+
     log(`[DepositApiMock] Withdraw ${formatAmount(amount)} for token ${tokenAddress}. User ${userAddress}`)
     return { data: undefined, receipt: RECEIPT }
   }
--- a/src/components/DepositWidget/Row.tsx
+++ b/src/components/DepositWidget/Row.tsx
@@ -50,10 +50,10 @@ const txOptionalParams: TxOptionalParams = {
   onSentTransaction: (receipt: Receipt): void => {
     toast.info(
       <div>
-        The transaction has been sent! Check
+        The transaction has been sent! Check{' '}
         <a href={`https://etherscan.io/tx/${receipt.transactionHash}`} target="_blank" rel="noopener noreferrer">
           {receipt.transactionHash.slice(0, 6)}...{receipt.transactionHash.slice(-4)}
-        </a>
+        </a>{' '}
         for details
       </div>,
     )
--- a/src/components/DepositWidget/Row.tsx
+++ b/src/components/DepositWidget/Row.tsx
@@ -33,6 +33,7 @@ const TokenTr = styled.tr`
 
   div.success {
     color: #63ab52;
+    margin-top: -0.5rem;
   }
 `
 
--- a/src/components/DepositWidget/Row.tsx
+++ b/src/components/DepositWidget/Row.tsx
@@ -61,9 +61,9 @@ const txOptionalParams: TxOptionalParams = {
   },
 }
 
-export const Row: React.FC<RowProps> = (params: RowProps) => {
-  const { balances, updateBalances } = useTokenBalances(params.initialTokenBalances.address)
-  const tokenBalances = balances === null ? params.initialTokenBalances : balances[0]
+export const Row: React.FC<RowProps> = (props: RowProps) => {
+  const { balances, updateBalances } = useTokenBalances(props.initialTokenBalances.address)
+  const tokenBalances = balances === null ? props.initialTokenBalances : balances[0]
   const {
     address,
     addressMainnet,
--- a/src/components/DepositWidget/Row.tsx
+++ b/src/components/DepositWidget/Row.tsx
@@ -11,7 +11,6 @@ import { useEnableTokens } from 'hooks/useEnableToken'
 import Form from './Form'
 import { useWithdrawTokens } from 'hooks/useWithdrawTokens'
 import { ZERO } from 'const'
-import { useTokenBalances } from 'hooks/useTokenBalances'
 
 const TokenTr = styled.tr`
   img {
@@ -62,8 +61,7 @@ const txOptionalParams: TxOptionalParams = {
 }
 
 export const Row: React.FC<RowProps> = (props: RowProps) => {
-  const { balances, updateBalances } = useTokenBalances(props.tokenBalances.address)
-  const tokenBalances = balances === null ? props.tokenBalances : balances[0]
+  const [tokenBalances, setTokenBalances] = useState<TokenBalanceDetails>(props.tokenBalances)
   const {
     address,
     addressMainnet,
@@ -100,11 +98,21 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
       console.debug(`Starting the withdraw for ${formatAmountFull(withdrawingBalance, decimals)} of ${symbol}`)
 
       const result = await withdraw()
-      await updateBalances()
+
+      setTokenBalances(
+        (current: TokenBalanceDetails): TokenBalanceDetails => {
+          return {
+            ...current,
+            exchangeBalance: current.exchangeBalance.sub(withdrawingBalance),
+            withdrawingBalance: ZERO,
+            walletBalance: current.walletBalance.add(withdrawingBalance),
+          }
+        },
+      )
 
       console.log(`The transaction has been mined: ${result.receipt.transactionHash}`)
 
-      toast.success(`Withdraw completed`)
+      toast.success(`Withdraw of ${withdrawingBalance} ${symbol} completed`)
     } catch (error) {
       console.error('Error executing the withdraw request', error)
       toast.error(`Error executing the withdraw request: ${error.message}`)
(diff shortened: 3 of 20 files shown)


src/components/DepositWidget/Row.tsx
```diff
@@ -89,7 +114,26 @@ export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
       </td>
       <td>{name}</td>
       <td title={formatAmountFull(exchangeBalanceTotal, decimals)}>{formatAmount(exchangeBalanceTotal, decimals)}</td>
-      <td title={formatAmountFull(withdrawingBalance, decimals)}>{formatAmount(withdrawingBalance, decimals)}</td>
+      <td title={formatAmountFull(withdrawingBalance, decimals)}>
+        {withdrawable ? (
+          <>
+            <button className="success" onClick={_withdraw} disabled={withdrawing}>
```
alfetopito: For some reason, `withdrawing` state is not changing to `true` when `withdraw` is invoked, thus button is never disabled.

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

.eslintrc.js
```diff
@@ -1,31 +1,37 @@
...
+    sourceType: 'module', // Allows for the use of imports
+    ecmaFeatures: {
+      jsx: true, // Allows for the parsing of JSX
     },
-    settings: {
-        react: {
-            version: 'detect',  // Tells eslint-plugin-react to automatically detect the version of React to use
-        },
+  },
+  plugins: ['react-hooks'],
```
anxolin: Cool!

src/components/DepositWidget/Row.tsx
```diff
@@ -35,6 +42,20 @@ function _loadFallbackTokenImage(event: React.FormEvent<HTMLImageElement>): void
   image.src = unknownTokenImg
 }
 
+const txOptionalParams: TxOptionalParams = {
+  onSentTransaction: (receipt: Receipt): void => {
+    toast.info(
+      <div>
+        The transaction has been sent! Check{' '}
+        <a href={`https://etherscan.io/tx/${receipt.transactionHash}`} target="_blank" rel="noopener noreferrer">
+          {receipt.transactionHash.slice(0, 6)}...{receipt.transactionHash.slice(-4)}
```
anxolin: This slide it's better done in a util. 
It can be sth like:
- You provide a string and a max size
- If the string is short enough, returns the original
- Otherwise, adds trims it and add the ellipsis

src/hooks/useEnableToken.ts
```diff
@@ -54,10 +54,10 @@ export const useEnableTokens = (params: Params): Result => {
       setEnabling(false)
 
       // Highlight the token for some seconds
-      setHightlight(true)
+      setHighlight(true)
```
anxolin: thanks for this 👍

src/hooks/useWithdrawTokens.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    setWithdrawing(true)
+    console.debug(`Is withdrawing? ${withdrawing}`)
+    // TODO: Remove connect once login is done
+    await walletApi.connect()
+
+    const userAddress = await walletApi.getAddress()
+    const result = depositApi.withdraw(userAddress, tokenAddress, params.txOptionalParams)
+
+    setWithdrawing(false)
```
anxolin: Note that I made this mistake and had to fix it in the Memory Leak PR (#74)

Every time you have an await, things can happen in the middle. For example, the user can navigate, and the componet get's distroyed while executing line `64`.

Then, line `66` will do something that will trigger a redraw of the component. But if there's no component, we'll have a small issue.

mmmm, actually maybe this is the issue you where asking, I don't see the `await` keyword in line `64`

src/hooks/useWithdrawTokens.ts
```diff
@@ -0,0 +1,72 @@
...
+    const userAddress = await walletApi.getAddress()
+    const [withdrawBatchId, currentBatchId] = await Promise.all([
+      depositApi.getPendingWithdrawBatchId(userAddress, tokenAddress),
+      depositApi.getCurrentBatchId(),
+    ])
+
+    return withdrawBatchId < currentBatchId
+  }
+
+  useEffect(() => {
```
anxolin: Careful with `useEffect`, when you don't provide a second argument it'll be triggered on every redraw.

I think in this case you need to pass an empty array, signalign you only want to be triggered on the first render.

src/hooks/useWithdrawTokens.ts
```diff
@@ -0,0 +1,72 @@
...
+    _checkWithdrawable()
+      .then(withdrawable => setWithdrawable(withdrawable))
+      .catch(error => {
+        console.error('Error checking withdraw state')
+        setError(error)
```
anxolin: error has type boolean, but you are passing an `Error`.
If you just need a boolean, I think it's better to do just `setError(true)`

src/hooks/useWithdrawTokens.ts
```diff
@@ -0,0 +1,72 @@
...
+    ])
+
+    return withdrawBatchId < currentBatchId
+  }
+
+  useEffect(() => {
+    _checkWithdrawable()
+      .then(withdrawable => setWithdrawable(withdrawable))
+      .catch(error => {
+        console.error('Error checking withdraw state')
```
anxolin: It's important in this case to print the original error:

`console.error('Error checking withdraw state', error)`

This way, if something goes wrong we'll also see the stack trace in the console.

fleupold:
Should we maybe call it claimable instead of withdrawable? Given that the other button us is already called withdraw?

src/hooks/useWithdrawTokens.ts
alfetopito: That was it (facepalm)

src/hooks/useWithdrawTokens.ts
alfetopito: Thanks. I'll move the function inside and only depend on the states I actually need.

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

src/components/DepositWidget/Row.tsx
alfetopito: Done.

alfetopito:
> Should we maybe call it claimable instead of withdrawable? Given that the other button us is already called withdraw?

I got that from `RF.3.5 Deposit Widget: Withdraw` in the mocks https://docs.google.com/document/d/1lW3SQV5FAI-gDrZZAD3REunrx4TSa9nNjxHljpCFOd4/edit#heading=h.tct8x21ykhg6

I agree `withdrawable` sounds a bit weird, but `claimable` sounds like something I won and just need to go get it. 

@anxolin since you came up with those mocks, have any opinion here?

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

src/components/DepositWidget/Row.tsx
```diff
@@ -28,19 +31,39 @@ const TokenTr = styled.tr`
...
+
+  div.success {
+    color: #63ab52;
+    margin-top: -0.5rem;
   }
 `
 
 export interface RowProps {
-  tokenBalances: TokenBalanceDetails
+  initialTokenBalances: TokenBalanceDetails
```
anxolin: I wouldn't change the name. `tokenBalances` was fine. From the parent point of view he just hand the tokenBalances, then the child use them just to init them ones he can modify

src/components/DepositWidget/Row.tsx
```diff
@@ -28,19 +31,39 @@ const TokenTr = styled.tr`
...
 }
 
-export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
+const txOptionalParams: TxOptionalParams = {
+  onSentTransaction: (receipt: Receipt): void => {
+    toast.info(
+      <div>
+        The transaction has been sent! Check{' '}
+        <a href={`https://etherscan.io/tx/${receipt.transactionHash}`} target="_blank" rel="noopener noreferrer">
+          {abbreviateString(receipt.transactionHash, 6, 4)}
```
anxolin: Nice 👍

src/components/DepositWidget/Row.tsx
alfetopito: Reverted

src/components/DepositWidget/Row.tsx
```diff
@@ -28,19 +31,39 @@ const TokenTr = styled.tr`
...
+        </a>{' '}
+        for details
+      </div>,
+    )
+  },
+}
+
+export const Row: React.FC<RowProps> = (props: RowProps) => {
+  const { balances, updateBalances } = useTokenBalances(props.initialTokenBalances.address)
```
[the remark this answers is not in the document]
alfetopito: What I did was to implement approach `2`, although might not have been clear.
I start with the copy of `tokenBalances` from the parent.

> why you listen for props.initialTokenBalances.address changes. Do you expect that to change?

Using `props.tokenBalances.address`, the hook is initialized and only fetch info related to the given token from the `depositApi`.
My main goal here was to avoid querying balances for all tokens when only 1 had any update.
So, whenever there's an action that modifies the balances, I call `updateBalances` to force the state to be updated and trigger a redraw.

But let's say I go for approach `1`. 
On `Row`, the deposit action is performed, then `onDeposit` from the parent is invoked.
What needs to happen on the parent in order to trigger the balances re-fetch and redraw?
Isn't it needed to update the `balances` state somehow?

On my mind, even if I don't make `useTokenBalances` update a single token at a time, I would need the `updateBalances` function.

src/hooks/useEnableToken.ts
```diff
@@ -20,12 +20,12 @@ export const useEnableTokens = (params: Params): Result => {
   const { enabled: enabledInitial, address: tokenAddress } = params.tokenBalances
   const [enabled, setEnabled] = useState(enabledInitial)
   const [enabling, setEnabling] = useState(false)
-  const [highlight, setHightlight] = useState(false)
-  let unmounted = false
+  const [highlight, setHighlight] = useState(false)
+  const mounted = useRef(true)
```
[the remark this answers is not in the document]
alfetopito: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

Also, the react hooks eslist rule given the following warning when not using a ref:

> Assignments to the 'mounted' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect.eslint(react-hooks/exhaustive-deps)

src/hooks/useTokenBalances.ts
```diff
@@ -1,55 +1,77 @@
...
 
-  return { balances, error }
+    return function cleanUp(): void {
+      mounted.current = false
+    }
+  }, [tokenAddress])
+
+  const updateBalances = async (): Promise<void> => {
+    if (mounted.current) {
+      setBalances(await _getBalances(tokenAddress))
```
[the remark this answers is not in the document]
alfetopito: Weird, I just tested on https://pr77--dexreact.review.gnosisdev.com/deposit/ and it works fine:

Before:

After:


Regardless, even though we already know the amount, after the withdraw the amount has to be updated, correct?
You mean just update it locally in the Row component? I'll give that a try...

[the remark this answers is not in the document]
alfetopito: Ok, I think I finally got what you mean 

/alleluia

Check the new version

src/components/DepositWidget/Row.tsx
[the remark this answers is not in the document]
alfetopito: Never mind all that, next comment helped me get it. The new version should address it :)

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

src/components/DepositWidget/Row.tsx
```diff
@@ -112,7 +141,26 @@ export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
...
         <td title={formatAmountFull(exchangeBalanceTotal, decimals)}>{formatAmount(exchangeBalanceTotal, decimals)}</td>
-        <td title={formatAmountFull(withdrawingBalance, decimals)}>{formatAmount(withdrawingBalance, decimals)}</td>
+        <td title={formatAmountFull(withdrawingBalance, decimals)}>
+          {withdrawable ? (
+            <>
+              <button className="success" onClick={_withdraw} disabled={withdrawing}>
+                {withdrawing && <FontAwesomeIcon icon={faSpinner} spin />}
+                &nbsp; {formatAmount(withdrawingBalance, decimals)}
+              </button>
+              <div className="success">
```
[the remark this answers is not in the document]
alfetopito: It had, I lost it on the merge. Will add it back.

src/hooks/useTokenBalances.ts
```diff
@@ -1,55 +1,77 @@
...
-      walletBalance,
-      enabled: allowance.eq(ALLOWANCE_VALUE),
-    }
-  })
+  const balancePromises = tokens.map(async token => fetchBalancesForToken(token, userAddress, contractAddress))
   return Promise.all(balancePromises)
 }
 
-export const useTokenBalances = (): UseTokenBalanceResult => {
+export const useTokenBalances = (tokenAddress?: string): UseTokenBalanceResult => {
```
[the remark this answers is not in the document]
alfetopito: Reverted

src/components/DepositWidget/Row.tsx
```diff
@@ -56,20 +79,9 @@ export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
...
-              {receipt.transactionHash.slice(0, 10)}...
-            </a>{' '}
-            for details
-          </div>,
-        )
-      },
-    },
+    txOptionalParams,
   })
+  const { withdrawable, withdrawing, withdraw } = useWithdrawTokens({ tokenBalances, txOptionalParams })
```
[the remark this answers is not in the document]
alfetopito: Yeah, forgot to use this error. How would be best to display this loading error?

src/components/DepositWidget/Row.tsx
```diff
@@ -112,7 +141,26 @@ export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
         </td>
         <td>{name}</td>
         <td title={formatAmountFull(exchangeBalanceTotal, decimals)}>{formatAmount(exchangeBalanceTotal, decimals)}</td>
-        <td title={formatAmountFull(withdrawingBalance, decimals)}>{formatAmount(withdrawingBalance, decimals)}</td>
+        <td title={formatAmountFull(withdrawingBalance, decimals)}>
+          {withdrawable ? (
+            <>
+              <button className="success" onClick={_withdraw} disabled={withdrawing}>
```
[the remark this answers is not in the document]
alfetopito: Done

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

anxolin:
> Should we maybe call it claimable instead of withdrawable? Given that the other button us is already called withdraw?
@fleupold @alfetopito , sorry I forgot to answer this one. Withdrawable is weird, and claim I agree that is weird too in this context. What about just "Withdraw"

src/components/DepositWidget/Row.tsx
```diff
@@ -40,7 +58,22 @@ function _loadFallbackTokenImage(event: React.FormEvent<HTMLImageElement>): void
   image.src = unknownTokenImg
 }
 
-export const Row: React.FC<RowProps> = ({ tokenBalances }: RowProps) => {
+const txOptionalParams: TxOptionalParams = {
+  onSentTransaction: (receipt: Receipt): void => {
+    toast.info(
+      <div>
```
Velenir: This much code looks like it could be its own component
`<TxPopup hash={}/>` maybe?

alfetopito: Sure. I can do that in another PR.

src/api/exchange/DepositApiMock.ts
```diff
@@ -146,6 +148,9 @@ export class DepositApiMock implements DepositApi {
     pendingWithdraws.amount = ZERO
     balanceState.balance = balanceState.balance.sub(amount)
 
+    // mock transfer tokens to user's mock `wallet`
+    await this._erc20Api.transfer(tokenAddress, this.getContractAddress(), userAddress, amount)
```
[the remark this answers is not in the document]
alfetopito: Yes, they are missing indeed. I thought about doing it now, but decided to leave it for when dealing with that, to not increase the scope of this PR further.

alfetopito:
> Well, I think loading withdrawability make sense in the widget, not in the row, so we get all data together.
Do you mean to lift this up to the parent component?

src/components/DepositWidget/Row.tsx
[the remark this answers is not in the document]
alfetopito: Hmmm, then the hook would take in all balances and calculate it for all tokens in the same call?
Or you mean only the `withdrawable` flag is calculated on the parent, while the `withdrawing` (and the new `highlight`) are on the Row?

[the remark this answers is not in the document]
alfetopito: Done as suggested, please take a look at the latest revision

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

anxolin:
@alfetopito  Reviewed and liked last commits, especially https://github.com/gnosis/dex-react/pull/77/commits/25320c5b78f90cb275707616b3557e219883dfdc

Regarding the hightlighting, I like the change, but I would rename the other highlight var too. Also note that you don't need to have a specific name in the returning object of `useWithdrawTokens`

You can rename when extracting in JS:
```
const { highlight : highlighWithdraw, ... } = useWithdrawTokens(...)
```
This would apply to both hooks.

gnosis-info:
Travis automatic deployment:
 https://pr77--dexreact.review.gnosisdev.com

alfetopito:
> @alfetopito Reviewed and liked last commits, especially [25320c5](https://github.com/gnosis/dex-react/commit/25320c5b78f90cb275707616b3557e219883dfdc)
> 
> Regarding the hightlighting, I like the change, but I would rename the other highlight var too. Also note that you don't need to have a specific name in the returning object of `useWithdrawTokens`
> 
> image
> 
> You can rename when extracting in JS:
> 
> ```
> const { highlight : highlighWithdraw, ... } = useWithdrawTokens(...)
> ```
> 
> This would apply to both hooks.

Done. I'll merge straight away then.

Issue #81, opened 2019-09-30
Connect wallet using dapp-ui


anxolin:
* [X] Review/Fix PRs: https://github.com/gnosis/dapp-ui/pulls
* [X] Integrate wallet with the connect button in the web

It may have a dependency with:
- #80 
- #79

PR #83, opened 2019-10-01
Connect/Disconnect wallet and provide Wallet info

anxolin:
This PR allows to connect and disconnect a wallet, provides a simple hook with the wallet information, and allows to protect pages that require a connected wallet.

Apologies cause the PR is a big bigger than expected. All the changes are related, I hope that with the explanations it's a bit clear all the changes involved.

Includes:
- [X] Provides a hook that can be easily used to received the connected wallet info and keep the components updated on every wallet change: `useWalletConnection.tsx`
- [X] subscribe/desubscribe functions in WalletApi to notify any change in the wallet information
- [X] Extracted and createad a component our of the "Connect Wallet" button from the Header. It's called `Wallet.tsx`
- [X] The `Wallet` detects if the user is connected or not and depending on that, it will:
  * [X] Show "Connect Wallet" button that triggers the connect attempt
  * [X] Show uses address and a "Disconnect Wallet" button that triggers the disconnect attempt
- [X] When the user is connecting/disconnecting it shows a loading state and the right label in the button
- [X] `ConnectWallet.tsx`: New page that will ask you to connect. It would be the equivalent to a login page.
- [X] Define `PrivateRoute` for protecting the pages that require a connected wallet. This way the deposit page can assume there's a connected wallet and doesn't need to add any validation. For the trade page, the approach will be different since we ask the user to connect only when the user tries to trade.

**Out of the scope of the PR**
- It doesn't style a lot the wallet information (#89 )
- It doesn't implement the Web3 implementation. Only connects using the mock. (#81 )
- It doesn't remove the `TODOS` in the paces where there was a `// TODO: Review after implementing connect wallet.` because this will be handled in a separate PR (#105)
- I think it's convenient to do a autoconnect for the mock implementation, so we don't need to connect for quick testing and on every reload. This will be handled in a separate PR too (#93 )

**Test**
- https://pr83--dexreact.review.gnosisdev.com

### DIFF
--- a/src/api/wallet/WalletApiMock.ts
+++ b/src/api/wallet/WalletApiMock.ts
@@ -1,4 +1,4 @@
-import { WalletApi, Network } from 'types'
+import { WalletApi, Network, WalletInfo } from 'types'
 import BN from 'bn.js'
 import assert from 'assert'
 
@@ -11,22 +11,24 @@ import { USER_1 } from '../../../test/data'
 export class WalletApiMock implements WalletApi {
   private _connected = false
   private _balance = toWei(new BN(2.75), 'ether')
+  private _listeners: ((walletInfo: WalletInfo) => void)[] = []
 
   public isConnected(): boolean {
     return this._connected
   }
 
   public async connect(): Promise<void> {
-    // TODO: Uncomment after doing login
-    // await wait()
+    await wait(500)
     this._connected = true
     log('[WalletApiMock] Connected')
+    await this._notifyListeners()
   }
 
   public async disconnect(): Promise<void> {
-    await wait()
+    await wait(500)
     this._connected = false
     log('[WalletApiMock] Disconnected')
+    await this._notifyListeners()
   }
 
   public async getAddress(): Promise<string> {
@@ -45,6 +47,23 @@ export class WalletApiMock implements WalletApi {
     assert(this._connected, 'The wallet is not connected')
 
     return Network.Rinkeby
+  }
+
+  public addOnChangeWalletInfo(callback: (walletInfo: WalletInfo) => void): void {
+    this._listeners.push(callback)
+  }
+
+  public removeOnChangeWalletInfo(callback: (walletInfo: WalletInfo) => void): void {
+    this._listeners = this._listeners.filter(c => c !== callback)
+  }
+
+  private async _notifyListeners(): Promise<void> {
+    const walletInfo: WalletInfo = {
+      isConnected: this._connected,
+      userAddress: this._connected ? await this.getAddress() : undefined,
+      networkId: this._connected ? await this.getNetworkId() : undefined,
+    }
+    this._listeners.forEach(listener => listener(walletInfo))
   }
 }
 
--- a/src/api/wallet/WalletApiMock.ts
+++ b/src/api/wallet/WalletApiMock.ts
@@ -60,7 +60,6 @@ export class WalletApiMock implements WalletApi {
 
   public removeOnChangeWalletInfo(callback: (walletInfo: WalletInfo) => void): void {
     this._listeners = this._listeners.filter(c => c !== callback)
-    console.log('removeOnChangeWalletInfo', this._listeners.length)
   }
 
   private _getWalletInfo(): WalletInfo {
--- a/src/api/wallet/WalletApiMock.ts
+++ b/src/api/wallet/WalletApiMock.ts
@@ -70,7 +70,7 @@ export class WalletApiMock implements WalletApi {
     }
   }
 
-  private async _notifyListeners(): Promise<void> {
+  private _notifyListeners(): void {
     const walletInfo: WalletInfo = this._getWalletInfo()
     this._listeners.forEach(listener => listener(walletInfo))
   }
--- a/src/api/wallet/WalletApiMock.ts
+++ b/src/api/wallet/WalletApiMock.ts
@@ -10,6 +10,8 @@ import { USER_1 } from '../../../test/data'
  */
 export class WalletApiMock implements WalletApi {
   private _connected = false
+  private _user = USER_1
+  private _networkId = Network.Rinkeby
   private _balance = toWei(new BN(2.75), 'ether')
   private _listeners: ((walletInfo: WalletInfo) => void)[] = []
 
@@ -18,14 +20,14 @@ export class WalletApiMock implements WalletApi {
   }
 
   public async connect(): Promise<void> {
-    await wait(500)
+    await wait(1000)
     this._connected = true
     log('[WalletApiMock] Connected')
     await this._notifyListeners()
   }
 
   public async disconnect(): Promise<void> {
-    await wait(500)
+    await wait(1000)
     this._connected = false
     log('[WalletApiMock] Disconnected')
     await this._notifyListeners()
@@ -34,7 +36,7 @@ export class WalletApiMock implements WalletApi {
   public async getAddress(): Promise<string> {
     assert(this._connected, 'The wallet is not connected')
 
-    return USER_1
+    return this._user
   }
 
   public async getBalance(): Promise<BN> {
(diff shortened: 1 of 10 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr83--dexreact.review.gnosisdev.com

src/api/wallet/WalletApiMock.ts
```diff
@@ -44,7 +48,31 @@ export class WalletApiMock implements WalletApi {
...
+
+  private _getWalletInfo(): WalletInfo {
+    return {
+      isConnected: this._connected,
+      userAddress: this._connected ? this._user : undefined,
+      networkId: this._connected ? this._networkId : undefined,
+    }
+  }
+
+  private async _notifyListeners(): Promise<void> {
```
Velenir: This doesn't need to be async

src/api/wallet/WalletApiMock.ts
```diff
@@ -44,7 +48,31 @@ export class WalletApiMock implements WalletApi {
...
+    const walletInfo: WalletInfo = this._getWalletInfo()
+    this._listeners.forEach(listener => listener(walletInfo))
```
Velenir: So your idea is to trigger listeners on every change, without specifying the event type `notifyListers('accounChanged')`?
And then listeners will have to re-query the WalletInfo state

src/components/Wallet.tsx
```diff
@@ -0,0 +1,88 @@
...
+  } else if (isConnected) {
+    onClick = disconnectWallet
+    content = (
+      <>
+        {userAddress.substr(0, 9)}...{userAddress.substr(35, userAddress.length)} &nbsp;<small>(disconnect)</small>
+      </>
+    )
+  } else {
+    onClick = connectWallet
+    content = <>Connect to Wallet</>
```
Velenir: This can just be a string `content = 'Connect to Wallet'`
Unless TS complains...

src/components/Wallet.tsx
```diff
@@ -0,0 +1,88 @@
...
+  text-decoration: none;
+  min-width: 16em;
+  text-align: center;
+
+  &:hover {
+    color: #ff5097;
+  }
+
+  &:hover {
+    background-color: #ffe3ee;
```
Velenir: Put them inside one `&:hover`

src/components/layout/Header.tsx
```diff
@@ -63,11 +63,16 @@ const Wrapper = styled.header`
     margin-top: 0;
     color: #e0aacf;
   }
-`
 
-function connectWallet(): void {
-  alert('TODO: Connect to Wallet')
-}
+  a${Wallet} {
```
Velenir: Why specify `a` tag?
If anything you are breaking compatibility with usage as `<Wallet as="button"/>` if we ever needed that

src/components/layout/Header.tsx
```diff
@@ -30,8 +31,7 @@ const Wrapper = styled.header`
     }
   }
 
-  .logo,
-  .connect-wallet {
+  .logo {
```
Velenir: Put 
```
.logo {
    font-size: 1.2em;
  }
```
from below inside this one too

src/api/wallet/WalletApiMock.ts
```diff
@@ -44,7 +48,31 @@ export class WalletApiMock implements WalletApi {
...
 
-    return Network.Rinkeby
+    return this._networkId
+  }
+
+  public addOnChangeWalletInfo(callback: (walletInfo: WalletInfo) => void, trigger?: boolean): void {
+    this._listeners.push(callback)
+    if (trigger) {
+      callback(this._getWalletInfo())
+    }
```
Velenir: consider returning `removeOnChangeWalletInfo(callback)` for ease of unsubscribe, especially inside `useEffect`

src/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,16 @@
...
+import { WalletInfo } from 'types'
+
+export const useWalletConnection = (): WalletInfo => {
+  const [walletInfo, setWalletInfo] = useState<WalletInfo>({ isConnected: false })
+
+  useEffect((): (() => void) => {
+    walletApi.addOnChangeWalletInfo(setWalletInfo, true)
+
+    return function cleanup(): void {
+      walletApi.removeOnChangeWalletInfo(setWalletInfo)
```
Velenir: this whole passage could just be
```
useEffect(() => walletApi.addOnChangeWalletInfo(setWalletInfo), [])
```
if you implement my suggestion for `WalletApi::addOnChangeWalletInfo`

src/components/Wallet.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <>
+        <FontAwesomeIcon icon={faSpinner} />
+        {' ' + loadingLabel}
+      </>
+    )
+  } else if (isConnected) {
+    onClick = disconnectWallet
+    content = (
+      <>
+        {userAddress.substr(0, 9)}...{userAddress.substr(35, userAddress.length)} &nbsp;<small>(disconnect)</small>
```
alfetopito: Use the `abbreviateString` helper

src/hooks/useEnableToken.ts
```diff
@@ -33,9 +33,10 @@ export const useEnableTokens = (params: Params): Result => {
     assert(!enabled, 'The token was already enabled')
 
     setEnabling(true)
+
     // TODO: Review after implementing connect wallet.
     //   Probably some APIs should have an implicit user and it should be login aware
-    walletApi.connect()
+    // walletApi.connect()
```
alfetopito: Remove this entirely?

src/api/wallet/WalletApiMock.ts
anxolin: Right. It was before when `this._getWalletInfo()` was async. It was using the getters instead of the attributes.

Fixed

src/api/wallet/WalletApiMock.ts
anxolin: I was thinking on diferencienting but went for this simple approach.
We could provide the name the events, but notice that a connect event would modify the address and network.

Not a strong opinion, if you see the hook the implementation is super simple.
It may be improvable cause it can trigger more re-renders than it should, but I don't see it as a big issue.

src/components/Wallet.tsx
anxolin: I don't understand, I added them in `&:hover`, didn't I?

src/components/layout/Header.tsx
anxolin: Fixed, I think I had an issue with the specificity, but it was solved and that `a` was not removed

src/components/layout/Header.tsx
anxolin: right

src/api/wallet/WalletApiMock.ts
anxolin: Not sure if you meant exactly [this](https://github.com/gnosis/dex-react/pull/83/commits/36053915fee69995bffd98e0c553a114f695ae33), but it was a good suggestion:

src/hooks/useWalletConnection.ts
anxolin: yep, I just read this now after implementing it for your previous comment.

I did that, but just not in one line. 
Usually when the line does several things, I prefer to separate them so each line has it’s porpouse:
image

1. useEffect line + definition of the argument function
2. Impl of the effect
3. dependencies

gnosis-info:
Travis automatic deployment:
 https://pr83--dexreact.review.gnosisdev.com

src/components/Wallet.tsx
anxolin: I was in my TODO, because it wasn't merged yet.

It was merged yesterday super late :) 
I'll rebase and add it

src/components/Wallet.tsx
Velenir: I mean L82 and L86 both have `:hover`

gnosis-info:
Travis automatic deployment:
 https://pr83--dexreact.review.gnosisdev.com

src/api/wallet/WalletApiMock.ts
Velenir: Not a strong opinion either :+1:

src/hooks/useEnableToken.ts
anxolin: There's a comment regarding this in the description of the PR.

I just created a issue for this #105

src/components/Wallet.tsx
alfetopito: Oh, forgot it was part of my super long PR :D

src/api/wallet/WalletApiMock.ts
anxolin: Okay, I think as we include more APIs and components common sense will tell us what should be done :)

src/components/Wallet.tsx
anxolin: Aha, u are right. I'll correct that in my next PR

Issue #106, opened 2019-10-02
Deposit: reduce allowance on success


alfetopito:
I think you need to also reduce the allowance on top of sending the tokens. And I believe this will uncover a mistake I made when detecting if a token is enabled (I made the comment in another PR)

_Originally posted by @anxolin in https://github.com/gnosis/dex-react/pull/104_

PR #108, opened 2019-10-02
Change check for enabled tokens

anxolin:
Address one of the problems mentioned in #106 

- When enabling the tokens we set the maximun amount. That's a huge number `2**256-1`
- When checking if the token is enabled, we relax the conditin, but still an arbitrarily huge number that is much smaller than the max: It uses `2**128`


gnosis-info:
Travis automatic deployment:
 https://pr108--dexreact.review.gnosisdev.com

Issue #105, opened 2019-10-02
Review use of wallet information in APIs


anxolin:
With PR #83, now some components that needs to get the address and network should be reviewed.

Theres's some TODOS with `// TODO: Review after implementing connect wallet` that should be reviewed.

This PR aims to find a simple way to use the address/network information.

PR #110, opened 2019-10-02
106/reduce allowance on deposit

alfetopito:
Implements https://github.com/gnosis/dex-react/issues/106

- Added optional `senderAddress` to `TxOptionalParams` to pass along the sender, so we can emulate `msg.sender` inside the mocks
- Reducing the allowance upon transfers made by an address that's not the owner of the tokens
- Added log message when transferring
- Removing format from `approve` log message because format requires token's `decimals` to properly format output


gnosis-info:
Travis automatic deployment:
 https://pr110--dexreact.review.gnosisdev.com

src/api/erc20/Erc20ApiMock.ts
```diff
@@ -82,16 +80,47 @@ export class Erc20ApiMock implements Erc20Api {
     this._initBalances(fromAddress, tokenAddress)
     this._initBalances(toAddress, tokenAddress)
 
+    // Since we don't have a msg.sender, we mimic it by using the optional field
+    const { senderAddress: spenderAddress } = txOptionalParams
```
Velenir: I'm not sure it's a good idea to add this parameter which will have to go when switching to real contracts, anyway.
As `msg.sender` is the internal responsibility of Mmask/other provider, how about grabbing it from a const
```
class Erc20ApiMock ...
  static sender = ....
  ...
  const sender = this.sender || userAddress
```
or something

Also in real web3 TxParams it's called `from`, not that it matters yet.

anxolin: I thought about this, and in the one hand I agree with @DmitryBespalov , web3 usually have a default account setup and you can define any other by changing the `from`. But in the other, i thought it's more general if we provide the address (this is precisely what I was going to review in #105 ). If you see the other APIs, I send always the `fromAddress`.

image

The reason I did this is to make APIs more general. This also allows in the future for example to reuse this APIs easily for any backend service, bots or CLI that handle multiple accounts.

Regarding the name, I like `senderAddress` and could leave it like that perfectly, but I like slightly more the `fromAddress`.

anxolin: Mmm @alfetopito, on a deeper look, I saw there was already a `fromAddress` that it's exactly that:
image

I do expect to add more optional params in `TxOptionalParams`, like maybe `gasLimit` or `gasPrice`. The `fromUser` could go, but for now I think it has enough relevance to have it's own mandatory param (I might change my mind after my #105 proposal :P )

src/api/erc20/Erc20ApiMock.ts
```diff
@@ -82,16 +80,47 @@ export class Erc20ApiMock implements Erc20Api {
...
+    assert(spenderAddress, 'TxOptionalParams.senderAddress is required for this mock work')
+    this._initAllowances(fromAddress, tokenAddress, spenderAddress)
+
     const balance = this._balances[fromAddress][tokenAddress]
     assert(balance.gte(amount), "The user doesn't have enough balance")
 
+    assert(
+      this._checkAndUpdateAllowance(spenderAddress, fromAddress, tokenAddress, amount),
```
anxolin: I think should never update anything. I understand the method should be called `_hasAllowance`, then after the assert, we update the allowances.

src/api/exchange/DepositApiMock.ts
```diff
@@ -104,7 +104,9 @@ export class DepositApiMock implements DepositApi {
     pendingDeposits.batchId = currentBatchId
 
     // mock transfer tokens from user's mock `wallet`
-    await this._erc20Api.transfer(tokenAddress, userAddress, this.getContractAddress(), amount)
+    await this._erc20Api.transfer(tokenAddress, userAddress, this.getContractAddress(), amount, {
+      senderAddress: this.getContractAddress(),
```
anxolin: As said in the comment, I understand we don't need this, the `fromAddress` (userAddress here) makes the `from`.
In addition, the sender cannot be the contract address :)

src/api/erc20/Erc20ApiMock.ts
alfetopito: First, replying to the last comment by @anxolin 

> Mmm @alfetopito, on a deeper look, I saw there was already a fromAddress that it's exactly that:

This `fromAddress`, AFAIU, is inline with the `transferFrom` contract interface defined for ERC20: `function transferFrom(address _from, address _to, uint256 _value) public returns (bool success)
`

That means, I'm transferring from account `fromAddress` to account `toAddress`. The sender is not part of the parameters.

alfetopito: @Velenir if I understand correctly, you mean to have the `msg.sender` as a member variable o Erc20ApiMock class?
If so, yes, that would simplify things. The sender will always be the same (the mock contract address).

I'll changed it.

src/api/erc20/Erc20ApiMock.ts
alfetopito: Ok, will split it.

src/api/exchange/DepositApiMock.ts
alfetopito: @anxolin Wait, who should be the sender then?

anxolin: The user is depositing tokens in the exchange, so they are tokens that are deducted from the user's balanace and goes to the contract balance:
I don't see why we need the sender:
```js
await this._erc20Api.transfer(tokenAddress, userAddress, this.getContractAddress(), amount)
```

src/api/erc20/Erc20ApiMock.ts
anxolin: @alfetopito , no I mean the ERC20 function called `transfer`, not `transferFrom`.
`transferFrom` although is in the ERC20 standard, I think we don't need it, so I didn't create it.

So my field `fromAddress` is what dima is calling `from`.
I understand that he is meaning that that is implicit, cause most of the times in the `web3` has already a defined default account. But I model the interfaces more explicit, so you provide always the user. This is part of the review I want to do in #105

gnosis-info:
Travis automatic deployment:
 https://pr110--dexreact.review.gnosisdev.com

src/api/exchange/DepositApiMock.ts
alfetopito: Let's go through the process to check if we are on the same line:

1. User enables token:
  a. A `approve` tx is created with `spender` == contract address, `from` == user address, `token` == token address, `amount` == max amount.
1. User deposits token:
  a. A `transferFrom` tx is executed from the ERC20 contract (?) with `sender` == (user or contract?), `from` == user address, `to` == contract address
  b. Depending on who was the `sender` in the previous step, reduce the allowance.
1. Withdraw request -- not relevant
1. User withdraws
  a. A `transferFrom` tx is executed from the ERC20 contract (?) with `sender` == (user or contract?), `from` == contract address, `to` == user address
  b. Depending on who was the `sender` in the previous step, reduce the allowance.

Now that I was writing this I got confusing regarding where the tx originates from and at what point the allowance will be deduced.

@anxolin can you point where I got this wrong?

src/api/erc20/Erc20ApiMock.ts
alfetopito: @anxolin  Ah, so we are not using `transferFrom` at all. In this case, why do we need the allowance? 
I understood this is only used when another account is transferring from your balance on your behalf. If you yourself is calling `transfer`, there's no need to check the allowance at all, no?

anxolin: >  I think we don't need it, so I didn't create it.
Okay, I thought that, but you are actually right. It's true that we don't need it in the app itself, **BUT** you were using it for implementing the deposit mock function. This function DO need to use `transferFrom` so I guess we do have to implement it after all.

Sorry for missleading you here. I guess we need a new method in `Erc20Api` similar to:
```
transferFrom(
    tokenAddress: string,
    senderAddress: string,
    fromAddress: string,
    toAddress: string,
    amount: BN,
    txOptionalParams?: TxOptionalParams,
  ): Promise<TxResult<boolean>>
```

src/api/exchange/DepositApiMock.ts
anxolin: I think the confusion was what I explained in the other thread. In case of doubt check the contract, cause the mock tries to mimic it, so as you see is what you say, it's a `transferFrom(user, contract, amount)`  where the sender is in fact the contract. We won't be needing to call this function in the app, but in the mock implementation makes sense.

image

src/api/erc20/Erc20ApiMock.ts
alfetopito: Checked the base contract. It does use `transferFrom` on deposit https://github.com/gnosis/dex-contracts/blob/8c93796971b02853b2dc7dfb8332f7b97d846777/contracts/EpochTokenLocker.sol#L49-L54

I'll add it

gnosis-info:
Travis automatic deployment:
 https://pr110--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr110--dexreact.review.gnosisdev.com

PR #113, opened 2019-10-02
86/extract etherscan component

### LINKED ISSUE #86: Component: Etherscan Link
Watch Anxo's PR about connect wallet (useWalletConnection) https://github.com/gnosis/dex-react/pull/83

**Estimation:** 0.5

alfetopito:
Closes #86 

- Extracted the etherescan notification into the `TxNotification` and `EtherscanLink` components
- Link is set according to network used
- Added component tests

### DIFF
--- a/src/components/EtherscanLink.tsx
+++ b/src/components/EtherscanLink.tsx
@@ -1,0 +1,52 @@
+import React, { ReactElement } from 'react'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import { abbreviateString } from 'utils'
+import { Network } from 'types'
+
+export enum EtherscanLinkType {
+  tx,
+  address,
+  contract,
+  token,
+}
+
+export interface EtherscanLinkProps {
+  type: EtherscanLinkType
+  identifier: string
+  label?: string | ReactElement | void
+}
+
+function getEtherscanDomainPrefix(networkId: Network): string {
+  return !networkId || networkId === Network.Mainnet ? '' : Network[networkId].toLowerCase() + '.'
+}
+
+function getEtherscanDomainSuffix(type: EtherscanLinkType, identifier: string): string {
+  switch (type) {
+    case EtherscanLinkType.tx:
+      return `tx/${identifier}`
+    case EtherscanLinkType.address:
+      return `address/${identifier}`
+    case EtherscanLinkType.contract:
+      return `address/${identifier}#code`
+    case EtherscanLinkType.token:
+      return `token/${identifier}`
+  }
+}
+
+export const EtherscanLink: React.FC<EtherscanLinkProps> = ({ type, identifier, label }) => {
+  const { networkId } = useWalletConnection()
+
+  if (!networkId) {
+    return null
+  }
+
+  const href = `https://${getEtherscanDomainPrefix(networkId)}etherscan.io/${getEtherscanDomainSuffix(
+    type,
+    identifier,
+  )}`
+  return (
+    <a href={href} target="_blank" rel="noopener noreferrer">
+      {label ? label : abbreviateString(identifier, 6, 4)}
+    </a>
+  )
+}
--- a/src/components/EtherscanLink.tsx
+++ b/src/components/EtherscanLink.tsx
@@ -3,12 +3,7 @@ import { useWalletConnection } from 'hooks/useWalletConnection'
 import { abbreviateString } from 'utils'
 import { Network } from 'types'
 
-export enum EtherscanLinkType {
-  tx,
-  address,
-  contract,
-  token,
-}
+type EtherscanLinkType = 'tx' | 'address' | 'contract' | 'token'
 
 export interface EtherscanLinkProps {
   type: EtherscanLinkType
@@ -22,13 +17,13 @@ function getEtherscanDomainPrefix(networkId: Network): string {
 
 function getEtherscanDomainSuffix(type: EtherscanLinkType, identifier: string): string {
   switch (type) {
-    case EtherscanLinkType.tx:
+    case 'tx':
       return `tx/${identifier}`
-    case EtherscanLinkType.address:
+    case 'address':
       return `address/${identifier}`
-    case EtherscanLinkType.contract:
+    case 'contract':
       return `address/${identifier}#code`
-    case EtherscanLinkType.token:
+    case 'token':
       return `token/${identifier}`
   }
 }
--- a/src/components/TxNotification.tsx
+++ b/src/components/TxNotification.tsx
@@ -1,0 +1,27 @@
+import React from 'react'
+import { abbreviateString } from 'utils'
+import { Network } from 'types'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+
+interface TxNotificationProps {
+  txHash: string
+}
+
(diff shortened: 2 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr113--dexreact.review.gnosisdev.com

src/components/TxNotification.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { abbreviateString } from 'utils'
+import { Network } from 'types'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+
+interface TxNotificationProps {
+  txHash: string
+}
+
+function getDomainPrefix(networkId: Network): string {
```
anxolin: move to utils?

src/components/TxNotification.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const TxNotification: React.FC<TxNotificationProps> = ({ txHash }: TxNotificationProps) => {
+  const { networkId } = useWalletConnection()
+  const href = `https://${getDomainPrefix(networkId)}etherscan.io/tx/${txHash}`
+
+  return (
+    <div>
+      The transaction has been sent! Check{' '}
+      <a href={href} target="_blank" rel="noopener noreferrer">
```
anxolin: What about this being a component too?
Props:
* type: 'tx' (default) | 'address' | 'code'
* tx: string
* label?: string (by default uses the abbreviated tx, but we can use it for example for the contract code link in the deposit widget)

src/components/TxNotification.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface TxNotificationProps {
+  txHash: string
+}
+
+function getDomainPrefix(networkId: Network): string {
+  return !networkId || networkId === Network.Mainnet ? '' : Network[networkId].toLowerCase() + '.'
+}
+
+export const TxNotification: React.FC<TxNotificationProps> = ({ txHash }: TxNotificationProps) => {
+  const { networkId } = useWalletConnection()
```
anxolin: This can be undefined/null

I would add a:
```js
if (!networkId){
  return null
}
```

...so it doesn't draw anything if there's no connection

anxolin: @DmitryBespalov  what is best returning null/undefined/empty_string or it doesn't matter?

test/components/TxNotification.components.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { render } from 'enzyme'
+
+import { TxNotification } from 'components/TxNotification'
+import { TX_HASH } from '../data'
+import { abbreviateString } from 'utils'
+import { Network, WalletInfo } from 'types'
+
+let networkId: Network = undefined
+
+jest.mock('hooks/useWalletConnection', () => {
```
anxolin: Is it posible to create a test also with `return { isConnected: false }`?

src/components/TxNotification.tsx
Velenir: doesn't matter

alfetopito: @anxolin so you rather not return anything in case the wallet is not connected?

Right, this is not likely to happen, but thinking in terms of interface, considering that we'll use this initially inside a toast notification, the notification will be completely empty.

Unless you mean to only display the notification if there is one, such as:

```
const notification = <TxNotification .../>
if (notification) {
  toast.info(notification...)
}
```

Is that what you have in mind?

anxolin: Yes I meant that, but, that's a good point. So then I would say that this would affect if a user:

1. Create a tx
2. Disconnects
3. Receives the confirmation

You are right, showing a empty notification in that case is a bid bad. The best I can come up with is to "remember" the last network id. In the described event, the user had previously a network and then disconnected. We can try to have always the last networkw we remember :)

alfetopito: That (cache networkId) I believe makes sense to have in the `useWalletInfo` hook, so from that point of view, no changes in this component regarding to that.

I'll make the caching and return null from this component.

test/components/TxNotification.components.test.tsx
alfetopito: Added.

Still, does it matter in that case?
From this component point of view, `networkId` is what we care about.
We know from the logic that if `networId` == `undefined` then `isConnected` is very likely `false`.

The reason for adding it in the mock is because this is a required property of `WalletInfo`.

src/components/TxNotification.tsx
alfetopito: Well, caching in the hook is not as trivial as I thought. The hook doesn't update the walletinfo itself, but instead pass the update funtion as a callback. If this is a must, we should deal with it in another issue/PR

src/components/TxNotification.tsx
alfetopito: I decided to move it back because it's a pretty contained function. I don't see where this would be reused other than on the new component.

src/components/TxNotification.tsx
alfetopito: Check the latest revision :+1:

gnosis-info:
Travis automatic deployment:
 https://pr113--dexreact.review.gnosisdev.com

src/components/TxNotification.tsx
anxolin: It's not a must 👍 
Good suggestion. I would leave it for now

test/components/TxNotification.components.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { EtherscanLink } from 'components/EtherscanLink'
+
+describe('<TxNotification />', () => {
+  it('renders with link component', () => {
+    const wrapper = shallow(<TxNotification txHash={TX_HASH} />)
+    expect(wrapper.text()).toMatch(/^The transaction has been sent! Check .* for details$/)
+    expect(wrapper.contains(<EtherscanLink type="tx" identifier={TX_HASH} />)).toBeTruthy()
+  })
+
+  it('does not render component', () => {
```
anxolin: I think leaving the TODO is just enough

test/components/TxNotification.components.test.tsx
anxolin: Yep, I meant to make some expectancies for when there's no connection (therefore networkId), but I see now u have them

src/components/EtherscanLink.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { ReactElement } from 'react'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import { abbreviateString } from 'utils'
+import { Network } from 'types'
+
+export enum EtherscanLinkType {
```
anxolin: nit: maybe is simpler just to use a type like `'tx' | 'address' | ...`, so is still checked, but easier to use:
* Don't need to import the type, just the component
* less verbose to use
* The IDE still does the check and you even get a suggestion of posible values

alfetopito: Addressed in the latest revision :)

gnosis-info:
Travis automatic deployment:
 https://pr113--dexreact.review.gnosisdev.com

Issue #96, opened 2019-10-01
Keep balance updated: Simple pooling strategy


alfetopito:
DepositWidget query every 30s, update data

**Estimation:** 1

anxolin:
I said 30 sec during the planning, cause the UI updates itself already when you deposit/withdraw, no need to do pooling. 

However things can happen in the contracts: A simple example is we receive a deposit or a trade is executed.

This strategy is meant to be as a simple approach, and good enough for our first iteration in the PoC, later on we can research about different ways we can keep the data more updated (listen for events, sockets, etc..)

anxolin:
Won't do
doing reactive approach instead

PR #120, opened 2019-10-04
Move responsability from row to widget

### LINKED ISSUE #118: Remove local state in Form, move logic up
This would fix the issue described in the comments of: https://github.com/gnosis/dex-react/issues/105

Also, it's a better pattern to notify the owner of the data and let the data flow down as props.

It might even solve the issue that we see with one effect in the console... let's see :)

anxolin:
Fixes #118 

Moves responsibility for updating the data to the `DepositWidget` from row/form.
Expect now the row logic to be much smaller.

**Includes**:
- [X] Models the highlight/select as props for row. Also the events for deposit, withdraw, claim or enable. This way is the parent responsibility now to provide them
- [X] Implement highlight/select logic in `DepositWidget` instead of using the hook

**Aditionally**:
- [X] Adapt tests to the new properties. Add some tests. Small refactor on them.
- [X] Define const for highlight time (used in hook and in widget for now)
- [X] Small refactor for `withdrawing` prop. Renaming it to `claiming` to be coherent with the UI

**Test**:
- Run unit testing
- Check: https://pr120--dexreact.review.gnosisdev.com


gnosis-info:
Travis automatic deployment:
 https://pr120--dexreact.review.gnosisdev.com

src/components/DepositWidget/index.tsx
```diff
@@ -72,6 +99,152 @@ const DepositWidget: React.FC = () => {
...
+          enabling: true,
+        }
+      })
+      const result = await erc20Api.approve(
+        tokenAddress,
+        userAddress,
+        contractAddress,
+        ALLOWANCE_MAX_VALUE,
+        txOptionalParams,
+      )
```
alfetopito: Note to self: I changed the order of this function to be consistent with others and have the `from`/`senderAddress` as first parameter https://github.com/gnosis/dex-react/pull/110/files#diff-db4249d04c7a8259d80ed280f6bfef22L6

No need to change it in this PR, unless that PR is merged first.

src/hooks/useTokenBalances.ts
```diff
@@ -68,10 +72,7 @@ export const useTokenBalances = (): UseTokenBalanceResult => {
   useEffect(() => {
     _getBalances(walletInfo)
       .then(balances => {
-        console.log(
-          '[useTokenBalances] Wallet balances',
-          balances ? balances.map(b => formatAmount(b.walletBalance)) : null,
-        )
+        log('[useTokenBalances] Wallet balances', balances ? balances.map(b => formatAmount(b.walletBalance)) : null)
```
alfetopito: Use the `decimals` amount when formating otherwise this won't be consistent for all tokens

test/components/DepositWidget.components.test.tsx
```diff
@@ -3,59 +3,111 @@ import { render } from 'enzyme'
...
+  it.only('is highlighted', () => {
+    const wrapper = render(
+      _createRow({
+        highlighted: true,
+      }),
+    )
+    expect(wrapper.attr('class')).toMatch(/highlight/)
+  })
+
+  it.only('is enabling', () => {
```
alfetopito: BTW, `it.only` skips the other tests in this suite. Any specific reason? Otherwise, just remove the `.only` from them.

With `only`:

PASS  test/components/DepositWidget.components.test.tsx
  <Row /> not enabled token
    ○ skipped renders single <tr> element
    ○ skipped contains 6 <td> elements
    ○ skipped contains 1 <button> (enable)
  <Row /> enabled token
    ○ skipped renders single <tr> element
    ○ skipped contains 6 <td> elements
    ○ skipped contains 2 <button> elements (deposit and withdraw)
  <Row /> claimable token
    ○ skipped contains 2 <button> elements (claim, deposit, withdraw)
  <Row /> style
    ✓ is highlighted (102ms)
    ✓ is enabling (9ms)

Without:

 PASS  test/components/DepositWidget.components.test.tsx
  <Row /> not enabled token
    ✓ renders single <tr> element (113ms)
    ✓ contains 6 <td> elements (16ms)
    ✓ contains 1 <button> (enable) (14ms)
  <Row /> enabled token
    ✓ renders single <tr> element (9ms)
    ✓ contains 6 <td> elements (13ms)
    ✓ contains 2 <button> elements (deposit and withdraw) (5ms)
  <Row /> claimable token
    ✓ contains 2 <button> elements (claim, deposit, withdraw) (9ms)
  <Row /> style
    ✓ is highlighted (5ms)
    ✓ is enabling (4ms)

src/components/DepositWidget/index.tsx
```diff
@@ -60,9 +68,28 @@ const LinkWrapper = styled.div`
   }
 `
 
+const txOptionalParams: TxOptionalParams = {
+  onSentTransaction: (receipt: Receipt): void => {
+    const notification = <TxNotification txHash={receipt.transactionHash} />
+    if (notification) {
+      toast.info(notification)
```
Velenir: move `const notification` here, no need it create for `else {`

src/components/DepositWidget/index.tsx
```diff
@@ -60,9 +68,28 @@ const LinkWrapper = styled.div`
...
 const DepositWidget: React.FC = () => {
-  const { balances, error } = useTokenBalances()
+  const { userAddress } = useWalletConnection()
+  const { balances, setBalances, error } = useTokenBalances()
+
   const contractAddress = depositApi.getContractAddress()
+  const mounted = useRef(true)
+  useEffect(() => {
+    return function cleanUp(): void {
+      mounted.current = false
```
Velenir: Question: are we using this mounted-check logic often? I may have a proposal for a generic hook `useIfMounted` if we need it

src/components/DepositWidget/index.tsx
```diff
@@ -60,9 +68,28 @@ const LinkWrapper = styled.div`
...
+    }
+  }, [])
 
   if (balances === undefined) {
     // Loading: Do not show the widget
```
Velenir: To render nothing, better return `null`, not a Fragment

src/components/DepositWidget/index.tsx
```diff
@@ -72,6 +99,152 @@ const DepositWidget: React.FC = () => {
...
+
+      if (mounted.current) {
+        _updateToken(tokenAddress, ({ exchangeBalance, walletBalance, ...otherParams }) => {
+          return {
+            ...otherParams,
+            claiming: false,
+            exchangeBalance: exchangeBalance.sub(withdrawingBalance),
+            withdrawingBalance: ZERO,
+            claimable: false,
+            walletBalance: walletBalance.add(withdrawingBalance),
```
Velenir: calls to `_updateToken` that change balances make this otherwise api-controlled component uncontrolled. The way I see it -- the source of truth for balances comes from `useTokenBalances`, which will get updated when `depositApi.withdraw` gets its tx processed, or just when blockchain state is changed. But when you setState with new balances, best case -- it acts as an optimistic update, worst - due to possible race conditions (in async function), balances get subtracted twice.

Imho, get balances from api only, make requests to the api, get new balances from the api -- simpler and better.

src/components/DepositWidget/index.tsx
```diff
@@ -72,6 +99,152 @@ const DepositWidget: React.FC = () => {
...
   )
 
+  function _getToken(tokenAddress: string): TokenBalanceDetails {
+    return balances.find(({ address }) => address === tokenAddress)
+  }
+
+  function _updateToken(tokenAddress: string, updateBalances: Mutation<TokenBalanceDetails>): void {
+    setBalances(balances =>
+      balances.map(tokenBalancesAux => {
+        const { address: tokenAddressAux } = tokenBalancesAux
```
Velenir: This is getting complicated for `setState` . At this point I would `useReducer`, though not strictly necessary, with actions for `highlight | deposit | claim | withdraw`

src/components/DepositWidget/index.tsx
```diff
@@ -99,7 +272,16 @@ const DepositWidget: React.FC = () => {
...
             {balances &&
-              balances.map(tokenBalances => <Row key={tokenBalances.addressMainnet} tokenBalances={tokenBalances} />)}
+              balances.map(tokenBalances => (
+                <Row
+                  key={tokenBalances.addressMainnet}
+                  tokenBalances={tokenBalances}
+                  onEnableToken={(): Promise<void> => _enableToken(tokenBalances.address)}
+                  onSubmitDeposit={(balance): Promise<void> => _deposit(balance, tokenBalances.address)}
+                  onSubmitWithdraw={(balance): Promise<void> => _requestWithdraw(balance, tokenBalances.address)}
+                  onClaim={(): Promise<void> => _claim(tokenBalances.address)}
```
Velenir: There are a lot of functions in this one component, 280 lines already. Maybe an extracted custom hook, like
```
const create createFuncs = useCreateRowFuncs(..., [tokenBalances, whatever])

...
balances.map(tokenBalances => <Row key={...} {...createFuncs(tokenBalances.address)/>)
```

src/hooks/useHighlight.ts
```diff
@@ -22,7 +23,7 @@ export const useHighlight = (): Result => {
         if (mounted.current) {
           setHighlight(false)
         }
-      }, 5000)
+      }, HIGHLIGHT_TIME)
```
Velenir: While not strictly necessary with timeouts, as they clear themselves in time :stuck_out_tongue: , a good practice is to cleanup on unmount and when retriggering, something like:
```
const [highlight, setHighlight] = useState(false)
  const mounted = useRef(true)

  const timeoutId = useRef()
  
  const triggerHighlight = (): void => {
    clearTimeout(timeoutId.current)
    if (mounted.current) {
      setHighlight(true)
      timeoutId.current = setTimeout(() => {
          setHighlight(false)
      }, HIGHLIGHT_TIME)

useEffect(() => => {
    mounted.current = false
    clearTimeout(timeoutId.current)
  }, [])

File: src/components/DepositWidget/index.tsx (hunk 4 above)
Comment (Velenir) [reply src/components/DepositWidget/index.tsx]: The whole _if/else_ check doesn't do anything, for that matter

Comment (W3stside) [reply src/components/DepositWidget/index.tsx]: `notification` is always truthy here

File: src/components/DepositWidget/index.tsx (hunk 5 above)
Comment (W3stside) [reply src/components/DepositWidget/index.tsx]: I've seen it used a lot, go ahead and make an issue for it I'd be interested in seeing it's implementation

File: src/components/DepositWidget/index.tsx (hunk 1 above)
Comment (anxolin) [reply src/components/DepositWidget/index.tsx]: I think it's not merged for now. So resolving this one

File: test/components/DepositWidget.components.test.tsx (hunk 3 above)
Comment (anxolin) [reply test/components/DepositWidget.components.test.tsx]: Right! there was a `.only` there...

File: src/components/DepositWidget/index.tsx (hunk 4 above)
Comment (anxolin) [reply src/components/DepositWidget/index.tsx]: fixed

File: src/components/DepositWidget/index.tsx (hunk 6 above)
Comment (anxolin) [reply src/components/DepositWidget/index.tsx]: weird cause in my code I have:

image

File: src/components/DepositWidget/index.tsx (hunk 8 above)
Comment (anxolin) [reply src/components/DepositWidget/index.tsx]: I thought the same, this part was coded on the plane on my way to Devcon, and that same trip I made a simulation of using a reducer here, and I didn't like where it was heading so I didn't continue that path. 

I think it's not that horrible for now. My attempt to use the reducer was in a branch called https://github.com/gnosis/dex-react/tree/use-reducer

File: src/components/DepositWidget/index.tsx [hunk 11]
```diff
@@ -72,6 +99,152 @@ const DepositWidget: React.FC = () => {
...
+      toast.success(`The token ${symbol} has been enabled for trading`)
+    } catch (error) {
+      console.error('Error enabling the token', error)
+      toast.error('Error enabling the token')
+    }
+  }
+
   return (
     <Wrapper className="widget">
       {contractLink && <LinkWrapper>{contractLink}</LinkWrapper>}
```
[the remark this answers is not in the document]
anxolin: not from this PR and was solved in develop, so I don't see this code anymore

src/components/DepositWidget/index.tsx
anxolin: Very good point. I agree, however `useTokenBalances` doesn't listen for changes for now. I'll create an issue for this, as for now, there wouldn't be racing conditions since the balances are only updated using _updateToken. 

After we do https://github.com/gnosis/dex-react/issues/96 there might be a race condition, super unlikely.

Anyways, at some point we should react better to the state change and remove this imperative state updates in favour of responsive data flowing down.

WDYT?

anxolin: Also, this is not part of this PR since this is just moving the logic up. So whatever we do, better to address it in a separate PR.

src/components/DepositWidget/index.tsx
anxolin: Although is not related, I think this PR #142 simplifies the widget, so it's not so strictly necessary the reducer imo.

src/components/DepositWidget/index.tsx
alfetopito: I updated it in another PR :)

src/components/DepositWidget/index.tsx
Velenir: That's perfectly good. It will be listening eventually.

PR #121, opened 2019-10-12
Fix infinity loop

anxolin:
This PR fixes the infinity loop that the validation is generating.

It solves the issue that can be seen in the console when doing a deposit or withdraw:
image


gnosis-info:
Travis automatic deployment:
 https://pr121--dexreact.review.gnosisdev.com

PR #126, opened 2019-10-16
85/second withdraw confirmation

### LINKED ISSUE #85: Deposit Widget: Second withdraw
See RF.3.6
Detect situation of an user trying to do a second withdraw and ask for confirmation.

Integrate modals: Use it for confirm a second withdraw

Use modals found here https://github.com/gnosis/dex-react/pull/62 and/or get input from Dmitry regarding better approaches.

**Estimation:** 2

alfetopito:
Closes #85 

Dependent of changes on #118. Will keep it as a draft until it is merged.

---

- Add [Modali](https://github.com/upmostly/modali) as dependency
- Add type definitions for Modali
- Move `types` into a separated folder
- Display confirmation modal when user already has a pending withdraw

request-overwrite


gnosis-info:
Travis automatic deployment:
 https://pr126--dexreact.review.gnosisdev.com

src/components/DepositWidget/index.tsx
```diff
@@ -83,6 +84,39 @@ const DepositWidget: React.FC = () => {
   const { userAddress } = useWalletConnection()
   const { balances, setBalances, error } = useTokenBalances()
 
+  const withdrawRequest = useRef({ amount: null, tokenAddress: null, pendingAmount: null, symbol: null })
```
Velenir: Why add indirect _state_ with `useRef`, just do `useState`, it'll update before Modal buttons are created anyway.

alfetopito: My idea was to avoid triggering a redraw by using those values as `instance` variables. I'll try as state as you suggest.

alfetopito:
> Currently the Deposit widget has a lot of closure functions, which are new on each render. That makes all the Row components rerender every time, which may be a problem in the future.
> For example, the action of opening the Modal (when just clicking Withdraw) triggers render 2 times.

@Velenir What do you propose to address this issue with regards to the modal?

src/components/DepositWidget/index.tsx
alfetopito: Updated.

gnosis-info:
Travis automatic deployment:
 https://pr126--dexreact.review.gnosisdev.com

Velenir:
> > Currently the Deposit widget has a lot of closure functions, which are new on each render. That makes all the Row components rerender every time, which may be a problem in the future.
> > For example, the action of opening the Modal (when just clicking Withdraw) triggers render 2 times.
> 
> @Velenir What do you propose to address this issue with regards to the modal?

Memoize all the things :smile: 
```js
const Rows = useMemo(() => {
    function _getToken

    function _updateToken

    function _clearHighlight
    async function _deposit

    async function _requestWithdraw

    async function _claim
    async function _enableToken
    return (
      balances &&
      balances.map(tokenBalances => (
        <Row
          key={tokenBalances.addressMainnet}
          tokenBalances={tokenBalances}
          onEnableToken={(): Promise<void> => _enableToken(tokenBalances.address)}
          onSubmitDeposit={(balance): Promise<void> => _deposit(balance, tokenBalances.address)}
          onSubmitWithdraw={(balance): Promise<void> => _requestWithdraw(balance, tokenBalances.address)}
          onClaim={(): Promise<void> => _claim(tokenBalances.address)}
        />
      ))
    )
  }, [balances, contractAddress, userAddress])
```

src/components/DepositWidget/index.tsx
Velenir: `toggleModal` does `setState` internally, so rerender is triggered either way. And as your `setWithdrawBalance` is close together with `toggleModal` and they are one sync operation (even if they are inside an async function, because await appears later), state is updated once only.
Or that's how I understand it works.

alfetopito:
> > > Currently the Deposit widget has a lot of closure functions, which are new on each render. That makes all the Row components rerender every time, which may be a problem in the future.
> > > For example, the action of opening the Modal (when just clicking Withdraw) triggers render 2 times.
> > 
> > 
> > @Velenir What do you propose to address this issue with regards to the modal?
> 
> Memoize all the things smile
> 
> ```js
> const Rows = useMemo(() => {
>     function _getToken
> 
>     function _updateToken
> 
>     function _clearHighlight
>     async function _deposit
> 
>     async function _requestWithdraw
> 
>     async function _claim
>     async function _enableToken
>     return (
>       balances &&
>       balances.map(tokenBalances => (
>         <Row
>           key={tokenBalances.addressMainnet}
>           tokenBalances={tokenBalances}
>           onEnableToken={(): Promise<void> => _enableToken(tokenBalances.address)}
>           onSubmitDeposit={(balance): Promise<void> => _deposit(balance, tokenBalances.address)}
>           onSubmitWithdraw={(balance): Promise<void> => _requestWithdraw(balance, tokenBalances.address)}
>           onClaim={(): Promise<void> => _claim(tokenBalances.address)}
>         />
>       ))
>     )
>   }, [balances, contractAddress, userAddress])
> ```

Created a new issue (https://github.com/gnosis/dex-react/issues/131) to address that after #120 is finished.

alfetopito:
> I really like it!
> 
> As a small suggestion, I would make the question stand out.
> 
> One proposal is making it look like:
> image
> 
> I've added some styles to the text, and then some margin to the question:
> 
> ```css
>     padding: 0 1em;
>     color: #828282;
>     font-size: 0.85em;
> ```

I like the suggestion. Updated.

Will merge now, I can open a new PR if any changes are required.

PR #128, opened 2019-10-18
100/trading page widget

### LINKED ISSUE #100: Trading page / widget: layout
Style the widget, components

**RF.2.2 Swap widget: Layout**

**Estimation:** 2

alfetopito:
Closes #100 

Implements the basic Trade widget layout:

- [x] Refactoring out a base `Widget` component
  - Moved global `.widget` styles into `Widget` component
  - Made `Widget` `display: flex`
- [x] Refactored out `TokenImg` component
- [x] Implemented basic `TradeWdiget`

trade-widget

---

Not part of this PR:
- Implement button actions
- Interact with user wallet
- Interact with contract
- Show notifications

### DIFF
--- a/src/components/TokenImg.tsx
+++ b/src/components/TokenImg.tsx
@@ -1,0 +1,28 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import unknownTokenImg from 'img/unknown-token.png'
+
+const Wrapper = styled.img`
+  width: 30px;
+  height: 30px;
+`
+
+interface Props {
+  image: string
+  name: string
+  // required to be able to apply custom styles
+  // see https://www.styled-components.com/docs/basics#styling-any-component
+  className?: string
+}
+
+function _loadFallbackTokenImage(event: React.SyntheticEvent<HTMLImageElement>): void {
+  const image = event.currentTarget
+  image.src = unknownTokenImg
+}
+
+const TokenImg: React.FC<Props> = ({ image, name, className }: Props) => {
+  return <Wrapper src={image} alt={name} className={className} onError={_loadFallbackTokenImage} />
+}
+
+export default TokenImg
--- a/src/components/TokenImg.tsx
+++ b/src/components/TokenImg.tsx
@@ -3,26 +3,12 @@ import styled from 'styled-components'
 
 import unknownTokenImg from 'img/unknown-token.png'
 
-const Wrapper = styled.img`
-  width: 30px;
-  height: 30px;
-`
-
-interface Props {
-  image: string
-  name: string
-  // required to be able to apply custom styles
-  // see https://www.styled-components.com/docs/basics#styling-any-component
-  className?: string
-}
-
 function _loadFallbackTokenImage(event: React.SyntheticEvent<HTMLImageElement>): void {
   const image = event.currentTarget
   image.src = unknownTokenImg
 }
 
-const TokenImg: React.FC<Props> = ({ image, name, className }: Props) => {
-  return <Wrapper src={image} alt={name} className={className} onError={_loadFallbackTokenImage} />
-}
-
-export default TokenImg
+export default styled.img.attrs(() => ({ onError: _loadFallbackTokenImage }))`
+  width: 30px;
+  height: 30px;
+`
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -58,7 +58,7 @@ const TokenRow: React.FC<Props> = ({ selectLabel, tokenDetails: { name, image } 
 
   return (
     <Wrapper>
-      <TokenImgWrapper name={name} image={image} />
+      <TokenImgWrapper alt={name} src={image} />
       <SelectBox>
         <label htmlFor={selectId}>{selectLabel}</label>
         <select name="tokenSelector" id={selectId}>
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -54,7 +54,7 @@ interface Props {
 }
 
 const TokenRow: React.FC<Props> = ({ selectLabel, tokenDetails: { name, image } }: Props) => {
-  const selectId = selectLabel.replace(/[^a-zA-Z]+/g, '')
+  const selectId = selectLabel.replace(/[^a-zA-Z]/g, '')
 
   return (
     <Wrapper>
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import TokenImg from 'components/TokenImg'
+import { TokenDetails } from 'types'
+import { Link } from 'react-router-dom'
+
+const Wrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+`
+
+const TokenImgWrapper = styled(TokenImg)`
+  margin-right: 1em;
+  width: 50px;
+  height: 50px;
+`
+
+const SelectBox = styled.div`
+  margin: 0 1em 0 1em;
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+
+  label {
+    text-transform: uppercase;
+  }
+`
+
+const InputBox = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  flex-grow: 1;
+  margin-left: 1em;
+
+  input {
+    margin: 0 0 0.5em 0;
+  }
+`
+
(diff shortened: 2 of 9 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr128--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import TokenRow from './TokenRow'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExchangeAlt, faPaperPlane } from '@fortawesome/free-solid-svg-icons'
+
+function _getTokenDetails(symbol: string, networkId: number): TokenDetails {
+  const _networkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
+  return tokenListApi
+    .getTokens(_networkId)
+    .filter(tokenDetails => tokenDetails.symbol == symbol)
+    .reduce((_, tokenDetails) => tokenDetails)
```
Velenir: I don't quite get the intention of all the chaining. This returns the last TokenDetails with a matching symbol.
Just do `tokenListApi.getTokens(_networkId).find(tokenDetails => tokenDetails.symbol == symbol)` or if you really need the last one, `slice().reverse()` first

src/components/layout/Widget.tsx
```diff
@@ -0,0 +1,42 @@
...
+const Widget: React.FC = ({ children }) => {
+  const contractAddress = depositApi.getContractAddress()
+
+  const contractLink = (
+    <LinkWrapper type="contract" identifier={contractAddress} label={<small>View verified contract</small>} />
+  )
+
+  return (
+    <Wrapper>
+      {contractLink && contractLink}
```
Velenir: `contractLink` is always true

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+`
+
+interface Props {
+  tokenDetails: TokenDetails
+  selectLabel: string
+}
+
+const TokenRow: React.FC<Props> = ({ selectLabel, tokenDetails: { name, image } }: Props) => {
+  const selectId = selectLabel.replace(/[^a-zA-Z]+/g, '')
```
Velenir: No need for `+` in regex if you use global flag

src/components/TokenImg.tsx
```diff
@@ -0,0 +1,28 @@
...
+  className?: string
+}
+
+function _loadFallbackTokenImage(event: React.SyntheticEvent<HTMLImageElement>): void {
+  const image = event.currentTarget
+  image.src = unknownTokenImg
+}
+
+const TokenImg: React.FC<Props> = ({ image, name, className }: Props) => {
+  return <Wrapper src={image} alt={name} className={className} onError={_loadFallbackTokenImage} />
```
Velenir: Really no need to create an extra Wrapper

Either do this
```
const TokenImg = styled.img.attrs(
  {name, image} => ({src:image, alt:name, onError:_loadFallbackTokenImage}
)`
  width: 30px;
  height: 30px;
`
```

Or this
```
export default styled(TokenImg)`...`
```
There are some differences in when className gets overridden between these approaches, but in this case it doesn't matter

src/components/TradeWidget/index.tsx
alfetopito: I want to have a single `TokenDetails` instance. Filter returns a new array. I'm using `reduce` to extract the single filtered match, if any.
I opted to not use `.filter(tokenDetails => tokenDetails.symbol == symbol).get(0)` for the case where there's no match and that causes an exception. Although I'm not handling when there's no match outside of the function...

Anyway, do you get the point? 
With this in mind, what would you suggest?

src/components/layout/Widget.tsx
alfetopito: Depending on the network, `EtherscanLink` component might be null: https://github.com/gnosis/dex-react/blob/develop/src/components/EtherscanLink.tsx#L31-L36

gnosis-info:
Travis automatic deployment:
 https://pr128--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
Velenir:
```js
tokenListApi.getTokens(_networkId).find(tokenDetails => tokenDetails.symbol == symbol)
```
does exactly what you described, without creating temp arrays

src/components/layout/Widget.tsx
Velenir: No, `EtherscanLink` may render to null, not be null. Component is always an object
```js
<EtherscanLink/> = React.createElement(EtherscanLink) = {type: EtherscanLink, ref, key, props: {}, ...metaStuff}

```

src/components/TokenImg.tsx
```diff
@@ -3,26 +3,12 @@ import styled from 'styled-components'
...
   const image = event.currentTarget
   image.src = unknownTokenImg
 }
 
-const TokenImg: React.FC<Props> = ({ image, name, className }: Props) => {
-  return <Wrapper src={image} alt={name} className={className} onError={_loadFallbackTokenImage} />
-}
-
-export default TokenImg
+export default styled.img.attrs(() => ({ onError: _loadFallbackTokenImage }))`
```
Velenir: If you opt not to depend on props in `attrs`, can just do an object instead of func
```
styled.img.attrs({ onError: _loadFallbackTokenImage })`...`
```

alfetopito: Well, that should work as stated in the docs https://www.styled-components.com/docs/api#attrs, but for some reason it doesn't.

When I use the object instead of function, the `image` instance is `undefined` on the fallback function:

```
    TypeError: Cannot set property 'src' of undefined

       6 | function _loadFallbackTokenImage(event: React.SyntheticEvent<HTMLImageElement>): void {
       7 |   const image = event.currentTarget
    >  8 |   image.src = unknownTokenImg
         |            ^
       9 | }
      10 | 
      11 | export default styled.img.attrs({ onError: _loadFallbackTokenImage })`
```
Same for unit tests and the browser

src/components/layout/Widget.tsx
alfetopito: Ah, so this is effectively the same as rendering the component inside the `Wrapper`. When `Etherscan` renders to `null` it simply won't show up:

```
    <Wrapper>
      <LinkWrapper type="contract" identifier={contractAddress} label={<small>View verified contract</small>} />
      {children}
    </Wrapper>
```

src/components/TradeWidget/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Widget>
+      <TokenRow tokenDetails={DAI} selectLabel="pay with" />
+      <IconWrapper>
+        <FontAwesomeIcon icon={faExchangeAlt} rotation={90} size="2x" />
+      </IconWrapper>
+      <TokenRow tokenDetails={USDC} selectLabel="receive" />
+      <SubmitButton>
+        <FontAwesomeIcon icon={faPaperPlane} size="lg" /> Send limit offer
```
Velenir: Should this be `limit order`?

src/components/TokenImg.tsx
Velenir: I experimented with this a little bit and found out
if a field of the attrs object is a function, then that property becomes a computed property, like
```
styled.img.attrs({ onError: (props) => _loadFallbackTokenImage })
// onError = _loadFallbackTokenImage

and with just
styled.img.attrs({ onError: _loadFallbackTokenImage })
// onError = _loadFallbackTokenImage(props)
```

Totally not obvious from the docs.

src/components/TradeWidget/index.tsx
alfetopito: Correct! Oops :)

src/components/TokenImg.tsx
alfetopito: Thanks for digging into this.
In that case, I'll stick to the current version.

alfetopito:
[Latest build is failing](https://travis-ci.org/gnosis/dex-react/jobs/601866378?utm_medium=notification&utm_source=github_status) to install the latest node version. I got the same results locally.
So far, it seems unrelated to the changes made in this PR.
Reaching out to the dev-ops people for a hand.

alfetopito:
> [Latest build is failing](https://travis-ci.org/gnosis/dex-react/jobs/601866378?utm_medium=notification&utm_source=github_status) to install the latest node version. I got the same results locally.
> So far, it seems unrelated to the changes made in this PR.
> Reaching out to the dev-ops people for a hand.

Temporary issue downloading latest Node (currently v13.0.1) with `nvm install stable`.

Worked after a retry.

gnosis-info:
Travis automatic deployment:
 https://pr128--dexreact.review.gnosisdev.com

PR #133, opened 2019-10-24
95/token selector

### LINKED ISSUE #95: Token Selector
Combo select + filter + selection

Depends on https://github.com/gnosis/dex-react/issues/94

**Estimation:** 2

alfetopito:
Closes #95 

Dependent on #128 / #100 

---

Implements token selector (RF.2.3 Token Selector) using [React Select](https://react-select.com/home).

- [x] List available tokens
- [x] Allow filtering by token symbol/name
- [x] I click swap between pay with/receive tokens
- [x] Do not allow same token to be selected on both
- [x] Renamed label from `PAY WITH` to `SELL`. This was the winner choice in the poll ran on team channel.

token-selector

Try it out on https://pr133--dexreact.review.gnosisdev.com

---

Not part of this PR:
- Add arbitrary tokens (out of the scope of POC, AFAIK)
- Remaining trade logic
  - Load balances
  - Submit order

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -14,9 +14,10 @@ const Wrapper = styled.div`
 `
 
 const TokenImgWrapper = styled(TokenImg)`
-  margin-right: 1em;
   width: 50px;
   height: 50px;
+
+  margin-right: 1em;
 `
 
 const SelectBox = styled.div`
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -1,9 +1,11 @@
-import React from 'react'
+import React, { CSSProperties } from 'react'
+import { Link } from 'react-router-dom'
 import styled from 'styled-components'
+import Select from 'react-select'
+import { FormatOptionLabelContext } from 'react-select/src/Select'
 
 import TokenImg from 'components/TokenImg'
 import { TokenDetails } from 'types'
-import { Link } from 'react-router-dom'
 
 const Wrapper = styled.div`
   display: flex;
@@ -18,13 +20,19 @@ const TokenImgWrapper = styled(TokenImg)`
 `
 
 const SelectBox = styled.div`
-  margin: 0 1em 0 1em;
   display: flex;
   flex-direction: column;
   align-items: stretch;
 
+  margin: 0em 1em;
+
   label {
     text-transform: uppercase;
+    padding-left: 8px; // to align with Select input padding
+  }
+
+  input {
+    margin-left: 0; // to fix extra space on Select search box
   }
 `
 
@@ -33,6 +41,7 @@ const InputBox = styled.div`
   flex-direction: column;
   align-items: stretch;
   flex-grow: 1;
+
   margin-left: 1em;
 
   input {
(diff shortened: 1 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr133--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -26,20 +30,50 @@ const SubmitButton = styled.button`
 
 const TradeWidget: React.FC = () => {
   const { networkId } = useWalletConnection()
-  const DAI = _getTokenDetails('DAI', networkId)
-  const USDC = _getTokenDetails('USDC', networkId)
+  const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
```
W3stside: should we fallback to Mainnet? Traditionally it's been local

src/components/TradeWidget/index.tsx
```diff
@@ -26,20 +30,50 @@ const SubmitButton = styled.button`
...
+  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
```
W3stside: I'm curious - wouldn't it be better to use `networkId` from the `useWalletConnection` effect instead? Because the `const` `fallBackNetworkId` won't change if `networkId` changes thus meaning `tokens` will never change

wdyt?

src/components/TradeWidget/index.tsx
alfetopito: My idea is to get the mainnet token addresses because for those we are more likely to have all the tokens with the proper images listed. 

Also, we currently have only Mainnet and Rinkeby networks https://github.com/gnosis/dex-react/blob/develop/src/api/index.ts#L29, no local so far.

Maybe I'm confusing the concepts of connecting to a local network vs loading token addresses from different networks. If connected to local, the node could be connected to either Mainnet or Rinkeby.
In that case, we have two sets of options:
1. Connect to a local node OR a web3provider such as metamask
2. Pick the network we'll operate on, such as Mainnet or Rinkeby

This fallback only addresses `2`.

Does that make sense?

src/components/TradeWidget/index.tsx
alfetopito: That's a very good point! I have not realized it won't change. 

Complementing what I said in the previous comment, my goal for having a fallback is to always have the trade widget fully rendered, even if the user haven't connected a wallet.
When no wallet is connected, the `networkId` is null, and thus no tokens will be rendered.

Currently in our mock implementation it doesn't matter because regardless of the `networkId`, we'll always return the same tokens, but that won't be the case for the non-mock scenario.

Anyway, thanks for catching that, I'll update it.

bh2smith:
Hey I notice that above the token selector is says "Pay With" and find it confusing. Is this supposed to be the "Sell Token" in the trade? It almost makes me think that I am choosing what to pay my fees with or something.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -33,6 +42,7 @@ const InputBox = styled.div`
   flex-direction: column;
   align-items: stretch;
   flex-grow: 1;
+
   margin-left: 1em;
 
   input {
```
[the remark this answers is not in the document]
alfetopito: Done

alfetopito:
> Hey I notice that above the token selector is says "Pay With" and find it confusing. Is this supposed to be the "Sell Token" in the trade? It almost makes me think that I am choosing what to pay my fees with or something.

I'm looking at the [requirements doc](https://docs.google.com/document/d/1lW3SQV5FAI-gDrZZAD3REunrx4TSa9nNjxHljpCFOd4/edit#heading=h.o1r25ca7o3uq) and I don't see any point mentioning a rationale for this.

I'll do a poll in the team channel to gather more opinions.

src/components/TradeWidget/index.tsx
Velenir: Also, don't memoize primitive data types, it's useless

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -48,22 +59,85 @@ const WalletDetail = styled.div`
...
+  token: TokenDetails
+  tokens: TokenDetails[]
   selectLabel: string
+  onSelectChange: (selected: TokenDetails) => void
 }
 
-const TokenRow: React.FC<Props> = ({ selectLabel, tokenDetails: { name, image } }: Props) => {
-  const selectId = selectLabel.replace(/[^a-zA-Z]/g, '')
+const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange }: Props) => {
+  const options = tokens.map(token => ({ token, value: token.symbol, label: token.name }))
```
Velenir: Memoize this on `tokens`

src/components/TradeWidget/index.tsx
```diff
@@ -26,20 +30,51 @@ const SubmitButton = styled.button`
 
 const TradeWidget: React.FC = () => {
   const { networkId } = useWalletConnection()
-  const DAI = _getTokenDetails('DAI', networkId)
-  const USDC = _getTokenDetails('USDC', networkId)
+  // Avoid displaying an empty list of tokens when the wallet is not connected
+  const fallBackNetworkId = useMemo(() => (networkId ? networkId : Network.Mainnet), [networkId])
+  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
+  const [payWithToken, setPayWithToken] = useState(_getToken('DAI', tokens))
+  const [receiveToken, setReceiveToken] = useState(_getToken('USDC', tokens))
```
Velenir: Do `useState(() => _getToken('USDC', tokens))`. No need to run on each render

src/components/TradeWidget/index.tsx
alfetopito: Reverted, since this is a primitive data type.

src/components/TradeWidget/index.tsx
anxolin: Hi! 

I understand this faultback is only relevant while the wallet is not connected and we don't know the network yet.

In this case options are:
- *Do not show anything*: I think it's undesirable in this case, cause we want to allow the user to play with the trade page although he is not connected
- *Show a default network*: This is what I understand Leandro did and I agree that mainnet can be a good default

However, I think we can do something else. It can be address in a different PR (if you agree we can add it to our backlog). 
The app will be deployed in a server, the URL for rinkeby will be different as the URL for mainnet, i.e:
* `batch.trade`
* `rinkeby.batch.trade`
I think we should know if we are targeting mainnet or rinkeby, and use it as our default. One way to do this is setup a constant on app initialization using the url of the app:
* `batch.trade` --> mainnet
* `rinkeby.batch.trade` ---> rinkeby
* null --> in this case, we can decide in the component what to do. 

WDYT?

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -48,22 +59,85 @@ const WalletDetail = styled.div`
   }
 `
 
+function renderOptionLabel(token: TokenDetails): React.ReactNode {
+  return (
+    <div
+      style={{
+        display: 'flex',
+        alignItems: 'center',
```
anxolin: Shouldn't we avoid to inline styles like this?

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -48,22 +59,85 @@ const WalletDetail = styled.div`
...
     <Wrapper>
-      <TokenImgWrapper alt={name} src={image} />
+      <TokenImgWrapper alt={token.name} src={token.image} />
       <SelectBox>
-        <label htmlFor={selectId}>{selectLabel}</label>
-        <select name="tokenSelector" id={selectId}>
-          <option value="DAI">DAI</option>
-        </select>
+        <label>{selectLabel}</label>
+        <Select
```
anxolin: I think this Select will be used in more parts of the App. It wouldn't be a bad idea to create a component out of it

src/components/TradeWidget/index.tsx
alfetopito: I like the idea of the constant in the initialization, given the two different deployments. In that case though, the logic for detecting which is the default network should not be in the component, but instead in whatever provides the `networkId`, and this value should never be `undefined`.

Other point somewhat related: what should happen in case the user is using our Rinkeby instance and (using Metamask for example) switches the network to Mainnet? Should that be allowed? How would that look like from a legal stand point?

src/components/TradeWidget/TokenRow.tsx
alfetopito: Yeah I did not have a good feeling about this. I'll refactor the styles in a new PR.

src/components/TradeWidget/TokenRow.tsx
alfetopito: Good idea, I can have a simple PR just for that.

alfetopito: Will address on #148

src/components/TradeWidget/TokenRow.tsx
alfetopito: Will address on #147

PR #134, opened 2019-10-25
swap inline styles for styled components

W3stside:
removed inline for styled components, small pr


gnosis-info:
Travis automatic deployment:
 https://pr134--dexreact.review.gnosisdev.com

PR #140, opened 2019-10-29
Real ERC20 api

Velenir:
very much depends on #136 

`Row` component doesn't always reflect the changes, because it's very uncontrolled.


gnosis-info:
Travis automatic deployment:
 https://pr140--dexreact.review.gnosisdev.com

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -0,0 +1,123 @@
...
+export class Erc20ApiImpl implements Erc20Api {
+  private _ReferenceERC20: ERC20
+
+  private static _address2ERC20cache: { [K: string]: ERC20 } = {}
+
+  public constructor(web3: Web3) {
+    this._ReferenceERC20 = new web3.eth.Contract(Erc20ABI)
+
+    // TODO remove later
+    ;(window as any).erc20 = this._ReferenceERC20
```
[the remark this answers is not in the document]
W3stside: TS `()` typing on the window messes with linter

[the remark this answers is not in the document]
Velenir: yep, either `prettier` or `eslint --fix` moved it automatically

src/api/index.ts
```diff
@@ -58,8 +60,13 @@ function createDepositApi(erc20Api: Erc20Api): DepositApi {
   return depositApi
 }
 
+// TODO connect to mainnet if we need AUTOCONNECT at all
+const InfuraEndpoint = 'rinkeby.infura.io/v3/8b4d9b4306294d2e92e0775ff1075066'
```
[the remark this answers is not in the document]
Velenir: This should be a `process.env` variable from a non-committed `.env` file eventually
Let it stay for now

src/api/wallet/WalletApiImpl.ts
```diff
@@ -142,8 +143,9 @@ export class WalletApiImpl implements WalletApi {
     this._provider = provider
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this._web3 = new Web3(provider as any)
+    this._web3.setProvider(provider as any)
     log('[WalletApi] Connected')
+    ;(window as any).web3 = this._web3
```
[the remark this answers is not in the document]
Velenir: For testing purposes to have it available in the console
But you are right should rename it to something else not to override `window.web3`, `web3c` for Custom

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -0,0 +1,123 @@
...
+    tokenAddress: string,
+    fromAddress: string,
+    toAddress: string,
+    amount: BN,
+    txOptionalParams?: TxOptionalParams,
+  ): Promise<Receipt> {
+    const erc20 = this._getERC20AtAddress(tokenAddress)
+
+    const tx = erc20.methods.transferFrom(senderAddress, toAddress, amount).send({
+      from: fromAddress,
```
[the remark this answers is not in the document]
Velenir: But `fromAddress` in `transfer` is the sender address. Unless the sender is from address. And actually, it usually is the same address. Ugh, now I'm confused.

@W3stside @alfetopito @anxolin 
```
erc20.methods.transferFrom(senderAddress, toAddress, amount).send({
      from: fromAddress,
})
// or

erc20.methods.transferFrom(fromAddress, toAddress, amount).send({
      from: senderAddress,
})
```

gnosis-info:
Travis automatic deployment:
 https://pr140--dexreact.review.gnosisdev.com

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /********************************    private methods   ********************************/
+
+  private _getERC20AtAddress(address: string): ERC20 {
+    const erc20 = Erc20ApiImpl._address2ERC20cache[address]
+
+    if (erc20) return erc20
+
+    const newERC20 = this._ReferenceERC20.clone()
```
anxolin: It more natural to not keep a reference implementation and in this line do a:
```js
const erc20Contract = new web3.eth.Contract(Erc20ABI, address)
this._contractsCache[address] = erc20Contract
```

nit: Regarding `_address2ERC20cache`. In the app `ERC20` when used in camel case, we use it like `erc20`, or `whateverErc20`. Anyways, `_contractsCache` seems like a simpler name

src/api/exchange/DepositApiMock.ts
```diff
@@ -107,15 +107,15 @@ export class DepositApiMock implements DepositApi {
...
+    return RECEIPT
   }
 
   public async requestWithdraw(
     userAddress: string,
     tokenAddress: string,
     amount: BN,
     txOptionalParams?: TxOptionalParams,
-  ): Promise<TxResult<void>> {
+  ): Promise<Receipt> {
```
anxolin: This change is good, BUT it's not part of this PR. I think in this situations is more clear to create on PR with this change and the refactor of all the app. Then this #140 would use that PR's branch as it's base

Please, keep this in mind so PRs are easier to review.

src/api/index.ts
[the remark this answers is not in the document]
anxolin: Why web3 construction uses rinkeby infura http connection string?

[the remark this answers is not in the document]
Velenir: You are looking at an old commit

Velenir:
> It has an issue when you switch accounts.
> It loads the right address in the wallet, BUT the balances in the DepositWidget don't update.
> 
> The PR is bigger than it should, please try to keep them smaller. You don't need to tackle all the things in a single PR. In this we can leave it like this

I did write in the description
> Row component doesn't always reflect the changes, because it's very uncontrolled.
DepositWidget updates all right, Rows don't

src/api/erc20/Erc20ApiImpl.ts
Velenir: `clone` retains defaults (which granted we don't use here), but also skips unnecessary stuff like abi decoding, etc. So it's a recommended way
Yes, it would be ideal if it accepted a new address (`clone(address)`), but it doesn't

Velenir: Will change to `_contractsCache`

src/api/erc20/Erc20ApiImpl.ts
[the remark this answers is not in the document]
alfetopito: That depends on which pattern we'll stick to.

In my view, the later makes more sense because it reads in English:
 > Transfer from `address` to `address` the `amount`, sent by `address`.
The confusion is because the parameter on `send` is called `from`.

Also, the address the amount will be sent from is not necessarily the address who will be sending the transaction.

Another suggestion, to remove the ambiguity:

```
erc20.methods.transferFrom(sourceAddress, destinationAddress, amount).send({
      from: senderAddress,
})
```

src/api/erc20/Erc20ApiImpl.ts
anxolin: > clone retains defaults (which granted we don't use here), but also skips unnecessary stuff like abi decoding

Cool, didn't know!

src/api/exchange/ExchangeApiMock.ts
```diff
@@ -91,24 +88,23 @@ export class ExchangeApiMock extends DepositApiMock implements ExchangeApi {
       priceDenominator: orderParams.sellAmount,
       remainingAmount: orderParams.sellAmount,
     })
-    const orderId = this.orders[orderParams.userAddress].length - 1
 
-    return { data: orderId, receipt: RECEIPT }
+    return RECEIPT
```
alfetopito: Well, the returned data is actually relevant I believe: https://github.com/gnosis/dex-contracts/blob/eb058c006e5ba929382c01851a25f3307e38c893/contracts/StablecoinConverter.sol#L179-L187

Or we'll just ignore it on the frontend?

src/api/index.ts
```diff
@@ -72,9 +74,13 @@ function createExchangeApi(erc20Api: Erc20Api): ExchangeApi {
   }
 }
 
+// TODO connect to mainnet if we need AUTOCONNECT at all
+const InfuraEndpoint = 'rinkeby.infura.io/ws/v3/8b4d9b4306294d2e92e0775ff1075066'
```
alfetopito: Config?

src/types/index.ts
```diff
@@ -171,5 +157,66 @@ export interface Erc20Api {
...
+
+  transferFrom(
+    senderAddress: string,
+    tokenAddress: string,
+    fromAddress: string,
+    toAddress: string,
+    amount: BN,
+    txOptionalParams?: TxOptionalParams,
+  ): Promise<Receipt>
+}
```
alfetopito: Duplicate `transferFrom`?

src/api/erc20/Erc20ABI.ts
```diff
@@ -0,0 +1,108 @@
...
+    inputs: [],
+    name: 'name',
+    outputs: [{ name: '', type: 'string' }],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function',
+  },
+  {
+    constant: false,
+    inputs: [{ name: 'guy', type: 'address' }, { name: 'wad', type: 'uint256' }],
```
W3stside: what is `name: guy`?

src/api/exchange/ExchangeApiMock.ts
W3stside: You don't get back `return` values from `solidity` contracts unless it's a `call` no?

src/hooks/useTokenBalances.ts
```diff
@@ -59,8 +59,15 @@ async function _getBalances(walletInfo: WalletInfo): Promise<TokenBalanceDetails
...
-  return Promise.all(balancePromises)
+  const balancePromises: Promise<TokenBalanceDetails | null>[] = tokens.map(async token =>
+    fetchBalancesForToken(token, userAddress, contractAddress).catch(e => {
+      log('Error for', token, userAddress, contractAddress)
+      log(e)
+      return null
+    }),
+  )
+  const balances = await Promise.all(balancePromises)
+  return balances.filter(Boolean)
```
W3stside: what does Boolean do here? just check if it's a boolean response and filter those out?

src/api/exchange/ExchangeApiMock.ts
Velenir: We can't get it on the frontend. 
When a contract calls a contract it does get the proper returned value.
When JS or anyone else but solidity calls a contract producing a change, that is transaction, what we get back is a transaction receipt.
If we need to then check the state after that transaction we either must make a call (not tx) to the contract or parse the tx receipt for events if there are any relevant to what data we need.

src/hooks/useTokenBalances.ts
Velenir: filters out falsy values
Boolean is equivalent to `val => !!val`

src/api/index.ts
Velenir: Config?
Config!

src/api/erc20/Erc20ABI.ts
Velenir: :rofl: 
That was in the ERC20 abi I grabbed
I'll switch to open-zeppeling one with better names

gnosis-info:
Travis automatic deployment:
 https://pr140--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr140--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApiMock.ts
alfetopito: Understood. Never mind then :)

src/api/index.ts
```diff
@@ -72,9 +75,12 @@ function createExchangeApi(erc20Api: Erc20Api): ExchangeApi {
   }
 }
 
+// TODO connect to mainnet if we need AUTOCONNECT at all
+const web3 = new Web3(process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT)
```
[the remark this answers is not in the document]
Velenir: We decide that we need to show something even before user connected any wallet or unlocked MMask, or doesn't have MMask. Therefore infura, which we use anyway for WalletConnect (internally creates HttpConnection).

Do we no longer want any non-wallet connection?

Velenir:
> > I specifically don't tackle all the things, as in, I didn't fix the `Row` component. I only tackle stuff away from the main focus when I want travis build to pass. Like changing types or similar.
> 
> I don't understand. The account switching functionality was working, at least for the mock. Every time you change wallet, the balances are reloaded. Not sure if you mean that it's broken now, or it has a design problem
> 
> It's there an issue for the problem you say that the Row has so I can understand better?

Was broken at some point. After some PRs were merged no longer broken.

src/api/exchange/DepositApiMock.ts
Velenir:
```
real_wallet PR 
                            \
                              \
                                   this PR
                             /
                            /
type change PR
```
I chose to avoid diamond problem with inheritance. also avoid a mid-step in the waterfall

src/api/index.ts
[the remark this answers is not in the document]
anxolin: Okay, I understand. This is for showing the tokens. And what happen if the user connects, does this API use different web3 object, or you use infura as a fallback?

src/api/exchange/DepositApiMock.ts
anxolin: I think I would have done a waterfall here.
`real_wallet PR` --> `type change PR` --> `this PR`

or if you prefer:
`type change PR` --> `real_wallet PR`  --> `this PR`

We are all trying to do this to make PRs easier to review for others. 
For this one is fine now, wasn't that bad, but AGAIN, we need to keep the scope of a PR into one single thing please for next PRs. Don't need to take it to an extreme, but just something that makes sense.

Now we are more in the project, so difficulting the review has a multiplying effect. Even for the one making the PR. And I mean not just people in front end, I think it's important for people working in other parts of dfusion.
 
Thanks for understanding!

Velenir: Sure, shoudl have gone with either of those. Hindsight is 20/20 :stuck_out_tongue_closed_eyes:

src/api/index.ts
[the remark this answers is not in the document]
Velenir: [web3.setProvider](https://github.com/gnosis/dex-react/pull/140/files#diff-82aa698065493d8d6c07cc1c693f1407R160)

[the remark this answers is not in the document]
anxolin: Okay, that's fine, but then I think something is wrong when you disconnect in the current version:

Try to follow this steps:
1. Connect
2. Go to the deposits
3. Disconnect
4. Try to go back to the trade page (it doesn't go in my case)

src/api/exchange/DepositApiMock.ts
anxolin: Also, it would have been good to invest in bitcoin in 2009

alfetopito: > Also, it would have been good to invest in bitcoin in 2009

The list of emoji reactions don't include a :cry: nor a :sob:

PR #143, opened 2019-10-30
139/fix responsive design

### LINKED ISSUE #139: Fix responsive design
Some changes need to be made, notable:

1. table on iphone view
2. potential size of content in certain rows

**Estimation**: 0.5

W3stside:
Fixes #139 responsive design problems as discussed in slack

1. vertical stretch in mobile
2. switch from HTML table to CSS Grid
3. minor cosmetic changes

TODO:
*I still need to remove some test code before approval so draft for now*

- [x] Fix Row when clicking on Withdraw/Deposit for new grid layout


gnosis-info:
Travis automatic deployment:
 https://pr143--dexreact.review.gnosisdev.com

anxolin:
Haven't reviewed the CSS, but I saw the demo and it's a great improvement!
I like the trade widget taking the 100% of the horizontal space.
I also really like the the layotout for the deposit widget, how is like a table for desktop and cards for mobile!

My only suggestion would be to avoid using so many colors in the desktop table, hover and colouring the odd rows overloads the layout a bit in my opinion

image

I prefer the old version cause it's more discrete. The highlighted items are buttons so it's easier to understand:
image

The responsive design I find it ok with the two colors, since they are cards. I would just soften a bit the color making it more light.

image

src/components/DepositWidget/Row.tsx
```diff
@@ -31,6 +67,64 @@ const TokenTr = styled.tr`
 
   &.selected {
     background-color: #ecdcff;
+  }
+
+  // &:nth-child(odd) {
+  //   background: #8080801c;
+  // }
```
alfetopito: Remove?

src/components/DepositWidget/grid.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  useEffect(() => {
+    window.addEventListener('resize', handleResize)
+
+    return (): void => window.removeEventListener('resize', handleResize)
+  }, [])
+
+  return windowWidth
+} */
```
alfetopito: Remove?

src/pages/Deposit.tsx
```diff
@@ -1 +1 @@
-export { default } from 'components/DepositWidget'
+export { default } from 'components/DepositWidget/grid'
```
alfetopito: I'm curious as to why create a new grid component replacing the default index but still keeping the index file.

If intending to keep both for loading a different page in different situations, I would suggest to create a new `components/DepositWidget/table` for the existing component and export both from `components/DepositWidget/index`.
Or, if we are gonna have only `grid`, put that instead on `index`

W3stside: in the main description i wrote:
```
TODO:
I still need to remove some test code before approval so draft for now
```
which was referencing my need to replace `index` with `grid` should the style be accepted

alfetopito: Oops, guess I should pay closer attention to the PR description =P

anxolin:
I tried in TrustWallet on my iPhone, and it works mostly well!
I found some small issues that I will describe next.

(1)
I'm not super convinced of how it looks the withdraw/deposit form. In the table format it's clear that is a form belonging to the stable coin above, but in the phone it just looks like two cards one after each other. This is not a huge deal for me right now, but does anyone have an idea on how to better present it in mobile?
IMG_2146

(2)
In the same picture you see a issue with the "Use max" link.

(3)
The header is two big. It takes all the space of the window. This makes that when we navigate nothing changes, and also doesn't use the space very efficiently. Additionally, the navigation menu  seems lost in the middle of the header.
IMG_2147


(4)
This one is super small, not sure why it happens, but when I click on the input of the trade widget, it zooms and then the app looks like this (even after I lose the focus):

Before focus:
IMG_2145

After focus:
IMG_2144

alfetopito:
When the screen height is smaller than the content, something breaks the layout, making the widget overlap with the header:

Overlap

Also, I observed this awhile ago and it's still present. When the screen heigh is greater than the content size, the widget is no longer above the header:

no-overlap

Would you mind fixing that in this PR?

gnosis-info:
Travis automatic deployment:
 https://pr143--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr143--dexreact.review.gnosisdev.com

src/hooks/useWindowSpecs.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react'
+
+const useWindowSpecs = (): { innerWidth: number; innerHeight: number } => {
+  const [innerWindowSpecs, setInnerWindowSpecs] = useState({
+    innerWidth: window.innerWidth,
+    innerHeight: window.innerHeight,
+  })
+  function handleResize(): void | number {
+    // check that client can handle this sexiness
```
alfetopito: :D

src/components/DepositWidget/Row.tsx
```diff
@@ -75,6 +162,20 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
   const [visibleForm, showForm] = useState<'deposit' | 'withdraw' | void>()
   const exchangeBalanceTotal = exchangeBalance.add(depositingBalance)
 
+  // Checks innerWidth
+  let showResponsive = innerWidth < 500
```
alfetopito: I understand `innerWidth` is coming from `windowSpecs` spread onto `Row` props, but I don't see it unpacked here.
Where's `innerWidth` defined?

W3stside: good catch, damn. just learned that `innerWidth` is in the global scope of Node... from `window.innerWidth` i guess. maybe i should change the name to avoid issues like this. tho it does work properly the way it was i changed it to explicit from props

src/hooks/useWindowSpecs.ts
W3stside: lmao. forgot about this

W3stside: lmao forgot about this

src/components/DepositWidget/Form.tsx
```diff
@@ -7,31 +7,57 @@ import { TokenBalanceDetails } from 'types'
...
 import BN from 'bn.js'
 
-const Wrapper = styled.tr`
+const DynamicWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &.fixedContainer {
+    position: fixed;
```
Velenir: Since we are using styled-components:
```js
${props => props.responsive && `
    position: fixed;
    top: 0;
    left: 0;
    bottom: 0;
    right: 0;
    background: #000000b5;
`
}
```
without classes

src/components/DepositWidget/Row.tsx
```diff
@@ -75,6 +163,20 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
...
+
+  useMemo((): void => {
+    const bodyClassList: string[] | DOMTokenList = (window && window.document.body.classList) || []
+    const noScrollActive = Array.from(bodyClassList).find((className: string): boolean => className === 'noScroll')
```
Velenir: This is a bit wrong. If there's a window, there's a document and a body and classList is always available even if empty.
If there's no window, then `window` will throw a `RefrenceError` and the right check is `typeof window !=== 'undefined`

Consider instead
```ts
const noScrollActive = document.body.classList.contains('noScroll')
```
with window check if you want

src/hooks/useWindowSpecs.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react'
+
+const useWindowSpecs = (): { innerWidth: number; innerHeight: number } => {
+  const [innerWindowSpecs, setInnerWindowSpecs] = useState({
+    innerWidth: window.innerWidth,
+    innerHeight: window.innerHeight,
```
Velenir: Accessing, `window.innerWidth` on each render causes a [reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#window), which is unnecessary evil.
Do like [this](https://usehooks.com/useWindowSize/)

src/components/DepositWidget/Row.tsx
```diff
@@ -87,15 +189,25 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
   const isDepositFormVisible = visibleForm == 'deposit'
   const isWithdrawFormVisible = visibleForm == 'withdraw'
 
+  const _showHideForm = (type?: 'deposit' | 'withdraw' | null): void => {
+    if (window && window.document && innerWidth < 500) {
+      type ? window.document.body.classList.add('noScroll') : window.document.body.classList.remove('noScroll')
```
Velenir:
```js
window.document.body.classList.toggle('noScroll', type)
```

src/components/DepositWidget/Row.tsx
```diff
@@ -75,6 +163,20 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
   const [visibleForm, showForm] = useState<'deposit' | 'withdraw' | void>()
   const exchangeBalanceTotal = exchangeBalance.add(depositingBalance)
 
+  // Checks innerWidth
+  let showResponsive = innerWidth < 500
```
Velenir: Same concern as way way below, reflow happens here.
Use your hook.

src/components/DepositWidget/Row.tsx
```diff
@@ -75,6 +163,20 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
...
+
+  useMemo((): void => {
+    const bodyClassList: string[] | DOMTokenList = (window && window.document.body.classList) || []
+    const noScrollActive = Array.from(bodyClassList).find((className: string): boolean => className === 'noScroll')
+    // if innerWidth > 500 && body has noScroll active - REMOVE it
+    if (noScrollActive && !visibleForm && !showResponsive) {
+      ;(bodyClassList as DOMTokenList).remove('noScroll')
```
anxolin: How is modifying the `classList` of the body responsibility of a `DepositWidget row` component?

W3stside: accessing `visibleForm` is a bit difficult elsewhere

W3stside: normally I'd do this via a `middleware` but not sure how here with our structure of data and setters being rather local

src/components/DepositWidget/Row.tsx
W3stside: even tho innerWidth here is a prop? (it's not from window.innerWidth)

src/components/DepositWidget/Row.tsx
W3stside: made into a hook

gnosis-info:
Travis automatic deployment:
 https://pr143--dexreact.review.gnosisdev.com

src/components/DepositWidget/Row.tsx
```diff
@@ -75,6 +46,10 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
   const [visibleForm, showForm] = useState<'deposit' | 'withdraw' | void>()
   const exchangeBalanceTotal = exchangeBalance.add(depositingBalance)
 
+  // Checks innerWidth
+  let showResponsive = innerWidth < 500
```
alfetopito: Can you move the `500` to the constants file?

src/hooks/useWindowSpecs.ts
```diff
@@ -0,0 +1,27 @@
...
+      innerWidth: isClient ? window.innerWidth : undefined,
+      innerHeight: isClient ? window.innerHeight : undefined,
+    }
+  }
+
+  const [innerWindowSpecs, setInnerWindowSpecs] = useState(getSize)
+  function handleResize(): void | number {
+    // check that client can handle this sexiness
+    const innerWidth = typeof window !== 'object' ? null : window.innerWidth
+    const innerHeight = typeof window !== 'object' ? null : window.innerHeight
```
alfetopito: Why not use `getSize()` here? AFAICT, lines 16,17 are just a duplicate of 8,7.

W3stside: good call, forgot about optimising it, thanks

Issue #101, opened 2019-10-01
Trading page / widget: validation


alfetopito:
Validation + form lib

**Estimation:** 2

PR #163, opened 2019-10-31
Extracting utils function getTokens

alfetopito:
Closes #162 

---

Small refactor to generalize a helper function and move it to utils.

Added unit tests.

## Testing
- [x] `yarn test`
- [x] `yarn build`

PR #164, opened 2019-10-31
101/populate balances

alfetopito:
Part of #101 

---

Populating balances from user/contract

##Testing
- [x] `yarn test && yarn build`
- [x] visually checked

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -5,7 +5,8 @@ import Select from 'react-select'
 import { FormatOptionLabelContext } from 'react-select/src/Select'
 
 import TokenImg from 'components/TokenImg'
-import { TokenDetails } from 'types'
+import { TokenDetails, TokenBalanceDetails } from 'types'
+import { formatAmount } from 'utils'
 
 const Wrapper = styled.div`
   display: flex;
@@ -108,11 +109,12 @@ const customSelectStyles = {
 interface Props {
   token: TokenDetails
   tokens: TokenDetails[]
+  balance: TokenBalanceDetails
   selectLabel: string
   onSelectChange: (selected: TokenDetails) => void
 }
 
-const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange }: Props) => {
+const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange, balance }) => {
   const options = useMemo(() => tokens.map(token => ({ token, value: token.symbol, label: token.name })), [tokens])
 
   return (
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr164--dexreact.review.gnosisdev.com

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -145,10 +147,10 @@ const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange
           <strong>
             <Link to="/deposit">Exchange wallet:</Link>
           </strong>{' '}
-          <span className="success">312312.33</span>
+          <span className="success">{balance ? formatAmount(balance.exchangeBalance, balance.decimals) : 0}</span>
```
W3stside: is `balance` here always an object? if its empty (but still `{}`) it will always be truthy. Maybe better to check for `balance.exchangeBalance`

src/components/TradeWidget/index.tsx
```diff
@@ -8,9 +8,10 @@ import { tokenListApi } from 'api'
 import TokenRow from './TokenRow'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faExchangeAlt, faPaperPlane } from '@fortawesome/free-solid-svg-icons'
+import { useTokenBalances } from 'hooks/useTokenBalances'
 
-function _getToken(symbol: string, tokens: TokenDetails[]): TokenDetails {
-  return tokens.find(token => token.symbol == symbol)
+function _getToken<T extends TokenDetails>(symbol: string, tokens: T[]): T {
```
W3stside: I would just do:

`function _getToken<T extends TokenDetails>(symbol: string, tokens: T[] = []): T { ... }` (default param)`

instead of the `(token || []).find(...)`

alfetopito: Refactored in https://github.com/gnosis/dex-react/pull/163, I'll address this there.

src/components/TradeWidget/TokenRow.tsx
alfetopito: Actually, `formatAmount` returns `null` if amount is null.
And when the wallet is disconnected the `balance` is `undefined`.

Extracted into a new function, keeping the check for `undefined` and added a check for `null` return from `formatAmount`

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -110,14 +111,22 @@ const customSelectStyles = {
   valueContainer: (provided: CSSProperties): CSSProperties => ({ ...provided, minWidth: '4.5em' }),
 }
 
+function displayBalance(balance: TokenBalanceDetails | undefined, key: string): string {
+  if (balance === undefined) {
```
anxolin: nit: I think we don't have null checks in typescript yet, so I think we could get an error if balance is null (`null[key]`). I think it's fine and shorter just testing for falsy values:

```js
if (!balance) {
  return '0' 
}
```

src/components/TradeWidget/index.tsx
```diff
@@ -35,6 +36,9 @@ const TradeWidget: React.FC = () => {
   const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
   const [sellToken, setSellToken] = useState(() => _getToken('DAI', tokens))
   const [receiveToken, setReceiveToken] = useState(() => _getToken('USDC', tokens))
+  const { balances } = useTokenBalances()
+  const sellTokenBalance = useMemo(() => _getToken(sellToken.symbol, balances), [balances, sellToken.symbol])
```
anxolin: You've made another PR with a refactor for `getToken`, I guess we have to write down to adapt this fn too once it's merged

alfetopito: Will do :+1:

PR #161, opened 2019-10-31
101/minor adjustments

alfetopito:
Part of #101 

----

- [x] Adjust some stylings
- [x] Removed unnecessary `<div>`

## Testing

- [x] `yarn test`
- [x] `yarn build`
- [x] `visually verified

PR #165, opened 2019-10-31
101/trading widget form

### LINKED ISSUE #101: Trading page / widget: validation

alfetopito:
Closes #101 

Contains changes from #161 and #164. 
Recommend reviewing them first.
Once they are merged there will be less code to review.

---

### Note:
Styling is very basic. I welcome suggestions to how better display the error messages:

---

- [x] Using `react-hook-form`
- [x] Making total exchange amount clickable and filling the input with its value
- [x] Validating a float number > 0 is provided
- [x] Displaying warning when sell token amount is > exchange wallet balance for given token
- [x] Disabling the submit button when form is not valid

token-selector-validation

## Testing

- [x] `yarn test && yarn build`
- [x] visually tested in the browser

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -1,8 +1,9 @@
-import React, { CSSProperties, useMemo } from 'react'
+import React, { CSSProperties, useMemo, useRef, useEffect } from 'react'
 import { Link } from 'react-router-dom'
 import styled from 'styled-components'
 import Select from 'react-select'
 import { FormatOptionLabelContext } from 'react-select/src/Select'
+import { useFormContext } from 'react-hook-form'
 
 import TokenImg from 'components/TokenImg'
 import { TokenDetails, TokenBalanceDetails } from 'types'
@@ -113,16 +114,46 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
   return formatAmount(balance[key], balance.decimals) || '0'
 }
 
+function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
+  return formatAmount((balance || {}).exchangeBalance, token.decimals, 4, false) || '0'
+}
+
+// TODO: move into a validators file?
+function validatePositive(value: string): true | string {
+  return Number(value) > 0 || 'Invalid amount'
+}
+
+function validateMax(value: string, validateMaxAmount: boolean, max: string): true | string {
+  return validateMaxAmount ? Number(value) <= Number(max) || 'Insufficient funds' : true
+}
+
 interface Props {
   token: TokenDetails
   tokens: TokenDetails[]
   balance: TokenBalanceDetails
   selectLabel: string
   onSelectChange: (selected: TokenDetails) => void
+  inputId: string
+  validateMaxAmount?: true
 }
 
-const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange, balance }) => {
+const TokenRow: React.FC<Props> = ({
+  token,
+  tokens,
+  selectLabel,
+  onSelectChange,
+  balance,
+  inputId,
+  validateMaxAmount,
+}) => {
   const options = useMemo(() => tokens.map(token => ({ token, value: token.symbol, label: token.name })), [tokens])
+
+  const { register, errors } = useFormContext()
+
+  const maxRef = useRef(getMax(balance, token))
+  useEffect(() => {
+    maxRef.current = getMax(balance, token)
+  }, [balance, token])
 
   return (
     <Wrapper>
@@ -144,7 +175,19 @@ const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange,
         />
       </SelectBox>
       <InputBox>
-        <input type="text" placeholder="0" />
+        <input
+          className={errors[inputId] && 'error'}
+          placeholder="0"
+          name={inputId}
+          type="text"
+          required
+          ref={register({
+            validate: {
+              positive: (value: string): true | string => validatePositive(value),
+              max: (value: string): true | string => validateMax(value, validateMaxAmount, maxRef.current),
+            },
+          })}
+        />
         <WalletDetail>
           <strong>
             <Link to="/deposit">Exchange wallet:</Link>
@@ -154,6 +197,7 @@ const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange,
         <WalletDetail>
           <strong>Wallet:</strong> {displayBalance(balance, 'walletBalance')}
         </WalletDetail>
+        {errors[inputId] && <WalletDetail className="error">{errors[inputId].message}</WalletDetail>}
       </InputBox>
     </Wrapper>
   )
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -72,8 +72,9 @@ const WalletDetail = styled.div`
 
   &.error,
   &.warning {
-    text-align: right;
+    margin: 0 0 1em 0;
   }
+
   &.error {
     color: red;
   }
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr165--dexreact.review.gnosisdev.com

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -110,16 +128,55 @@ const customSelectStyles = {
   valueContainer: (provided: CSSProperties): CSSProperties => ({ ...provided, minWidth: '4.5em' }),
 }
 
+function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
```
W3stside: I would use a default param for `balance` here instead of using an `||` check in the function body.

e.g `function getMax(balance: TokenBalanceDetails | null = {}, token: TokenDetails): string {`

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -140,15 +197,29 @@ const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange
...
+          })}
+        />
+        {errorOrWarning}
         <WalletDetail>
           <strong>
             <Link to="/deposit">Exchange wallet:</Link>
           </strong>{' '}
-          <span className="success">312312.33</span>
+          <a className="success" onClick={(): void => setValue(inputId, max.toFixed(4))}>
+            {formatAmount((balance || {}).exchangeBalance, token.decimals) || 0}
```
W3stside: Should clicking the exchange wallet amount really fill max? seems a bit dangerous imo

src/components/TradeWidget/index.tsx
```diff
@@ -1,25 +1,32 @@
...
+import { useTokenBalances } from 'hooks/useTokenBalances'
+import { useWalletConnection } from 'hooks/useWalletConnection'
 import TokenRow from './TokenRow'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faExchangeAlt, faPaperPlane } from '@fortawesome/free-solid-svg-icons'
+import Widget from 'components/layout/Widget'
 
-function _getToken(symbol: string, tokens: TokenDetails[]): TokenDetails {
-  return tokens.find(token => token.symbol == symbol)
+function _getToken<T extends TokenDetails>(symbol: string, tokens: T[]): T {
```
W3stside: same as above - may be better to use default param for `tokens`

W3stside:
Last comment: I would recommend disabling the TradeWidget submission button if the form is invalid. Better UX imo

src/components/TradeWidget/TokenRow.tsx
alfetopito: Not sure. In my view it's convenient but no strong opinion here. I'm fine not having it either.
This is not part of the requirements, but since we do have a way to fill in the max when depositing or withdrawing, I thought would be useful to have it here as well.

@anxolin what do you say?

src/components/TradeWidget/TokenRow.tsx
alfetopito: Hmm, `balance: TokenBalanceDetails | null = {}` results in:

```Type '{}' is missing the following properties from type 'TokenBalanceDetails': exchangeBalance, depositingBalance, withdrawingBalance, claimable, and 6 more.ts(2740)```

Couldn't yet find a way to make TS compiler happy about that.

Suggestions?

File: src/components/TradeWidget/index.tsx (hunk 3 above)
Comment (alfetopito) [reply src/components/TradeWidget/index.tsx]: Addressed on https://github.com/gnosis/dex-react/pull/163

**Comment by alfetopito** [comment]
> Last comment: I would recommend disabling the TradeWidget submission button if the form is invalid. Better UX imo

Good idea. Updated.

File: src/components/TradeWidget/TokenRow.tsx (hunk 1 above)
Comment (W3stside) [reply src/components/TradeWidget/TokenRow.tsx]: then it's probably better not to take the object as a param and the property only

File: src/components/TradeWidget/TokenRow.tsx [hunk 4]
```diff
@@ -113,17 +135,55 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
...
+  validateMaxAmount,
+}) => {
   const options = useMemo(() => tokens.map(token => ({ token, value: token.symbol, label: token.name })), [tokens])
 
+  const { register, errors, setValue, watch } = useFormContext()
+  const error = errors[inputId]
+
+  const max = Number(getMax(balance, token))
+  const inputValue = Number(watch(inputId)) || 0
+  const overMax = validateMaxAmount && inputValue > max ? inputValue - max : 0
```
Velenir: As `overMax` error isn't from the hook-form, I don't think it will prevent form submission
If I'm right we need either to especially check for it in `submitForm`, or make it independent from `useFormContext` instance.

I have this suggestion:
```ts
const TokenRow =(...) => {
   const max = getMax(balance, token)

  return <Wrapper>
      <TokenImgWrapper alt={token.name} src={token.image} />
      <SelectBox>
        ...
      </SelectBox>
      <InputBox>
        <InputComp key={token.name + max} max={max} balance={balance}/>
...
}

const InputComp = ({max, ...}) => {
  return       <InputBox>
        <input type="text" placeholder="0" />
        <input
          className={className}
          placeholder="0"
          name={inputId}
          type="text"
          required
          ref={register({
            validate: {
              positive: (value: string): true | string => value < max || 'Not enough balance'
            },
          })}
        />
        <WalletDetail>
            ...
        </WalletDetail>
}
```

Including `key` in props will remount component on key change, so register will always get an up-to-date validation rule and there'll be no need for out-of-form error checks.

Or that's the idea, not sure if it'll work.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -144,12 +204,26 @@ const TokenRow: React.FC<Props> = ({ token, tokens, selectLabel, onSelectChange,
...
+            },
+          })}
+        />
+        {errorOrWarning}
         <WalletDetail>
           <strong>
             <Link to="/deposit">Exchange wallet:</Link>
           </strong>{' '}
-          <span className="success">{displayBalance(balance, 'exchangeBalance')}</span>
+          <a className="success" onClick={(): void => setValue(inputId, max.toFixed(4))}>
```
Velenir: `toFixed` rounds the number, `max.toFixed(4)` can be `> max`, e.g. `5.55555.toFixed(4) = '5.5556'` so it'll error by default

src/components/TradeWidget/TokenRow.tsx
Velenir: Same can be done with `register/unregister` in `useEffect(..., [max])` but it's ugly.

src/components/TradeWidget/TokenRow.tsx
Velenir: `Pick<TokenBalanceDetails, 'which-prop-do-we-need-again'>`?

src/components/TradeWidget/TokenRow.tsx
W3stside: which is why I find max dangerous in the first place tbh

gnosis-info:
Travis automatic deployment:
 https://pr165--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -62,26 +76,33 @@ const TradeWidget: React.FC = () => {
...
-        <FontAwesomeIcon icon={faPaperPlane} size="lg" /> Send limit order
-      </SubmitButton>
+      <FormContext {...methods}>
+        <WrapperForm onSubmit={methods.handleSubmit(data => console.log('data', data))}>
+          <TokenRow
+            token={sellToken}
+            tokens={tokens}
+            balance={sellTokenBalance}
+            selectLabel="sell"
+            onSelectChange={onSelectChangeFactory(setSellToken, receiveToken)}
```
anxolin: My question is not not from this PR I guess, but why the word `Factory` here?

alfetopito: That's the metaphor I could thing of a function that returns a function. Is there anything better?

PR #166, opened 2019-11-01
Removing package-lock.json so we stick to yarn

alfetopito:
As discussed with the team, this PR removes the `package-lock.json`.
This means we'll use only `yarn` for all development to be consistent across the team.


gnosis-info:
Travis automatic deployment:
 https://pr166--dexreact.review.gnosisdev.com

PR #173, opened 2019-11-06
Fix get token function

anxolin:
`getToken` util has an issue and is making the app to break in development.

The issue has to do with how the case insensitivity was implemented.
Also, added some test to prevent regression bugs for this.

### DIFF
--- a/src/components/DepositWidget/index.tsx
+++ b/src/components/DepositWidget/index.tsx
@@ -179,17 +179,15 @@ const DepositWidget: React.FC = () => {
       const result = await depositApi.deposit(userAddress, tokenAddress, amount, txOptionalParams)
       log(`The transaction has been mined: ${result.receipt.transactionHash}`)
 
-      if (mounted.current) {
-        _updateToken(tokenAddress, ({ depositingBalance, walletBalance, ...otherParams }) => {
-          return {
-            ...otherParams,
-            depositingBalance: depositingBalance.add(amount),
-            walletBalance: walletBalance.sub(amount),
-            highlighted: true,
-          }
-        })
-        _clearHighlight(tokenAddress)
-      }
+      _updateToken(tokenAddress, ({ depositingBalance, walletBalance, ...otherParams }) => {
+        return {
+          ...otherParams,
+          depositingBalance: depositingBalance.add(amount),
+          walletBalance: walletBalance.sub(amount),
+          highlighted: true,
+        }
+      })
+      _clearHighlight(tokenAddress)
 
       toast.success(`Successfully deposited ${formatAmount(amount, decimals)} ${symbol}`)
     } catch (error) {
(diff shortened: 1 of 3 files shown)


src/utils/miscellaneous.ts
```diff
@@ -10,5 +10,13 @@ export function getToken<T extends TokenDetails>(
...
 ): T | undefined {
-  return (tokens || []).find(token => token[key] === value.toUpperCase())
+  const valueUppercase = value.toUpperCase()
+  return (tokens || []).find(token => {
+    const value = token[key]
+    if (value) {
+      return value.toString().toUpperCase() === valueUppercase
+    } else {
+      return false
+    }
```
alfetopito: What about:

```suggest
return (tokens || []).find(token => (token[key] || '').toString().toUpperCase() === value.toUpperCase())
```

Too much clutter?

gnosis-info:
Travis automatic deployment:
 https://pr173--dexreact.review.gnosisdev.com

src/utils/miscellaneous.ts
```diff
@@ -10,5 +10,13 @@ export function getToken<T extends TokenDetails>(
...
   tokens: T[] | undefined | null,
 ): T | undefined {
-  return (tokens || []).find(token => token[key] === value.toUpperCase())
+  const valueUppercase = value.toUpperCase()
+  return (tokens || []).find(token => {
+    const value = token[key]
+    if (value) {
+      return value.toString().toUpperCase() === valueUppercase
+    } else {
+      return false
```
W3stside: because this is generic it's not complaining but this return should be undefined. actually you can just leave out the `else` as it'll auto return `undefined` if not found.

src/utils/miscellaneous.ts
anxolin: I find to difficult to read these lines

Also, not that it matters much but you are converting to uppercase in each iteration of the token the `value` when it doesn't change

src/utils/miscellaneous.ts
anxolin: I think the doc says that it should return a `truthy` value when you find the element and a `falsy` value if the current element is not what you look for.

It's true that `undefined` is `falsy` and I can remove the else, but I find it more declarative to use a function that returns always a boolean. `false` it's also a good `falsy` value :) 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

gnosis-info:
Travis automatic deployment:
 https://pr173--dexreact.review.gnosisdev.com

src/utils/miscellaneous.ts
alfetopito: In that case, update the function return value to include `false`

src/components/DepositWidget/index.tsx
```diff
@@ -151,6 +151,10 @@ const DepositWidget: React.FC = () => {
   }
 
   function _updateToken(tokenAddress: string, updateBalances: Mutation<TokenBalanceDetails>): void {
+    if (!mounted.current) {
```
Velenir: If there's a mounted check inside `_updateToken`, then there's absolutely no need for any mounted checks outside, like L182 and some others

src/utils/miscellaneous.ts
W3stside: ok, then type the function return as `boolean`

src/components/DepositWidget/index.tsx
anxolin: True! changed

src/utils/miscellaneous.ts
anxolin: The linter didn't enforce me to do that, but if you think it's more clear, I'll add it explicitly

PR #175, opened 2019-11-07
102/submit order

alfetopito:
Closes #102 

Depends on #36 

---

Placing order

- [x] New hook `usePlaceOrder`
- [x] Submits order
- [x] Disables all inputs while submitting
- [x] Displays success message
- [x] Resets form

Does not:
1. Redirect and remembers all parameters. ~~Waiting for https://github.com/gnosis/dex-react/pull/138/~~ Addressed on https://github.com/gnosis/dex-react/pull/184
1. Handles state when no longer mounted. Waiting for #132

PR #176, opened 2019-11-07
Use BN over Number

alfetopito:
Depends on #175 

---

Proposal to use BN instead of Number on Trading widget.

To be able to work with the proper precision for each token, this change enforces only max precision chars are used for each respective token, preventing the user from inserting more than given precision.

Try it out and let me know what you think

---

**Edit**: Based on comments, did a small refactor on the validation:

- [x] No longer alowing chars other than 0-9 and .
- [x] Validating input based on regex rather than converting value to number
- [x] Removing redundant zeros/dot onBlur
- [x] Refactored displayBalance to use keyof

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -7,7 +7,7 @@ import { useFormContext } from 'react-hook-form'
 import TokenImg from 'components/TokenImg'
 import TokenSelector from 'components/TokenSelector'
 import { TokenDetails, TokenBalanceDetails } from 'types'
-import { formatAmount, formatAmountFull, parseAmount } from 'utils'
+import { formatAmount, formatAmountFull, parseAmount, adjustPrecision } from 'utils'
 import { ZERO } from 'const'
 
 const Wrapper = styled.div`
@@ -78,16 +78,6 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
     return '0'
   }
   return formatAmount(balance[key], balance.decimals) || '0'
-}
-
-function adjustPrecision(value: string | undefined | null, precision: number): string {
-  const match = (value || '').match(/(^\d+\.)(\d+)$/)
-  if (match && match[2].length > precision) {
-    let [, intPart, decimalPart] = match
-    decimalPart = decimalPart.slice(0, precision)
-    return intPart + decimalPart
-  }
-  return value
 }
 
 // TODO: move into a validators file?
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -1,4 +1,5 @@
-import React from 'react'
+import React, { useEffect, useCallback } from 'react'
+import BN from 'bn.js'
 import { Link } from 'react-router-dom'
 import styled from 'styled-components'
 import { useFormContext } from 'react-hook-form'
@@ -6,7 +7,8 @@ import { useFormContext } from 'react-hook-form'
 import TokenImg from 'components/TokenImg'
 import TokenSelector from 'components/TokenSelector'
 import { TokenDetails, TokenBalanceDetails } from 'types'
-import { formatAmount, formatAmountFull } from 'utils'
+import { formatAmount, formatAmountFull, parseAmount } from 'utils'
+import { ZERO } from 'const'
 
 const Wrapper = styled.div`
   display: flex;
@@ -78,13 +80,19 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
   return formatAmount(balance[key], balance.decimals) || '0'
 }
 
-function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
-  return formatAmountFull((balance || {}).exchangeBalance, token.decimals, false) || '0'
+function adjustPrecision(value: string | undefined | null, precision: number): string {
+  const match = (value || '').match(/(^\d+\.)(\d+)$/)
+  if (match && match[2].length > precision) {
+    let [, intPart, decimalPart] = match
+    decimalPart = decimalPart.slice(0, precision)
+    return intPart + decimalPart
+  }
+  return value
 }
 
 // TODO: move into a validators file?
-function validatePositive(value: string): true | string {
-  return Number(value) > 0 || 'Invalid amount'
+function validatePositive(value: string, precision: number): true | string {
+  return new BN(parseAmount(value, precision) || '0').gt(ZERO) || 'Invalid amount'
 }
 
 interface Props {
@@ -110,19 +118,23 @@ const TokenRow: React.FC<Props> = ({
 }) => {
   const { register, errors, setValue, watch } = useFormContext()
   const error = errors[inputId]
+  const inputValue = watch(inputId)
 
-  const max = Number(getMax(balance, selectedToken))
-  const inputValue = Number(watch(inputId)) || 0
-  const overMax = validateMaxAmount && inputValue > max ? inputValue - max : 0
+  let overMax = ZERO
+  if (balance && validateMaxAmount) {
+    const max = balance.exchangeBalance
+    const value = new BN(parseAmount(inputValue, selectedToken.decimals) || '0')
+    overMax = value.gt(max) ? value.sub(max) : ZERO
+  }
 
-  const className = error ? 'error' : !!overMax ? 'warning' : ''
+  const className = error ? 'error' : overMax.gt(ZERO) ? 'warning' : ''
 
   const errorOrWarning = error ? (
     <WalletDetail className="error">{error.message}</WalletDetail>
   ) : (
-    !!overMax && (
+    overMax.gt(ZERO) && (
       <WalletDetail className="warning">
-        Selling {overMax.toFixed(4)} {selectedToken.symbol} over your current balance
+        Selling {formatAmountFull(overMax, selectedToken.decimals)} {selectedToken.symbol} over your current balance
       </WalletDetail>
     )
   )
@@ -130,6 +142,17 @@ const TokenRow: React.FC<Props> = ({
   function useMax(): void {
     setValue(inputId, formatAmountFull(balance.exchangeBalance, balance.decimals, false))
   }
+
+  const enforcePrecision = useCallback(() => {
+    const newValue = adjustPrecision(inputValue, selectedToken.decimals)
+    if (inputValue !== newValue) {
+      setValue(inputId, newValue, true)
+    }
+  }, [inputValue, selectedToken.decimals, setValue, inputId])
+
+  useEffect(() => {
+    enforcePrecision()
+  }, [enforcePrecision])
 
   return (
     <Wrapper>
@@ -151,9 +174,10 @@ const TokenRow: React.FC<Props> = ({
           required
           ref={register({
             validate: {
-              positive: (value: string): true | string => validatePositive(value),
+              positive: (value: string): true | string => validatePositive(value, selectedToken.decimals),
             },
           })}
+          onChange={enforcePrecision}
         />
         {errorOrWarning}
         <WalletDetail>
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -73,11 +73,14 @@ const WalletDetail = styled.div`
   }
 `
 
-function displayBalance(balance: TokenBalanceDetails | undefined | null, key: string): string {
+function displayBalance<K extends keyof TokenBalanceDetails>(
+  balance: TokenBalanceDetails | undefined | null,
+  key: K,
+): string {
   if (!balance) {
     return '0'
   }
-  return formatAmount(balance[key], balance.decimals) || '0'
+  return formatAmount(balance[key] as BN, balance.decimals) || '0'
 }
 
 // TODO: move into a validators file?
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr176--dexreact.review.gnosisdev.com

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -137,13 +139,19 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
   return formatAmount(balance[key], balance.decimals) || '0'
 }
 
-function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
-  return formatAmountFull((balance || {}).exchangeBalance, token.decimals, false) || '0'
+function adjustPrecision(value: string | undefined | null, precision: number): string {
```
anxolin: What does it do exactly?
shouldn't this be using the utils we have. We have kind of similar code we could use, right?

alfetopito: This function `rounds` the value to the required precision, such as `1.15` on precision `1` becomes `1.1`.

The existing functionality (`parseAmount`) returns `null` when decimal places > precision: https://github.com/gnosis/dex-react/blob/develop/src/utils/format.ts#L89

But this is a good point. I can just adapt `parseAmount` instead.

alfetopito:
> I think using BN it's the right approach, but tbh, I didn't understand what is the exactly issue addressed:
> 
> > only max precision chars are used for each respective token
> 
> Is this that if the use makes a mistake and add an extra decimal, we only take the precision of the token?

The point for this is, on the contract interface and contract itself we cannot submit an order with higher precision than the token allows. Or can we?

In order words, using GUSD as example (token decimals == 2), we cannot sell/buy 1.001 GUSD. Either 1.00 or 1.01.

So in this change I'm enforcing it on the interface.

Other reason is that, working with BN only allows integers. To represent the decimal string input values I convert them to BN using the token precision. The value above the precision is lost.

Does that make sense?

src/components/TradeWidget/TokenRow.tsx
anxolin: Aha, I understand 👍 

Yes, I agree, a refactor there would be a good idea

alfetopito: While working on it I realized that it won't be very efficient if I use existing helpers, even after adjusting `parseAmount` to truncate on given precision.
That's because I'll have to do 2 conversions: input string -> BN -> string, something like:

```typescript
function adjustPrecision2(value: string | undefined | null, precision: number): string {
  const bn = parseAmount(value, precision) || ZERO
  return formatAmountFull(bn, precision, false) || value
}
```

I believe the version implemented might be more efficient. I would instead move it to utils and extract common functionality there.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -190,6 +202,17 @@ const TokenRow: React.FC<Props> = ({
     setValue(inputId, formatAmountFull(balance.exchangeBalance, balance.decimals, false))
   }
 
+  const enforcePrecision = useCallback(() => {
```
Velenir: No need to create this outside of `useEffect`

alfetopito: I need it because this function is also used for `onChange` on the input.

Velenir: Missed that, my bad :sweat:

alfetopito: Unforgivable!

gnosis-info:
Travis automatic deployment:
 https://pr176--dexreact.review.gnosisdev.com

src/utils/format.ts
```diff
@@ -98,6 +100,30 @@ export function parseAmount(amountFmt: string, amountPrecision = DEFAULT_PRECISI
...
+ *
+ * @param value The decimal value to be adjusted as a string
+ * @param precision How many decimals should be kept
+ */
+export function adjustPrecision(value: string | undefined | null, precision: number): string {
+  if (!value) {
+    return ''
+  }
+
+  const match = value.match(/(^\d+\.)(\d+)$/)
```
Velenir: or
```js
const regexp = new RegExp(`(\\.\\d{${precision}})\\d+$`)
return value.replace(regexp, '$1')
```

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -226,9 +239,10 @@ const TokenRow: React.FC<Props> = ({
           required
           ref={register({
             validate: {
-              positive: (value: string): true | string => validatePositive(value),
+              positive: (value: string): true | string => validatePositive(value, selectedToken.decimals),
```
Velenir: I can't be sure, but
As `validate` inside register is memoized (isn't it?) and `selectedToken` comes from props, which can change. If `decimals` change, `validate` would still have older decimals.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -141,13 +143,9 @@ function displayBalance(balance: TokenBalanceDetails | undefined | null, key: st
...
 
-function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
-  return formatAmountFull((balance || {}).exchangeBalance, token.decimals, false) || '0'
-}
-
 // TODO: move into a validators file?
-function validatePositive(value: string): true | string {
-  return Number(value) > 0 || 'Invalid amount'
+function validatePositive(value: string, precision: number): true | string {
+  return new BN(parseAmount(value, precision) || '0').gt(ZERO) || 'Invalid amount'
```
Velenir: Do we really care about precision when comparing to 0? `1.123456789045678 > 0`  same as `1.1234 > 0`

src/components/TradeWidget/TokenRow.tsx
Velenir: That is why I mentioned in some other comment just remounting the whole thing on token change.

test/utils/adjustPrecision.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+
+describe('bellow precision', () => {
+  test('with decimals', () => {
+    expect(adjustPrecision('1.1', 3)).toBe('1.1')
+  })
+  test('only integers', () => {
+    expect(adjustPrecision('1', 3)).toBe('1')
+  })
+  test('partial number', () => {
+    expect(adjustPrecision('1.', 3)).toBe('1.')
```
anxolin: This is a bit odd, isn't it?

I see JS accept `1.` as input, but would we want that?

src/utils/format.ts
```diff
@@ -85,9 +85,11 @@ export function parseAmount(amountFmt: string, amountPrecision = DEFAULT_PRECISI
 
   const groups = /^(\d+)(?:\.(\d+))?$/.exec(amountFmt)
   if (groups) {
-    const [, integerPart, decimalPart = ''] = groups
+    const integerPart = groups[1]
+    let decimalPart = groups[2] || ''
     if (decimalPart.length > amountPrecision) {
-      return null
+      // truncate whatever is over precision
+      decimalPart = decimalPart.slice(0, amountPrecision)
```
anxolin: So we want to change the behaviour for parse amount too, is that right?

src/utils/format.ts
```diff
@@ -98,6 +100,30 @@ export function parseAmount(amountFmt: string, amountPrecision = DEFAULT_PRECISI
...
+ * @param value The decimal value to be adjusted as a string
+ * @param precision How many decimals should be kept
+ */
+export function adjustPrecision(value: string | undefined | null, precision: number): string {
+  if (!value) {
+    return ''
+  }
+
+  const match = value.match(/(^\d+\.)(\d+)$/)
+  if (match && match[2].length > precision) {
```
anxolin: It looks very similar to me to `parseAmount`

Wouldn't be worth to split `parseAmount` in two. It would be similar to `_decomposeBn` but accepting a string instead of a BN. One idea would be to overload the method and provide the two signatures

test/utils/adjustPrecision.spec.ts
alfetopito: Because the validation runs on every keystroke. I don't want to prevent the user from inserting a decimal value by preventing him to place a dot.

src/utils/format.ts
alfetopito: Even though I'm not depending on this, we had agreed to change. I don't mind removing the change though, just think it makes more sense to truncate rather than return `null`.

What do you say?

src/components/TradeWidget/TokenRow.tsx
alfetopito: We do if using BN. In fact we also do if dealing with Number due to issues with float point.
Example for BN:

```typescript
parseAmount('0.001', 2) === new BN('0.00'); === ZERO
parseAmount('0.001', 3) === new BN('0.001'); > ZERO
```

src/components/TradeWidget/TokenRow.tsx
alfetopito: You are right. This needs fixing.

test/utils/adjustPrecision.spec.ts
anxolin: aha, okay!

src/utils/format.ts
anxolin: `null` signals, I can't parse. However, the truncation is a fair approach. And if we agreed, I don't want to change that :)

src/components/TradeWidget/TokenRow.tsx
alfetopito: Well, actually, if we are enforcing the precision, the example where I do need precision will never happen. In fact, I can don't need BN at all here. As long as the value is a valid number > 0 it should return true.

I'll revert this part.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -131,6 +141,36 @@ const TokenRow: React.FC<Props> = ({
...
+    () => (event: React.SyntheticEvent<HTMLInputElement>): void => {
+      // Q: Why do we need this function instead of relying on `preventInvalidChars` or `enforcePrecision`?
+      // A: Because on those functions we still want the user to be able to input partial values. E.g.:
+      //    0 -> 0. -> 0.1 -> 0.10 -> 0.105
+      //    When losing focus though, we remove everything that's redundant, such as leading zeros, trailing dots and/or zeros
+      // Q: Why not use formatAmount/parseAmount that already take care of this?
+      // A: Too many steps (convert to and from BN) and binds the function to selectedToken.decimals
+
+      const { value } = event.currentTarget
+      const newValue = value.replace(leadingAndTrailingZeros, '').replace(trailingZerosAfterDot, '$1')
```
Velenir: One replacement with `value.replace(/^0{2,}(?=\d*\.?)|\.0*$|(?<=\.\d*)0*$/g, '')`
Check [here](https://regex101.com/r/MismO5/2)

Velenir:
Approved with some suggestions

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -71,20 +73,24 @@ const WalletDetail = styled.div`
...
     return '0'
   }
-  return formatAmount(balance[key], balance.decimals) || '0'
+  return formatAmount(balance[key] as BN, balance.decimals) || '0'
 }
 
-function getMax(balance: TokenBalanceDetails | null, token: TokenDetails): string {
-  return formatAmountFull((balance || {}).exchangeBalance, token.decimals, false) || '0'
-}
+const validInputPattern = new RegExp(/^\d+(\.(\d+)?)?$/) // allows leading and trailing zeros
```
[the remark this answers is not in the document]
alfetopito: Yeah, that simplifies things, thanks :)

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -131,6 +141,36 @@ const TokenRow: React.FC<Props> = ({
...
+
+  useEffect(() => {
+    enforcePrecision()
+  }, [enforcePrecision])
+
+  const removeExcessZeros = useMemo(
+    () => (event: React.SyntheticEvent<HTMLInputElement>): void => {
+      // Q: Why do we need this function instead of relying on `preventInvalidChars` or `enforcePrecision`?
+      // A: Because on those functions we still want the user to be able to input partial values. E.g.:
+      //    0 -> 0. -> 0.1 -> 0.10 -> 0.105
```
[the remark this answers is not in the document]
alfetopito: The second case, this function is triggered only `onBlur`

src/components/TradeWidget/TokenRow.tsx
alfetopito: Yes, I tried that, but the positive look behind with wildcards doesn't work.
From https://regex101.com/: `* A quantifier inside a lookbehind makes it non-fixed width`

Even then one you linked has an error message: `? The preceding token is not quantifiable`

alfetopito:
> I just noticed that now the web gives you this messages now before you input anything:
> 
> image

That's because by default there's no value filled in.
Hopefully when #184 is merged this should be fixed, by always having something from the url redirect.
I rather address it when both are in place to avoid re-work.

src/components/TradeWidget/TokenRow.tsx
Velenir: Worked for me in chrome, but actually doesn't in Firefox now that I tried. Don't bother then

PR #177, opened 2019-11-08
Extracting TokenList into json

fleupold:
First step for tackling https://github.com/gnosis/dex-contracts/issues/252

Extracting the relevant token information in a json the we can query in the dex-services repo and then run `registerToken()` with all the addresses for the specified network.

I don't really know what the best way to get type unsafe json into typescript. Happy to get pointers. Note, that right now eslint seems to fail with three unrelated errors:

```
/Users/felixleupold/Gnosis/dex-react/src/components/DepositWidget/Row.tsx
  114:22  error  Replace `⏎················onClick={():·void·=>·showForm('withdraw')}⏎················disabled={isWithdrawFormVisible}⏎················className="danger"⏎··············` with `·onClick={():·void·=>·showForm('withdraw')}·disabled={isWithdrawFormVisible}·className="danger"`  prettier/prettier

/Users/felixleupold/Gnosis/dex-react/src/hooks/useWindowSizes.ts
  17:1  error  Delete `····`  prettier/prettier
  25:1  error  Delete `··`    prettier/prettier
```

### Test Plan

`npm start`

Make sure page loads and shows tokens. What else should I test?


gnosis-info:
Travis automatic deployment:
 https://pr177--dexreact.review.gnosisdev.com

anxolin:
Thanks for the PR! 

About the error you mentioned, was merged to develop already. You can rebase and you should get that fix:
https://github.com/gnosis/dex-react/pull/183

One small thing. Why not keep it as a JS instead of JSON so we can keep the comments? I find them useful. It's a pity to see them gone. TS can import JS in the same way it can import the JSON too.

josojo:
I am trying to finish this PR, as Felix is on vacation.

>One small thing. Why not keep it as a JS instead of JSON so we can keep the comments? I find them useful. It's a pity to see them gone. TS can import JS in the same way it can import the JSON too.

@anxolin  Json is a more universal protocol. We also want to import in our rust base, and there it is much easier to use Json than js code :)

>About the error you mentioned, was merged to develop already. You can rebase and you should get that fix:
#183

perfect! did it and got rid of it.

How do you guys deal with coverage? Should I improve coverage, before I merge it?

anxolin:
> Json is a more universal
Agree 👍 
Maybe we could just move the info we wrote in the comments to a markdown somewhere, so we don't lose it?

> How do you guys deal with coverage

There's no a strong policy in place. We want to raise coverage over time. We focus more on testing utils and custom libs than React components.

I don't think we need to test a JSON file imo

@Velenir might have some input on how to add types for the JSON that could be useful for you. 

Last point, I see the branch is in sync with development cause it was force-pushed. There's 2 approvals already, so I guess the new code should be reviewed again.

josojo:
>Maybe we could just move the info we wrote in the comments to a markdown somewhere, so we don't lose it?

Done

> I don't think we need to test a JSON file imo

I was asking more in general. fortunately, .md and .json files are getting excluded automatically. But now, I see how you guys are doing it. All good.

>Last point, I see the branch is in sync with development cause it was force-pushed. There's 2 approvals already, so I guess the new code should be reviewed again.

Yeah, I rebased it based on your suggestion. If you guys don't like rebasing and force pushes, as it requires a little bit more trust in the coders, I can do merge instead the next time.

I think the PR is ready now. Let me know!

anxolin:
>  I see how you guys are doing it. All good.

Yep we do

>  If you guys don't like rebasing and force pushes

It's fine. It's your branch. Rebase with `develop` should be fine. I asked because in the moment I checked develop was identical to this branch, so there were no changes. But I see there's now. I'll review!

Thanks!!

src/api/tokenList/Faucet_info.md
```diff
@@ -0,0 +1,53 @@
+Ether, Tether, TrueUSD, USD Coin, Paxos, Gemini, Dai
```
anxolin: Can you add a bit of format and create a link the file in the main README.md? it's being displayed like this: https://github.com/gnosis/dex-react/blob/token_list_json/src/api/tokenList/Faucet_info.md

For example:
```md
# Faucelet information
## Wrapped Ether
* [https://weth.io](https://weth.io)
* Wrapper of Ether to make it ERC-20 compliant
* **Rinkeby faucet**: [https://faucet.rinkeby.io/](https://faucet.rinkeby.io)
...
```

josojo:
Nice! thanks @Velenir 
Please review again.

src/api/tokenList/Faucet_info.md
```diff
@@ -0,0 +1,51 @@
+# Ether, Tether, TrueUSD, USD Coin, Paxos, Gemini, Dai
```
fleupold: Note that I tried to capture this information inside the json as well by adding *website*, *description* and *faucet* fields.

PR #184, opened 2019-11-08
102/url params

alfetopito:
Depends on #175 

---

- [x] Updating urls params with amounts on change
- [x] Keeping values when connecting/disconnecting wallet

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -157,7 +157,8 @@ const TradeWidget: React.FC = () => {
         reset()
       }
     } else {
-      history.push('/connect-wallet', { from: history.location.pathname })
+      const from = history.location.pathname + history.location.search
+      history.push('/connect-wallet', { from })
     }
   }
 
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -64,7 +64,7 @@ const TradeWidget: React.FC = () => {
   const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
 
   // Listen on manual changes to URL search query
-  const { sell: sellTokenSymbol, receive: receiveTokenSymbol } = useParams()
+  const { sell: sellTokenSymbol, sellAmt, receive: receiveTokenSymbol, receiveAmt } = useParams()
 
   const [sellToken, setSellToken] = useState(
     () => getToken('symbol', sellTokenSymbol, tokens) || getToken('symbol', 'DAI', tokens),
@@ -72,9 +72,14 @@ const TradeWidget: React.FC = () => {
   const [receiveToken, setReceiveToken] = useState(
     () => getToken('symbol', receiveTokenSymbol, tokens) || getToken('symbol', 'USDC', tokens),
   )
+  const sellInputId = 'sellToken'
+  const receiveInputId = 'receiveToken'
+
+  const methods = useForm({ mode: 'onBlur', defaultValues: { [sellInputId]: sellAmt, [receiveInputId]: receiveAmt } })
+  const { handleSubmit, watch, reset } = methods
 
   // Change URL on internal token change
-  useURLParams(`sell=${sellToken.symbol}-0&receive=${receiveToken.symbol}-20`)
+  useURLParams(`sell=${sellToken.symbol}-${watch(sellInputId)}&receive=${receiveToken.symbol}-${watch(receiveInputId)}`)
 
   const { balances } = useTokenBalances()
 
@@ -83,12 +88,6 @@ const TradeWidget: React.FC = () => {
     balances,
     receiveToken.symbol,
   ])
-
-  const methods = useForm({ mode: 'onBlur' })
-  const { handleSubmit, watch, reset } = methods
-
-  const sellInputId = 'sellToken'
-  const receiveInputId = 'receiveToken'
 
   const { isSubmitting, placeOrder } = usePlaceOrder()
   const history = useHistory()
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -20,7 +20,7 @@ import { tokenListApi } from 'api'
 
 import { Network, TokenDetails } from 'types'
 
-import { getToken, safeTokenName, parseAmount } from 'utils'
+import { getToken, safeTokenName, parseAmount, formatAmountFull } from 'utils'
 
 const WrappedWidget = styled(Widget)`
   overflow-x: visible;
@@ -56,6 +56,19 @@ const SubmitButton = styled.button`
   }
 `
 
+// This is sort of a duplication from the validation done on TokenRow.
+// Still, required because the initial value comes from the URL params,
+//   and we have no control over what the user inserts there.
+// Thus, we make it a nicely formated value, or 0 if we can't
+// And only on first render. Subsequent validation is taken care by TokenRow.
+function sanitizeInputAmount(value: string, precision: number): string {
+  const number = parseAmount(value, precision)
+  if (!number) {
+    return '0'
+  }
+  return formatAmountFull(number, precision, false)
+}
+
 const TradeWidget: React.FC = () => {
   const { networkId, isConnected } = useWalletConnection()
   // Avoid displaying an empty list of tokens when the wallet is not connected
@@ -75,9 +88,17 @@ const TradeWidget: React.FC = () => {
   const sellInputId = 'sellToken'
   const receiveInputId = 'receiveToken'
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const sanitizedSellAmount = useMemo(() => sanitizeInputAmount(sellAmount, sellToken.decimals), [])
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const sanitizedReceiveAmount = useMemo(() => sanitizeInputAmount(receiveAmount, receiveToken.decimals), [])
+
   const methods = useForm({
     mode: 'onBlur',
-    defaultValues: { [sellInputId]: sellAmount, [receiveInputId]: receiveAmount },
+    defaultValues: {
+      [sellInputId]: sanitizedSellAmount,
+      [receiveInputId]: sanitizedReceiveAmount,
+    },
   })
   const { handleSubmit, watch, reset } = methods
 
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -64,7 +64,7 @@ const TradeWidget: React.FC = () => {
   const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
 
   // Listen on manual changes to URL search query
-  const { sell: sellTokenSymbol, sellAmt, receive: receiveTokenSymbol, receiveAmt } = useParams()
+  const { sell: sellTokenSymbol, sellAmount, receive: receiveTokenSymbol, receiveAmount } = useParams()
 
   const [sellToken, setSellToken] = useState(
     () => getToken('symbol', sellTokenSymbol, tokens) || getToken('symbol', 'DAI', tokens),
@@ -75,7 +75,10 @@ const TradeWidget: React.FC = () => {
   const sellInputId = 'sellToken'
   const receiveInputId = 'receiveToken'
 
-  const methods = useForm({ mode: 'onBlur', defaultValues: { [sellInputId]: sellAmt, [receiveInputId]: receiveAmt } })
+  const methods = useForm({
+    mode: 'onBlur',
+    defaultValues: { [sellInputId]: sellAmount, [receiveInputId]: receiveAmount },
+  })
   const { handleSubmit, watch, reset } = methods
 
   // Change URL on internal token change
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -82,7 +82,13 @@ const TradeWidget: React.FC = () => {
   const { handleSubmit, watch, reset } = methods
 
   // Change URL on internal token change
-  useURLParams(`sell=${sellToken.symbol}-${watch(sellInputId)}&receive=${receiveToken.symbol}-${watch(receiveInputId)}`)
+  const url = new URLSearchParams({
+    sell: sellToken.symbol,
+    'sell-amount': watch(sellInputId),
+    buy: receiveToken.symbol,
+    'buy-amount': watch(receiveInputId),
+  })
+  useURLParams(url.toString())
 
   const { balances } = useTokenBalances()
 
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -125,16 +125,7 @@ const TradeWidget: React.FC = () => {
         reset()
       }
     } else {
-      history.push('/connect-wallet')
-      // TODO: connect wallet then come back, ideally with the data pre-filled
-      // might require David changes to be merged first
-      // when redirect back we need:
-      // 1. fill the form
-      // 2. trigger validation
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr184--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr184--dexreact.review.gnosisdev.com

src/components/Wallet.tsx
```diff
@@ -150,7 +150,12 @@ const Wallet: React.FC<RouteComponentProps> = (props: WalletProps) => {
       toast.error('Error disconnecting wallet')
     } finally {
       setLoadingLabel(undefined)
-      props.history.push('/')
+
+      // If on trade page already, keep pre-selected values when logging out
+      const pathname = props.location.pathname
```
W3stside: nit: `const { pathname } = props.location`

src/components/Wallet.tsx
```diff
@@ -150,7 +150,12 @@ const Wallet: React.FC<RouteComponentProps> = (props: WalletProps) => {
...
+      const redirectTo = typeof pathname === 'string' && pathname.match(/^\/trade/) ? pathname : '/'
+
+      props.history.push(redirectTo)
```
Velenir: Are you still pushing even if on the same path? Why?

Velenir: Also there's a recommended way to check if you are where you think you are location-wise
https://reacttraining.com/react-router/web/api/Hooks/useroutematch

src/components/Wallet.tsx
Velenir: Since we are slowly switching to hooks for react-router. Mind changing Wallet from `withRouter` to hooks?

src/components/TradeWidget/index.tsx
```diff
@@ -64,17 +64,22 @@ const TradeWidget: React.FC = () => {
   const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
 
   // Listen on manual changes to URL search query
-  const { sell: sellTokenSymbol, receive: receiveTokenSymbol } = useParams()
+  const { sell: sellTokenSymbol, sellAmt, receive: receiveTokenSymbol, receiveAmt } = useParams()
```
anxolin: I wouldn't shorten the name. `sellAmount`

Also, although we could change `sell` for `sellSymbol` too, maybe is best to leave it. It seams reasonable to make `sell` to accept in the future also addresses. So you can virtually encode any trade of ANY token pair in the url.

src/components/TradeWidget/index.tsx
```diff
@@ -64,17 +64,22 @@ const TradeWidget: React.FC = () => {
...
   )
+  const sellInputId = 'sellToken'
+  const receiveInputId = 'receiveToken'
+
+  const methods = useForm({ mode: 'onBlur', defaultValues: { [sellInputId]: sellAmt, [receiveInputId]: receiveAmt } })
+  const { handleSubmit, watch, reset } = methods
 
   // Change URL on internal token change
-  useURLParams(`sell=${sellToken.symbol}-0&receive=${receiveToken.symbol}-20`)
+  useURLParams(`sell=${sellToken.symbol}-${watch(sellInputId)}&receive=${receiveToken.symbol}-${watch(receiveInputId)}`)
```
anxolin: ubernit: should we define a couple of constants for `sellAmount` and `receiveAmount` so the line 82 is more declarative.

Also, the standard way of encoding the amounts should be something like:
`?sell=DAI&sell-amount=200&buy=WETH&buy-amount=1`

src/components/TradeWidget/index.tsx
```diff
@@ -122,16 +121,7 @@ const TradeWidget: React.FC = () => {
...
-      // TODO: connect wallet then come back, ideally with the data pre-filled
-      // might require David changes to be merged first
-      // when redirect back we need:
-      // 1. fill the form
-      // 2. trigger validation
-      // 3. call submit again
-      // OR... bypass all that and call this function with given parameters?
-      // we'll potentially need a flag from the redirect to indicate whether we are now connected
-      //   and should proceed from where we left off
+      history.push('/connect-wallet', { from: history.location.pathname })
```
anxolin: This is cool!

I think I made a comment in another PR just about this, I see you solved it 👍

src/components/TradeWidget/index.tsx
Velenir: We are currently using a path with slugs. You want to switch to search params?

Velenir: More declarative? Like this?
```ts
  const url = new URLSearchParams({
    sell: `${sellToken.symbol}-${watch(sellInputId)}`,
    receive: `${receiveToken.symbol}-${watch(receiveInputId)}`,
  })

  // Change URL on internal token change
  useURLParams(url.toString())
```

src/components/Wallet.tsx
alfetopito: > nit: const { pathname } = props.location

Done

> Since we are slowly switching to hooks for react-router. Mind changing Wallet from withRouter to hooks?

Should be in a different PR. Will create a task for it.

src/components/Wallet.tsx
alfetopito: Not entirely sure I understood how to use this hook. Can you tell me if that's correct?

src/components/TradeWidget/index.tsx
alfetopito: Updated

src/components/TradeWidget/index.tsx
alfetopito: @anxolin updated as you suggested, but I don't know.

Feels a bit weird: `/trade/sell=GUSD&sell-amount=50048.29&buy=TUSD&buy-amount=980`

anxolin: I find it ok, I would feel comfortable leaving as `/trade/sell=GUSD&sell-amount=50048.29&buy=TUSD&buy-amount=980`

@DmitryBespalov the part i don't like that much, is having a param with two params on it with our own logic. 

An alternative approach would be to give more entity to the token pair as in many REST APIs, so these URL would be valid:
* `/trade`
* `/trade/GUSD-TUSD`
* `/trade/GUSD-TUSD?sell-amount=50048.29
* `/trade/GUSD-TUSD?sell-amount=50048.29&buy-amount=980`

Would you be more comfortable with this approach?

anxolin: Problem would be, what do we do when the user selects JUST the sell token or buy token?

It's not weird for example sth like?
* /trade/GUSD-none
* /trade/none-TUSD

also, hope that some doesn't create the `NONE` token 

Equally weird to do:
* /trade/GUSD-
* /trade/-TUSD

Or not modifying the URL until the user selected the token...

alfetopito: I do not like the idea of GUSD-none, but this is not a case that will ever happen.

We have pre-selected by default DAI-USDC, and the selector only allows valid tokens.

When the user manually types a non-existing token in the url we can just redirect it to the default path.

So if we want to go for this approach, the url will always be fully filled like `/trade/GUSD-TUSD?sell-amount=50048.29&buy-amount=980`

anxolin: Okay, sounds food then to me!

bh2smith: I would personally scrap the word `amount` as part of the URL all together
```
/trade/sell-50048.29-TUSD&buy-980-DAI
```

alfetopito: Although this idea shortens the url, the initial argument was against too little url params.

Building on top of it, we can shorten the url keeping individual params without losing meaning with:

```
/trade/GUSD-TUSD?sell=4234.43&buy=87312.1
```

What do you guys think of this? @bh2smith @anxolin @DmitryBespalov

bh2smith: I think this looks MUCH better! However, if you are worried about uniqueness... you might want to include the user's address/id or `auctionIndex` (if necessary)

alfetopito: @bh2smith I don't think we need that.

Anyway, a few hours later, I managed to get the url working, just not quite the way I expected.

Couldn't figure out how to use a `?` in the url like usual url search params :/
The routing just wouldn't work.

So, updated url to:  

```
/trade/TUSD-GUSD/sell=1.0&buy=1.0
```

Velenir: `history.push('?whatever')` should work, no?

anxolin: @bh2smith for encoding params in the URL there's some rules, I think although `/trade/sell-50048.29-TUSD&buy-980-DAI` sounds good in terms of you can almost read it as if it's english, this URL assumes some collocation order of the parameters. Some of them are even optional

I think Leandro proposal is good, cause:
* `/trade`: It's an entry point for doing any trade
* `/trade/TUSD-GUSD`: Allows to trade TUSD for GUSD
* `/trade/TUSD-GUSD?sell=1.0&buy=1.0`: Allows to provide any or both of the optional parameters. Since the token is removed from the query params and becomes part of the document it self, I agree is a good call to drop the word `amount`

Only BUT is that, `sell` and `buy`should be to the right of the question mark imo

anxolin: @alfetopito, maybe this would help: https://github.com/sindresorhus/query-string

alfetopito: @Velenir 
> `history.push('?whatever')` should work, no?

The issue is in the route matching. I can push the `?` into history, but the Trade widget is never rendered when I use any `?` in the `path`. Examples:

```
<Route path="/trade/:sell-:buy?sell=:sellAmount&buy=:buyAmount" component={Trade} />
<Route path="/trade/:sell-:buy\?sell=:sellAmount&buy=:buyAmount" component={Trade} />
<Route path="/trade/:sell-:buy\\?sell=:sellAmount&buy=:buyAmount" component={Trade} />
```

From [react-router docs](https://reacttraining.com/react-router/web/api/Route/path-string-string), [path-to-regexp](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0) deals with this.
My guess is that the `?` is being interpreted in their sudo regex rather than as a literal in the url path.

gnosis-info:
Travis automatic deployment:
 https://pr184--dexreact.review.gnosisdev.com

alfetopito:
> Something stranges happens if you visit:
> https://pr184--dexreact.review.gnosisdev.com/trade/sell=DAI&sell-amount=0&buy=USDC&buy-amount=1
> 
> the URL changes automatically to:
> https://pr184--dexreact.review.gnosisdev.com/trade/sell=DAI&sell-amount=0&buy=USDC&buy-amount=1/sell=DAI&sell-amount=0&buy=USDC&buy-amount=1
> 
> I made a small suggestion for the naming convention, also a suggestion for parsing the query params

Fixed in the latest revision

src/components/TradeWidget/index.tsx
alfetopito: Tried again a minimal test:

```tsx
const TestQuestionMark: React.FC = () => {
  const match = useRouteMatch()
  return <div>Matched {match.path}</div>
}

const App: React.FC = () => (
  <>
    <GlobalStyles />
    <Router basename={process.env.BASE_URL}>
      <Layout>
        <Switch>
          <Route path="/test?" component={TestQuestionMark} />
          <Route path="/test\?" component={TestQuestionMark} />
          <Route path="/test\\?" component={TestQuestionMark} />
          <Route path="/test/?" component={TestQuestionMark} />
          <Route path="/test/\?" component={TestQuestionMark} />
          <Route path="/test/\\?" component={TestQuestionMark} />
          <Route component={NotFound} />
        </Switch>
      </Layout>
    </Router>
  </>
)
```

Nothing matches the url `/test?` nor `/test/?`

---

Aaaaand as I was writing this I realized it will never match because that's not part of the url
:man_facepalming:. I need to deal with it as query-params, like Anxo suggested.

Working on it...

alfetopito: Yay, now `/trade/TUSD-GUSD?sell=1.0&buy=1.0` works as expected

src/components/TradeWidget/index.tsx
```diff
@@ -56,6 +56,23 @@ const SubmitButton = styled.button`
   }
 `
 
+// This is sort of a duplication from the validation done on TokenRow.
+// Still, required because the initial value comes from the URL params,
+//   and we have no control over what the user inserts there.
+// Thus, we make it a nicely formated value, or 0 if we can't
+// And only on first render. Subsequent validation is taken care by TokenRow.
+function sanitizeInputAmount(value: string, precision: number): string {
```
anxolin: Would it make sense to be moved to our utils?

src/components/TradeWidget/index.tsx
```diff
@@ -56,6 +56,23 @@ const SubmitButton = styled.button`
...
+  const number = parseAmount(value, precision)
+  if (!number) {
+    return '0'
+  }
+  return formatAmountFull(number, precision, false)
+}
+
+function useQuery(): URLSearchParams {
```
anxolin: Clean and nice!

src/components/TradeWidget/index.tsx
```diff
@@ -64,17 +81,37 @@ const TradeWidget: React.FC = () => {
...
   const [receiveToken, setReceiveToken] = useState(
     () => getToken('symbol', receiveTokenSymbol, tokens) || getToken('symbol', 'USDC', tokens),
   )
+  const sellInputId = 'sellToken'
+  const receiveInputId = 'receiveToken'
 
-  // Change URL on internal token change
-  useURLParams(`sell=${sellToken.symbol}-0&receive=${receiveToken.symbol}-20`)
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const sanitizedSellAmount = useMemo(() => sanitizeInputAmount(sellAmount, sellToken.decimals), [])
```
anxolin: What do useMemo without dependencies? shouldn't this depend on the `sellAmount, sellToken.decimals`?
Also, is only for optimizing? Is it really worth it in this case?

src/components/TradeWidget/index.tsx
```diff
@@ -64,17 +81,37 @@ const TradeWidget: React.FC = () => {
...
+    mode: 'onBlur',
+    defaultValues: {
+      [sellInputId]: sanitizedSellAmount,
+      [receiveInputId]: sanitizedReceiveAmount,
+    },
+  })
+  const { handleSubmit, watch, reset } = methods
+
+  const urlParams = new URLSearchParams({ sell: watch(sellInputId), buy: watch(receiveInputId) })
+  const url = `/trade/${sellToken.symbol}-${receiveToken.symbol}?${urlParams.toString()}`
```
anxolin: nit: I might be wrong, but I would think that if you drop the `toString()`, it has the same effect

src/components/TradeWidget/index.tsx
alfetopito: Yes, for optimization. I had a comment explaining this but I guess during refactoring I got rid of it.

The point of this sanitization is to prevent invalid numbers to be inserted into the input, when coming from user input url. For example: `/trade/USDC-DAI?sell=kjasrhsker&buy=1.1.1`will be converted to `/trade/USDC-DAI?sell=0&buy=0`

Non-numeric values are prevented to be inserted in the input, but the values coming from the url are used as default values, thus bypassing validation.
For this reason, the sanitization is ran only once (the `[]` dependency). Subsequent updates will be validated by the form and we can trust the value won't contain weird things.

src/components/TradeWidget/index.tsx
alfetopito: Did a small refactoring, let me know if you still think this should be in utils.

src/components/TradeWidget/index.tsx
alfetopito: Yeah it works! Refactored.

PR #190, opened 2019-11-12
137/strict null check

### LINKED ISSUE #137: strictNullCheck for TS
We may need to turn `strictNullChecks` on in `tsconfig.json`
Otherwise it doesn't notify about a lot of vars that can and will break when `null | undefined`
With `strictNullChecks: true`  Ts would require a typeguard
A couple places where it bit me: 

```js
// EtherscanLink.tsx:L15
Network[networkId].toLowerCase()

// many provider checks in WalletApiImpl.tsx

```

But just casually turning it on reveals many TS error, so need a PR to deal with them

**Estimation**: 1

W3stside:
Closes #137 

May need several revisions, not sure if you guys like the guarding I did.

`Null-Check`


src/components/DepositWidget/useRowActions.tsx
```diff
@@ -65,8 +66,13 @@ export const useRowActions = (params: Params): Result => {
   }
 
   async function enableToken(tokenAddress: string): Promise<void> {
-    const { symbol } = getToken('address', tokenAddress, balances)
+    const { symbol } = getToken('address', tokenAddress, balances) as Required<TokenBalanceDetails>
```
alfetopito: If I understood the docs (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#improved-control-over-mapped-type-modifiers), `Required<TokenBalanceDetails>` removes the `?` properties.
Does that mean things like `symbol`, `name` and `decimals` will be stripped away?

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -65,8 +66,13 @@ export const useRowActions = (params: Params): Result => {
...
+
     try {
+      if (!userAddress) {
+        throw new Error('No logged in user found. Please check wallet connectivity status and try again.')
+      }
```
alfetopito: Could this be extracted as a function?

src/components/TradeWidget/index.tsx
```diff
@@ -64,22 +65,40 @@ const TradeWidget: React.FC = () => {
   const { sell: sellTokenSymbol, receive: receiveTokenSymbol } = useParams()
 
   const [sellToken, setSellToken] = useState(
-    () => getToken('symbol', sellTokenSymbol, tokens) || getToken('symbol', 'DAI', tokens),
+    () => getToken('symbol', sellTokenSymbol, tokens) || (getToken('symbol', 'DAI', tokens) as Required<TokenDetails>),
```
alfetopito: Shouldn't the cast be applied to the first expression as well?

src/utils/miscellaneous.ts
```diff
@@ -5,8 +5,8 @@ function noop(..._args: any[]): void {}
 
 export const log = process.env.NODE_ENV === 'test' ? noop : console.log
 
-export function getToken<T extends TokenDetails, K extends keyof T>(
-  key: K,
+export function getToken<T extends TokenDetails>(
+  key: string,
```
alfetopito: Why change this back?

src/components/TradeWidget/index.tsx
W3stside: no because the first may not be there - the second, DAI or USDC, is

src/utils/miscellaneous.ts
W3stside: when is the key of TokenDetails not a string? it made the line below, `value.toString()` hard to type

src/components/DepositWidget/useRowActions.tsx
W3stside: it's saying that the returned `TokenBalanceDetails` is the entire object with all `?` properties filled. Meaning the rest of the code relying on say, `sellTokenBalance.symbol` will never be `undefined`

src/components/DepositWidget/useRowActions.tsx
W3stside: tried this before changing it back and pushing as I couldn't figure out how to type the function out - it was:

```
function checkUserAddress(userAddress?: string): void {
  if (!userAddress) {
        throw new Error('No logged in user found. Please check wallet connectivity status and try 
        again.')
      }
}
```

but the code following line 74 was throwing issues with `userAddress` still - I should just write a typeguard function instead. ill take a look

src/components/DepositWidget/useRowActions.tsx
Velenir: `Required<>` makes optional fields required, hence the name

src/components/DepositWidget/useRowActions.tsx
Velenir: type [never](https://www.typescriptlang.org/play/?target=99&ssl=12&ssc=1&pln=12&pc=19#code/FAMwrgdgxgLglgewgAigCwKZQNYFUDOGATgIIAmZRG++AFAJQBcyEGAbsaJLIiulnkKkKVGrTBDylavmb4YROBADmTZGwRwyXaPCSpMOAsSmi6EkyJkB+OQqWrmGrcgA+LdsWQBvYMmRwIMi0AIQWwtI09D5+-nH+MGhECADuHmkAokTJRLQA5AByCMgANgjKyhhkASjhyCAIkGQAdMgACiUYAIaEBgLIKV0lnTCoSKw8bHAwAJ7I8l0wEshdENUKc13KXUrNefSxAL7Ax8BQSPLIian4AErUYCWjALx9RpJWYgfnEJdsQ1p7vhHi83oJLJE6HkugAjKD7YBAA)

src/utils/miscellaneous.ts
alfetopito: The change was not because the key isn't string, but because the typing enforces the parameters will always be existing keys in TokenDetails. See for reference https://mariusschulz.com/blog/keyof-and-lookup-types-in-typescript#indexed-access-types. This is where I found out about it.

src/components/DepositWidget/useRowActions.tsx
W3stside: this wasn't working - still wasn't able to get TS to accept the `userAddress` const after the function call - so leaving as is for now

gnosis-info:
Travis automatic deployment:
 https://pr190--dexreact.review.gnosisdev.com

src/utils/miscellaneous.ts
```diff
@@ -12,9 +12,9 @@ export function getToken<T extends TokenDetails, K extends keyof T>(
 ): T | undefined {
   const valueUppercase = value.toUpperCase()
   return (tokens || []).find((token: T): boolean => {
-    const value = token[key]
-    if (value) {
-      return value.toString().toUpperCase() === valueUppercase
+    const tokenKeyValue = token[key]
+    if (tokenKeyValue) {
+      return typeof tokenKeyValue === 'string' && tokenKeyValue.toString().toUpperCase() === valueUppercase
```
alfetopito: Can you do the type guard in the if?
In that case I think the `toString()` is not needed, right?

```typescript
    if (tokenKeyValue && typeof tokenKeyValue === 'string') {
￼      return tokenKeyValue.toUpperCase() === valueUppercase
```

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -55,8 +58,13 @@ export const useRowActions = (params: Params): Result => {
   }
 
   async function enableToken(tokenAddress: string): Promise<void> {
-    const { symbol } = getToken('address', tokenAddress, balances)
+    const { symbol } = getToken('address', tokenAddress, balances) as Required<TokenBalanceDetails>
```
anxolin: ubernit: I think it's good when we add a `Required` to include a comment on why we know it contains the properties. Also, I think in that case, it was used as a shortcut for saying that it has a symbol for sure.

Important: Why we assume here that it has a symbol? I think this might not be true.

I think I understand why you did it, but in this case, the symbol is used only for giving some feedback to the user. In this case I would do:

```js
const { symbol } = getToken('address', tokenAddress, balances)
const label = symbol || tokenAddress
toast.success(`The token ${label} has been enabled for trading`)
```

Also, if we see this pattern repeated 10 times in the app, we might consider to add this `label` param as mandatory, and we compute it when we load the tokens. This way, the children don't need to decide.

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -113,8 +125,12 @@ export const useRowActions = (params: Params): Result => {
   }
 
   async function requestWithdraw(amount: BN, tokenAddress: string): Promise<void> {
-    const { symbol, decimals } = getToken('address', tokenAddress, balances)
+    const { symbol, decimals } = getToken('address', tokenAddress, balances) as Required<TokenBalanceDetails>
```
anxolin: I would reconsider all the `Required` modifiers

src/hooks/useEnableToken.ts
```diff
@@ -29,29 +29,40 @@ export const useEnableTokens = (params: Params): Result => {
...
-    return result
+      if (mounted.current) {
+        // Update the state
+        setEnabled(true)
+        setEnabling(false)
+      }
+
+      return result
+    } catch (error) {
+      console.error(error)
```
anxolin: We cannot hide errors, catch and log an error is almost never the right thing to do

src/hooks/useEnableToken.ts
```diff
@@ -29,29 +29,40 @@ export const useEnableTokens = (params: Params): Result => {
     }
   }, [])
 
-  async function enableToken(): Promise<TxResult<boolean>> {
+  async function enableToken(): Promise<TxResult<boolean> | boolean> {
```
anxolin: Why this complex signature, either is a reject, or resolves giving you the receipt (well the receipt is in another branch out there, here is a TxResult)

src/hooks/useTokenBalances.ts
```diff
@@ -37,6 +37,7 @@ async function fetchBalancesForToken(
 
   return {
     ...token,
+    decimals: token.decimals || 18,
```
anxolin: Use the constant you created in this PR

src/hooks/useTokenBalances.ts
```diff
@@ -49,11 +50,11 @@ async function fetchBalancesForToken(
   }
 }
 
-async function _getBalances(walletInfo: WalletInfo): Promise<TokenBalanceDetails[] | null> {
+async function _getBalances(walletInfo: WalletInfo): Promise<TokenBalanceDetails[]> {
   const { userAddress, networkId } = walletInfo
   log('[useTokenBalances] getBalances for %s in network %s', userAddress, networkId)
   if (!userAddress || !networkId) {
-    return null
+    return []
```
anxolin: Are we sure we want to replace `null` with an empty string?

`null` meant, I don't know
`[]` I know but there aren't any tokens

Maybe in this case it's not too relevant, I would be OK leaving it like this, but just want to hear your thoughts

src/hooks/useWindowSizes.ts
```diff
@@ -8,7 +8,7 @@ function getSize(): { innerWidth: number | undefined; innerHeight: number | unde
   }
 }
 
-const useWindowSizes = (): { innerWidth: number; innerHeight: number } => {
+const useWindowSizes = (): { innerWidth: number | undefined; innerHeight: number | undefined } => {
```
anxolin: What's the difference with?
```js
 { innerWidth?: number; innerHeight?: number }
```

src/types/index.ts
```diff
@@ -15,8 +15,8 @@ export enum Network {
 
 export interface TokenDetails {
   name?: string
-  symbol?: string
-  decimals?: number
+  symbol: string
```
anxolin: Why is symbol required? 
I would understand `decimals` if we want to assume `18` as the default.

Also, in another change, you added `Required` to `TokenDetails` and then only extracted the `symbol`, so also not coherent with this change.

src/utils/mock.ts
```diff
@@ -9,7 +9,9 @@ export async function waitAndSendReceipt(params: {
 }): Promise<void> {
   const { waitTime = 2500, waitTimeForReceipt = 1000, txOptionalParams } = params
   if (txOptionalParams && txOptionalParams.onSentTransaction) {
-    wait(waitTimeForReceipt).then(() => txOptionalParams.onSentTransaction(RECEIPT))
+    wait(waitTimeForReceipt).then(
+      () => txOptionalParams.onSentTransaction && txOptionalParams.onSentTransaction(RECEIPT),
```
anxolin: Doesn't matter that much, but the conditional part should be done before:

```js
if (txOptionalParams.onSentTransaction) {
 wait(...)
}

File: test/components/EtherscanLink.components.test.tsx [hunk 15]
```diff
@@ -30,6 +30,7 @@ describe('<EtherscanLink /> general', () => {
   })
 
   it('does not render when networkId is missing', () => {
+    // @ts-ignore
```
anxolin: Why is this required?

src/components/DepositWidget/useRowActions.tsx
alfetopito: On that note, I created this util function awhile ago for exactly this use case: https://github.com/gnosis/dex-react/blob/2ae24e9dab20646d583796bead91664fa4cf849a/src/utils/format.ts#L133

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -95,7 +95,7 @@ export class Erc20ApiImpl implements Erc20Api {
       from: fromAddress,
     })
 
-    if (txOptionalParams) {
+    if (txOptionalParams && txOptionalParams.onSentTransaction) {
```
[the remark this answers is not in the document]
W3stside: yeah i had to update and change things back cause it wasnt working but thats the idea.

src/api/tokenList/tokenList.ts
```diff
@@ -1,23 +1,24 @@
...
     if (address) {
       // There's an address for the current network
-      const { name, symbol, decimals } = token
+      const { name, symbol, decimals = DEFAULT_PRECISION } = token
       const addressMainnet = token.addressByNetwork[Network.Mainnet]
 
-      return { name, symbol, decimals, address, addressMainnet }
-    } else {
-      return null
+      return [...acc, { name, symbol, decimals, address, addressMainnet }]
```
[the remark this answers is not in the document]
W3stside: how so?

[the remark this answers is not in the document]
W3stside: anyways changed to this but curious on as to why

src/api/wallet/WalletApiImpl.ts
```diff
@@ -120,7 +120,7 @@ const closeOpenWebSocketConnection = (web3: Web3): void => {
   if (
     web3 &&
     typeof web3.currentProvider === 'object' &&
-    web3.currentProvider.connected &&
+    web3.currentProvider?.connected &&
     'disconnect' in web3.currentProvider
```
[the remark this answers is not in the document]
W3stside: yeah it hates it, normally that would work

test/components/EtherscanLink.components.test.tsx
W3stside: does this test even use `networkId` doesn't make sense to me looking at it. NetworkId is resolved in `useWalletConnection` and changing it there shouldnt do anything

src/hooks/useTokenBalances.ts
W3stside: It may be best, let me check the implementation again

W3stside: should it throw?

W3stside: one approach is to assert every important required param (`userAddress`, `contractAddress`) etc and handle life this way. Thoughts? @Velenir @alfetopito @anxolin

W3stside: the other approach, as was tackled in dx-react for example, was not blocking UI and rendering empty data sets (`[]` in this approach which renders the screen but with no usable params (`userADdress` etc) and then presenting an error somewhere to read. This could be approached with the `toast` things or a banner when a certain disconnect happens (like `dx-react`)

src/utils/miscellaneous.ts
```diff
@@ -14,7 +27,7 @@ export function getToken<T extends TokenDetails, K extends keyof T>(
   return (tokens || []).find((token: T): boolean => {
     const tokenKeyValue = token[key]
     if (tokenKeyValue) {
-      return typeof tokenKeyValue === 'string' && tokenKeyValue.toString().toUpperCase() === valueUppercase
+      return tokenKeyValue?.['toString']().toUpperCase() === valueUppercase
```
Velenir: Not `tokenKeyValue?.toString().toUpperCase()`?

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -47,11 +47,15 @@ export const useRowActions = (params: Params): Result => {
...
-
     try {
-      assert(contractAddress, 'Contract address missing. Aborting.')
       assert(userAddress, 'User address missing. Aborting.')
+      assert(contractAddress, 'Contract address missing. Aborting.')
+
+      const token = getToken('address', tokenAddress, balances)
+      assert(token, 'No token, aborting.')
+
+      const { symbol: tokenDisplayName } = safeFilledToken(token)
```
Velenir: Btw, completely unnecessary here as you employ `safeFilledToken` for that, but
a nifty trick to get `tokenDisplayName = name || symbol || address` out of a token is
```ts
const {
  name: tokenDisplayName,
  symbol: tokenDisplayName = tokenDisplayName,
  address: tokenDisplayName = tokenDisplayName
} = token
```

Cool?

src/utils/miscellaneous.ts
W3stside: if it's `toString` pass it auto applies `toUpperCase` - i tested it but maybe u right

src/components/DepositWidget/useRowActions.tsx
W3stside: hmmm super cool, what the heck

src/utils/miscellaneous.ts
W3stside: ah i see ur comment, i tried the former but it broke all the things. This was the only way I could get it to work and my google foo energy was low

src/components/DepositWidget/useRowActions.tsx
W3stside:
```
let token = {
	name: undefined,
	symbol: undefined,
	address: '0x12312312312',
}

const {
  name: tokenDisplayName,
  symbol: tokenDisplayName = tokenDisplayName,
  address: tokenDisplayName = tokenDisplayName
} = token
VM313:9 Uncaught SyntaxError: Identifier 'tokenDisplayName' has already been declared
```

didnt work para mi :(

src/utils/miscellaneous.ts
Velenir: but, but they are [the same](https://www.typescriptlang.org/play/?ssl=5&ssc=25&pln=1&pc=1#code/LAKFBcHsGsFMDsDSsCeA1AhgGwK6wPwB0UAyuAE4CW8A5gBQCUxkAqgA5uzkDCGAzrEagIMBMnTY8RANoByUhWo1ZAXUbN2nHv0EMgA)

weird

W3stside: yep i know
/shrug

W3stside: resolving even tho it sucks

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -109,7 +109,7 @@ export class Erc20ApiImpl implements Erc20Api {
...
```
alfetopito: Isn't this the same as line 67 and 89? Can't you just do the same here? 

```suggestion
    if (txOptionalParams?.onSentTransaction) {
```

src/components/DepositWidget/useRowActions.tsx
alfetopito: May I suggest safeTokenName from https://github.com/gnosis/dex-react/blob/develop/src/utils/format.ts#L133 ?

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -45,8 +47,16 @@ export const useRowActions = (params: Params): Result => {
...
-    const { symbol } = getToken('address', tokenAddress, balances)
     try {
+      assert(userAddress, 'User address missing. Aborting.')
+      assert(contractAddress, 'Contract address missing. Aborting.')
+
+      const token = getToken('address', tokenAddress, balances)
+      assert(token, 'No token, aborting.')
+
+      const { symbol: tokenDisplayName } = safeFilledToken(token)
+      assert(tokenDisplayName, 'No token name found - symbol, name, or address. Aborting.')
```
alfetopito: Address is required and should never fail. I mean, should fail sooner in case token has no address...

src/components/layout/Widget.tsx
```diff
@@ -41,11 +41,15 @@ interface Props {
...
   const { networkId } = useWalletConnection()
-  const contractAddress = depositApi.getContractAddress(networkId)
+  const contractAddress = networkId ? depositApi.getContractAddress(networkId) : null
 
   return (
     <Wrapper className={className}>
-      <LinkWrapper type="contract" identifier={contractAddress} label={<small>View verified contract</small>} />
+      <LinkWrapper
+        type="contract"
+        identifier={contractAddress || 'No address detected'}
```
alfetopito: In case there's no contract, why render this component at all?

src/utils/format.ts
```diff
@@ -133,3 +129,13 @@ export function abbreviateString(inputString: string, prefixLength: number, suff
 export function safeTokenName(token: TokenDetails): string {
   return token.symbol || token.name || abbreviateString(token.address, 6, 4)
 }
+
+export function safeFilledToken<T extends TokenDetails>(token: T): T {
+  // if (!token) return null
+
+  return {
+    ...token,
+    name: token.name || token.symbol || abbreviateString(token.address, 6, 4),
```
alfetopito: === line 130 above. Reuse/refactor?

src/hooks/useTokenBalances.ts
anxolin: Imo, not having a network or account is not an exception, it's a normal case, where the user doesn't have the wallet connected, this is why I think it's fine to signal that returning a `null` (would say, I don't know).

Note that the alternative of throwing wouldn't be good for a hook, since if you throw you are moving the responsibility to the caller the hook, and hooks cannot be conditionally rendered (one of their core restrictions)

In dx-react I think we weren't handing very well this states, this is why sometimes you could see balance 0 for some seconds, and then changing to something else (confusing for an user)

src/components/DepositWidget/useRowActions.tsx
anxolin: Agree with Leandro. Looks to me better handled, easier to read, and we don't copy paste this logic around.

anxolin:
Regarding: https://github.com/gnosis/dex-react/pull/190#discussion_r349002900
> anyways changed to this but curious on as to why

Dima says this cause you are reducing an object, and in every iteration you are creating it again and coping the elements, where in the other approach, is reusing the same array. This array starts as an empty one, and we then conditionally `push` elements into it.

src/api/erc20/Erc20ApiImpl.ts
W3stside: missed it

src/components/DepositWidget/useRowActions.tsx
Velenir: > didnt work para mi :(

src/components/DepositWidget/useRowActions.tsx
W3stside: yes true

src/components/layout/Widget.tsx
W3stside: :+1:

src/utils/format.ts
W3stside: i think it's fine like this as the name and symbol have different logic

src/components/DepositWidget/useRowActions.tsx
W3stside: I prefer my implementation as I get both the name and symbol back with the token object and I don't need to do an extra param and making it one bigger complicated function.

Velenir:
```js
let tokenDisplayName;
{
  name: tokenDisplayName,
  symbol: tokenDisplayName = tokenDisplayName,
  address: tokenDisplayName = tokenDisplayName
} = token
```

Issue #193, opened 2019-11-13
Switch contract API methods to accept Object arguments


W3stside:
Seen an issue in `erc20API.approve(senderAddr, tokenAddr, spenderAddr)` where the func params were mixed up leaving a hard-to-debug issue.

I propose switching to Object parameters i.e

`erc20API.approve({ senderAddr: string, tokenAddr: string, spenderAddr: string })`

W3stside:
Would prevent hard to debug code errors

bh2smith:
I like this idea and would be in favour of such an improvement!

anxolin:
Good point.

Yes, I agree. We had a chat the other day. The object-like params approach would be applicable for any non-trivial util/lib/api since although typescript helps a bit to check and document the params, if they have the same type, it would allow to switch the positions messing it up and as David says, making it very difficult to debug.

PR #194, opened 2019-11-13
switched improper arg placement in enableToken

W3stside:
Fixes `enableToken` bug

But on a bigger note, we _really_ should change the API functions to accept `Object args` - see issue #193 to avoid issues as these.


gnosis-info:
Travis automatic deployment:
 https://pr194--dexreact.review.gnosisdev.com

PR #196, opened 2019-11-13
155/mount check refactor

### LINKED ISSUE #155: Refactor for mounted check
Choose one or offer a different way from #132

**Estimation**: 1

W3stside:
Closes #155 

Switches `useState` with `useSafeState` where applicable

- [ ] Tests

### DIFF
--- a/src/components/DepositWidget/Form.tsx
+++ b/src/components/DepositWidget/Form.tsx
@@ -1,9 +1,11 @@
-import React, { useState, useEffect, ChangeEvent, ReactNode, useRef } from 'react'
+import React, { useEffect, ChangeEvent, ReactNode } from 'react'
 import BN from 'bn.js'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { IconDefinition, faSpinner } from '@fortawesome/free-solid-svg-icons'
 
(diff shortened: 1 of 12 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr196--dexreact.review.gnosisdev.com

src/components/DepositWidget/Form.tsx
```diff
@@ -85,11 +78,9 @@ export const Form: React.FC<FormProps> = (props: FormProps) => {
...
       setLoading(true)
       props.onSubmit(parseAmount(amountInput, decimals)).then(() => {
-        if (mounted.current) {
-          setLoading(false)
-          setValidatorActive(false)
-          cancelForm()
-        }
+        setLoading(false)
+        setValidatorActive(false)
+        cancelForm()
```
alfetopito: The new hook doesn't cover the side effects of this function, or does it?
`props.onClose()` will now be called even when unmounted.

Velenir: Yes, but I think in this case it's ok, since it triggers `setForm` in `Row` . It will only be a problem if a different form was opened and would force close as a result.

Maybe bring in also the `useIfMounted` hook for such cases.

PR #214, opened 2019-11-18
99/real exchange api

Velenir:
ExchangeApi implementation with web3

Currently won't work because of:

- ~Only WETH returns a valid tokenID (we may have stale token addresses)~
- web3 Contract methods encode abi with BN incorrectly (will create an issue at web3 tomorrow)
workaround in #221 
- ~can't query dynamic arrays from Contract (will try to use [decoding func from dex-contracts](https://github.com/gnosis/dex-contracts/blob/4f1183d3b583cc44741f50bf85af00263e108c7d/test/utilities.js#L177-L202))~

### DIFF
--- a/src/api/exchange/DepositApiImpl.ts
+++ b/src/api/exchange/DepositApiImpl.ts
@@ -175,7 +175,7 @@ export class DepositApiImpl implements DepositApi {
   /********************************    private methods   ********************************/
 
   protected async _getNetworkId(): Promise<number> {
-    let networkId = getNetworkIdFromWeb3(this._web3)
+    const networkId = getNetworkIdFromWeb3(this._web3)
 
     return networkId === null ? this._web3.eth.net.getId() : networkId
   }
--- a/src/api/exchange/DepositApiImpl.ts
+++ b/src/api/exchange/DepositApiImpl.ts
@@ -25,7 +25,7 @@ export class DepositApiImpl implements DepositApi {
   private _ReferenceEpochTokenLocker: EpochTokenLocker
   private _web3: Web3
 
-  private static _address2ContractCache: { [K: string]: EpochTokenLocker } = {}
+  protected static _address2ContractCache: { [K: string]: EpochTokenLocker } = {}
 
   public constructor(web3: Web3) {
     this._ReferenceEpochTokenLocker = new web3.eth.Contract(EpochTokenLockerAbi)
@@ -178,14 +178,15 @@ export class DepositApiImpl implements DepositApi {
     let networkId = getNetworkIdFromWeb3(this._web3)
 
     return networkId === null ? this._web3.eth.net.getId() : networkId
-    }
+  }
 
+  protected async _getContract(): Promise<EpochTokenLocker> {
     const networkId = await this._getNetworkId()
 
     return this._getContractForNetwork(networkId)
   }
 
-  private _getContractForNetwork(networkId: number): EpochTokenLocker {
+  protected _getContractForNetwork(networkId: number): EpochTokenLocker {
     const address = this.getContractAddress(networkId)
 
     assert(address, `EpochTokenLocker was not deployed to network ${networkId}`)
@@ -193,7 +194,7 @@ export class DepositApiImpl implements DepositApi {
     return this._getContractAtAddress(address)
   }
 
-  private _getContractAtAddress(address: string): EpochTokenLocker {
+  protected _getContractAtAddress(address: string): EpochTokenLocker {
     const contract = DepositApiImpl._address2ContractCache[address]
 
     if (contract) return contract
--- a/src/api/exchange/DepositApiImpl.ts
+++ b/src/api/exchange/DepositApiImpl.ts
@@ -174,12 +174,13 @@ export class DepositApiImpl implements DepositApi {
 
   /********************************    private methods   ********************************/
 
-  private async _getContract(): Promise<EpochTokenLocker> {
+  protected async _getNetworkId(): Promise<number> {
     let networkId = getNetworkIdFromWeb3(this._web3)
 
-    if (networkId === null) {
-      networkId = await this._web3.eth.net.getId()
+    return networkId === null ? this._web3.eth.net.getId() : networkId
     }
+
+    const networkId = await this._getNetworkId()
 
     return this._getContractForNetwork(networkId)
   }
--- a/src/api/exchange/DepositApiImpl.ts
+++ b/src/api/exchange/DepositApiImpl.ts
@@ -138,7 +138,7 @@ export class DepositApiImpl implements DepositApi {
       tx.once('transactionHash', txOptionalParams.onSentTransaction)
     }
 
-    log(`[DepositApiImpl] Deposited ${amount} for token ${tokenAddress}. User ${userAddress}`)
+    log(`[DepositApiImpl] Deposited ${amount.toString()} for token ${tokenAddress}. User ${userAddress}`)
     return tx
   }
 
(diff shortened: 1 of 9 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr214--dexreact.review.gnosisdev.com

anxolin:
Thanks for pointing out the issues and uploading the branch early!
We can review the problems tomorrow. I think the critical thing we need asap is the submit order, would it help splitting the PR in 2, where in first one we only implement the MINIMAL required functions? 

Also, related to the API, is not immutable :) 
We made one that would allow us to parallelize things and work in the widgets already, but if something doesn't feel natural, it's super possible to discuss it and modify it (same as what it was done with the receipt confirmation)

src/api/exchange/DepositApiImpl.ts
```diff
@@ -155,25 +155,27 @@ export class DepositApiImpl implements DepositApi {
...
   /********************************    private methods   ********************************/
 
-  private async _getContract(): Promise<EpochTokenLocker> {
+  protected async _getNetworkId(): Promise<number> {
     let networkId = getNetworkIdFromWeb3(this._web3)
 
-    if (networkId === null) {
-      networkId = await this._web3.eth.net.getId()
-    }
+    return networkId === null ? this._web3.eth.net.getId() : networkId
```
alfetopito: Don't you need to await `this._web3.eth.net.getId()`?

src/api/exchange/DepositApiImpl.ts
```diff
@@ -155,25 +155,27 @@ export class DepositApiImpl implements DepositApi {
 
   /********************************    private methods   ********************************/
 
-  private async _getContract(): Promise<EpochTokenLocker> {
+  protected async _getNetworkId(): Promise<number> {
     let networkId = getNetworkIdFromWeb3(this._web3)
```
alfetopito:
```suggestion
    const networkId = getNetworkIdFromWeb3(this._web3)
```

src/types/index.ts
```diff
@@ -243,3 +243,160 @@ export interface ERC20 extends Contract {
...
+
+    getBalance(user: string, token: string): TransactionObject<string, [string, string]>
+
+    getCurrentBatchId(): TransactionObject<string>
+
+    requestFutureWithdraw(
+      token: string,
+      amount: number | string | BN,
+      batchId: number | string | BN,
+    ): TransactionObject<void, [string, number | string | BN, number | string | BN]>
```
alfetopito: Shouldn't this be in the EpochTokenLocker contract definition?

src/types/index.ts
```diff
@@ -243,3 +243,160 @@ export interface ERC20 extends Contract {
...
+
+    hasValidWithdrawRequest(user: string, token: string): TransactionObject<boolean, [string, string]>
+
+    MAX_TOKENS(): TransactionObject<string>
+
+    getPendingDepositBatchNumber(user: string, token: string): TransactionObject<string, [string, string]>
+
+    withdraw(user: string, token: string): TransactionObject<void, [string, string]>
```
alfetopito: Shouldn't this be in the EpochTokenLocker contract definition as well?

src/types/index.ts
```diff
@@ -243,3 +243,160 @@ export interface ERC20 extends Contract {
...
+      [
+        number | string | BN,
+        number | string | BN,
+        string[],
+        (number | string | BN)[],
+        (number | string | BN)[],
+        (number | string | BN)[],
+        (number | string | BN)[],
+      ]
+    >
```
alfetopito: Do we need to have these methods mapped in the fronted?

src/api/exchange/DepositApiImpl.ts
W3stside: no, `async` function return promises, so you don't return an `await` you just `return`

src/types/index.ts
Velenir: you are right, typechain confused me with which methods are where
I'll sort them out in a separate PR as it doesn't break anything here

src/types/index.ts
Velenir: They are only mapped as far as types go. So that we know we have them on `Contract.methods`

gnosis-info:
Travis automatic deployment:
 https://pr214--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApiImpl.ts
```diff
@@ -1,12 +1,86 @@
...
+  UINT16_WIDTH,
+)})(?<sellTokenId>${hn(UINT16_WIDTH)})(?<validFrom>${hn(UINT32_WIDTH)})(?<validUntil>${hn(
+  UINT32_WIDTH,
+)})(?<priceNumerator>${hn(UINT128_WIDTH)})(?<priceDenominator>${hn(UINT128_WIDTH)})(?<remainingAmount>${hn(
+  UINT128_WIDTH,
+)})`
+
+// decodes Orders for a userAddress if passed one
+// otherwise decodes all Orders for all users
+const decodeAuctionElements = (bytes: string, userAddress?: string): AuctionElement[] => {
```
alfetopito: Hopefully we won't need this :crossed_fingers:

W3stside:
Should I wait before reviewing this?

Velenir:
@W3stside go ahead, review it

src/api/exchange/ExchangeApiImpl.ts
```diff
@@ -0,0 +1,221 @@
...
```
anxolin: Can we move this to the utils?

eventually we are also moving it back to a common lib. By the way, they already had something for this in dex-contracts

src/api/exchange/DepositApiImpl.ts
```diff
@@ -155,25 +155,27 @@ export class DepositApiImpl implements DepositApi {
...
+  }
+
+  protected async _getContract(): Promise<EpochTokenLocker> {
+    const networkId = await this._getNetworkId()
 
     return this._getContractForNetwork(networkId)
   }
 
-  private _getContractForNetwork(networkId: number): EpochTokenLocker {
+  protected _getContractForNetwork(networkId: number): EpochTokenLocker {
```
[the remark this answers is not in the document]
Velenir: Private can be accessed from class itself only, protected -- from class itself and children. Need protected here

src/api/exchange/ExchangeApiImpl.ts
Velenir: Eventually I hope we get `getOrdersForUser` on the contract and this will go away completely.
Also this function is only used in this file.
So let it stay, and then we get rid of it entirely.

src/api/exchange/ExchangeApiImpl.ts
```diff
@@ -0,0 +1,221 @@
...
+const UINT256_WIDTH = 32 * 2
+const UINT16_WIDTH = 2 * 2
+const UINT32_WIDTH = 4 * 2
+const UINT128_WIDTH = 16 * 2
+
+const hexPattern = '[0-9a-fA-F]'
+
+const hn = (n: number): string => hexPattern + `{${n}}`
+
+// /(address)(sellTokenBalance)(buyTokenId)(sellTokenId)(validFrom)(validUntil)(priceNumerator)(priceDenominator)(remainingAmount)/g
```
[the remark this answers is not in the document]
Velenir: :arrow_up: :arrow_up: :arrow_down: :arrow_down: :arrow_left: :arrow_right: :arrow_left: :arrow_right: :b: :a:

src/api/exchange/ExchangeApiImpl.ts
anxolin: Okay, although I believe it's possible that we include it in the common js libs we want to do

anxolin:
@Velenir I rebased with develop, then fixed a small issue after getting the new changes (changed the cancelOrder signature). I'll merge now

PR #219, opened 2019-11-19
Enhancement/ts assertion

W3stside:
TS Friendly `assert`

depends on #190

### DIFF
--- a/src/api/exchange/DepositApiMock.ts
+++ b/src/api/exchange/DepositApiMock.ts
@@ -94,7 +94,7 @@ export class DepositApiMock implements DepositApi {
 
     // Create the balance state if it's the first deposit
     const currentBatchId = await this.getCurrentBatchId()
-    let balanceState = this._initBalanceState(userAddress, tokenAddress, currentBatchId)
+    const balanceState = this._initBalanceState(userAddress, tokenAddress, currentBatchId)
 
     // Update any unapplied deposit
     this._updateDepositsBalance(userAddress, tokenAddress, currentBatchId)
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr219--dexreact.review.gnosisdev.com

src/utils/miscellaneous.ts
```diff
@@ -1,4 +1,17 @@
 import { TokenDetails } from 'types'
+import { AssertionError } from 'assert'
+
+export function assert<T>(val: T, message: string): asserts val is NonNullable<T> {
```
Velenir: Consider dropping this function when DefinitelyTyped/DefinitelyTyped#40415 is fixed

Velenir: This actually asserts not only that val is NonNullable, but that it is truthy. No idea if it's possible to type that.

src/utils/miscellaneous.ts
```diff
@@ -1,4 +1,17 @@
...
+  if (!val || val === undefined || val === null) {
```
Velenir: `!val` is enough to check for `null and undefined`

src/api/exchange/DepositApiMock.ts
```diff
@@ -94,7 +94,7 @@ export class DepositApiMock implements DepositApi {
 
     // Create the balance state if it's the first deposit
     const currentBatchId = await this.getCurrentBatchId()
-    let balanceState = this._initBalanceState(userAddress, tokenAddress, currentBatchId)
+    const balanceState = this._initBalanceState(userAddress, tokenAddress, currentBatchId)
```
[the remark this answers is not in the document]
W3stside: think that was the linter that updated

PR #222, opened 2019-11-20
Reducing logging

alfetopito:
As title states

### DIFF
--- a/src/api/wallet/WalletApiImpl.ts
+++ b/src/api/wallet/WalletApiImpl.ts
@@ -248,7 +248,6 @@ export class WalletApiImpl implements WalletApi {
   private async _notifyListeners(): Promise<void> {
     await Promise.resolve()
     const walletInfo: WalletInfo = this._getWalletInfo()
-    console.table(walletInfo)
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr222--dexreact.review.gnosisdev.com

src/api/wallet/WalletApiImpl.ts
```diff
@@ -248,7 +248,6 @@ export class WalletApiImpl implements WalletApi {
   private async _notifyListeners(): Promise<void> {
     await Promise.resolve()
     const walletInfo: WalletInfo = this._getWalletInfo()
-    console.table(walletInfo)
```
Velenir: But, but we never get to use `console.table` :cry:

W3stside: @Velenir I always know when u find some new thing when it pops up EVERYWHERE lol.

alfetopito: I'm happy to keep it if we can manage to log it ONLY when it changes

PR #225, opened 2019-11-21
Fixed MD issues and updated Gemini faucet instructions

alfetopito:
As title states.

### DIFF
--- a/src/api/tokenList/Faucet_info.md
+++ b/src/api/tokenList/Faucet_info.md
@@ -1,51 +1,61 @@
 # Ether, Tether, TrueUSD, USD Coin, Paxos, Gemini, Dai
 
 ## Wrapped Ether
-* [https://weth.io](https://weth.io)
-* Wrapper of Ether to make it ERC-20 compliant
-* **Rinkeby faucet**: [https://faucet.rinkeby.io/](https://faucet.rinkeby.io)
 
-## Tether:
-* [https://tether.to](https://tether.to)
-* Fiat enabled collateralized stable coin, that is backed by the most popular fiat currency, USD (US Dollar) in a 1:1 ratio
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr225--dexreact.review.gnosisdev.com

alfetopito:
> I see most of the lines changed because of removing `*`. Made it very hard to see the differences.

Sorry. Should have split into two commits: 1 for linter changes and 1 for actual changes.

Still, you can diff the MD formatted version as well. Nice feature!

src/api/tokenList/Faucet_info.md
```diff
@@ -1,51 +1,61 @@
...
-* **Rinkeby faucet**: [https://faucet.rinkeby.io/](https://faucet.rinkeby.io)
 
-## Tether:
-* [https://tether.to](https://tether.to)
-* Fiat enabled collateralized stable coin, that is backed by the most popular fiat currency, USD (US Dollar) in a 1:1 ratio
-* **Rinkeby faucet**: contract created with Team's shared account 0xf85D1a0E1b71e72013Db51139f285C6d5356B712.
-Use it to mint new tokens
+- [https://weth.io](https://weth.io)
+- Wrapper of Ether to make it ERC-20 compliant
+- **Rinkeby faucet**: [https://faucet.rinkeby.io/](https://faucet.rinkeby.io)
```
Velenir: You don't need `[url](url)` when it's the same url. https://faucet.rinkeby.io/ would have been enough

Just an fyi, no need to change

bh2smith: Nice to know!

anxolin: True, although not all MD implementation support that. I think Github does, so I agree it should be enough in this case.

alfetopito: I only `fixed` it because VScode MD linter complained about it.

PR #227, opened 2019-11-21
226/keep url search params on page change

### LINKED ISSUE #226: [bug] Amounts are not kept on page change
### Reproducing

1. on trade page, select tokens and insert amounts on both fields
2. click on deposit
3. click on trade

tokens are kept, but amounts are not

alfetopito:
Closes #226 

---

Merged changes proposed by @Velenir on #236


gnosis-info:
Travis automatic deployment:
 https://pr227--dexreact.review.gnosisdev.com

src/components/LinkToDepositPage.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { useLocation } from 'react-router'
+import { Link } from 'react-router-dom'
+
+const LinkToDepositPage: React.FC = ({ children }) => {
+  const location = useLocation()
+  const to = {
```
Velenir: I'm not sure it's a good idea to create a whole component just for `ToDepositPage` use case.
Too specific, and we use it only a couple times.
If we really need a `LinkWithPastLocation` or something, that is another matter.
Also should pass [all the possible props along](https://reacttraining.com/react-router/web/api/Link/others)

```js
const LinkWithPastLoc = (props) => {
  const location = useLocation()
  const to = {...props.to, state : {...props.to?.state, from: location}}
  return <Link {...props} to={to}/>
}
```

Velenir: Can then use as:
```xml
<LinkWithPastLoc to="/deposit">Exchange wallet:</Link>
```

anxolin: I wouldn't shorten the words:  `LinkWithPastLocation` or even `LinkWithFromLocation`

alfetopito: I can't make TS happy with your proposal @Velenir.
I pushed some changes that looks kinda like that, but not quite.
It works, but doesn't look nice. Help?

gnosis-info:
Travis automatic deployment:
 https://pr227--dexreact.review.gnosisdev.com

src/components/LinkToDepositPage.tsx
Velenir: Have a look at #236

alfetopito: Thank you! Merged changes into this PR

PR #232, opened 2019-11-22
fix withdraw method params order

anxolin:
Invert the orders of params for withdraw to match a change in the SC


gnosis-info:
Travis automatic deployment:
 https://pr232--dexreact.review.gnosisdev.com

PR #237, opened 2019-11-25
[on hold]Show error if one token is not registered

anxolin:
Detect the case where a token is not registered in the exchange, and show the pertinent error

*update*:
I think this fix doesn't apply anymore. We should check for existence of the token. token 0 will be valid for fee token.

I'll leave it open, but needs to change  into: "Detect if the tokens are enabled". Also depends on #279

### DIFF
--- a/src/hooks/usePlaceOrder.ts
+++ b/src/hooks/usePlaceOrder.ts
@@ -47,23 +47,41 @@ export const usePlaceOrder = (): Result => {
           exchangeApi.getCurrentBatchId(),
         ])
 
-        const validUntil = batchId + DEFAULT_ORDER_DURATION
+        if (sellTokenId !== 0 || buyTokenId !== 0) {
+          log('sellTokenId, buyTokenId, batchId', sellTokenId, buyTokenId, batchId, sellToken.address, buyToken.address)
 
-        const params: ExchangeApiPlaceOrderParams = {
-          userAddress,
-          buyTokenId,
-          sellTokenId,
-          validUntil,
-          buyAmount,
-          sellAmount,
+          const validUntil = batchId + DEFAULT_ORDER_DURATION
+
+          const params: ExchangeApiPlaceOrderParams = {
+            userAddress,
+            buyTokenId,
+            sellTokenId,
+            validUntil,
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr237--dexreact.review.gnosisdev.com

src/hooks/usePlaceOrder.ts
```diff
@@ -47,23 +47,41 @@ export const usePlaceOrder = (): Result => {
           exchangeApi.getCurrentBatchId(),
         ])
 
-        const validUntil = batchId + DEFAULT_ORDER_DURATION
+        if (sellTokenId !== 0 || buyTokenId !== 0) {
```
W3stside: would this `0` be worth making into a const like `OWL_TOKEN_ID` or is it always zero?

W3stside: also a comment note would be useful coming back to it later we'll likely forget.

alfetopito: As we discussed offline, this code uses `idToAddressBiMap` lib which requires the id to exist https://github.com/gnosis/solidity-data-structures/blob/9d0c8c05b85ba6c28fa5717ca1ac9bdb60ee117d/contracts/libraries/IdToAddressBiMap.sol#L19

Thus, it should throw an error in that case, rather than return 0.

We need to check the contract you are testing against to figure out what tokes are registered.

anxolin: totaly, good point.

I this case, this 0 means, that is not a valid token. I need to review a bit why the contract is returning 0 for tokens that don't know, apparently it should be reverting according to the SC implementation, so sth weird worth reviewing is going on there

anxolin: Yep, but for some reason we are getting 0s there. We need to investigate a bit more why

anxolin:
I think this fix doesn't apply anymore. We should check for existance of the token. token 0 will be valid for fee token.

I'll leave it open, but needs to change  into: "Detect if the tokens are enabled". Also depends on #279

PR #238, opened 2019-11-25
Update contract types and ABI

Velenir:
* Contract types and abi updated to match the current state of contracts.
* Used **typechain** to generate Contract types and then modified them a little:

Q: Why `abstract class` instead of `interface`?
A: Less of a modification from typechain

Q: Why no inheriting `methods: EpochTokenLocker['methods'] & {` for `StablecoinConverter`
A: I don't want to manually pick what methods are inherited. Typechain generates every method together.

* `decodeAuctionElements` is simplified as we don't need to filer by `userAddress` any more

Left old contract types in `types/index.ts` for now, as they work well. Need to see if userflow works with the new abi first. Preliminary everything works though.


gnosis-info:
Travis automatic deployment:
 https://pr238--dexreact.review.gnosisdev.com

PR #239, opened 2019-11-25
Update smart contract

anxolin:


gnosis-info:
Travis automatic deployment:
 https://pr239--dexreact.review.gnosisdev.com

src/hooks/usePlaceOrder.ts
```diff
@@ -47,23 +47,41 @@ export const usePlaceOrder = (): Result => {
           exchangeApi.getCurrentBatchId(),
         ])
 
-        const validUntil = batchId + DEFAULT_ORDER_DURATION
+        if (sellTokenId !== 0 || buyTokenId !== 0) {
```
Velenir: shouldn't this be `&&`?

anxolin: totally! good catch!

alfetopito: Same comment I have https://github.com/gnosis/dex-react/pull/237 where this change is coming from.

anxolin: yep, I've merged that into this branch to use for the demo, I'll remove the commit from the other branch and leave this PR with just the address update

PR #240, opened 2019-11-25
fix wrong address on sell|buy token

Velenir:
This bug is again because TradeWidget saves tokens state locally and believes it to be the source of truth (not chain state), which comes from the fact we don't differentiate enough between local state (token list in this case)  and network-dependent token balances from the blockchain.

This is a quick fix for now.


gnosis-info:
Travis automatic deployment:
 https://pr240--dexreact.review.gnosisdev.com

PR #243, opened 2019-11-26
229/fix claim bug

### LINKED ISSUE #229: [bug] can claim multiple times
[claiming](https://github.com/gnosis/dex-react/blob/a8a2ae0765c3e9a97956fda6414ad5f6260f63e2/src/components/DepositWidget/Row.tsx#L78) prop becomes false soon after clicking it, before claim tx actually goes through
That allows to click it several times.

This happens because we don't separate on-chain state from local state very well when it comes to token balances:
 

1. `claiming` first comes from from [`useTokenBalance`](https://github.com/gnosis/dex-react/blob/10b1f16d7c0465c3955c628903794d26fb4ff329/src/hooks/useTokenBalances.ts#L51) with `false` by default
2.  then it's set to true in [`useRowActions`](https://github.com/gnosis/dex-react/blob/10b1f16d7c0465c3955c628903794d26fb4ff329/src/components/DepositWidget/useRowActions.tsx#L137) on `claim()`
3.  then on new block new balances come from `useTokenBalances` which override `claiming`

Ways to solve as I see them:

- don't update balances before tx is mined -- BAD, balances should be updated dynamically, irrespective of any tx
- don't keep `claiming` together with balances, even putting it super close to the metal, so to say, would work, in `<Row/>`
```ts
// Row.tsx::L78
<RowClaimButton
  className="success"
  onClick={() => onClaim.then(toggleLocalClaiming, toggleLocalClaiming)}
  disabled={claiming}
>
```

Disadvantage of the latter way, though I believe it's the right way, -- rewriting state logic for `DepositWidget, Row, useTokenBalances, useRowActions, who knows what else`

Thoughts?

W3stside:
Closes #229 

Uses `useReducer` inside `useRowActions` to keep token state locally


gnosis-info:
Travis automatic deployment:
 https://pr243--dexreact.review.gnosisdev.com

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -16,13 +17,71 @@ interface Params {
...
+const reducer = (state: TokenLocalState, action: Actions): TokenLocalState => {
+  switch (action.type) {
+    case SET_ENABLING:
+    case SET_CLAIMING:
+    case SET_HIGHLIGHTED:
+      return {
+        ...state,
+        [action.type]: state[action.type].has(action.payload)
+          ? state[action.type].delete(action.payload) && state[action.type]
+          : state[action.type].set(action.payload, action.payload),
```
alfetopito: Since it seems that `key: value` in the map are always equal, can you use a `Set()` instead?

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -16,13 +17,71 @@ interface Params {
...
+  highlighted: Map<string, string>
+  claiming: Map<string, string>
+}
+
+const SET_ENABLING = 'enabling'
+const SET_CLAIMING = 'claiming'
+const SET_HIGHLIGHTED = 'highlighted'
+const SET_HIGHLIGHTED_AND_CLAIMING = 'highlighted_and_claiming'
+
+type Actions =
```
Velenir: This just asks for an enum
```ts
const enum ActionTypes  {
    SET_ENABLING = 'enabling',
    SET_CLAIMING = 'claiming',
    SET_HIGHLIGHTED = 'highlighted',
    SET_HIGHLIGHTED_AND_CLAIMING = 'highlighted_and_claiming',
}

interface Action {
    type: ActionTypes
    payload: string
}
```

though it's very much individual preference

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -16,13 +17,71 @@ interface Params {
...
+      }
+    default:
+      return state
+  }
 }
 
 export const useRowActions = (params: Params): Result => {
   const { balances, setBalances } = params
+
+  const [state, dispatch] = useReducer(reducer, initialState, initState)
```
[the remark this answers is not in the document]
W3stside: yep sorry had an initState reset thing but scrapped it. good catch

src/components/DepositWidget/useRowActions.tsx
W3stside: me likey

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -16,13 +17,71 @@ interface Params {
...
+        ...state,
+        [action.type]: state[action.type].has(action.payload)
+          ? state[action.type].delete(action.payload) && state[action.type]
+          : state[action.type].set(action.payload, action.payload),
+      }
+    case SET_HIGHLIGHTED_AND_CLAIMING:
+      return {
+        ...state,
+        claiming: state.claiming.has(action.payload)
+          ? state.claiming.delete(action.payload) && state.claiming
```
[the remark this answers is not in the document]
W3stside: yes!

src/components/DepositWidget/useRowActions.tsx
W3stside: good idea, better idea, great idea

src/components/DepositWidget/useRowActions.tsx
```diff
@@ -1,30 +1,104 @@
...
+
+const enum ActionTypes {
+  SET_ENABLING = 'enabling',
+  SET_CLAIMING = 'claiming',
+  SET_HIGHLIGHTED = 'highlighted',
+  SET_HIGHLIGHTED_AND_CLAIMING = 'highlighted_and_claiming',
+}
+
+interface Actions {
+  type: ActionTypes
```
anxolin: I think this approach is totally fine, although I'm more in favour of  declaring how we do in other parts of the app :
`type: 'enabling' | 'claiming' | ...`  instead of an enum cause it's less verbose and typescript does the checks also, and even the the intellisense/suggestions work better

anxolin:
nvm about my last comment about reseting the state when you navigate! I'm reading now https://github.com/gnosis/dex-react/pull/259

PR #259, opened 2019-11-28
250/global state

### LINKED ISSUE #250: Interface state is lost when changing pages
1. enable token
2. before completing, move to Trade page
3. navigate back to Deposit page

Creates a bug of allowing two enablings:
1. enable token
2. before completing, move to Trade page
3. navigate back to Deposit page
4. enable same token again

both transactions succeed

**Est**: 0.5

W3stside:
closes #250 

## Global State
useGlobalState hook and HOC that wraps App
reducers-actions folder with reducer/actions logic + types from reducers-actions/index
tests fixed
### TESTING STUFF
enable a token, change pages immediately (you can but don't have to approve MM tx) - then go back
deposit/withdraw/claim - same thing, switch pages back
### BROKEN
types obviously - having hard time with hooks/useGlobalState types:
```
/* eslint-disable @typescript-eslint/no-explicit-any */
/* eslint-disable @typescript-eslint/explicit-function-return-type */
```


gnosis-info:
Travis automatic deployment:
 https://pr259--dexreact.review.gnosisdev.com

src/hooks/useGlobalState.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const [state, dispatch] = useReducer(reducer, initialStateFunc())
+
+  return (
+    <GlobalStateContext.Provider value={[state, dispatch]}>
+      <WrappedComponent {...props} />
+    </GlobalStateContext.Provider>
+  )
+}
+
+const useGlobalState = (): [GlobalState, Function] => useContext(GlobalStateContext) as [GlobalState, Function]
```
Velenir: Consider
```ts
const defaultSelector = st => st
const useGlobalState = (selector = defaultSelector) => {
  const [GlobalState, dispatch] => useContext(GlobalStateContext)

 const localState = useMemo(() => selector(GlobalState), [GlobalState, selector])

 const actions = useMemo(() => ({
  setClaiming: addr => dispatch(setClaimingAction(addr)),
  ...
 }), [])
 
 return [localState, actions]
}
```

just an idea

Velenir: Will allow not to care about dispatching the right thing

src/reducers-actions/combineReducers.ts
```diff
@@ -0,0 +1,77 @@
+import { Reducer } from 'react'
```
anxolin: Got a bit lost in this file. Is this something copied from a library or something? Is there a dependency we can use?

src/App.tsx
```diff
@@ -66,4 +72,18 @@ const App: React.FC = () => (
...
+export default hot(
+  withGlobalContext(
+    App,
+    // Initial State
+    () => {
+      // Make sure key names here match that of src/reducers-actions/combineReducers keys
+      // These key names are the state slices accessible from the useGlobalState hook via
+      // const [{ TokenRow }, dispatch] = useGlobalState()
+      return {
+        TokenRow,
```
anxolin: I don't understand, why the global state is modeled as a TokenRow

anxolin: Okay, I see `TokenRow` is not what I think, we have a component called `TokenRow` but this seems is totally different thing. it's a `TokenRowInitialState` so I wouldn't change the name of that when you import it.

Also, I wouldn't capitalize the state, right? 
```
return {
    tokenRow: TokenRowInitialState
}
```

Actually, the name it's kind of misleading too. Maybe you call that state prop `TokenRow` because you know the data is used in the token row components? I think it's a better name `tokens` for containing the global state related to the tokens (for now, only flags for enabling, highlighting and so on)

anxolin:
@W3stside @DmitryBespalov @alfetopito changes like this one, that is an important new approach, or add something that all the team should be aligned, I would suggest to try to get more approvals and give a decent time to review (as it was the case)

Sometimes I get my PRs clogged so I appreciate the extra time for changes like this one. What do you think?

src/App.tsx
```diff
@@ -1,10 +1,12 @@
 import 'types'
 
-import { hot } from 'react-hot-loader/root'
 import React from 'react'
-import GlobalStyles from './components/layout/GlobalStyles'
+import { hot } from 'react-hot-loader/root'
```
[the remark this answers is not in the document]
W3stside: I was under the impression that 

```
resolve: {
    alias: {
      'react-dom': '@hot-loader/react-dom',
    },
```

made that unnecessary. But I can change it back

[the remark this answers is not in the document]
W3stside: (in webpack config)

W3stside:
> @W3stside @DmitryBespalov @alfetopito changes like this one, that is an important new approach, or add something that all the team should be aligned, I would suggest to try to get more approvals (3/3 ideally) and give a decent time to review (as it was the case)
> 
> Sometimes I get my PRs clogged so I appreciate the extra time for changes like this one. What do you think?

@Velenir - sorry @DmitryBespalov

src/App.tsx
W3stside: fixed

src/App.tsx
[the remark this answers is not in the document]
W3stside: fixed

src/reducers-actions/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**********************************
+ * Initial Global State
+ *
+ * Sets app's initial global state
+ * make sure the name of the state key(s) is/are the same as the reducer key(s) below
+ */
+
+export const INITIAL_STATE: GlobalState = {
+  TOKENS,
```
anxolin: why capitalized?

W3stside: ah i mis-read a previous requested change as "state key should be capitalised" not "it shouldn't be capitalised" - I'll consider the change

PR #281, opened 2019-12-04
Menu names

anxolin:
* [X] These are some changes in the navigation menu names. 
* [X] Also remove the word PoC from the title
* [X] Changes the subtitle, to give a better context of the page you are in
* [X] Shows a dFusion basic text 
* [X] Navigation menu: Show the current page that we are in underlined. Do not underline any other
* [X] Adapt a bit the styles for responsive. Also make the purple color less shiny

Didn't include:
* Remove the logo on the top left corner. It breaks a bit the layout, so I think it's better addressed in other PR. Alternatively we can add the logo there and try to make the header smaller
* It seems to me we are not supper efficient with the space of the header. Maybe add the menu in the "warning message" on top, and reduce drastically the header.


gnosis-info:
Travis automatic deployment:
 https://pr281--dexreact.review.gnosisdev.com

src/components/layout/Header.tsx
```diff
@@ -110,27 +122,54 @@ const Wrapper = styled.header`
...
+            <Link to="/orders">Orders</Link>
           </li>
         </ul>
         <Wallet />
       </nav>
       <div className="header-title">
-        <h1>Swap stable coins</h1>
-        <h3>Fair, Efficient prices, Onchain</h3>
+        <h1>
+          <em>𝚏</em>𝓊𝗌𝓮
```
Velenir: cool

c3rnst:
Screenshot 2019-12-04 at 16 36 19

Would suggest these changes on top of this :) but otherwise, great, thank you!

src/components/Wallet.tsx
```diff
@@ -223,15 +223,15 @@ const Wallet: React.FC<RouteComponentProps> = (props: WalletProps) => {
             {/* Copy Confirmation */}
             {copiedToClipboard ? (
               <CopyDiv>
-                <FontAwesomeIcon color="#ff5097" icon={faCheck} /> <span>Copied!</span>
+                <FontAwesomeIcon color="#ff62a2;" icon={faCheck} /> <span>Copied!</span>
```
alfetopito: TODO: refactor out inline styling

src/components/layout/GlobalStyles.ts
```diff
@@ -105,8 +105,8 @@ const GlobalStyles = createGlobalStyle`
     }
 
     &.danger {
-      border-color: #ff5097;
-      color: #ff5097;
+      border-color: #ff62a2;;
```
alfetopito: TODO: create global css vars with default colors, so it can be used on wallet for instance

W3stside:
> https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fmedia.giphy.com%2Fmedia%2FtdN4TpyM3dzYA%2Fgiphy.gif&f=1&nofb=1
> 
> This logo is ugly. (No offense)

Agreed. But i think it's all a place holder anyways until we get Lea's styling changes

anxolin:
@DmitryBespalov , I know u have some concerns about namin Wallet to what we now call "Deposit". This thread is the right place to discuss it. The reason of that is that the page represents the wallet of the user (exchange wallet, and wallet) showing all his balances. Deposit can be confusing, since you also do other actions, so you could be for example withdrawing in the deposit page. I think it was also a suggestion from Lea who didn't like "deposit" either.

Happy to review the best name for it. 

@c3rnst thanks for your suggestions. I think they are good. The one of the wallet I think is best to review on it's own. The other ones, I agree, I thoght about doing them, but it would take me a bit more of time that the changes I want to include in this PR:
- Review menu names
- Underscore the current section
- Delete the subtitle, have a context aware subtitle with the name of the page
- Remove "PoC" from the header. I made a few MINIMAL changes in style, but I agree with @alfetopito the web looks a bit ugly.

Further changes in design I think would require another PR and maybe the help of @W3stside @DmitryBespalov  as masters of CSS

What should we do? Do we split this in more PRs and plan them next wednesday? Should we merge this and later apply further changes or this changes are not wanted at all?

c3rnst:
> @c3rnst thanks for your suggestions. I think they are good. The one of the wallet I think is best to review on it's own PR (#300 ). The other ones, I agree, I thought about doing them, but it would take me a bit more of time that the changes I want to include in this PR:

Totally makes sense!

anxolin:
> Totally makes sense! Though I think you might have tagged the wrong Dmitry!

It happens too much!! My present my deeper apologies to both

src/components/Wallet.tsx
anxolin: https://github.com/gnosis/dex-react/issues/313

src/components/layout/GlobalStyles.ts
anxolin: https://github.com/gnosis/dex-react/issues/313

PR #284, opened 2019-12-05
Say no to empty divs/

Velenir:
I really dislike empty divs. Sorry
Plus it's [not a good practice](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Palpable_content) to have empty elements


gnosis-info:
Travis automatic deployment:
 https://pr284--dexreact.review.gnosisdev.com

PR #285, opened 2019-12-05
Trying stuff for WC connection..v4.final

Velenir:
Hopefully final
Same as v3 but with `prototype.getWalletConnector` override instead, so that we don't have to disable class transpilation

Please test and then it's good to go

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,10 @@
 {
   "presets": [[
     "@babel/preset-env",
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr285--dexreact.review.gnosisdev.com

W3stside:
Login, Approve WETH, and trade WETH for DAI worked for me on test.walletconnect.

very nice work, approved!

Issue #228, opened 2019-11-21
[bug] Showing wrong amounts for exchange balance on /deposit


alfetopito:
### Summary

Exchange wallet showing duplicated amount after overwriting withdraw request

### Steps to reproduce 
1. on a token where there's balance on the exchange of **B**, request withdraw of **X**
1. after submitting the first withdraw request of **X** and before the tx being mined, overwrite it with a request of **Y**
1. wait until withdraw is claimable (~5min)

After second request is mined, balance is: B - Y + B
It persists even after reload/re-connect the wallet

### Details
Version: https://pr224--dexreact.review.gnosisdev.com/
Address: `0x45acf82f1B9E12890e5c1854e1250a90B1E3836e`
Token: WETH (`0xc778417E063141139Fce010982780140Aa0cD5Ab`)
Total on exchange: 7
Requested withdraw of: 5
Wallet amount: 0

![weird-amounts.png](https://images.zenhubusercontent.com/54161e1014b417b8716bc5ad/5bd9e3ce-6c7d-4cbd-bf4f-2e8903af0a4e)

### Comment
That's possibly a feature, not a bug. Contract code includes ""claimable"" amount on balance, once it's ready to claim https://github.com/gnosis/dex-contracts/blame/master/contracts/EpochTokenLocker.sol#L197

We'll have to adjust what we display in the interface in this case

Although that doesn't happen when a single request is made. Probably a race condition in the contract?

anxolin:
Would it make sense to move this issue into dex-contracts?

If this is as you describe, It looks like you found a way to increase your balance :) 
Maybe is unrelated, but I learned yesterday that they changed the requestWithdraw, now instead of overriding is doing before the previous claim. @josojo briefly explained me why, but I didn't fully understand, apparently is to prevent an attack. Not sure if this is somehow related.

bh2smith:
I think we need more refined details about the values of `B` and `Y` here espesially in relation with their respective magnitudes. What you are suggesting here is that I could have a balance of 10, request a withdraw of 1 and then have a balance of 19 (10 - 1 + 10). Is this really the case? I will play around with this myself and look into the contract code...

anxolin:
I understood that there's 7 in the exchange. Did a first request, then a second of 5  without waiting for the 1st one to be mined.

As a result, and this is the reported weird behaviour, the balance was 9: he thinks is 7 - 5 + 7
This was my understanding, not sure if I got it right...

alfetopito:
[...]

PR #286, opened 2019-12-05
Adding useOrders hook

alfetopito:
Part of #208 

---

New hook to get user orders

### Note:
Build failed because I split changes into multiple PRs.
Following PR build works: #287

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -10,6 +10,7 @@ import { tokenListApi } from 'api'
 import { getToken } from 'utils'
 import { Network, TokenDetails } from 'types'
 import OrderRow from './OrderRow'
+import { useOrders } from 'hooks/useOrders'
 
 const OrdersWrapper = styled(Widget)`
   > a {
@@ -130,6 +131,9 @@ const OrdersWidget: React.FC = () => {
     }
   }, [])
   // end temporary layout vars
+
+  const orders = useOrders()
+  const noOrders = orders.length === 0
 
   return (
     <OrdersWrapper>
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -1,0 +1,18 @@
+import { AuctionElement } from 'types'
+import { useWalletConnection } from './useWalletConnection'
+import useSafeState from './useSafeState'
+import { exchangeApi } from 'api'
+import { useEffect } from 'react'
+
+export function useOrders(): AuctionElement[] {
+  const { userAddress } = useWalletConnection()
+  const [orders, setOrders] = useSafeState<AuctionElement[]>([])
+
+  useEffect(() => {
+    exchangeApi.getOrders(userAddress as string).then(_orders => {
+      setOrders(_orders)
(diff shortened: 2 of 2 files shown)


src/hooks/useOrders.ts
```diff
@@ -0,0 +1,18 @@
...
+import useSafeState from './useSafeState'
+import { exchangeApi } from 'api'
+import { useEffect } from 'react'
+
+export function useOrders(): AuctionElement[] {
+  const { userAddress } = useWalletConnection()
+  const [orders, setOrders] = useSafeState<AuctionElement[]>([])
+
+  useEffect(() => {
+    exchangeApi.getOrders(userAddress as string).then(_orders => {
```
Velenir:
```suggestion
    userAddress && exchangeApi.getOrders(userAddress).then(setOrders)
```

src/components/OrdersWidget/index.tsx
```diff
@@ -131,6 +132,9 @@ const OrdersWidget: React.FC = () => {
   }, [])
   // end temporary layout vars
 
+  const orders = useOrders()
+  const noOrders = orders.length === 0
```
Velenir: Unused breaks, TS build

Velenir:
`contract.methods.getEncodedUserOrders` thows
```
Error: invalid address (arg="user", coderType="address", value=undefined)
Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.
```

Need to update contract address

alfetopito:
> `contract.methods.getEncodedUserOrders` thows
> 
> ```
> Error: invalid address (arg="user", coderType="address", value=undefined)
> Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.
> ```
> 
> Need to update contract address

We also need to update the contract and exchangeApi for this task https://github.com/gnosis/dex-react/issues/228

I think we could do it all at once

src/hooks/useOrders.ts
```diff
@@ -0,0 +1,18 @@
...
+import { exchangeApi } from 'api'
+import { useEffect } from 'react'
+
+export function useOrders(): AuctionElement[] {
+  const { userAddress } = useWalletConnection()
+  const [orders, setOrders] = useSafeState<AuctionElement[]>([])
+
+  useEffect(() => {
+    userAddress && exchangeApi.getOrders(userAddress).then(setOrders)
+      setOrders(_orders)
```
Velenir: You just committed my suggestion ? :smile: Right here in github?

Multiline suggestions aren't here yet, but are on the github roadmap

alfetopito: I did! Though it was ready to go, no?

alfetopito: Ok now I see it. Fixed in latest revision :)

W3stside:
APPROVED!

PR #287, opened 2019-12-05
208/no orders layout

alfetopito:
Part of #208 
Depends on #286 

- Adding layout for empty orders
- Refactoring buttons to use `grid`

no-orders

PR #288, opened 2019-12-05
208/mock orders

alfetopito:
Part of #208 
Depends on #286 and #287 

- Adding mock orders for local testing.
- Added utils time functions to calculate batchIds based on Date objects OR timestamps
- Reordered ExchangeApiMock constructor parameters for convenience

NOT loaded into the Orders page yet


gnosis-info:
Travis automatic deployment:
 https://pr288--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApiMock.ts
```diff
@@ -23,8 +23,8 @@ export class ExchangeApiMock extends DepositApiMock implements ExchangeApi {
     balanceStates: BalancesByUserAndToken,
     erc20Api: Erc20Api,
     registeredTokens: string[] = [],
-    maxTokens = 10000,
     ordersByUser: OrdersByUser = {},
+    maxTokens = 10000,
```
Velenir: Is it time to switch to object as params? :smile:

alfetopito: Well, can't say that I thought about that. But I like the suggestion. Will do!

src/utils/time.ts
```diff
@@ -1,10 +1,83 @@
...
+ * Example: dateAdd(new Date(), 30, 'minute')  //returns 30 minutes from now.
+ * https://stackoverflow.com/a/1214753/18511
+ *
+ * Adapted from https://stackoverflow.com/a/1214753/1272513
+ *
+ * @param date  Date to start with
+ * @param time  How much of given time unit to add to date.
+ * @param unit  One of: year, month, week, day, hour, minute, second
+ */
+export function addTimeToDate(
```
anxolin: Yes, this ideally uses a library. Here it looks there's some alternatives to momentjs if we feel moment might add to much weight.
https://github.com/you-dont-need/You-Dont-Need-Momentjs

anxolin: For this function I would consider OK not to use a lib, but I think we'll need to format dates easily, like for showing relative dates "fromNow" functionality

alfetopito: Will indeed need that for the next step. Will look into it.

PR #291, opened 2019-12-08
260/eth wrapper

### LINKED ISSUE #260: [on hold] Wrap ETH
We should allow to WRAP and unwrap ether easily...

Ideally integrated in the app. At least a link to somewhere we can wrap/unwrap.

W3stside:
closes #260  
ETH Wrapper module

Uses: [Wraptor](https://github.com/W3stside/wraptor)

1. updates auto on new blockheader
2. catches bad amounts/ does data validation

### DIFF
--- a/src/api/wallet/WalletApiImpl.ts
+++ b/src/api/wallet/WalletApiImpl.ts
@@ -251,10 +251,10 @@ export class WalletApiImpl implements WalletApi {
     }
   }
 
-  private async _notifyListeners(): Promise<void> {
-    const blockNumber = await this._blockNumber
+  private async _notifyListeners(options?: { blockHeader: { number: number } }): Promise<void> {
     const walletInfo: WalletInfo = this._getWalletInfo()
-    this._listeners.forEach(listener => listener({ ...walletInfo, blockNumber }))
+    const wInfoExtended = { ...walletInfo, blockNumber: options?.blockHeader?.number }
+    this._listeners.forEach(listener => listener(wInfoExtended))
   }
 
   private get _connected(): boolean {
@@ -267,10 +267,6 @@ export class WalletApiImpl implements WalletApi {
   private get _balance(): Promise<string> {
     if (!this._web3) return Promise.resolve('0')
     return this._web3.eth.getBalance(this._user)
-  }
-  private get _blockNumber(): Promise<number> {
-    if (!this._web3) return Promise.resolve(0)
-    return this._web3.eth.getBlockNumber()
   }
   private get _networkId(): Network {
     const { chainId = 0 } = getProviderState(this._provider) || {}
--- a/src/api/wallet/WalletApiImpl.ts
+++ b/src/api/wallet/WalletApiImpl.ts
@@ -134,7 +134,7 @@ const closeOpenWebSocketConnection = (web3: Web3): void => {
 export class WalletApiImpl implements WalletApi {
   private _listeners: ((walletInfo: WalletInfo) => void)[]
   private _provider: Provider | null
-  public _web3: Web3
+  private _web3: Web3
 
   private _unsubscribe: Command = () => {
     // Empty comment to indicate this is on purpose: https://github.com/eslint/eslint/commit/c1c4f4d
--- a/src/api/wallet/WalletApiImpl.ts
+++ b/src/api/wallet/WalletApiImpl.ts
@@ -134,7 +134,7 @@ const closeOpenWebSocketConnection = (web3: Web3): void => {
 export class WalletApiImpl implements WalletApi {
   private _listeners: ((walletInfo: WalletInfo) => void)[]
   private _provider: Provider | null
-  private _web3: Web3
+  public _web3: Web3
 
   private _unsubscribe: Command = () => {
     // Empty comment to indicate this is on purpose: https://github.com/eslint/eslint/commit/c1c4f4d
@@ -163,7 +163,6 @@ export class WalletApiImpl implements WalletApi {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     this._web3.setProvider(provider as any)
     log('[WalletApi] Connected')
-
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ;(window as any).web3c = this._web3
 
@@ -248,6 +247,7 @@ export class WalletApiImpl implements WalletApi {
       isConnected,
       userAddress: accounts[0],
       networkId: isConnected ? +chainId : undefined,
+      web3: this._web3,
     }
   }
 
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr291--dexreact.review.gnosisdev.com

src/components/WrapEtherWidget/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { tokenListApi } from 'api'
+import { Network } from 'types'
+
+const WrapEther: React.FC = () => {
+  const { userAddress, networkId, web3 } = useWalletConnection()
+
+  const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
+  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
+
+  const contractAddress = useMemo(() => tokens.find(token => token.symbol === 'WETH')?.address, [tokens])
```
Velenir: As we only use `tokens` here, put it all into one `useMemo`

src/api/wallet/WalletApiImpl.ts
```diff
@@ -134,7 +134,7 @@ const closeOpenWebSocketConnection = (web3: Web3): void => {
 export class WalletApiImpl implements WalletApi {
   private _listeners: ((walletInfo: WalletInfo) => void)[]
   private _provider: Provider | null
-  private _web3: Web3
+  public _web3: Web3
```
Velenir: Why `public`?

W3stside: I am returning it inside the `useWalletInfo` hook. Thoughts?

src/components/WrapEtherWidget/index.tsx
W3stside: ok!

src/api/wallet/WalletApiImpl.ts
Velenir: That shouldn't affect anything
Public is for `new WalletApiImpl().web3`
Returns are unrelated

W3stside: :+1:

src/api/wallet/WalletApiImpl.ts
```diff
@@ -248,13 +247,14 @@ export class WalletApiImpl implements WalletApi {
       isConnected,
       userAddress: accounts[0],
       networkId: isConnected ? +chainId : undefined,
+      web3: this._web3,
     }
   }
 
   private async _notifyListeners(): Promise<void> {
-    await Promise.resolve()
+    const blockNumber = await this._blockNumber
```
alfetopito: No need for parenthesis when calling `this._blockNumber`?

src/components/WrapEtherWidget/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Network } from 'types'
+
+const WrapEther: React.FC = () => {
+  const { blockNumber, userAddress, networkId, web3 } = useWalletConnection()
+
+  const contractAddress = useMemo(() => {
+    const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
+    const tokens = tokenListApi.getTokens(fallBackNetworkId)
+
+    return tokens.find(token => token.symbol === 'WETH')?.address
```
alfetopito: Use `getToken` function from utils?

src/api/wallet/WalletApiImpl.ts
W3stside: Getter, so nope!

alfetopito: Cool, good to know!

Velenir: Try
```ts
private async _notifyListeners({blockHeader}): Promise<void> {
    const walletInfo: WalletInfo = this._getWalletInfo()
    const wInfoExtended = {...walletInfo, blockNumber: blockHeader?.number}
    this._listeners.forEach(listener => listener(wInfoExtended))
  }

File: src/api/wallet/WalletApiImpl.ts [hunk 5]
```diff
@@ -248,13 +247,14 @@ export class WalletApiImpl implements WalletApi {
...
     const walletInfo: WalletInfo = this._getWalletInfo()
-    this._listeners.forEach(listener => listener(walletInfo))
+    this._listeners.forEach(listener => listener({ ...walletInfo, blockNumber }))
```
Velenir:
```js
{ ...walletInfo, blockNumber }
```
will get recreated for each listener, construct outside the loop

W3stside: :+1:

src/components/WrapEtherWidget/index.tsx
W3stside: :+1:

W3stside:
> Cool!!
> 
> I have a few functional comments to make it more UX friendly.
> 
> * I wouldn't add a link in the navigation menu for it
> * I think it's ok and seems simpler to follow your approach of having on it's own page. I think we can keep it like that.
> * We should navigate to the page by clicking this button:
>   image
> * For the allowance, we should be coherent with the other allowances handling. Detect if WET is enabled, and if it's not, show only a "Enable WETH" button
> * Display the WETH balance automatically, on top of the unwrap input, ideally with a set max function
> * Display the ETH balance automatically, on top of the wrap input, ideally with a set max function
> * If we are implementing the wrapper ourselves, do you think is worth it to use the component. I mean, most of the logic, formatters, approve and use-max functionality and so on it's already in this project. I'm not 100% using the component would make this development simpler
> 
> Screenshot at Dec 09 19-12-49

good ideas, will implement.

in regards to using the component - there is also a hook: [`useWraptor`](https://github.com/W3stside/wraptor/blob/master/src/hooks/useWraptor.ts) that provides the logic and is exported - we can use that

src/api/wallet/WalletApiImpl.ts
W3stside: cool, done. thanks!

anxolin:
> in regards to using the component - there is also a hook: useWraptor that provides the logic and is exported - we can use that

Nicee

W3stside:
Good point by @anxolin:

> If we are implementing the wrapper ourselves, do you think is worth it to use the component. I mean, most of the logic, formatters, approve and use-max functionality and so on it's already in this project. I'm not 100% using the component would make this development simpler

What do you guys think @alfetopito @Velenir ? Vote :+1: if we should add it, or :-1: if we shouldn't

anxolin:
> What do you guys think @anxolin @alfetopito @Velenir ? Vote 👍 if we should add it, or 👎 if we shouldn't

Imo. I would leave this decision to you. I just want to make whatever is simpler. If you feel is not very difficult to adapt the component to our needs and use it, go for it. If you think it's easier to do it as we've done the other components, maybe in the future we can extract part of that functionality and move it to the component. You are the best to take this decision, use your best judgement :)

PR #297, opened 2019-12-09
208/display orders on page

alfetopito:
Part of #208 

---

This is a bigger than usual PR:

- Adding id to AutionElement when decoding encoded order from contract
- Made parameter optional on ExchangeApiMock for testing
- Moved url helper to tokenList
- Adjusting fake order data ids and amounts
- New Util function to convert from batchId to Date
- New constants for thresholds on no limit/no expiration for orders
- Replacing static orders with data from useOrders hook
- Added bignumber.js for calculating prices (decimal big number division)
- Rendering OrderRow with the real order data

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@types/react": "^16.9.13",
     "@types/react-select": "^3.0.5",
     "@walletconnect/web3-provider": "^1.0.0-beta.38",
-    "bignumber": "^1.1.0",
+    "bignumber.js": "^9.0.0",
     "bn.js": "^5.0.0",
     "combine-reducers": "^1.0.0",
     "modali": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "@types/react": "^16.9.13",
     "@types/react-select": "^3.0.5",
     "@walletconnect/web3-provider": "^1.0.0-beta.38",
+    "bignumber": "^1.1.0",
     "bn.js": "^5.0.0",
     "combine-reducers": "^1.0.0",
     "modali": "^1.2.0",
--- a/src/api/exchange/ExchangeApiImpl.ts
+++ b/src/api/exchange/ExchangeApiImpl.ts
@@ -32,6 +32,8 @@ const decodeAuctionElements = (bytes: string): AuctionElement[] => {
   const result: AuctionElement[] = []
   const oneOrder = new RegExp(orderPattern, 'g')
   let order
+  let index = 0 // order ID is given by position and it's not part of the encoded data
+
   while ((order = oneOrder.exec(bytes))) {
     const {
       user,
@@ -58,6 +60,7 @@ const decodeAuctionElements = (bytes: string): AuctionElement[] => {
     result.push({
       user: '0x' + user,
       sellTokenBalance: new BN(sellTokenBalance, 16),
+      id: index++,
       buyTokenId: parseInt(buyTokenId, 16),
       sellTokenId: parseInt(sellTokenId, 16),
       validFrom: parseInt(validFrom, 16),
--- a/src/api/exchange/ExchangeApiImpl.ts
+++ b/src/api/exchange/ExchangeApiImpl.ts
@@ -60,7 +60,7 @@ const decodeAuctionElements = (bytes: string): AuctionElement[] => {
     result.push({
       user: '0x' + user,
       sellTokenBalance: new BN(sellTokenBalance, 16),
-      id: index++,
+      id: (index++).toString(),
       buyTokenId: parseInt(buyTokenId, 16),
       sellTokenId: parseInt(sellTokenId, 16),
       validFrom: parseInt(validFrom, 16),
--- a/src/api/tokenList/TokenListApiImpl.ts
+++ b/src/api/tokenList/TokenListApiImpl.ts
@@ -22,8 +22,6 @@ export class TokenListApiImpl implements TokenList {
     networkIds.forEach(networkId => {
       // initial value
       this._tokensByNetwork[networkId] = getTokensByNetwork(networkId)
-      // update async
-      this.fetchTokenIdsFromExchange(networkId)
     })
   }
 
--- a/src/api/tokenList/TokenListApiImpl.ts
+++ b/src/api/tokenList/TokenListApiImpl.ts
@@ -1,21 +1,17 @@
-import { TokenList, TokenDetails, ExchangeApi } from 'types'
+import { TokenList, TokenDetails } from 'types'
 import { getTokensByNetwork } from './tokenList'
-import { log } from 'utils'
 
 /**
  * Basic implementation of Token API
  *
  * Has a pre-define list of tokens.
- * Gets the images from https://github.com/trustwallet/tokens
  */
 export class TokenListApiImpl implements TokenList {
   public networkIds: number[]
   private _tokensByNetwork: { [networkId: number]: TokenDetails[] }
-  private _exchangeApi: ExchangeApi
 
-  public constructor(networkIds: number[], exchangeApi: ExchangeApi) {
+  public constructor(networkIds: number[]) {
     this.networkIds = networkIds
-    this._exchangeApi = exchangeApi
 
     // Init the tokens by network
     this._tokensByNetwork = {}
@@ -27,30 +23,6 @@ export class TokenListApiImpl implements TokenList {
 
   public getTokens(networkId: number): TokenDetails[] {
     return this._tokensByNetwork[networkId] || []
-  }
-
-  public async updateTokenIdsForNetwork(networkId: number): Promise<void> {
-    // on every call fetch initial list again as tokens previously discarded might have been added to exchange
-    const tokens = getTokensByNetwork(networkId)
-    if (!tokens) {
-      return
-    }
-
-    const tokenPromises = tokens.map(token => this.injectExchangeIdIntoToken(token))
-
-    // Resolve promises and remove tokens which are not registered in the exchange
-    this._tokensByNetwork[networkId] = (await Promise.all(tokenPromises)).filter(Boolean) as TokenDetails[]
-  }
-
-  private async injectExchangeIdIntoToken(token: TokenDetails): Promise<TokenDetails | null> {
-    try {
-      // Should throw when address not registered
-      const exchangeId = await this._exchangeApi.getTokenIdByAddress(token.address)
-      return { ...token, id: exchangeId }
-    } catch (e) {
-      log('Token not registered on the exchange %s', token.address, e)
-      return null
-    }
   }
 }
 
--- a/src/api/tokenList/TokenListApiImpl.ts
+++ b/src/api/tokenList/TokenListApiImpl.ts
@@ -1,11 +1,6 @@
 import { TokenList, TokenDetails } from 'types'
 import { getTokensByNetwork } from './tokenList'
 
-function _getImageUrl(tokenAddress?: string): string | undefined {
-  if (!tokenAddress) return undefined
-  return `https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${tokenAddress}/logo.png`
-}
-
 /**
  * Basic implementation of Token API
  *
@@ -22,16 +17,8 @@ export class TokenListApiImpl implements TokenList {
     // Init the tokens by network
     this._tokensByNetwork = {}
     networkIds.forEach(networkId => {
-      // Get list of tokens
-      const tokens = getTokensByNetwork(networkId).map(token => {
-        // Add image to the tokens
-        return {
-          ...token,
-          image: _getImageUrl(token.addressMainnet),
-        }
-      })
-
-      this._tokensByNetwork[networkId] = tokens
+      // initial value
+      this._tokensByNetwork[networkId] = getTokensByNetwork(networkId)
     })
   }
 
--- a/src/api/tokenList/TokenListApiImpl.ts
+++ b/src/api/tokenList/TokenListApiImpl.ts
@@ -40,10 +40,8 @@ export class TokenListApiImpl implements TokenList {
 
     const tokenPromises = tokens.map(token => this.injectExchangeIdIntoToken(token))
 
-    this._tokensByNetwork[networkId] = (await Promise.all(tokenPromises)).filter(
-      // remove tokens which are not registered in the exchange
-      token => !!token,
-    ) as TokenDetails[]
+    // Resolve promises and remove tokens which are not registered in the exchange
+    this._tokensByNetwork[networkId] = (await Promise.all(tokenPromises)).filter(Boolean) as TokenDetails[]
   }
 
   private async injectExchangeIdIntoToken(token: TokenDetails): Promise<TokenDetails | null> {
--- a/src/api/tokenList/TokenListApiImpl.ts
+++ b/src/api/tokenList/TokenListApiImpl.ts
@@ -1,5 +1,6 @@
-import { TokenList, TokenDetails } from 'types'
+import { TokenList, TokenDetails, ExchangeApi } from 'types'
 import { getTokensByNetwork } from './tokenList'
+import { log } from 'utils'
 
 /**
  * Basic implementation of Token API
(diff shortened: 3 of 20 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr297--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -95,12 +97,10 @@ const OrderDetails: React.FC<Pick<Props, 'price' | 'buyToken' | 'sellToken' | 'p
   )
 }
 
-const UnfilledAmount: React.FC<Pick<Props, 'sellToken' | 'unfilledAmount' | 'unlimited' | 'pending'>> = ({
-  sellToken,
-  unfilledAmount,
-  unlimited,
-  pending,
-}) => {
+const UnfilledAmount: React.FC<Pick<Props, 'sellToken' | 'pending'> & {
```
Velenir: It is a bit hard to reson about Props in all these `Pick<Props, 'whatever'> & {}`
Maybe create interfaces `interface ** extends Pick<Props, 'whatever'> {`

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -131,33 +130,51 @@ const AvailableAmount: React.FC<Pick<Props, 'sellToken' | 'availableAmount' | 'o
...
+    formatAmountFull(order.priceDenominator, sellToken.decimals, false),
+  )
+  const unfilledAmount = formatAmount(order.remainingAmount, sellToken.decimals) || '0'
+  const availableAmount = formatAmount(order.sellTokenBalance, sellToken.decimals) || '0'
+  const unlimitedAmount = order.priceDenominator.gt(MIN_UNLIMITED_SELL_ORDER)
+  const overBalance = order.remainingAmount.gt(order.sellTokenBalance)
+  const unlimitedTime =
+    dateFromBatchId(order.validUntil).getTime() >=
+    addTimeToDate(new Date(), MIN_UNLIMITED_SELL_ORDER_EXPIRATION_TIME, 'minute').getTime()
+  const expiresOn = order.validUntil.toLocaleString() // TODO: make it nice like, 3 min, 1 day, etc
```
Velenir: Please memoize everything here

src/components/OrdersWidget/index.tsx
```diff
@@ -222,36 +234,14 @@ const OrdersWidget: React.FC = () => {
...
-                sellToken={PAX}
-                buyToken={DAI}
-                price="1.10"
-                unfilledAmount="5,876.8429"
-                availableAmount="500"
-                expiresOn="In 2 days"
-              />
+              {orders.map(order => (
+                <OrderRow
+                  key={order.id}
```
Velenir: Is this the only place you use `order.id`? Then really no need for that, just use array index.

src/types/index.ts
```diff
@@ -31,6 +32,7 @@ export interface TokenBalanceDetails extends TokenDetails {
 
 export interface TokenList {
   getTokens: (networkId: number) => TokenDetails[]
+  fetchTokenIdsFromExchange: (networkId: number) => Promise<void>
```
Velenir: With this kind of name I'd expect this function to return an array of you know what.

src/components/OrdersWidget/index.tsx
```diff
@@ -153,23 +156,32 @@ const OrdersForm = styled.div`
...
+}
 
+const OrdersWidget: React.FC = () => {
   const orders = useOrders()
   const noOrders = orders.length === 0
 
+  // TODO: find a way to update list of tokens based on address returned by contract if we don't have it in our list
+  const { networkId } = useWalletConnection()
+  // this page is behind login wall so networkId should always be set
+  const tokens = tokenListApi.getTokens(networkId as number)
```
Velenir: Try `tokenListApi.getTokens(networkId!)`, more generic that way

package.json
```diff
@@ -63,6 +63,7 @@
     "@types/react": "^16.9.13",
     "@types/react-select": "^3.0.5",
     "@walletconnect/web3-provider": "^1.0.0-beta.38",
+    "bignumber": "^1.1.0",
```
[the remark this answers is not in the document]
alfetopito: oh, is there a difference? :O
Guess I picked the wrong one then...

src/api/exchange/ExchangeApiImpl.ts
```diff
@@ -58,6 +60,7 @@ const decodeAuctionElements = (bytes: string): AuctionElement[] => {
     result.push({
       user: '0x' + user,
       sellTokenBalance: new BN(sellTokenBalance, 16),
+      id: index++,
```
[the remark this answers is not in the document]
alfetopito: My idea is to have it from the API, because later we might re-order the list based on different sorting factors and lose the initial index. Also, we might eventually add pagination

src/api/tokenList/TokenListApiImpl.ts
```diff
@@ -15,29 +11,51 @@ function _getImageUrl(tokenAddress?: string): string | undefined {
...
-          ...token,
-          image: _getImageUrl(token.addressMainnet),
-        }
-      })
-
-      this._tokensByNetwork[networkId] = tokens
+      // initial value
+      this._tokensByNetwork[networkId] = getTokensByNetwork(networkId)
+      // update async
+      this.fetchTokenIdsFromExchange(networkId)
```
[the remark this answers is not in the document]
alfetopito: Good point. Will move it into the API constructor instead

src/api/tokenList/TokenListApiImpl.ts
```diff
@@ -15,29 +11,51 @@ function _getImageUrl(tokenAddress?: string): string | undefined {
...
     })
   }
 
   public getTokens(networkId: number): TokenDetails[] {
     return this._tokensByNetwork[networkId] || []
   }
+
+  public async fetchTokenIdsFromExchange(networkId: number): Promise<void> {
+    // on every call fetch initial list again as tokens previously discarded might have been added to exchange
+    const tokens = getTokensByNetwork(networkId)
```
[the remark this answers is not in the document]
alfetopito: I agree. First, I'll call this from the API constructor which will initialize both networks as I said in the previous comment.

Second, indeed, I intended to have this updated on network change/new block.
Still haven't got there neither am completely sure what's the best way to do it.

src/components/OrdersWidget/index.tsx
alfetopito: As I said in another comment, the main reason for having the id is for sorting, and later possibly pagination. Let's say `orders` is sorted by expiration date, how else could we keep track of the initial ids?

src/types/index.ts
alfetopito: What name would you suggest instead?

src/api/tokenList/TokenListApiImpl.ts
```diff
@@ -15,29 +11,51 @@ function _getImageUrl(tokenAddress?: string): string | undefined {
...
+    this._tokensByNetwork[networkId] = (await Promise.all(tokenPromises)).filter(
+      // remove tokens which are not registered in the exchange
+      token => !!token,
+    ) as TokenDetails[]
+  }
+
+  private async injectExchangeIdIntoToken(token: TokenDetails): Promise<TokenDetails | null> {
+    try {
+      // Should throw when address not registered
+      const exchangeId = await this._exchangeApi.getTokenIdByAddress(token.address)
```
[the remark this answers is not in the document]
alfetopito: Yeah, I have mixed feelings. As I said in the team chat, I'm not sure where would be the best place to load the token ids.
This is my first try, but happy to try another way.

Also, I have not implemented yet, but we need a way to add tokens to the list when we get an order with an unknown id.

src/types/index.ts
alfetopito: Renamed to `updateTokenIdsForNetwork`. Let me know whether that's better.

gnosis-info:
Travis automatic deployment:
 https://pr297--dexreact.review.gnosisdev.com

src/api/tokenList/TokenListApiImpl.ts
```diff
@@ -15,29 +11,47 @@ function _getImageUrl(tokenAddress?: string): string | undefined {
 export class TokenListApiImpl implements TokenList {
   public networkIds: number[]
   private _tokensByNetwork: { [networkId: number]: TokenDetails[] }
+  private _exchangeApi: ExchangeApi
 
-  public constructor(networkIds: number[]) {
+  public constructor(networkIds: number[], exchangeApi: ExchangeApi) {
```
anxolin: This is a bit tricky, I would think it's not bad for now, but the issue is "Api" is a logical layer we created to abstract certain things from the UI logic since the UI logic is complicated enough. 

Ideally, a layer relies **only in the immediate layer below** which for simplicity we don't have in this project for now. Only utils and libs.

In this case we have an Api depending on another Api. This would likely bring problems in the future.

Another approach would be:
- We abstract the access to blockchain, the file list, or whatever undelying persistence layer with a `Repository` layer. 
- We abstract our logics, denormalizing, joining data from different datasources (lists, TCRs, different smart contracts, etc..) into more UI like objects, etc... This would be a service layer (we can call it `Service` or `Api`). This layer would receive Repositories and config params in it's constructor, never other elements from the same layer.

Issue: I don't want to overcomplicate, so I'm happy to leave it like this if we don't have the issue yet. Also, another issue, would be that in may cases, our service layer might be just re-exposing some/all of the repository methods.

Happy to hear everyone's thoughts on alternatives.

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
anxolin: I opened a discussion to define responsibilities of the layers. Maybe it would help us agree on the best way to organize them.

I don't think we should have a declarative `updateTokenIdsForNetwork `. In my opinion we should just get the list of tokens from the api providing the id, the API should know for how long we cache the list and abstract this away to the UI (hooks or components). However, if we end up using a repository layer, I would have the cache logic there.

I have the same opinion for the other question you have. How to deal with "unknown" ids. I think we should have a getter in an api/repository that caches:
* forever if the id is known
* 10-30min for example, if the token is unknown. It's fine for me having a delay of some minutes (or make the user REFRESH to get a brand new token :P ) 

Potencially the cache can be even persisted in the browser for future versions if we see a BOOST with that ;)

src/api/tokenList/TokenListApiImpl.ts
```diff
@@ -15,29 +11,47 @@ function _getImageUrl(tokenAddress?: string): string | undefined {
...
+  }
+
+  private async injectExchangeIdIntoToken(token: TokenDetails): Promise<TokenDetails | null> {
+    try {
+      // Should throw when address not registered
+      const exchangeId = await this._exchangeApi.getTokenIdByAddress(token.address)
+      return { ...token, id: exchangeId }
+    } catch (e) {
+      log('Token not registered on the exchange %s', token.address, e)
+      return null
```
anxolin: At first I thought this was not right, cause we could return the token details but without the ID.

But on a second thought, maybe is fine to do this for now, since I cannot came up with an use case we have in the app where we have a token without an ID. 
This behaviour would become a problem if we do a interface for adding tokens to the exchange.

Alternatively we do the filtering in the UI. But again, I don't foresee the need in the short-middle term

src/components/OrdersWidget/index.tsx
anxolin: in the real implementation, what will be the id? I think to uniquely refer to an order you need something like:
`key={order.owner + '_' + order.orderId}`

src/api/exchange/ExchangeApiImpl.ts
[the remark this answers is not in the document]
Velenir: valid point

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
Velenir: Crazy idea in line with
> Potentially the cache can be even persisted in the browser

prefill known ids before compilation, as in in webpack config

src/types/index.ts
Velenir: :+1:

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
alfetopito: I like this idea. Will do it in a follow up PR https://github.com/gnosis/dex-react/issues/306

src/components/OrdersWidget/index.tsx
alfetopito: Yes, the order the ids are returned is the orderId itself, if I understood the contract correctly: 
https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L375 and
https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L212

You can only cancel your own orders, so the ids always refer to yourself, so I don't think we need to prefix it with order.owner.

src/api/tokenList/TokenListApiImpl.ts
alfetopito: I'm not entirely sure what would be the immediate solution.

We already have a dependency between APIs, although they are a bit more related (Deposit and Exchange APIs require ERC20 api instance).

And if we want to update the token list with tokens on the exchange and not on our initial list, we'll need: tokenList (API), erc20 API and exchange API.

I agree and like the separation of concerns between the APIs, although I don't like the duplication it can create.

src/api/tokenList/TokenListApiImpl.ts
alfetopito: In case we are adding tokens from the UI, we would have the ID once it's added.
So I don't see a scenario where we would have a token in the UI without the ID. Either it's registered in the contract or it's added to it by the user. Never empty.

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
alfetopito: > prefill known ids before compilation, as in in webpack config

Now that our contracts are close to `stable` and we can expect less often deployments, if any deployments at all, that makes sense. We can just hard code the token ids as they are always inserted in the same order.

In this case, we could even postpone this whole id fetching logic to when we decide to deal with adding tokens to the UI by users.

I pretty much like this idea as it `1` makes it simpler and `2` cleans up this half baked code that everyone has concerns with.

I'll give it a try and you guys let me know if you like the new version.

src/components/OrdersWidget/index.tsx
anxolin: Mm, yes, but I would identify the orders using the unique id. Otherwise, you'll find an issue if you reuse your rows for displaying for example the orders of a batch

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
anxolin: But we will have an issue when we want to deal with tokens not in our list, right?

src/api/tokenList/tokenList.json
```diff
@@ -1,5 +1,6 @@
 [
   {
+    "id": 1,
```
Velenir: While it's ok when we add tokens ourselves in a fixed order, id won't necessarily match across networks

Velenir:
> I can approve this as a first PR, but I think ...

I am of like mind on this. Feels like this PR doesn't cover the full logic to display orders, but it's big enough already. Need a follow-up PR.

src/api/tokenList/TokenListApiImpl.ts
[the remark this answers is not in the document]
alfetopito: Correct. All I did was postpone this to remove a bit of complexity from this PR

src/components/OrdersWidget/index.tsx
alfetopito: What would be the use case of displaying orders of a batch? Is this a requirement for the short term?
Otherwise, I don't yet see the need to complicate the orderId.

Unless you mean to preemptively adapt to [getEncodedOrders](https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L386) (not implemented on frontend API yet), which returns the orders sorted by user.

Still, that's a different context and there's no problem the id being different in these two situations. What I would change for this is to make the `id` property a `string` instead of `number`.

Does that work for you @anxolin ?

src/api/tokenList/tokenList.json
alfetopito: The static token list will (potentially) always have the same index, since the deployment script uses this JSON to add tokens to the contract, after adding OWL as the first token.

So, as long as the contract is deployed using this list and the list is not re-ordered in any way, the ids remain the same for every network.

PR #302, opened 2019-12-10
208/low balance indicator

alfetopito:
Part of #208 

Depends on #297 

---

Adding logic to display/hide low balance indicator for all rows.


gnosis-info:
Travis automatic deployment:
 https://pr302--dexreact.review.gnosisdev.com

src/components/OrdersWidget/useLowBalance.ts
```diff
@@ -0,0 +1,42 @@
+import { useMemo, useCallback } from 'react'
+import useSafeState from 'hooks/useSafeState'
+
+interface Result {
+  isLowBalance: boolean
+  updateLowBalanceFactory: (id: number) => (isLow: boolean, remove?: true) => void
+}
+
+export function useLowBalance(): Result {
+  const [lowBalance, setLowBalance] = useSafeState<Set<number>>(new Set())
```
anxolin: low balance is a set, I guess then it should be `lowBalances?`

src/components/OrdersWidget/useLowBalance.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const updateLowBalanceFactory = useCallback(
+    (id: number) => (isLow: boolean, remove?: true): void => {
+      if (remove || !isLow) {
+        setLowBalance(curr => {
```
anxolin: it looks like this can be factorized a bit, the two branches are super similar and makes it a bit more difficult to understand the difference.

Also, I'm a bit lost with what this does. So maybe some comment. Is this `id` an order id? or is it `orderId + owner`?

src/components/OrdersWidget/index.tsx
```diff
@@ -241,6 +246,7 @@ const OrdersWidget: React.FC = () => {
                   sellToken={getTokenById(order.sellTokenId)}
                   buyToken={getTokenById(order.buyTokenId)}
                   order={order}
+                  updateLowBalance={updateLowBalanceFactory(order.id)}
```
anxolin: why factory?

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -184,6 +185,17 @@ const OrderRow: React.FC<Props> = props => {
   )
   const expiresOn = order.validUntil.toLocaleString() // TODO: make it nice like, 3 min, 1 day, etc
 
+  useEffect(() => {
```
anxolin: Shouldn't `overBalances` (set) be what the hook returns? This way you can use it in the rows, and also to display the legend if it's not an empty set.

Why we need the update logic? Maybe is just to use the orders in the hook constructor param?

src/components/OrdersWidget/useLowBalance.ts
alfetopito: id == orderId

This useCallback returns a function to be used on each OrderRow to update the state when that order has a low balance.
The set holds ids for all orders with low balance.
If there's anything in the set, the low balance indicator on the top is displayed.

Will make it more verbose.

src/components/OrdersWidget/index.tsx
alfetopito: Because it returns a function. We talked about this in another function I called a factory.
What would be a better name for a function that returns functions?

src/components/OrdersWidget/OrderRow.tsx
alfetopito: > Shouldn't overBalances (set) be what the hook returns? This way you can use it in the rows, and also to display the legend if it's not an empty set.

I don't think so, because I don't need to use the Set contents anywhere here. It's only used to track whether there's an order over balance.

> Why we need the update logic? Maybe is just to use the orders in the hook constructor param?

Because the `overBalance` calculation is done on `OrderRow`, while `OrderWidget` is the one holding the state and displaying the low balance indicator.

Velenir:
I feel like the logic is too convoluted for just showing :warning: when `available < unfilled`.
But can't put my finger on why.
Maybe it'll become clearer when there's the actual if condition for `available < unfilled`, where do we get `available, unfilled` from?
Will have another look later.

src/components/OrdersWidget/index.tsx
anxolin: Ah, I see know

alfetopito:
> I feel like the logic is too convoluted for just showing warning when `available < unfilled`.
> But can't put my finger on why.
> Maybe it'll become clearer when there's the actual if condition for `available < unfilled`, where do we get `available, unfilled` from?
> Will have another look later.

@Velenir  Both come from the AuctionElement.
- available === [sellTokenBalance](https://github.com/gnosis/dex-react/blob/develop/src/api/exchange/ExchangeApiImpl.ts#L62) (actually renamed to mean what it is on #308 )
- unfilled === [remainingAmount](https://github.com/gnosis/dex-react/blob/develop/src/api/exchange/ExchangeApiImpl.ts#L70)

And the flag is calculated on each OrderRow.

To be honest, I also think this is over engineered. In jQuery days I would just search for the low balance indicator on the children and if any is found, display one on top as well =P

Any react style pattern to simplify this approach comes to mind?

PR #307, opened 2019-12-11
208/nice date formating

alfetopito:
Part of #208 

Depends on #297 and #302 

---

- New dependency: `date-fns`
- Replacing utils time function with `date-fns`
- Displaying nice date formatting on orders page:

Screenshot_20191211_154541

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -179,7 +179,9 @@ const OrderRow: React.FC<Props> = props => {
       addMinutes(Date.now(), MIN_UNLIMITED_SELL_ORDER_EXPIRATION_TIME).getTime(),
     [order.validUntil],
   )
-  const expiresOn = order.validUntil.toLocaleString() // TODO: make it nice like, 3 min, 1 day, etc
+  const expiresOn = useMemo(() => formatDistanceToNow(dateFromBatchId(order.validUntil), { addSuffix: true }), [
+    order.validUntil,
+  ])
 
   return (
     <OrderRowWrapper className={'orderRow' + (pending ? ' pending' : '')}>
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -1,7 +1,6 @@
 import React, { useMemo } from 'react'
 import BigNumber from 'bignumber.js'
 import styled from 'styled-components'
-import { addMinutes, formatDistanceToNow } from 'date-fns'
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr307--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -180,10 +181,12 @@ const OrderRow: React.FC<Props> = props => {
   const unlimitedTime = useMemo(
     () =>
       dateFromBatchId(order.validUntil).getTime() >=
-      addTimeToDate(new Date(), MIN_UNLIMITED_SELL_ORDER_EXPIRATION_TIME, 'minute').getTime(),
+      addMinutes(Date.now(), MIN_UNLIMITED_SELL_ORDER_EXPIRATION_TIME).getTime(),
```
anxolin: not from this PR, but I think we don't need `getTime()` to compare two dates

also, not from this PR, but I would say MIN_UNLIMITED_SELL_ORDER_EXPIRATION_TIME of 1 year is too litle. I was thinking in an arbitrary high amount. We'll set the limit to the max value, so probably it will be an epoc of 2^256-1  I'm ok using for the comparison something like 100 years too

anxolin: It's likely that we will need this somewhere else, and it would make it more easy to test
Would it make sense to move the check to the utils/helpers?

also, this way, we don't add a direct dependency with `date-fns` in the UI

test/utils/calculateBatchId.spec.ts
```diff
@@ -11,15 +13,15 @@ describe('calculateBatchId', () => {
     expect(calculateBatchId()).toBe(BATCH_ID)
   })
   it('returns next batch id when exact time is set for next batch', () => {
-    const fiveMinutesInTheFuture = addTimeToDate(DATE, 5, 'minute')
+    const fiveMinutesInTheFuture = addMinutes(DATE, 5)
```
anxolin: I feel we are testing their lib :) 
Maybe consider reexporting the `addMinutes` and so on in our utils so we test our helper and remove direct dependencies to 'date-fns'

alfetopito: Not really, because `addMinutes` is just a helper for the test. `calculateBatchId` doesn't use `date-fns` at all.

src/components/OrdersWidget/OrderRow.tsx
alfetopito: Not a big deal, addressed all changes in this PR

PR #309, opened 2019-12-12
Update TokenList to use MCD DAI for mainnet

josojo:
The rinkeby address of DAI was already MCD.
But the mainnet address was not.

This PR corrects this.


gnosis-info:
Travis automatic deployment:
 https://pr309--dexreact.review.gnosisdev.com

PR #311, opened 2019-12-12
Autoconnect provider

Velenir:
Autoconnecting to last connected provider

I'm not sure I like the UX, so consider it draft for now. But do review already.
Not a Draft PR so travis deploys it
Drafty PR :wind_chime: 

Logging will be removed before merge, pay it no mind.


gnosis-info:
Travis automatic deployment:
 https://pr311--dexreact.review.gnosisdev.com

src/const.ts
```diff
@@ -34,4 +34,5 @@ export const DEFAULT_DECIMALS = 4
 export const DEFAULT_PRECISION = 18
 export const ELLIPSIS = '...'
 // TODO change infuraID for production
-export const INITIAL_INFURA_ENDPOINT = 'wss://rinkeby.infura.io/ws/v3/8b4d9b4306294d2e92e0775ff1075066'
+export const INFURA_ID = '8b4d9b4306294d2e92e0775ff1075066'
```
anxolin: At some point we should add a `process.env.INFURA_ID || '8b4d9b4306294d2e92e0775ff1075066' `

src/api/index.ts
```diff
@@ -99,3 +101,5 @@ export const tokenListApi: TokenList = createTokenListApi()
 export const erc20Api: Erc20Api = createErc20Api(web3)
 export const depositApi: DepositApi = createDepositApi(erc20Api, web3)
 export const exchangeApi: ExchangeApi = createExchangeApi(erc20Api, web3)
+
+setupAutoconnect(walletApi)
```
anxolin: I don't see this logic here. I think it makes more sense in the App.tsx for example

alfetopito: I feel like this should be in the app or another config place rather than here, although not sure where.

src/api/autoconnect.ts
```diff
@@ -0,0 +1,70 @@
...
+import { INFURA_ID } from 'const'
+
+const getWCIfConnected = async (): Promise<WCProvider | null> => {
+  const provider = new WalletConnectProvider({
+    infuraId: INFURA_ID,
+  })
+
+  // TODO: listen for specific websocket events when they are available
+  // hack to wait while WC tries connection with last session's key and uri
+  await delay(1000)
```
anxolin: Good enough :) 
let's create an issue for when https://github.com/WalletConnect/walletconnect-monorepo/pull/209 is merged

src/api/autoconnect.ts
```diff
@@ -0,0 +1,70 @@
...
+  } catch (error) {
+    console.warn('Error reestablishing previous WC connection', error)
+    return null
+  }
+
+  return provider
+}
+
+export const getLastProvider = (): Promise<Provider | null> => {
+  const lastProviderName = localStorage.getItem('lastProvider')
```
anxolin: Let's make a constat for this, so where you get and where you set it can use it

Velenir:
Incorporated suggested changes

Velenir:
So on the issue of UX/ showing connected wallet state.

As lastProvider logic is asynchronous (plus the artificial delay for WC), we have to wait somewhere:

1. Before the whole App renders -- terribad
2. Before Wallet component is shown -- maybe 
   1. Show Wallet comp in  `Loading` state -- I don't like
   2. Just switch to loaded state like it's doing now -- jarring
   3. Don't show component while loading provider

I think we can wait for WC update with new events. If the delay ends up minimal we can go with 2.iii

src/api/index.ts
W3stside: no inside wallet container or something?

src/utils/autoconnect.ts
```diff
@@ -0,0 +1,70 @@
...
+  } catch (error) {
+    console.warn('Error reestablishing previous WC connection', error)
+    return null
+  }
+
+  return provider
+}
+
+export const getLastProvider = (): Promise<Provider | null> => {
+  const lastProviderName = localStorage.getItem(STORAGE_KEY_LAST_PROVIDER)
```
W3stside: is this synchronous? I thought localStorage had an async api

W3stside: or wait is that `localForage` :thinking:

Velenir: IndexDB is async, localStorage is sync

src/api/index.ts
Velenir: I really don't want to mess with this anymore. Where it was before caused import race conditions for tests.

PR #315, opened 2019-12-13
Fix setState on unmounted

Velenir:
Fix setState on unmounted in Wallet component when logging out

### DIFF
--- a/src/hooks/useWalletConnection.ts
+++ b/src/hooks/useWalletConnection.ts
@@ -1,12 +1,13 @@
 import { walletApi } from 'api'
-import { useState, useEffect } from 'react'
+import { useEffect } from 'react'
 import { WalletInfo, Command } from 'types'
+import useSafeState from './useSafeState'
 
 export const useWalletConnection = (): WalletInfo => {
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr315--dexreact.review.gnosisdev.com

src/hooks/useWalletConnection.ts
```diff
@@ -1,12 +1,13 @@
...
+import { useEffect } from 'react'
 import { WalletInfo, Command } from 'types'
+import useSafeState from './useSafeState'
 
 export const useWalletConnection = (): WalletInfo => {
-  const [walletInfo, setWalletInfo] = useState<WalletInfo>({ isConnected: false })
+  const [walletInfo, setWalletInfo] = useSafeState<WalletInfo>({ isConnected: false })
 
   useEffect((): Command => {
     return walletApi.addOnChangeWalletInfo(setWalletInfo, true)
```
anxolin: when does `setWalletInfo` change? I mean, if that ever happens, shouldn't we also desubscribe?

Velenir: Doesn't change, linter wants it there. Either add or disable the rule.

PR #316, opened 2019-12-13
208/refactor service layer

alfetopito:
Part of #208, follow up to #297 

Implementing a proposal for a `services` layer to sit on top of low level APIs, providing functions which aggregate multiple APIs. 
This came out of a discussion with Anxo. Not sure I implemented the way we initially though of it.

Had to do a few adjustments implement it though:

- Make `getTokens` also filter by `number`
- Move `getImages` to `utils`
- Implement (symbol, name, decimals) view methods for ERC20

For now, added 3 functions, which are:

- `addTokenToExchange`
- `getTokenFromExchangeByAddress`
- `getTokenFromExchangeById`

There's a bunch of `TODOs` throughout the code where I'm asking for your opinion on what to do.

Once we have settled on the approach I'll proceed in putting this to use in the orders page.


gnosis-info:
Travis automatic deployment:
 https://pr316--dexreact.review.gnosisdev.com

Velenir:
About the general approach:
If we want to cache api response (exhcangeApi, erc20Api,...)
Then it makes sense to also not return new entities from services
```js
const token = {
    ...erc20Info,
    id: tokenId,
    address: tokenAddress,
    image: getImageUrl(tokenAddress),
  }
```

React component will thank you later

src/api/erc20/Erc20ApiMock.ts
```diff
@@ -36,6 +36,21 @@ export class Erc20ApiMock implements Erc20Api {
...
+    return 'token name'
+  }
+
+  public async symbol({ tokenAddress }: { tokenAddress: string }): Promise<string> {
+    log("Don't care about %s, just making TS shut up", tokenAddress)
+    return 'SMBL'
+  }
+
+  public async decimals({ tokenAddress }: { tokenAddress: string }): Promise<number> {
+    log("Don't care about %s, just making TS shut up", tokenAddress)
```
W3stside: :laughing:

W3stside:
> About the general approach:
> If we want to cache api response (exhcangeApi, erc20Api,...)
> Then it makes sense to also not return new entities from services
> 
> ```js
> const token = {
>     ...erc20Info,
>     id: tokenId,
>     address: tokenAddress,
>     image: getImageUrl(tokenAddress),
>   }
> ```
> 
> React component will thank you later

Can you expand on this? I dungetit

src/api/erc20/Erc20ApiMock.ts
```diff
@@ -36,6 +36,21 @@ export class Erc20ApiMock implements Erc20Api {
     return balance ? balance : ZERO
   }
 
+  public async name({ tokenAddress }: { tokenAddress: string }): Promise<string> {
+    log("Don't care about %s, just making TS shut up", tokenAddress)
```
anxolin: this is good enough, I think it's not bad at some point to add a mocked OWL, that is not in our list, but still return for his address the OWL symbol, name, and it's decimals.

If you want to try to create already some mock trades with OWL, you could see already if your service layer is resolving using the Erc20Api when it's not in our the list

src/services/addTokenToExchange.ts
```diff
@@ -0,0 +1,45 @@
...
+import { getErc20Info } from './getErc20Info'
+import { log, getImageUrl } from 'utils'
+
+interface Params {
+  address: string
+  exchangeApi: ExchangeApi
+  erc20Api: Erc20Api
+}
+
+export async function addTokenToExchange({ address, exchangeApi, erc20Api }: Params): Promise<boolean> {
```
anxolin: Do we plan to use this?

src/services/getErc20Info.ts
```diff
@@ -0,0 +1,30 @@
+import { Erc20Api } from 'types'
+import { log } from 'utils'
+
+interface Params {
+  address: string
+  erc20Api: Erc20Api
+}
+
+interface Erc20Info {
+  symbol: string
```
anxolin: this should be also optional, right?

src/services/getErc20Info.ts
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    const [symbol, name, decimals] = await Promise.all([
+      erc20Api.symbol({ tokenAddress }),
+      erc20Api.name({ tokenAddress }),
+      erc20Api.decimals({ tokenAddress }),
+    ])
+    return { symbol, name, decimals }
+  } catch (e) {
+    log('Failed to fetch ERC20 details for address %s', tokenAddress, e)
+    return null
```
anxolin: What does `null` mean?
If it is, there's no token in that address, I think it's fine. But for the implementation I would argue that we should allow to get info of tokens with no symbol/name/decimals. 

I suggest doing a trivial check, like verifying it is a smart contract (has code in that address), and querying the balance for example. If those are true, we assume is an Erc20 an we return an object with all fields optional (except the address if you want to include that). 
If it's not an ERC20 we return `null`

src/services/getTokenFromExchange.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  const token = {
+    ...erc20Info,
+    id: tokenId,
+    address: tokenAddress,
+    image: getImageUrl(tokenAddress),
+  }
+
+  // TODO: add to tokenList?
```
anxolin: why? 

A getter should never add anything. Do you mean to the cache? If that's the case, we can handle this in other PR, but I would suggest adding the cache layer using the proxy pattern in the API layer (Cache implementation of the interface that performs a simple delegation).

src/services/getTokenFromExchange.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  return token
+}
+
+interface BaseParams {
+  networkId: number
```
anxolin: this might change too often, but not the APIs. The factory is only initialized once.

src/services/getErc20Info.ts
```diff
@@ -0,0 +1,30 @@
...
+  decimals: number
+}
+
+/**
+ * Fetches info for an arbitrary ERC20 token from given address
+ */
+export async function getErc20Info({ address: tokenAddress, erc20Api }: Params): Promise<Erc20Info | null> {
+  try {
+    const [symbol, name, decimals] = await Promise.all([
+      erc20Api.symbol({ tokenAddress }),
```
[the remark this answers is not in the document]
alfetopito: In fact, all three items fetched are optional.

The question for me is: do we want to list this kind of token in the interface?

src/services/getTokenFromExchange.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  // Not there, get it from the ERC20 contract
+  return getTokenFromErc20({ tokenAddress, tokenId, erc20Api })
+}
+
+interface TokenFromExchangeByIdParams extends BaseParams {
+  tokenId: number
+}
+
+export async function getTokenFromExchangeById(params: TokenFromExchangeByIdParams): Promise<TokenDetails | null> {
```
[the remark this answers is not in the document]
alfetopito: Ah ok, then `services/index.ts` will be an even thinner layer, just exposing/renaming the functions. Will do.

src/api/erc20/Erc20ApiMock.ts
alfetopito: Yeah. I was not even going to implement it to get the PR out for feedback first, but TS was bugging me about it.
I intend to go back and add some mock functionality to it.

TODO

src/services/addTokenToExchange.ts
alfetopito: Since I was already playing around with the services layer and listing tokens is similar to adding tokens, I implemented it. 
This will not be immediately used. For now is more to get a feeling of the new layer with more than one use case.

src/services/getErc20Info.ts
alfetopito: From ERC20 point of view yes, but we already make it required together with `decimals` https://github.com/gnosis/dex-react/blob/develop/src/types/index.ts#L19-L20.

Should both be made optional again?

src/services/getErc20Info.ts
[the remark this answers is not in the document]
anxolin: I think we do, we want to list any token that is ERC20 compliant. So even without symbol we should be able to see the orders

src/services/getTokenFromExchange.ts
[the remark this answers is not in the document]
anxolin: yep, `index.ts` only provides the instances of the APIs and constructs the services using their factories

src/services/getErc20Info.ts
alfetopito: This goes a little into what I said in another comment. Do we want to list a token that doesn't have `symbol` or `decimals`?

If we do, how can I check if there's an ERC20 compliant token at a given address?

anxolin: One way, is like I commented above. You can do a basic test (u don't need to be exhaustive)
1. Check that it's a contract --> the address has code in it
2. Try to see if one of the ERC20 functions works. Use the one u like the most (approve, allowace, ...) I think chances are that it is a ERC20.

src/services/getTokenFromExchange.ts
alfetopito: Yes, that's what I meant. Now that you mentioned I remembered about the proxy/cache layer we discussed. I'll update the TODOs to not forget again.

On a second note, @Velenir is what you said related to this?

src/services/getTokenFromExchange.ts
alfetopito: The networkId is part of the external function parameters as well, so it won't be cached in the factory.

I need networkId for searching the local list of tokens before doing any external call.

The issue is that we keep a list of tokens for each network and the network id is required to use `tokenListApi.getTokens`.

Any ideas to make `tokenListApi.getTokens` network aware/agnostic?

gnosis-info:
Travis automatic deployment:
 https://pr316--dexreact.review.gnosisdev.com

src/services/getTokenFromExchange.ts
alfetopito: I guess now that I created the factories the code makes more sense.

src/services/getErc20Info.ts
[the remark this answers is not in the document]
alfetopito: Still pending to check whether the address is a contract (haven't figured how to do that yet), but made it so all fields returned (symbol, name and decimals) are optional.
Although I'm defaulting decimals to 18.

gnosis-info:
Travis automatic deployment:
 https://pr316--dexreact.review.gnosisdev.com

src/services/getErc20Info.ts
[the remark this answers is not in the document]
anxolin: maybe this helps:
```js
const code = web3.eth.getCode("0x085b0fdf115aa9e16ae1bddd396ce1f993c52220")
if(code === '0x') { not a contract, therefore not an ERC20 for sure   }
```

src/services/getErc20Info.ts
anxolin: ups. This is why it was optional at the begining. I know there was a discussion, but I didn't realized we made it mandatory.

I would understand a derived "fake" `label` prop or a `getLabel` function that takes a `TokenDetails` as an argument. But I think it's a mistake to make symbol mandatory

src/services/getTokenFromExchange.ts
anxolin: But remember we cache at API layer, not at service layer

src/services/getTokenFromExchange.ts
anxolin: What I mean, is that it should be like this imo:

```js
async function getTokenFromErc20Factory({  
  erc20Api,
}: TokenFromErc20Params): Promise<TokenDetails | null> {
    
   return ({ tokenAddress: string, tokenId: number } ) = {
        .... // logic
   }
}
```

anxolin: Forget my last comment, I see that that's what you did. I saw `getTokenFromErc20` and I got confused cause thought it was what it was exporting (and I saw mixed the API with the tokenId...)

src/services/addTokenToExchange.ts
```diff
@@ -0,0 +1,53 @@
+import { ExchangeApi, Erc20Api } from 'types'
+import { getErc20Info } from './getErc20Info'
```
anxolin: I think we should make a rule on never using other services to implement new services. We can easily get into cyclic dependency problems.

One solution is, to have reusable helper functions to refactor this pieces of code. I.e. Here you could make `./_getErc20Info` a function that requires all APIs+params. The service layers use the helper providing both things

alfetopito: Agree. Will play around with what might be best. Initial idea: split code into sub dirs:

```
services/
  index.ts
  factories/
    getTokenFromExchange.ts
    addTokenToExchange.ts
  helpers/
    getErc20Info.ts
```

src/services/getErc20Info.ts
alfetopito: Done

src/services/getErc20Info.ts
alfetopito: Symbol is optional again.

src/api/erc20/Erc20ApiMock.ts
alfetopito: Added the mock list of tokens to be `found`. Will add mock orders in a follow up PR

src/utils/miscellaneous.ts
```diff
@@ -20,18 +20,32 @@ export const log = process.env.NODE_ENV === 'test' ? noop : console.log
...
   tokens: T[] | undefined | null,
 ): T | undefined {
-  const valueUppercase = value.toUpperCase()
   return (tokens || []).find((token: T): boolean => {
     const tokenKeyValue = token[key]
     if (tokenKeyValue) {
-      return tokenKeyValue?.['toString']().toUpperCase() === valueUppercase
+      switch (typeof tokenKeyValue) {
+        case 'string':
+          return tokenKeyValue?.['toString']().toUpperCase() === (value as string).toUpperCase()
```
Velenir: why `toString()` if it's already string? TS complains?

Velenir: why get rid of `const valueUppercase = value.toUpperCase()`?

src/services/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  exchangeApi,
+  erc20Api,
+  web3,
+}
+
+export const getTokenFromExchangeByAddress = getTokenFromExchangeByAddressFactory(apis)
+
+export const getTokenFromExchangeById = getTokenFromExchangeByIdFactory(apis)
+
+export const addTokenToExchange = addTokenToExchangeFactory({ exchangeApi, erc20Api, web3 })
```
Velenir: Just pass the whole `apis` here, no need for a new object

src/services/helpers/getErc20Info.ts
```diff
@@ -0,0 +1,53 @@
...
+interface Params {
+  tokenAddress: string
+  erc20Api: Erc20Api
+  web3: Web3
+}
+
+/**
+ * Fetches info for an arbitrary ERC20 token from given address
+ */
+export async function getErc20Info({ tokenAddress, erc20Api, web3 }: Params): Promise<MinimalTokenDetails | null> {
```
Velenir: Would be good to cache all this on `tokenAddress, networkId`
Or will it be cached somewhere up the hierarchy?

src/services/getTokenFromExchange.ts
Velenir: Yes

Velenir:
> > About the general approach:
> > If we want to cache api response (exhcangeApi, erc20Api,...)
> > Then it makes sense to also not return new entities from services
> > ```js
> > const token = {
> >     ...erc20Info,
> >     id: tokenId,
> >     address: tokenAddress,
> >     image: getImageUrl(tokenAddress),
> >   }
> > ```
> > 
> > 
> > React component will thank you later
> 
> Can you expand on this? I dungetit



> > About the general approach:
> > If we want to cache api response (exhcangeApi, erc20Api,...)
> > Then it makes sense to also not return new entities from services
> > ```js
> > const token = {
> >     ...erc20Info,
> >     id: tokenId,
> >     address: tokenAddress,
> >     image: getImageUrl(tokenAddress),
> >   }
> > ```
> > 
> > 
> > React component will thank you later
> 
> Can you expand on this? I dungetit

New object each time, react rerenders, me sad
@alfetopito addressed it with `// TODO: cache new token`

src/api/erc20/Erc20ApiImpl.ts
```diff
@@ -33,6 +33,34 @@ export class Erc20ApiImpl implements Erc20Api {
     return toBN(result)
   }
 
+  public async name({ tokenAddress }: { tokenAddress: string }): Promise<string> {
+    const erc20 = this._getERC20AtAddress(tokenAddress)
+
+    return await erc20.methods.name().call()
```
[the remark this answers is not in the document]
anxolin: mm, but it's good if it throws, right? the handing logic belong to the one using the Api

src/utils/miscellaneous.ts
anxolin: yep, `valueUppercase` was done so we don't convert several times the same value to upercase.

src/api/erc20/Erc20ApiImpl.ts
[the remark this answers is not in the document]
Velenir: if it's handled, sure

anxolin:
mm, I think we have something strange with the network id. How does the API know the network id?

image

anxolin:
Okay, I think I can answer myself. We've made the APIs "context" aware, so they know how to ask web3 for the network address:

image

This is comfortable to remove the networkId from every method, but it makes it a bit less clean to implement the cache.

The alternative would be adding the param and use sth like this in every method call: `this._getContract(networkId: number)`

src/api/erc20/Erc20ApiImpl.ts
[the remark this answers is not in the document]
alfetopito: Yep, that's been handled on `services/helpers/getErc20Info.ts`

src/utils/miscellaneous.ts
alfetopito: > why toString() if it's already string? TS complains?

I guess it did complain before these change. Now it doesn't, will fix it.

> const valueUppercase = value.toUpperCase()?

Because the value might be a number, thus doing `.toUpperCase()` only when string inside the switch case.

src/services/helpers/getErc20Info.ts
alfetopito: Don't completely get your question.

If you mean why this function does not follow the same pattern as the others, it's because this is a helper meant to be used internally by the services rather than directly in the ui/hooks.

alfetopito:
> Okay, I think I can answer myself. We've made the APIs "context" aware, so they know how to ask web3 for the network address:
> 
> image
> 
> This is comfortable to remove the networkId from every method, but it makes it a bit less clean to implement the cache.
> 
> The alternative would be adding the param and use sth like this in every method call: `this._getContract(networkId: number)`

@anxolin 
While some APIs are network aware, tokenList API is not, thus I pass around the networkId.
Guess it would make sense to make tokenList API also network aware like the others?

src/api/erc20/Erc20ApiMock.ts
```diff
@@ -36,6 +50,38 @@ export class Erc20ApiMock implements Erc20Api {
     return balance ? balance : ZERO
   }
 
+  public async name({ tokenAddress }: { tokenAddress: string }): Promise<string> {
+    const erc20Info = this._initTokens(tokenAddress)
+
+    // Throws when token without `name` to mock contract behavior
+    this._assertTokenProperty(erc20Info, 'name')
+
+    return erc20Info.name as string
```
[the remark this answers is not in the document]
alfetopito: TS is unhappy when I do `!`, but happy when casting.
Let me give it another try.

anxolin:
> While some APIs are network aware, tokenList API is not, thus I pass around the networkId.
Guess it would make sense to make tokenList API also network aware like the others?

Being coherent makes sense yes. But I was wondering the opposite. If it makes sense to add `networkId` on every API method so they don't have "state/context"

When you cache, the symbol, you need to use both the `address` and the `networkId` for the key, if `networkId` is not in the params is more complicated

src/api/erc20/Erc20ApiMock.ts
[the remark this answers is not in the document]
alfetopito: No need for that at all if using `utils/assert`

alfetopito:
> > While some APIs are network aware, tokenList API is not, thus I pass around the networkId.
> > Guess it would make sense to make tokenList API also network aware like the others?
> 
> Being coherent makes sense yes. But I was wondering the opposite. If it makes sense to add `networkId` on every API method so they don't have "state/context"
> 
> When you cache, the symbol, you need to use both the `address` and the `networkId` for the key, if `networkId` is not in the params is more complicated

Ah, understood. Agree, makes sense the API not to be network aware.
I'll address that in the PR to introduce the caching layer.

src/services/getTokenFromExchange.ts
alfetopito: https://github.com/gnosis/dex-react/issues/329

PR #322, opened 2019-12-16
Fix issue navigating to trade page

anxolin:
Remove the `from` logic that it's causing navigation issues


gnosis-info:
Travis automatic deployment:
 https://pr322--dexreact.review.gnosisdev.com

PR #323, opened 2019-12-16
Refactor/new header

W3stside:
## Header Refactor

Smart Header - more generic, dynamic and responsive mobile first design.

1. Cleaner better separated CSS 
2. Responsive folded menu and highlight according to redesign spec

Benefits:
1. adding routes is easier: nav to `Layout/index.tsx` and add a route object. Then it just appears in the FE and in the responsive mode it is added to the folded nav and is highlighted when selected

Edit: the or shows a lot of few edited but it's mainly because of small CSS changes that were cross files. 

Please focus on the `Layout/Header` folder


gnosis-info:
Travis automatic deployment:
 https://pr323--dexreact.review.gnosisdev.com

src/components/Layout/Header/Navigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+      color: var(--color-text-primary);
+    }
+
+    &.active {
+      color: var(--color-text-primary);
+    }
+  }
+
+  ${({ $responsive, $open }): string | false =>
+    $responsive &&
```
alfetopito: Any special reason for prefixing the parameters with `$`?

src/components/Layout/Header/Navigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+  withPastLocation?: boolean
+  to: string | Location
+  order: number
+}
+
+const HeaderNavLink: React.FC<HeaderNavLinksInterface & {
+  responsive: boolean
+  showLinkSelector?: boolean
+  handleLinkSelect: (key: NavLinkKeys) => void
+}> = ({
```
alfetopito: Can you make an explicit interface for the input params instead of inline declaration?

src/components/Layout/Header/hooks.ts
```diff
@@ -0,0 +1,75 @@
...
+interface UseNavigationProps {
+  navigationArray: HeaderNavLinksInterface[]
+  handleLinkSelect: (key: string) => void
+}
+
+export const useNavigation = (initialState: HeaderProps['navigation'], responsive: boolean): UseNavigationProps => {
+  const [navObject, setNavObject] = useState(initialState)
+  const location = useLocation()
+
+  useMemo(() => {
```
alfetopito: Doubt about hooks. My understanding was that, `useMemo` is for when there's a need to memoize a value.
The way it's being used here looks to me more like a `useEffect` pattern instead. Not the case?

src/components/Layout/Header/hooks.ts
```diff
@@ -0,0 +1,75 @@
...
+  const navigationArray: HeaderNavLinksInterface[] = useMemo(
+    () =>
+      Object.keys(navObject).reduce((accum: HeaderNavLinksInterface[], currentKey) => {
+        const navArray = [...accum, navObject[currentKey]]
+        return navArray
+      }, []),
+    [navObject],
+  )
+
+  const handleLinkSelect = (key: string): void =>
```
alfetopito: `useCallback`?

src/components/Layout/Header/Navigation.tsx
W3stside: yeah as its styled and they are extending normal HTML comps, i dont want to clash there prop name wise

src/components/Layout/Header/Navigation.tsx
W3stside: yeah thought i did this :thinking: thanks for catching

src/components/Layout/Header/hooks.ts
W3stside: `useMemo` is fired before render is hit (https://reactjs.org/docs/hooks-reference.html#usememo) do you think it's bad to run it this way? @Velenir

src/components/Layout/Header/hooks.ts
W3stside: so in this case the savings are actually worse - seeing as using callback can be rewritten as:
```
const linkSelect = (key: string): void =>
    setNavObject(prevState => ({
      ...initialState,
      [key]: {
        ...prevState[key],
        order: 0,
      },
    }))

const handleLinkSelect = useCallback(linkSelect, [])
```
actually probably has to do more work here as it's calling `useCallback` and instantiating the array. not 100% but i think it's better as is since it's not such an expensive function

same thing for `useMemo`

src/components/Layout/Header/hooks.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const useOpenCloseNav = (): {
+  isResponsive: boolean
+  openNav: boolean
+  setOpenNav: (action: boolean) => void
+} => {
+  const [openNav, setOpenNav] = useState(false)
+
+  const { innerWidth } = useWindowSizes()
+  const isResponsive = !!(innerWidth && innerWidth < 720)
```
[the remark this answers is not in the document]
W3stside: yeah it's outlined in the design brief by Lea, it's not a super defacto number anymore as device sizes change but it's still widely accepted as tablet sizes. easily changable in the future though

src/components/Layout/Header/index.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
```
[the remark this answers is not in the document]
W3stside: thanks

src/components/Layout/Header/hooks.ts
```diff
@@ -0,0 +1,75 @@
+import { useMemo, useState } from 'react'
```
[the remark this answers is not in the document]
W3stside: ok ill separate them

src/components/Layout/index.tsx
```diff
@@ -26,7 +26,34 @@ const Wrapper = styled.div`
 const Layout: React.FC = ({ children }) => (
   <Wrapper>
     <LegalBanner startOpen={false} useFull={false} title="💀 This project is in beta. Use at your own risk." />
-    <Header />
+    <Header
+      navigation={{
+        Trade: {
```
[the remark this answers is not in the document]
W3stside: > Lot of good stuff here. I like the new look!
> 
> It was a bit difficult for me tor review cause there's a lot of things going on. I didn't do a deep review in all the CSS changes (just visual that part). I added a few comments.

Yeah I apologise for the number of files changed - it's always hard to do with design stuff. Should be easier now moving forward if we need to change stuff

W3stside:
> Lot of good stuff here. I like the new look!
> 
> It was a bit difficult for me tor review cause there's a lot of things going on. I didn't do a deep review in all the CSS changes (just visual that part). I added a few comments.

Yeah I apologise for the number of files changed - it's always hard to do with design stuff. Should be easier now moving forward if we need to change stuff

src/components/Layout/index.tsx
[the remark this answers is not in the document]
W3stside: changed to array

src/components/Layout/Header/hooks.ts
W3stside: changed to useEffect as it wasn't really necessary to make sideeffects during rendering, thanks!

PR #324, opened 2019-12-16
Add auto release on tag

Velenir:
This should make travis zip `./dist` folder and publish to releases when it sees a tag
Will see if it works when we make a new tag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,13 @@ before_install:
   # Needed to deploy pull request and releases
   - sudo apt-get -y install python-pip python-dev
   - pip install awscli --upgrade --user
+before_deploy:
+  # zip ./dist folder if tag IS present
+  - if [ $TRAVIS_TAG != "" ]; then
+      pushd dist;
+      zip -r  $OLDPWD/release_$TRAVIS_TAG.zip *;
+      popd;
+    fi
 after_success:
   # Pull Request - Deploy it to a review environment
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr324--dexreact.review.gnosisdev.com

.travis.yml
```diff
@@ -11,6 +11,13 @@ before_install:
   # Needed to deploy pull request and releases
   - sudo apt-get -y install python-pip python-dev
   - pip install awscli --upgrade --user
+before_deploy:
+  # zip ./dist folder if tag IS present
+  - if [ $TRAVIS_TAG != "" ]; then
+      pushd dist;
+      zip -r  $OLDPWD/release_$TRAVIS_TAG.zip *;
+      popd;
```
alfetopito: Was not familiar with pushd/popd. Is this essentially the same as `cd dist` then `cd -` ?

Velenir: Yep, was recommended where I copied it from

.travis.yml
```diff
@@ -50,3 +57,11 @@ deploy:
     script: bash travis/prepare_production_deployment.sh
     on:
       tags: true
+
+  # add release_v*.zip file to tagged Releases on github
+  - provider: releases
+    api_key: $GITHUB_GNOSIS_INFO_API_TOKEN
```
anxolin: Don't you need to add this secret?

Issue #301, opened 2019-12-10
Improve layout


anxolin:
Implement changes suggested by @c3rnst in here: https://github.com/gnosis/dex-react/pull/281#issuecomment-561714968

![image](https://user-images.githubusercontent.com/2352112/70545879-93e78e80-1b6e-11ea-9abe-054568fc4cbf.png)

If possible to include some minimal changes from Lea's input. It would be great.
For example, just change the colors to not have this blue header and so many "boxes".

Lea design notes:
[fuse_deisgn_updates.pdf](https://github.com/gnosis/dex-react/files/3945826/fuse_deisgn_updates.pdf)

PR #326, opened 2019-12-17
Refactor/wallet page

W3stside:
Another PR part of #301 - Wallet Page

1. deviates a bit from style guide, notably:
  a. green background for dropdown form on web size when depositing or withdrawing (opted for lighter bg but no big preference here)
  b. Line separators in form dropdowns (see screenshot)
image
2. changes button styles, card styles, etc


gnosis-info:
Travis automatic deployment:
 https://pr326--dexreact.review.gnosisdev.com

anxolin:
I love the new buttons a lot, but I think the yellow and green background is too much :) 

@leafilipo, we need your valuable input here. You can see the effect here: https://pr326--dexreact.review.gnosisdev.com/wallet

src/components/DepositWidget/Styled.tsx
```diff
@@ -1,28 +1,17 @@
...
 export const InnerWrapper = styled.div`
   position: relative;
-  background-color: #f7f0ff;
+  background-color: var(--color-background-actionCards);
   border-bottom: 2px solid #0000000f;
-  border-radius: var(--border-radius);
-  width: 96%;
+  border-radius: 0 0 var(--border-radius) var(--border-radius);
+  box-shadow: var(--box-shadow);
+  margin-top: -14px;
```
Velenir: As this value is so that the component is flush with the row above, it would make more sense to use the same units (`margin-top: -0.3rem;`) as in RowTokenDiv

src/components/DepositWidget/Styled.tsx
```diff
@@ -38,53 +27,49 @@ export const InnerWrapper = styled.div`
     text-align: center;
   }
 
-  ul {
+  .WalletItemContainer {
+    display: grid;
+    grid-template-rows: repeat(2, auto) 20px auto;
```
Velenir: `repeat(2, auto) 20px auto` vs `auto auto 20px auto`
I would understand if it was `repeat(10` or `repeat(2, auto 20 px 1 fr <complex pattern>)`
Not that it really matters

src/components/DepositWidget/Styled.tsx
```diff
@@ -38,53 +27,49 @@ export const InnerWrapper = styled.div`
...
-    list-style: none;
-    text-align: left;
-    margin: auto;
-    padding: 1rem 0 1rem 3rem;
-    max-width: 364px;
 
-    @media only screen and (max-width: 420px) {
-      padding: 1rem 0;
-    }
+    font-weight: 800;
```
Velenir: Any reason for being this specific instead of **bold** or **bolder**

W3stside: Short answer: no. Will change tomorrow, it's a good idea I didn't even think about it that way

src/components/DepositWidget/Styled.tsx
W3stside: It used to be repeat 4 and I was lazy ;)

anxolin:
I saw the new color, I think it's much better, but wouldn't be better something still lighter. This is my proposal:
image


Also, for another PR. I think we should review the colors of some texts:
image

image

anxolin:
I think also the header may benefit from being a bit darker, it would have a bit more of contrast and the button would be more visible. 

I'll do a minimal PR with this two changes and create an issue for a full review of the colors following all the user's workflow

image

Issue #97, opened 2019-10-01
Bundle size: webpack


alfetopito:
Bundle size: First review and fixes

**Estimation:** 4.5 (as of Sprint 9)

anxolin:
A bit of background, apparently the bundle right now is arround 5MB. The idea is just doing a basic review on why is so big with so few libs. Probably a dependency that shouldn't be there is being pooled.

We have to confirm that this 5MB was in the minimized version, and we do actually the tree shaking for deleting dead code. If we see something quick/basic that we can do, like load some core parts first and then have the optional parts run on demand. Also, maybe, if it's worth it, separating the vendor JS from the app so it can be cached between versions of our app.

anxolin:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F110gqI69qjVAkM%2Fgiphy.gif&token=fat" data-gitmeme-token="fat"><img src="https://media2.giphy.com/media/110gqI69qjVAkM/giphy.gif" title="Created by gitme.me with /fat"/></a>

alfetopito:
Check how Mock stuff is loaded: https://github.com/gnosis/dex-react/pull/327#discussion_r359981480

anxolin:
Something to take into account in this ISSUE too is:
- Make sure that when we load the ABIs, we don't add all the truffle contract info ---> huge
- Make sure none of the mock data is included and webpack removes that
- We don't include unused parts of code or libraries
- We don't include internationalization libraries from third party applications
- And of course study how to load our modules, maybe vendor, main, and then one async module per page... (study)

Velenir:
Through better tree-shaking, dependency changes and other obscure magic we are down 1/3 of bundle size.

Next step -- Lazy loading #485

PR #327, opened 2019-12-18
208/use service layer on orders page

alfetopito:
Depends on #316 

---

- [x] Refactoring Orders widget and OrderRow to use the service layer function `getTokenFromExchangeById`
- [x] Hiding order until both tokens are fetched.
- [x] Added mock order that requires token to be fetched and another that has a token without name/symbol
- [x] Simple mock of `web3.eth.getCode`
- [x] Displaying token address and link to etherscan when token has no name/symbol

Screenshot_20191218_131216

Not on this PR:
- Cache/proxy layer on top of API. Results are NOT yet cached.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -16,6 +16,7 @@ import {
   erc20Allowances,
   FEE_TOKEN,
   exchangeOrders,
+  TOKEN_8,
 } from '../../test/data'
 import Web3 from 'web3'
 import { INITIAL_INFURA_ENDPOINT } from 'const'
@@ -77,7 +78,7 @@ function createDepositApi(erc20Api: Erc20Api, web3: Web3): DepositApi {
 function createExchangeApi(erc20Api: Erc20Api, web3: Web3): ExchangeApi {
   let exchangeApi
   if (isExchangeMock) {
-    const tokens = [FEE_TOKEN, ...tokenList.map(token => token.address)]
+    const tokens = [FEE_TOKEN, ...tokenList.map(token => token.address), TOKEN_8]
     exchangeApi = new ExchangeApiMock({
       balanceStates: exchangeBalanceStates,
       erc20Api,
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -20,11 +20,26 @@ import {
 import Web3 from 'web3'
 import { INITIAL_INFURA_ENDPOINT } from 'const'
 
+const isWeb3Mock = process.env.MOCK_WEB3 === 'true'
 const isWalletMock = process.env.MOCK_WALLET === 'true'
 const isTokenListMock = process.env.MOCK_TOKEN_LIST === 'true'
 const isErc20Mock = process.env.MOCK_ERC20 === 'true'
 const isDepositMock = process.env.MOCK_DEPOSIT === 'true'
 const isExchangeMock = process.env.MOCK_EXCHANGE === 'true'
+
+// TODO connect to mainnet if we need AUTOCONNECT at all
+export const getDefaultProvider = (): string | null =>
+  process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT
+
+function createWeb3Api(): Web3 {
+  const web3 = new Web3(getDefaultProvider())
+
+  if (isWeb3Mock) {
+    // Only function that needs to be mocked so far. We can add more and add extra logic as needed
+    web3.eth.getCode = async (address: string): Promise<string> => address
+  }
+  return web3
+}
 
 function createWalletApi(web3: Web3): WalletApi {
   let walletApi
@@ -90,13 +105,8 @@ function createTokenListApi(): TokenList {
   return tokenListApi
 }
 
-// TODO connect to mainnet if we need AUTOCONNECT at all
-export const getDefaultProvider = (): string | null =>
-  process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT
-
-export const web3 = new Web3(getDefaultProvider())
-
 // Build APIs
+export const web3: Web3 = createWeb3Api()
 export const walletApi: WalletApi = createWalletApi(web3)
 export const erc20Api: Erc20Api = createErc20Api(web3)
 export const depositApi: DepositApi = createDepositApi(erc20Api, web3)
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -16,6 +16,7 @@ import {
   erc20Allowances,
   FEE_TOKEN,
   exchangeOrders,
+  unregisteredTokens,
   TOKEN_8,
 } from '../../test/data'
 import Web3 from 'web3'
@@ -56,7 +57,7 @@ function createWalletApi(web3: Web3): WalletApi {
 function createErc20Api(web3: Web3): Erc20Api {
   let erc20Api
   if (isErc20Mock) {
-    erc20Api = new Erc20ApiMock({ balances: erc20Balances, allowances: erc20Allowances })
+    erc20Api = new Erc20ApiMock({ balances: erc20Balances, allowances: erc20Allowances, tokens: unregisteredTokens })
   } else {
     erc20Api = new Erc20ApiImpl(web3)
   }
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -34,6 +34,7 @@ export const getDefaultProvider = (): string | null =>
   process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT
 
 function createWeb3Api(): Web3 {
+  // TODO: Create an `EthereumApi` https://github.com/gnosis/dex-react/issues/331
   const web3 = new Web3(getDefaultProvider())
 
   if (isWeb3Mock) {
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React, { useMemo, useState } from 'react'
 import BigNumber from 'bignumber.js'
 import styled from 'styled-components'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
@@ -6,6 +6,8 @@ import { faExclamationTriangle, faSpinner } from '@fortawesome/free-solid-svg-ic
 
 import Highlight from 'components/Highlight'
 import { EtherscanLink } from 'components/EtherscanLink'
+
+import { getTokenFromExchangeById } from 'services'
 
 import { TokenDetails, AuctionElement } from 'types'
 import { safeTokenName, formatAmount, formatAmountFull, isBatchIdFarInTheFuture, formatDateFromBatchId } from 'utils'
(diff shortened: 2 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr327--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -152,25 +157,56 @@ function calculatePrice(_numerator?: string | null, _denominator?: string | null
...
-    () =>
-      calculatePrice(
+const OrderRow: React.FC<Props> = props => {
+  const { order, networkId, pending = false } = props
+
+  // Fetching buy and sell tokens
+  const [sellToken, setSellToken] = useState<TokenDetails | null>(null)
+  const [buyToken, setBuyToken] = useState<TokenDetails | null>(null)
+
+  useMemo(async () => {
```
Velenir: Bad idea to do async logic in render, use `useEffect`

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -152,44 +87,147 @@ function calculatePrice(_numerator?: string | null, _denominator?: string | null
...
+  networkId: number
+  pending?: boolean
+}
+
+const OrderRow: React.FC<Props> = props => {
+  const { order, networkId, pending = false } = props
+
+  // Fetching buy and sell tokens
+  const [sellToken, setSellToken] = useState<TokenDetails | null>(null)
+  const [buyToken, setBuyToken] = useState<TokenDetails | null>(null)
```
Velenir: `useSafeState`?

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -152,44 +87,147 @@ function calculatePrice(_numerator?: string | null, _denominator?: string | null
...
+
+  useEffect(() => {
+    fetchToken(order.buyTokenId, networkId, setBuyToken)
+    fetchToken(order.sellTokenId, networkId, setSellToken)
```
anxolin: These two fetch tokens are async, if there's an error, how is it handled?

src/api/index.ts
```diff
@@ -16,16 +16,33 @@ import {
   erc20Allowances,
   FEE_TOKEN,
   exchangeOrders,
+  unregisteredTokens,
+  TOKEN_8,
 } from '../../test/data'
```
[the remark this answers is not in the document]
Velenir: I believe I brought it up at some point in the past.
Webpack drops unreachable code based on Env variables like `process.env.**` but only in that same statement
E.g.:
```ts
if (process.env.MOCK ==='true') //dropped

const isMock = process.env.MOCK ==='true'
if (isMock) // not dropped
```

so not removed currently

josojo:
We were discussing this error here:
https://github.com/gnosis/dex-contracts/issues/386

We would deploy a new node for our solver, who would allow us to make reads bigger than the gas limit.
For the front-end we imaged that you guys only need to pull in the orders for one owner, not the whole state. This restrictions should avoid the problem with the public infura nodes.

src/api/index.ts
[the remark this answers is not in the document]
anxolin: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FFGTVmzksb2j0k%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

src/api/index.ts
```diff
@@ -16,16 +16,33 @@ import {
...
+// TODO connect to mainnet if we need AUTOCONNECT at all
+export const getDefaultProvider = (): string | null =>
+  process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT
+
+function createWeb3Api(): Web3 {
+  const web3 = new Web3(getDefaultProvider())
+
+  if (isWeb3Mock) {
+    // Only function that needs to be mocked so far. We can add more and add extra logic as needed
+    web3.eth.getCode = async (address: string): Promise<string> => address
```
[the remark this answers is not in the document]
alfetopito: I like the idea. Will address this on https://github.com/gnosis/dex-react/issues/331

src/components/OrdersWidget/OrderRow.tsx
alfetopito: The order is only displayed when both tokens are known.
If there's a failure the order will remain hidden.

src/api/index.ts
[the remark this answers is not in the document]
alfetopito: Soooo, is there any action here?

alfetopito:
> Cool!
> 
> Please review the error handling, I think we don't take it into account.
> 
> Also, I saw this in the console when visiting the orders page:
> 
> image
> 
> @josojo I know you had the same issue, what was the outcome of your investigations about this and the limits of the contracts?

@anxolin @josojo I did see this before but did not give much attention because I know the contract/abi on the frontend is outdated. 
My assumption is that once the contract is updated, the issue will be gone. Or at least we'll be able to see the real failure reason.

> For the front-end we imaged that you guys only need to pull in the orders for one owner, not the whole state.

That's what we currently do, by using [`getEncodedUserOrders`](https://github.com/gnosis/dex-react/blob/develop/src/api/exchange/ExchangeApiImpl.ts#L97)

src/components/OrdersWidget/OrderRow.tsx
alfetopito: For some reason it doesn't work with `useSafeState`. With `useState` it does.

The ui does not update async when the token is fetched.

src/api/index.ts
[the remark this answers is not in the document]
anxolin: If you have it clear on how to do the ifs to instruct webpack and it obvious it good if we do it now. Otherwise, we can add a note in the issue for reviewing the bundle size.

src/components/OrdersWidget/OrderRow.tsx
anxolin: Mmm, I'm not sure about this. Everytime there's a promise that fetches info, there can be an exception, so it this exception happens, the user won't see any input, right?

alfetopito: Right. In this case, I can show a pop up notification on error.
Should we also notify in case the address referenced in the order is not a token or some other kind of edge case like this?

anxolin: Although it's weird I would say that this exception should trigger an alert modal or something. We hopefully won't see ever those ever. But if they show, we can review if we can handle better the case.

One issue is that it's likely that we see some repeated patterns and we will need in many places to show an error for weird cases, so maybe we can do something more or less general (like a general error handler we can give to all those promises).

src/components/OrdersWidget/OrderRow.tsx
alfetopito: Well, now it did work. Go figure...

src/api/index.ts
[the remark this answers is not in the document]
alfetopito: Yeah, I rather just leave a comment on https://github.com/gnosis/dex-react/issues/97

src/components/OrdersWidget/OrderRow.tsx
alfetopito: Created a generic utils error handler.

Also warning when (in the unlikely event) a token is not ERC20 compliant:

Screenshot_20191219_105907

src/api/index.ts
```diff
@@ -16,16 +16,33 @@ import {
...
 const isExchangeMock = process.env.MOCK_EXCHANGE === 'true'
 
+// TODO connect to mainnet if we need AUTOCONNECT at all
+export const getDefaultProvider = (): string | null =>
+  process.env.NODE_ENV === 'test' ? null : INITIAL_INFURA_ENDPOINT
+
+function createWeb3Api(): Web3 {
+  const web3 = new Web3(getDefaultProvider())
+
+  if (isWeb3Mock) {
```
Velenir: You didn't really need the whole `createWeb3Api` function. Can just move `if (isWeb3Mock) {` logic to L98

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -67,79 +73,12 @@ const PendingLink: React.FC<Pick<Props, 'pending'>> = ({ pending }) => {
...
-
-interface Props {
-  sellToken: TokenDetails
-  buyToken: TokenDetails
-  order: AuctionElement
-  isOverBalance: boolean
-  pending?: boolean
+function displayTokenSymbolOrLink(token: TokenDetails): React.ReactNode | string {
+  const displayName = safeTokenName(token)
+  if (displayName.startsWith('0x')) {
```
Velenir: If it's a check for address, then a better check would be a regex or `isAddress` from `web3-utils`

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -152,44 +91,156 @@ function calculatePrice(_numerator?: string | null, _denominator?: string | null
...
+interface UnfilledAmountProps extends Pick<Props, 'order' | 'pending'> {
+  sellToken: TokenDetails
+}
+
+const UnfilledAmount: React.FC<UnfilledAmountProps> = ({ sellToken, order, pending }) => {
   const unfilledAmount = useMemo(() => formatAmount(order.remainingAmount, sellToken.decimals) || '0', [
     order.remainingAmount,
     sellToken.decimals,
   ])
+  const unlimited = useMemo(() => order.priceDenominator.gt(MIN_UNLIMITED_SELL_ORDER), [order.priceDenominator])
```
Velenir: This kind of check is very small and produces primitive data, so I wouldn't memoize it

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -152,44 +91,156 @@ function calculatePrice(_numerator?: string | null, _denominator?: string | null
...
+    <>
+      {sellToken && buyToken && (
+        <OrderRowWrapper className={'orderRow' + (pending ? ' pending' : '')}>
+          <PendingLink {...props} />
+          <div className="checked">
+            <input type="checkbox" />
+          </div>
+          <OrderDetails {...props} sellToken={sellToken} buyToken={buyToken} />
+          <UnfilledAmount {...props} sellToken={sellToken} />
+          <AccountBalance {...props} sellToken={sellToken} />
```
Velenir: Are you sure about this kind of passing all the props wholesale to all the components? Usually you would pass props to one component or none at all. Otherwise many components will end up with props they don't need.
And if you know what props those components need, destructuring is better.

src/components/OrdersWidget/OrderRow.tsx
anxolin: Yep, agree with this comment. Memoizing makes it a bit more difficult to read, for fast operations that don't hit the network I wouldn't worry about performance issues until they are notizable

src/api/index.ts
alfetopito: I'd like to keep it here and refactor when doing https://github.com/gnosis/dex-react/issues/331.

I'll add a TODO in the code with the issue link as well.

src/components/OrdersWidget/OrderRow.tsx
alfetopito: Initially I was using this because I wasn't sure about the interface, in the first draft.
Then with `Pick<>`, the component only gets what it needs. Is it a big deal, performance wise?
I don't mind changing it, just looks cleaner this way.

src/components/OrdersWidget/OrderRow.tsx
alfetopito: `safeTokenName` returns the already abbreviated address (`0xXXXX...XXXX`) when no `name` nor `symbol` are present.
`isAddress` wouldn't work for this.

Issue #209, opened 2019-11-14
User's standing orders page: Cancelation of order


W3stside:
See [RF-4](https://docs.google.com/document/d/1lW3SQV5FAI-gDrZZAD3REunrx4TSa9nNjxHljpCFOd4/edit#heading=h.aqcghjakni1p)

PR #333, opened 2019-12-19
209/order checkboxes

alfetopito:
Part of #209 
Depends on #327 

---

- Adding logic to checkboxes

## Next PR:
- Order cancellation

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -79,7 +79,12 @@ const DeleteOrder: React.FC<Pick<Props, 'isMarkedForDeletion' | 'toggleMarkedFor
   pending,
 }) => (
   <div className="checked">
-    <input type="checkbox" onClick={toggleMarkedForDeletion} checked={isMarkedForDeletion} disabled={pending} />
+    <input
+      type="checkbox"
+      onClick={toggleMarkedForDeletion}
+      checked={isMarkedForDeletion && !pending}
+      disabled={pending}
+    />
   </div>
 )
 
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -233,7 +233,7 @@ interface Props {
 const onError = onErrorFactory('Failed to fetch token')
 
 const OrderRow: React.FC<Props> = props => {
-  const { order, isOverBalance, networkId, pending = false } = props
+  const { order, isOverBalance, networkId, pending = false, isMarkedForDeletion, toggleMarkedForDeletion } = props
 
   // Fetching buy and sell tokens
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr333--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -206,7 +227,7 @@ const OrdersWidget: React.FC = () => {
             <div className="ordersContainer">
               <div className="headerRow">
                 <div className="checked">
-                  <input type="checkbox" />
+                  <input type="checkbox" onClick={toggleSelectAll} checked={orders.length === markedForDeletion.size} />
```
Velenir: It's more common to use `onChange` instead of `onClick` for checkboxes and radio elements. I can't say if one is better than the other, just more common

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -73,6 +73,21 @@ const PendingLink: React.FC<Pick<Props, 'pending'>> = ({ pending }) => {
...
 
+const DeleteOrder: React.FC<Pick<Props, 'isMarkedForDeletion' | 'toggleMarkedForDeletion' | 'pending'>> = ({
+  isMarkedForDeletion,
+  toggleMarkedForDeletion,
+  pending,
+}) => (
+  <div className="checked">
+    <input
+      type="checkbox"
+      onClick={toggleMarkedForDeletion}
```
Velenir: See below about `onChange`

src/api/index.ts
```diff
@@ -16,16 +16,34 @@ import {
...
 import { INITIAL_INFURA_ENDPOINT } from 'const'
 
+const isWeb3Mock = process.env.MOCK_WEB3 === 'true'
 const isWalletMock = process.env.MOCK_WALLET === 'true'
 const isTokenListMock = process.env.MOCK_TOKEN_LIST === 'true'
 const isErc20Mock = process.env.MOCK_ERC20 === 'true'
 const isDepositMock = process.env.MOCK_DEPOSIT === 'true'
 const isExchangeMock = process.env.MOCK_EXCHANGE === 'true'
 
+// TODO connect to mainnet if we need AUTOCONNECT at all
```
anxolin: Why this change is in this PR?

anxolin:
Well, there's an issue with the layout. The checkboxes are moving from their position when you check-uncheck them, and that's a bit inconvenient:

image

Actually, this is why in my original proposal I was suggesting to always show the message, cause it was good enough imo and we prevent this cases. Anyways, now is done I would try to solve the layout.

One easy solution is using CSS (visibility), instead of display: none, or not rendering the message

src/api/index.ts
alfetopito: It's not. It was tracking another branch.

PR #334, opened 2019-12-20
Remove deposit API instance

anxolin:
Removed the instantiation of the deposit API, it's not required since it's the same contract as the deposit API.

It removes stuff, and the app apparently still works :P 

Aditionally, it renames some vars in ErcApi to be more consistent


gnosis-info:
Travis automatic deployment:
 https://pr334--dexreact.review.gnosisdev.com

alfetopito:
I need to update some contract API methods, so I'll merge it to `develop` now to avoid further conflicts.

alfetopito:
Damn, wrong branch. Figuring out how to fix this mess.

alfetopito:
Seems the option is to create a revert PR https://github.com/gnosis/dex-react/pull/350
That is merged I guess we can revert the revert and this time go for the right branch?

anxolin:
Yes, it makes sense to get rid of DepositApi if we see nor real value

PR #336, opened 2019-12-20
Move api definitions

anxolin:
This PR is part of the refactor, but since there's a lot of files involved, I didn't want to do it with the main changes for the production version relased today.

The idea is to simplify the definition of the Apis. For that, it brings the typescript definitions of the Api interfaces closer to the implementation.

Let's explain the change with `ExchangeApi`, but the same logic applies for all the other Apis. Before we had:
* `ExchangeApi` interface was declared in  `src/types`
*  All the complex types (like `PlaceOrderParams`), that are part of the interface were declared in `src/api/exchange/ExchangeApi`
* `src/api/exchange/ExchangeApiImpl` implemented `ExchangeApi`
* `src/api/exchange/ExchangeApiMock` implemented `ExchangeApi`

Now, we don't have the these types anymore in `src/types` and `ExchangeApiImpl` file was renamed to `ExchangeApi` (just the file, inside you'll find `ExchangeApiImpl`):
* `src/api/exchange/ExchangeApi`: Contains the interface, the complex types (like `PlaceOrderParams`), and the Implementation
* `src/api/exchange/ExchangeApiMock` implements `ExchangeApi`

The idea, is to have the interface closer to it's implementation keeping the files number small. Also to have the params near the interface and avoid `src/types` to grow out of control. 

Another reason is, this `API` might be a good candidate to be made a package on it's own. So it's own.

--

Additionally:
- [X] Removes the tokenList, it uses now the one in `dex-js`
- [X] Delete some remainings from the old typescript definition of the contracts.
- [X] Moves the `faucet-info.md` to a more suitable place

Apis:
image


gnosis-info:
Travis automatic deployment:
 https://pr336--dexreact.review.gnosisdev.com

src/utils/exchange.ts
```diff
@@ -0,0 +1,66 @@
...
+const UINT256_WIDTH = 32 * 2
+const UINT16_WIDTH = 2 * 2
+const UINT32_WIDTH = 4 * 2
+const UINT128_WIDTH = 16 * 2
+
+const hexPattern = '[0-9a-fA-F]'
+const hn = (n: number): string => hexPattern + `{${n}}`
+
+// /(address)(sellTokenBalance)(buyTokenId)(sellTokenId)(validFrom)(validUntil)(priceNumerator)(priceDenominator)(remainingAmount)/g
+const orderPattern = `(?<user>${hn(ADDRESS_WIDTH)})(?<sellTokenBalance>${hn(UINT256_WIDTH)})(?<buyTokenId>${hn(
```
alfetopito: I think the artifacts here belongs in the ExchangeApi or closer (in the `api/exchange` folder) rather than utils.
Not like anything else will need this, and the API depends on this to decode user orders.

anxolin: Good point, I'll move it closer

anxolin: Addressed: c2c148dfdc97fbb4e201890b02377daa887e09ea

PR #338, opened 2019-12-20
332/new balance apis from contract

alfetopito:
Closes #332
Closes https://github.com/gnosis/dex-react/issues/228

---

Adjusting frontend to contract changes regarding balance calculation/display.
Resolves the issue first reported on https://github.com/gnosis/dex-react/issues/228.

Exchange balance on the interface now depends on 3 things:
1. balance; from [`getBalance`](https://github.com/gnosis/dex-contracts/blob/master/contracts/EpochTokenLocker.sol#L163)
2. currentBatchId; from [`getCurrentBatchId`](https://github.com/gnosis/dex-contracts/blob/master/contracts/EpochTokenLocker.sol#L145)
3. pendingDeposit; from [`getPendingDeposit`](https://github.com/gnosis/dex-contracts/blob/master/contracts/EpochTokenLocker.sol#L127)

Given by:

``` typescript
// If deposit not matured yet, pending deposit is not part of balance. Add both
if (currentBatchId <= pendingDeposit.batchId) {
  totalExchangeBalance = balance + pendingDeposit.amount
} else {
// Otherwise, pending has matured and is included in the balance, nothing to do
  totalExchangeBalance = balance
}
```

Pending withdraw is not included in the calculation and does not in any way affects the displayed exchange balance, which gives the behaviour:

1. If pendingWithdraw is not matured, its amount is still part of `balance`.  Sum on exchange balance and pending withdraw == real user balance + pending withdraw > real user balance

Screenshot_20191220_141225

2. Otherwise, pendingWithdraw is deduced from `balance`. Sum on exchange balance and claimable withdraw == real user balance

Screenshot_20191220_141535

### Note:
PR size is as small as it gets. This is a core change touches many files.

PR #349, opened 2019-12-23
Remove update token

alfetopito:
Follow up from https://github.com/gnosis/dex-react/pull/338#pullrequestreview-335556679

Removing 'manual' interface updates.
Balances states now becomes read only, with the updates coming directly from the network.

Keep in mind this makes the mock behaviour stop working for the deposits page.

### DIFF
--- a/src/components/DepositWidget/index.tsx
+++ b/src/components/DepositWidget/index.tsx
@@ -23,7 +23,7 @@ interface WithdrawState {
 }
 
 const DepositWidget: React.FC = () => {
-  const { balances, setBalances, error } = useTokenBalances()
+  const { balances, error } = useTokenBalances()
   const {
     // Dispatchers
     enableToken,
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr349--dexreact.review.gnosisdev.com

alfetopito:
> It closes the form right after you click on the "withdraw button" in the form. You cannot see the "loading" in the same way as you see it for the deposit. 

Right, guess for this we'll still need to update some states. I thought the redux would take care of that but apparently not quite.

> Also, if you have already a pending claim, when you click the "claim button" it closes the form and then it shows the "confirm modal"

That's the current behaviour even without these changes. Just checked on https://dex.dev.gnosisdev.com/wallet. The modal is displayed after the form is closed, and no loading is shown after. Would you rather the form is not closed when the modal is displayed?

anxolin:
Mm, I would have sweared that I check dev and it was not happening. But if it's not from this PR we can leave that out. Still, we need to review it, cause before that part it was behaving as "deposit" do

alfetopito:
> Mm, I would have sweared that I check dev and it was not happening. But if it's not from this PR we can leave that out. Still, we need to review it, cause before that part it was behaving as "deposit" do

As discussed offline, the behaviour was not introduced with the changes here. Created an issue to deal with them https://github.com/gnosis/dex-react/issues/361

PR #351, opened 2019-12-24
hack for hosting on ipfs

Velenir:
This isn't ideal and modifies base.href before scripts are injected
A better solution would be to adopt HashRouter, which also is not ideal

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -14,7 +14,7 @@
         baseHref = ipfsMatch[0]
         if (!baseHref.endsWith('/')) baseHref += '/'
 
-        window.BASE_URL = ipfsMatch[0]
+        window.IS_IPFS = true
         
       } else {
         baseHref = '/'
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -5,13 +5,13 @@
   <%if (htmlWebpackPlugin.options.ipfsHack) { %>
   <script>
     (function () {
-      const { origin, pathname } = window.location
+      const { pathname } = window.location
       const ipfsMatch = /.*\/Qm\w{44}/.exec(pathname)
       const base = document.createElement('base')
       let baseHref
       if (ipfsMatch) {
 
-        baseHref = origin + ipfsMatch[0]
+        baseHref = ipfsMatch[0]
         if (!baseHref.endsWith('/')) baseHref += '/'
 
         window.BASE_URL = ipfsMatch[0]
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr351--dexreact.review.gnosisdev.com

src/html/index.html
```diff
@@ -1,11 +1,36 @@
...
+  <%if (htmlWebpackPlugin.options.ipfsHack) { %>
+  <script>
+    (function () {
+      const { origin, pathname } = window.location
+      const ipfsMatch = /.*\/Qm\w{44}/.exec(pathname)
+      const base = document.createElement('base')
+      let baseHref
+      if (ipfsMatch) {
+
+        baseHref = origin + ipfsMatch[0]
```
anxolin: Why is the origin required, cannot be relative for example to the domain root?

For example:
* **For url**: https://gateway.ipfs.io/ipfs/QmeYYwD4y4DgVVdAzhT7wW5vrvmbKPQj8wcV2pAzjbj886/docs/getting-started/
* **Base href would be**:`/ipfs/QmeYYwD4y4DgVVdAzhT7wW5vrvmbKPQj8wcV2pAzjbj886/`

src/html/index.html
```diff
@@ -1,11 +1,36 @@
...
+        if (!baseHref.endsWith('/')) baseHref += '/'
+
+        window.BASE_URL = ipfsMatch[0]
```
anxolin: Do we really need the global? can we use the `<base /> href` instead?

Velenir: `base.href` is same as `document.baseURI`, it prepends origin. We'd have to recalculate anyway

Velenir:
> * Don't do the hack using a helper/util that is used in `App.tsx` to initialize the app, so there's no need to modify and parametrize the template

Can't. `<scripts src=*>` in html are unaffected by runtime, need a hack applied before they are parsed by the browser

src/html/index.html
Velenir: Not required, no real reason I included it.

Velenir:
This is still not an ideal solution as you can't open a page relative to root. Because ipfs doesn't have 404 redirects or any other way to implement historyApi

i.e. https://gateway.ipfs.io/ipfs/Qm***/trade won't work

anxolin:
@Velenir what about, instead of this hack we decide to use a different router if we see that we are in a IPFS server?

PR #353, opened 2019-12-25
smaller wait for autoconnect with new events

Velenir:
New events allow for smaller delay between connection attempts


gnosis-info:
Travis automatic deployment:
 https://pr353--dexreact.review.gnosisdev.com

Velenir:
> So, we still need to wait even with the events?

There's still no event for _I tried to connect but couldn't_

PR #355, opened 2019-12-25
fix inline env.vars

Velenir:
`main.a4c5.js   781 KiB` >> `main.026c.js   778 KiB`
Change is minimal but this is the correct usage for tree shaking


gnosis-info:
Travis automatic deployment:
 https://pr355--dexreact.review.gnosisdev.com

PR #357, opened 2019-12-26
Fix deposit withdraw hidden on mount

Velenir:
Adds scroll into view when Deposit/Withdraw Form is opened below the viewport

localhost_8080_wallet


gnosis-info:
Travis automatic deployment:
 https://pr357--dexreact.review.gnosisdev.com

PR #358, opened 2019-12-26
Fix misaligned token

Velenir:
I chose alignment that looked the best to me
localhost_8080_wallet (1)

Another way would be to set `justify-content: space-between` on the containing div, if anyone wants to play with styles
localhost_8080_wallet (2)

If we want text left aligned or something different, we will need a div wrapper around the image of a fixed width

I wish subgrids had better support , this would be so much easier


gnosis-info:
Travis automatic deployment:
 https://pr358--dexreact.review.gnosisdev.com

Velenir:
> I think the issue is the image and label use the same flex item. I understand that if we use different flex items for them they would be aligned:

Well yes, that's what I wrote

> If we want text left aligned or something different, we will need a div wrapper around the image of a fixed width

alfetopito:
> I think it's better, but although the images are aligned, I think we need to align also the name of the tokens.
> 
> image
> 
> I think the issue is the image and label use the same flex item. I understand that if we use different flex items for them they would be aligned:
> image

I second Anxo. The names would be better left aligned.

W3stside:
so i played with this in the DEpositPage refactor, but found that I preferred the misaligned token images over the rest. But as everyone seems set on token image alignment, I'll also second left align name

Velenir:
Grid ftw

localhost_8080_wallet (3)

W3stside:
@Velenir would be cool to merge #369 first into here before merging this

PR #360, opened 2019-12-26
Update to last version of dex-js

anxolin:
Uses the updated version of dex-js: https://github.com/gnosis/dex-js/pull/44

For the web, not much change is needed, only that we can make use of the exported and typed ABI


gnosis-info:
Travis automatic deployment:
 https://pr360--dexreact.review.gnosisdev.com

PR #362, opened 2019-12-27
209/cancel orders

alfetopito:
Cancelling selected orders #209 

- Updating API to cancel multiple orders at once as opposed to 1 at a time.
- Triggering cancellation from Orders page
- Displaying order as pending while transaction is mined

### NOT in this PR:
- Hide orders that are cancelled/expired
- Button to display/hide cancelled/expired orders

PR #363, opened 2019-12-28
209/hide expired orders

alfetopito:
Dependent on #362 
Part of #209 

---

- Not displaying orders that are already expired.
- Disabling all checkboxes while deleting.
- Adding back Orders link to nav bar

### NOT part of this PR:
- Button to show/hide expired orders https://github.com/gnosis/dex-react/pull/370

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useCallback } from 'react'
+import React, { useMemo, useCallback, useEffect } from 'react'
 import { Link } from 'react-router-dom'
 import styled from 'styled-components'
 import { faExchangeAlt, faChartLine, faTrashAlt, faExclamationTriangle } from '@fortawesome/free-solid-svg-icons'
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr363--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -160,7 +164,16 @@ const OrdersForm = styled.div`
 `
 
 const OrdersWidget: React.FC = () => {
-  const orders = useOrders()
+  const allOrders = useOrders()
+
+  const [orders, setOrders] = useSafeState<AuctionElement[]>(allOrders)
+
+  useEffect(() => {
+    const filteredOrders = allOrders.filter(order => order.validUntil > dateToBatchId())
```
Velenir: Why do you filter in `useEffect`? I don't see why it needs to be done async

W3stside: my only comment, otherwise looks good

alfetopito: Because `useOrders` is async. At first it returns an empty list https://github.com/gnosis/dex-react/blob/develop/src/hooks/useOrders.ts#L7

alfetopito: And I'm using state because once delete is complete I want to remove the orders from the UI right away.

PR #370, opened 2019-12-30
209/show expired orders

alfetopito:
Part of #209 
Depends on #363 

---

**WARNING**: First iteration of the UI. Open to better ideas.

- [x] Allowing user to see expired orders
- [x] Showing count of active and expired orders
- [x] Filtering out deleted orders

### Active:
Screenshot_20191230_144120

### Expired:
Screenshot_20191230_144148

### What currently does not work
- After cancelling/deleting, the order is removed from the local list of orders. But when switching between active/expired it's shown again.
- Local list of orders is only updated after a page refresh. How can I update it based on blocks or batchId?

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -12,7 +12,14 @@ import { getTokenFromExchangeById } from 'services'
 import useSafeState from 'hooks/useSafeState'
 import { TokenDetails } from 'types'
 
-import { safeTokenName, formatAmount, formatAmountFull, isBatchIdFarInTheFuture, formatDateFromBatchId } from 'utils'
+import {
+  safeTokenName,
+  formatAmount,
+  formatAmountFull,
+  isBatchIdFarInTheFuture,
+  formatDateFromBatchId,
+  isOrderActive,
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr370--dexreact.review.gnosisdev.com

anxolin:
I would comment out and add a TODO for the "Create new strategy" part after this PR get's approved.

It's good to have the markup already, but better not to show this part until we implement it

anxolin:
@alfetopito  For the tabs, what do you think about using this: https://codepen.io/woranov/pen/NRqLWK/

alfetopito:
> @alfetopito For the tabs, what do you think about using this: https://codepen.io/woranov/pen/NRqLWK/

That does look pretty but also feels overly complicated with ~250 lines of CSS. Am I just being lazy? Very likely.

anxolin:
> That does look pretty but also feels overly complicated with ~250 lines of CSS

Just forget it, I just did a quick google search. There's nice components out there, I like this one visually. I saw a lot of code, but I assumed it we needed only a small portion of it. I'll write something as a super low prio just to not lose the link

PR #386, opened 2020-01-02
209/update orders on new block

alfetopito:
Subscribing to new blocks from wallet info.
Updating orders on every new block.

### DIFF
--- a/src/api/wallet/WalletApi.ts
+++ b/src/api/wallet/WalletApi.ts
@@ -46,6 +46,7 @@ export interface WalletInfo {
   isConnected: boolean
   userAddress?: string
   networkId?: number
+  blockNumber?: number
 }
 
 export type ProviderInfo = ReturnType<typeof Web3Connect.getProviderInfo>
@@ -347,7 +348,7 @@ export class WalletApiImpl implements WalletApi {
         isConnected: !!userAddress && !!networkId,
       }
     } catch (error) {
-      log('Error asynchrously getting WalletInfo', error)
+      log('Error asynchronously getting WalletInfo', error)
       return {
         userAddress: '',
         networkId: 0,
@@ -362,8 +363,9 @@ export class WalletApiImpl implements WalletApi {
     await Promise.resolve()
 
     const walletInfo = await (this.getWalletInfo() || this._getAsyncWalletInfo())
+    const wInfoExtended = { ...walletInfo, blockNumber: blockchainUpdate?.blockHeader?.number }
 
-    this._listeners.forEach(listener => listener(walletInfo))
+    this._listeners.forEach(listener => listener(wInfoExtended))
   }
 
   private get _connected(): boolean | Promise<boolean> {
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -28,7 +28,7 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
 }
 
 export function useOrders(): AuctionElement[] {
-  const { userAddress, networkId } = useWalletConnection()
+  const { userAddress, networkId, blockNumber } = useWalletConnection()
   const [orders, setOrders] = useSafeState<AuctionElement[]>([])
 
   useEffect(() => {
@@ -38,7 +38,8 @@ export function useOrders(): AuctionElement[] {
         .getOrders({ userAddress, networkId })
         .then(filterDeletedOrders)
         .then(setOrders)
-  }, [networkId, setOrders, userAddress])
+    // updating list of orders on every block by listening on `blockNumber`
+  }, [networkId, setOrders, userAddress, blockNumber])
 
   return orders
 }


gnosis-info:
Travis automatic deployment:
 https://pr386--dexreact.review.gnosisdev.com

src/api/wallet/WalletApi.ts
```diff
@@ -279,10 +280,10 @@ export class WalletApiImpl implements WalletApi {
 
   /* ****************      Private Functions      **************** */
 
-  private async _notifyListeners(): Promise<void> {
-    await Promise.resolve()
+  private async _notifyListeners(options?: { blockHeader: { number: number } }): Promise<void> {
```
anxolin: who is using this options? The only other call to this private function doesn't provide any params

src/api/wallet/WalletApi.ts
```diff
@@ -43,6 +43,7 @@ export interface WalletInfo {
   isConnected: boolean
   userAddress?: string
   networkId?: number
+  blockNumber?: number
```
anxolin: I don't know if it's a good idea to add the block number here. This would update this object every time there's a new block number. 

Normally most components just need the userAddress or networkId, so they are reevaluated if these changes. Adding here the blockNumber might make this components either to redraw or to add some code to prevent it explicitly

src/api/wallet/WalletApi.ts
```diff
@@ -279,10 +280,10 @@ export class WalletApiImpl implements WalletApi {
 
   /* ****************      Private Functions      **************** */
 
-  private async _notifyListeners(): Promise<void> {
-    await Promise.resolve()
+  private async _notifyListeners(options?: { blockHeader: { number: number } }): Promise<void> {
     const walletInfo: WalletInfo = this.getWalletInfo()
-    this._listeners.forEach(listener => listener(walletInfo))
+    const wInfoExtended = { ...walletInfo, blockNumber: options?.blockHeader?.number }
```
anxolin: Also, I don't get the logic cause I'm not sure how is this used, but in case if we wanted to send the blockNumber, why add the blockHeader in the arguments?

alfetopito: Actually, I just copied this form @W3stside pending PR https://github.com/gnosis/dex-react/pull/291

As for `blockHeader`, that's because it contains the `blockNumber`?

src/api/wallet/WalletApi.ts
alfetopito: I believe that won't be a problem because if components don't need `blockNumber` they can just not subscribe to it and React is smart enough to not update those components. Not the case?

src/api/wallet/WalletApi.ts
alfetopito: Yeah, couldn't find a use for this. Guess @W3stside had some use in mind before but the part that I shamelessly copied has no use for it.

Will remove.

src/api/wallet/WalletApi.ts
anxolin: But if it's there, then `addOnChangeWalletInfo` will register for any change in walletInfo, including the block

I feel it's strange that a block is part of the wallet info.

I would understand a separate event for the block

src/api/wallet/WalletApi.ts
anxolin: I'm a bit confused. Maybe we can talk it directly

src/api/wallet/WalletApi.ts
anxolin: Let's discuss it in person, cause I'm a bit confused by this

src/api/wallet/WalletApi.ts
W3stside: @anxolin 

we want to auto update state right? well that can be done via polling (e.g `setInterval(updateState(), 15seconds)` or via a websocket and some catalyst. In this case, we're using a Websocket connection to `infura` and updating the state when the catalyst updates, which in this case is a new `blockHeader` (`blockNumber`) - this tells us, hey the blockchain was updates, i.e a block was mined, so let's update state.

In conjunction with a `useEffect` hook (or `useMemo`) and setting the dependency as a new block header, we can auto update state that we want without having to manually poll.

anxolin: I understand, but I don't see where this logic is done, I don't find it in the PR. Maybe you can show us in our sync. Also, I don't get why this "blockNumber" needs to be part of the wallet information instead on it's own.

Velenir: All updates happen in [`subscribeToBlockchainUpdate`](https://github.com/gnosis/dex-react/blob/41b8995883b72781b6d0c4396e2ba2a987f38978/src/api/wallet/WalletApi.ts#L118)

src/hooks/useOrders.ts
```diff
@@ -37,7 +37,8 @@ export function useOrders(): AuctionElement[] {
         .getOrders(userAddress)
         .then(filterDeletedOrders)
         .then(setOrders)
-  }, [setOrders, userAddress])
+    // updating list of orders on every block by listening on `blockNumber`
+  }, [setOrders, userAddress, blockNumber])
```
Velenir: Really like this kind of responsibility delegation, form WalletInfo to effect deps to exchangeApi

anxolin: Here is a thought. Wouldn't this approach kill the battery of a phone if we get orders on every block, decode them etc?

For stuff with events, shouldn't be more like getPastevents, from last queried block until this `blockNumber`. Possibly this just means, that it would be a performance improvement scoped INSIDE the Api implementation and the hook don't care about that --> interesting for cache logic

cc/ @Velenir @alfetopito

alfetopito: From the battery perspective, will the page keep updating if not on focus? Don't know how mobile handles this, but my expectation is that whatever background task a page runs only does so when open.

For the second part, I don't really get what you mean.

Velenir: I don't think there's a difference client-side between calling `getPastEvents` and `getOrders`. Both those calls are encoded into rpc calls and sent to infura or wherever we are connected.

The only palpable difference as I see it, can be between calls and WebSocket subscriptions to events. And that is not something we can count on. Some providers do subscriptions through websockets, some do polling. MMask does polling only, as far as I can tell.

anxolin: My point was more like, your we getOrders on every block? it seams like a lot of work and I see nodes banning us.

WebSocket would solve it. 

My point is, only if there's new events for an user we have to load ALL his orders. If I know the block of the last event for the user, and I know the last block I queried the orders in the contract, I know if I have to do the call.

The battery thing, was just to point out that the app is using a lot the network, and lot of JS is being executed every few seconds, when there's no really new orders coming in for the user.

Issue #329, opened 2019-12-18
Implement cache/proxy layer for APIs


alfetopito:
- [ ] Implement cache layer
- [ ] Make APIs network agnostic

----

> > While some APIs are network aware, tokenList API is not, thus I pass around the networkId.
> Guess it would make sense to make tokenList API also network aware like the others?

> Being coherent makes sense yes. But I was wondering the opposite. If it makes sense to add `networkId` on every API method so they don't have "state/context"

> When you cache, the symbol, you need to use both the `address` and the `networkId` for the key, if `networkId` is not in the params is more complicated

> _Originally posted by @anxolin in https://github.com/gnosis/dex-react/pull/316#issuecomment-567117278_

PR #388, opened 2020-01-03
329/refactor api params

alfetopito:
Part of #329 

First step: refactoring API parameters.

### Reasons:
- Consistency across apis
- Remove inline type definitions
- Rename a few variables to be in line with other methods
- Make it easier to introduce additional parameters. Namely, `networkId`, for the next step
- Merge `txOptionalParams` for methods that accept it into a single params object

### Next step:
- Introduce `networkId` to DepositApi and ExchageApi to make them network agnostic
- Other apis (Erc20Api, TokenListApi) are network agnostic already, and Wallet api is the one providing the networkId in the first place.

### DIFF
--- a/src/api/deposit/DepositApi.ts
+++ b/src/api/deposit/DepositApi.ts
@@ -24,7 +24,7 @@ export interface DepositParams extends ReadOnlyParams, WithTxOptionalParams {
 
 export type RequestWithdrawParams = DepositParams
 
-export type WithdrawParams = ReadOnlyParams & WithTxOptionalParams
+export type WithdrawParams = Omit<RequestWithdrawParams, 'amount'>
 
 export interface DepositApi {
   getContractAddress(networkId: number): string | null
--- a/src/api/deposit/DepositApi.ts
+++ b/src/api/deposit/DepositApi.ts
@@ -9,46 +9,32 @@ import { Receipt, TxOptionalParams } from 'types'
 import Web3 from 'web3'
 import { getProviderState, Provider, ProviderState } from '@gnosis.pm/dapp-ui'
 
+export interface ReadOnlyParams {
+  userAddress: string
+  tokenAddress: string
+}
+
+export interface DepositParams extends ReadOnlyParams {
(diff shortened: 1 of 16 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr388--dexreact.review.gnosisdev.com

src/api/deposit/DepositApi.ts
```diff
@@ -9,46 +9,36 @@ import { Receipt, TxOptionalParams } from 'types'
 import Web3 from 'web3'
 import { getProviderState, Provider, ProviderState } from '@gnosis.pm/dapp-ui'
 
+export interface ReadOnlyParams {
```
anxolin: Consider not exporting this, and naming it `UserTokenPair` then you can define and export an alias for every `<method>Params` you want

alfetopito: Done.

src/api/erc20/Erc20Api.ts
```diff
@@ -9,15 +9,20 @@ import { toBN } from 'utils'
 
 import Web3 from 'web3'
 
-export interface ContractReadOnlyParams {
+interface BaseParams {
```
anxolin: nit: I'm ok with this, but don't you prefer a name that would tell you what's inside?

for example `TokenAddress`

The lines like this one would read easily:
`export type NameParams = TokenAddress`

alfetopito: Would the same style as this be better?

```typescript
interface WithTxOptionalParams {
  txOptionalParams?: TxOptionalParams
}
```

Thus we'd have:

```typescript
interface WithTokenAddress {
  tokenAddress: string
}
```

alfetopito: My concern with this is we might need to create one interface per parameter.
On the follow up PR (#395) I'm adding `tokenId` to the BaseParams interface.

The main reason for creating this base interface was so I could add stuff to it that is available to all methods without lots of changes.

PR #395, opened 2020-01-06
329/network agnostic apis

alfetopito:
Part of #329 
Follow up to/depends on #388 

Making Erc20, Deposit and Exchange APIs network agnostic.
All calls that do contract interaction now require `networkId`

### Next step
Start proxy/cache layer implementation.

### DIFF
--- a/src/api/deposit/DepositApi.ts
+++ b/src/api/deposit/DepositApi.ts
@@ -7,11 +7,11 @@ import { getAddressForNetwork } from './batchExchangeAddresses'
 import { Receipt, TxOptionalParams } from 'types'
 
 import Web3 from 'web3'
-import { getProviderState, Provider, ProviderState } from '@gnosis.pm/dapp-ui'
 
 interface ReadOnlyParams {
   userAddress: string
   tokenAddress: string
+  networkId: number
 }
 
 export type GetBalanceParams = ReadOnlyParams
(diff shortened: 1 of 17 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr395--dexreact.review.gnosisdev.com

W3stside:
Aren't networks automatically injected by Web3 when calling these contract methods? @Velenir @alfetopito @anxolin

src/api/deposit/DepositApi.ts
```diff
@@ -188,18 +184,25 @@ export class DepositApiImpl implements DepositApi {
 
     assert(address, `EpochTokenLocker was not deployed to network ${networkId}`)
 
-    // as string as assert is not detected by TS
-    return this._getContractAtAddress(address)
+    return this._getContractAtAddress(networkId, address)
   }
 
-  protected _getContractAtAddress(address: string): BatchExchangeContract {
-    const contract = DepositApiImpl._contractsCache[address]
```
anxolin: Interesting, we would have a collision in the previous implementation if the address where the same in rinkeby and mainnet --> it's possible if u use the same account

anxolin:
> Aren't networks automatically injected/detected by Web3 when calling these contract methods?

Yes. This approach is more general, it doesn't depend in the internal state. It ease to implement the cache layer, and reusability for other non-metamask based systems.

src/api/deposit/DepositApi.ts
alfetopito: We won't because I've updated the contract cache to be based on `networkId`. Check:
- https://github.com/gnosis/dex-react/blob/329/network-agnostic-apis/src/api/deposit/DepositApi.ts#L56
- https://github.com/gnosis/dex-react/blob/329/network-agnostic-apis/src/api/deposit/DepositApi.ts#L190-L207
- https://github.com/gnosis/dex-react/blob/329/network-agnostic-apis/src/api/erc20/Erc20Api.ts#L75
- https://github.com/gnosis/dex-react/blob/329/network-agnostic-apis/src/api/erc20/Erc20Api.ts#L201-L217

PR #397, opened 2020-01-08
Add more decimals to order details

anxolin:
Closes #347 

Doing this issue I realized that the formatting of the price was redundant, so there's a second PR for making a new util and refactoring between the trade widget and the orders:
* https://github.com/gnosis/dex-react/pull/398

PR #398, opened 2020-01-08
Calculate price

anxolin:
* Creates a format price function
* Refactor between orders and trade details regarding the format of the price
* It leaves for decimals as a default
* Improves https://github.com/gnosis/dex-react/pull/397 

Doesn't include: any styling. With more decimals, the rows are a big broken. I can easily fix this, but @W3stside is working on the styling of the orders and I want to separate the 2 PRs to avoid any conflict.

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -1,6 +1,7 @@
 import BN from 'bn.js'
 import { TEN, DEFAULT_PRECISION } from 'const'
 import { TokenDetails } from 'types'
+import BigNumber from 'bignumber.js'
 
 const DEFAULT_DECIMALS = 4
 const ELLIPSIS = '...'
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr398--dexreact.review.gnosisdev.com

src/utils/format.ts
```diff
@@ -140,3 +141,21 @@ export function safeFilledToken<T extends TokenDetails>(token: T): T {
     symbol: token.symbol || token.name || abbreviateString(token.address, 6, 4),
   }
 }
+
+export function calculatePriceBigNumber(numerator?: BigNumber, denominator?: BigNumber): BigNumber | null {
+  if (!numerator || !denominator) {
+    return null
+  }
+
+  return numerator.dividedBy(denominator)
```
W3stside: Can `denominator` ever be `0` here? I believe `BN` throws in this instance. Should we check `isZero(denominator)` here?

anxolin: I think they handle that case. Any number divided by 0 is infinity
It's the responsibility of the caller to check that if infinity is not a good number for the use-case.

image

anxolin: Also, you can even format that ;)

image

anxolin: On a second thought, I'm realizing it might not make any sense to return Infinity when showing prices. So it's more convenient to return null, otherwise we have to check `bn.isInfinity()` or ensure the denominator is not 0.

I've made this commit for convenience: cef5704a32e85302e819841248e387058be40002

PR #405, opened 2020-01-08
329/cache layer

### LINKED ISSUE #329: Implement cache/proxy layer for APIs

alfetopito:
Closes #329 

Adding a cache layer to the api layer.

Doing so by implementing a Proxy. Exact same api as regular implementation, with caching rules defined per API on the proxy.

For now, only implemented Erc20ApiProxy to get feedback.
Once everyone is happy I'll do the same for relevant APIs.

Some debug logging:

Screenshot_2020-01-08_15-10-12

### EDIT 1
Added another approach using `CacheMixin` on `Erc20ApiProxyV2`.
This one extends `Erc20ApiImpl` and holds an instance of the `CacheMixin` class.

On the constructor it configures which methods should be cached, overwriting parent class' methods.

Please take a look at both and let me know which you think is best.

I'm right now in favour of the second approach for making the cache impl cleaner.

### EDIT 2
Approach 2 was chosen.
Added cache to Erc20, Deposit and Exchange APIs, for methods where it makes sense at this point.
It's trivial to add new methods with custom `ttl` and `hash functions` if needed.


gnosis-info:
Travis automatic deployment:
 https://pr405--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr405--dexreact.review.gnosisdev.com

src/api/proxy/CacheProxy.ts
```diff
@@ -0,0 +1,86 @@
...
+    // cache was found
+    console.debug(`cache hit for ${hash}`)
+    return cacheObj.obj
+  }
+
+  private store<O>(hash: string, obj: O): void {
+    this.cache[hash] = { obj, createdOn: new Date() }
+  }
+
+  protected async fetchWithCache<P, R, M = keyof T>(method: M, params: P, expiration?: number): Promise<R> {
```
Velenir: Is this a good idea to mark method as async? That would make some sync cached functions unnecessarily sync

src/api/erc20/Erc20ApiProxy.ts
```diff
@@ -0,0 +1,58 @@
...
+  ApproveParams,
+  TransferParams,
+  TransferFromParams,
+} from './Erc20Api'
+import { CacheProxy } from 'api/proxy'
+
+const DEFAULT_CACHE_TIME = 30
+
+export class Erc20ApiProxy extends CacheProxy<Erc20Api> implements Erc20Api {
+  public name(params: NameParams): Promise<string> {
```
Velenir: Would it be better, instead of rewriting each function to use `this.fetchWithCache` inside, to accept an params object in constructor with a list of functions to cache and then in constructor:
```
for (const fname of toCache) {
  this[fname] = cacheHOF(api[fname])
}
```

same can be done with TS for types

src/api/proxy/CacheProxy.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Assumes all values being hashed can be converted to string
+   * Sorts parameters for determinism
+   *
+   * TODO: replace this with an actual hash function once testing is done
+   *
+   * @param params The params we want to hash
+   *
+   */
+  private hashParams<P>(methodName: string, params: P): string {
+    return Object.keys(params)
```
anxolin: cool, this is to generate the cache keys automatically using the params

We should be careful with the type of the params we use, I understand. Maybe they don't serialize properly into a nice "value"

src/api/erc20/Erc20ApiProxy.ts
```diff
@@ -0,0 +1,58 @@
...
+  TotalSupplyParams,
+  BalanceOfParams,
+  AllowanceParams,
+  ApproveParams,
+  TransferParams,
+  TransferFromParams,
+} from './Erc20Api'
+import { CacheProxy } from 'api/proxy'
+
+const DEFAULT_CACHE_TIME = 30
```
anxolin: Why 30? at least the symbol, and so on, we can catch forever, right?

alfetopito: I didn't get to test it, but one use case you mentioned before was that a token might be updated with those details?

Now that I'm thinking about it makes little sense, we could just cache those forever indeed.

src/api/proxy/CacheProxy.ts
alfetopito: Yeah, this can be a problem depending on the params. 
I haven't done so but considered allowing the proxy to pass to a custom hashing function, so the caller can exclude whatever might not be hashable.
The other alternative is to make this naive hashing function more robust and able to filter out complex/un-hashable types.

What would you prefer?

src/api/erc20/Erc20ApiProxy.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export class Erc20ApiProxy extends CacheProxy<Erc20Api> implements Erc20Api {
```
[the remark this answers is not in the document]
alfetopito: > Are you sure we want to use extension instead of composition?

Not really. This is the approach that made more sense at the time, but ok trying a different way.

src/api/proxy/CacheProxy.ts
```diff
@@ -0,0 +1,86 @@
+import { addSeconds } from 'date-fns'
```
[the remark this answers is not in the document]
alfetopito: Hmm cool, taking a look.

src/api/proxy/CacheProxy.ts
alfetopito: Did not think much about it now because the apis I want to cache are all async (contract calls).

But yeah, this did bother me, but can I use the same method for both sync and async versions?

src/api/proxy/CacheProxy.ts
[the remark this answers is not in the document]
anxolin: I don't know how big is the library anyways. That would be my only concern. But if we have a lib in general to use it. 

The good thing. With your approach, we only need to depend in a single place for the underlying cache library/technology

src/api/erc20/Erc20ApiProxy.ts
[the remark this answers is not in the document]
anxolin: I'm happy either way. I'm kind of thinking aloud

src/api/erc20/Erc20ApiProxy.ts
alfetopito: That looks neat!
Yes, this is something I would pretty much would like to have.
The reason for re-implementing all methods is due to the fact that I'm implementing the api interface and TS forced me to implement everything.
Will give this a try.

src/api/proxy/CacheProxy.ts
anxolin: Would it make sense an overloaded `fetchWithCache` with three params `cacheKey`,  and a `getterFn` function "() => T", and `expiration`?

src/api/proxy/CacheProxy.ts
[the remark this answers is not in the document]
alfetopito: From what I've seen, we can simply replace the private `cache` object instance with this library's instance and update the get/store methods and it's pretty much done.

At this point though, I don't see the need for an external lib for this simple caching.
It does provides some niceties like [events](https://www.npmjs.com/package/node-cache#events) and ttl, but as you mentioned, it'll increase our build size a bit more... hehe

Unless it does the caching in a magic way that makes it much more efficient/fast?

Will try regardless to see how better it is.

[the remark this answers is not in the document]
alfetopito: Added on da104f7dc5d0b80de5fe13872f4f02633b8d8abd

Build size Before:
```
Assets:
  main.0870.js (799 KiB)
  vendors~main.74ff.js (17.2 MiB)
Entrypoints:
  main (17.9 MiB)
```

After:
```
Assets:
  main.9b3d.js (800 KiB)
  vendors~main.806f.js (17.2 MiB)
Entrypoints:
  main (18 MiB)
```

Not a big deal, guess we can live with that.

src/api/erc20/Erc20ApiProxy.ts
alfetopito: Done

src/api/proxy/CacheProxy.ts
alfetopito: Added it on 352f9c316d6b10628dfc4867c7977dff32fca78e

src/api/erc20/Erc20ApiProxy.ts
alfetopito: Didn't quite get it working...

Pushed my attempt so far to https://github.com/gnosis/dex-react/commit/9855180f9fa46b5041a323807fb66ce3d0345f89

The proxy implementation doesn't recognize the class as having the methods implemented:

```typescript
class TestApiProxy extends CacheProxy<TestApi> implements TestApi {
  public constructor(api: TestApi) {
    super(api, [{ methodType: 'cachedMethod', expiration: 10 }])
  }

  // public async cachedMethod(params: Params): Promise<string> {
  //   return this.fetchWithCache('cachedMethod', params, 10)
  // }
  // public async nonCachedMethod(params: Params): Promise<string> {
  //   return this.fetchWithCache('nonCachedMethod', params)
  // }
}
```

TS warning: 
```
Class 'TestApiProxy' incorrectly implements interface 'TestApi'.
  Type 'TestApiProxy' is missing the following properties from type 'TestApi': cachedMethod, nonCachedMethod
```

src/api/proxy/CacheProxy.ts
anxolin: I think async is fine.
By definition, we want to cache stuff that takes time otherwise, and if it takes time should be async.

I'd leave it async until we find an issue, then we can think if we just create a `fetchWithCacheSync`  or whatever solution we came up with

src/api/proxy/CacheProxy.ts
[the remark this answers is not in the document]
anxolin: Nice!

Vendors is out of control. I'll increase PRIO for the review of that. We cannot ignore it another sprint 

Bundle size is out of control!!
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FxT8qBuhi7JJbdeqS64%2Fgiphy.gif&token=out_of_control" data-gitmeme-token="out_of_control"></a>

src/api/proxy/CacheProxy.ts
Velenir: > but can I use the same method for both sync and async versions?

don't await, cache a promise and return that promise

alfetopito: Ohhhh, cache the promise! Didn't know that would work. Trivial change, will try it out.

alfetopito: Done :)

src/api/erc20/Erc20ApiProxy.ts
alfetopito: Managed to get TS happy in a way that we don't have to re-implement all methods. Take a look at the latest revision.

src/api/erc20/Erc20ApiProxy.ts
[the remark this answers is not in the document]
alfetopito: I tried a different approach in the latest version. Take a look if you can, otherwise you can give your opinion when you get back from vacation :)

src/api/proxy/CacheProxy.ts
```diff
@@ -0,0 +1,70 @@
+import NodeCache from 'node-cache'
+
+export class CacheProxy<T> {
+  protected api: T
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
W3stside: is this still necessary?

src/api/proxy/CacheProxy.ts
```diff
@@ -0,0 +1,70 @@
...
+  private get<R>(hash: string): R | undefined {
+    const obj = this.cache.get(hash)
+
+    if (obj) {
+      console.debug(`cache hit for ${hash}`)
+      return obj as R
+    }
+
+    return
+  }
```
W3stside: could be rewritten as:

```ts
    ....
    if (!obj) return

    console.debug(`cache hit for ${hash}`)
    return obj
  }
```

to avoid `as blah` ts code

src/api/proxy/CacheProxy.ts
alfetopito: Nope, forgot to remove. There are probably more like that, will remove them.

src/api/proxy/CacheProxy.ts
alfetopito: I was missing the type on this.get()

But applying your suggestion regardless, looks better

src/api/proxy/CacheMixin.ts
```diff
@@ -0,0 +1,113 @@
...
+   * Injects into provided `instance` a version with cache according to the list `toCache`.
+   * Injected methods will obey `ttl` if provided, otherwise global cache config.
+   *
+   * @param instance Class instance to cache
+   * @param toCache List of configs for methods to be cached
+   */
+  public injectCache<T>(instance: T, toCache: CacheOptions<T>[]): void {
+    toCache.forEach(cacheConfig => {
+      const { method } = cacheConfig
+      const methodName = method.toString()
```
Velenir: Isn't method already a string, or TS complains?

alfetopito: No, method is a type. TS does complain.

PR #409, opened 2020-01-10
Gas station

### LINKED ISSUE #348: Gas station API
Create a new API that estimate the gas price so we avoid Metamask errors 

* Mainnet: https://safe-relay.gnosis.pm/api/v1/gas-station
* Rinkeby: https://safe-relay.staging.gnosisdev.com/api/v1/gas-station

For any other network, we can leave it undefined so it's estimated automatically by the wallet.

Velenir:
fetch **gasPrice** from gas station
only fetch when network or block has changed, cache otherwise

Maybe a better place for this logic would be in the upcoming Cache Layer. Then it could add `gasPrice` to `txOptionalParams` and pass onto the real `api.method`:

```ts
// in cache layer method
// only if method will send tx, not a call method

const txOptions = {
...txOptionalParams,
gasPrice: await fetchGasPrice(),
}

api.realMethod(args, txOptions)
```

Closes #348

### DIFF
--- a/src/api/gasStation.ts
+++ b/src/api/gasStation.ts
@@ -1,0 +1,65 @@
+import { walletApi } from './'
+
+const GAS_STATIONS = {
+  1: 'https://safe-relay.gnosis.pm/api/v1/gas-station/',
+  4: 'https://safe-relay.staging.gnosisdev.com/api/v1/gas-station/',
+}
+
+const GAS_PRICE_LEVEL: Exclude<keyof GasStationResponse, 'lastUpdate'> = 'standard'
+
+let cacheKey = ''
+let cachedGasPrice: undefined | string
+
+const constructKey = ({ blockNumber, chainId }: GasPriceCacheDeps): string => chainId + '@' + blockNumber
+
+interface GasPriceCacheDeps {
+  blockNumber: number | null
+  chainId: number
+}
+
+interface GasStationResponse {
+  lastUpdate: string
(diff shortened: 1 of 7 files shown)


src/api/wallet/WalletApi.ts
```diff
@@ -279,7 +293,9 @@ export class WalletApiImpl implements WalletApi {
 
   /* ****************      Private Functions      **************** */
 
-  private async _notifyListeners(): Promise<void> {
+  private async _notifyListeners(blockchainUpdate?: BlockchainUpdatePrompt): Promise<void> {
```
Velenir: same as update in #386

src/api/gasStation.ts
```diff
@@ -0,0 +1,65 @@
+import { walletApi } from './'
```
anxolin: It looks problematic this dependency. If you depend in `WalletApi`, the gas station should be a "factory function" with `WalletApi` as one of it's params.

Velenir:
Also not completely happy with almost-cyclic dependencies here. Injection in constructor does make , but wouldn't it be better as part of cache layer?

Will do injection for now.sense

gnosis-info:
Travis automatic deployment:
 https://pr409--dexreact.review.gnosisdev.com

src/api/gasStation.ts
Velenir: Done

PR #410, opened 2020-01-10
Hotfix/web3 check

### LINKED ISSUE #406: Error web3 not present
The app doesn't load in phone. 
It looks like it's assuming that web3 is injected.

Could we do an unit test that easily checks for errors of this kind? (i.e. mount the App component and render it)

Reproduce: Open Chrome in incognito mode

Velenir:
Closes #406


gnosis-info:
Travis automatic deployment:
 https://pr410--dexreact.review.gnosisdev.com

PR #417, opened 2020-01-14
Fix for TrustWallet in-app browser

Velenir:
Can't count on WalletInfo being available synchronously, so worked around that.

Also it seems to me TrustWallet and probably other mobile wallets don't support `newBlockHeaders` subscriptions.

Need to investigate more.

### DIFF
--- a/src/api/wallet/WalletApi.ts
+++ b/src/api/wallet/WalletApi.ts
@@ -26,6 +26,8 @@ import {
 import { log, toBN } from 'utils'
 import { INFURA_ID } from 'const'
 
+import { subscribeToWeb3Event } from './subscriptionHelpers'
+
 export interface WalletApi {
   isConnected(): boolean | Promise<boolean>
   connect(givenProvider?: Provider): Promise<boolean>
@@ -75,28 +77,43 @@ const subscribeToBlockchainUpdate = ({
   subscriptions?: Subscriptions
   web3: Web3
 }): BlockchainUpdatePromptCallback => {
-  console.log('provider', provider)
   const subs = subscriptions || createSubscriptions(provider)
-  console.log('subscriptions', subscriptions)
-  console.log('subs', subs)
 
   const blockUpdate = (cb: (blockHeader: BlockHeader) => void): Command => {
-    const blockSub = web3.eth.subscribe('newBlockHeaders').on('data', cb)
-    return (): void => {
-      blockSub.unsubscribe()
+    return subscribeToWeb3Event({
+      web3,
+      interval: 8000,
+      callback: cb,
+      getter: web3 => web3.eth.getBlock('latest'),
+      event: 'newBlockHeaders',
+    })
+  }
+
+  let blockchainPrompt: BlockchainUpdatePrompt
+
+  const providerState = getProviderState(provider)
+
+  if (providerState) {
+    const {
+      accounts: [account],
+      chainId,
+    } = providerState
+
+    blockchainPrompt = {
+      account,
+      chainId: +chainId,
+      blockHeader: null,
+    }
+  } else {
+    blockchainPrompt = {
+      account: '',
+      chainId: 0,
+      blockHeader: null,
     }
   }
 
-  if (!subs) {
-    let blockchainPrompt: BlockchainUpdatePrompt = {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      account: (provider as any).address,
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      chainId: +(provider as any).chainId,
-      blockHeader: null,
-    }
+  if (!subs || !providerState) {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    console.log('(provider as any).address', (provider as any).address)
 
     const subscriptionHOC: BlockchainUpdatePromptCallback = callback => {
       const unsubBlock = blockUpdate(blockHeader => {
@@ -105,9 +122,7 @@ const subscribeToBlockchainUpdate = ({
         callback(blockchainPrompt)
       })
 
-      return (): void => {
-        unsubBlock()
-      }
+      return unsubBlock
     }
 
     return subscriptionHOC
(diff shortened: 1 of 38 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr417--dexreact.review.gnosisdev.com

alfetopito:
Is this good for review already?

Velenir:
Good to review.
But there's sometimes a race condition in `PrivateRoute`. I might need to change `walletApi.isConnected` some more.

src/hooks/useTokenBalances.ts
```diff
@@ -90,6 +90,7 @@ export const useTokenBalances = (): UseTokenBalanceResult => {
 
   useEffect(() => {
     // can return NULL (if no address or network)
+    walletInfo.isConnected &&
```
alfetopito: Doesn't networkId and userAdress imply isConnected?

Velenir: yes, but there's no userAddress here
Actually in case with MMask we can have userAddress after we _disconnect_ wallet , because there's no real disconnect for MMask

src/api/wallet/WalletApi.ts
```diff
@@ -26,14 +26,16 @@ import {
 import { log, toBN } from 'utils'
 import { INFURA_ID } from 'const'
 
+import { subscribeToWeb3Event } from './subscriptionHelpers'
+
 export interface WalletApi {
-  isConnected(): boolean
+  isConnected(): boolean | Promise<boolean>
```
anxolin: @Velenir 
Why not ALWAYS `promise<boolean>`?  it's simpler interface

src/api/wallet/WalletApi.ts
```diff
@@ -77,31 +79,62 @@ const subscribeToBlockchainUpdate = ({
 }): BlockchainUpdatePromptCallback => {
   const subs = subscriptions || createSubscriptions(provider)
 
-  let networkUpdate: (callback: (chainId: number) => void) => Command
+  const blockUpdate = (cb: (blockHeader: BlockHeader) => void): Command => {
+    return subscribeToWeb3Event({
+      web3,
+      interval: 8000,
```
anxolin: Magic number? can we make it constant?

src/api/wallet/WalletApi.ts
```diff
@@ -293,28 +337,59 @@ export class WalletApiImpl implements WalletApi {
...
+    }
+  }
+
   private async _notifyListeners(blockchainUpdate?: BlockchainUpdatePrompt): Promise<void> {
     if (blockchainUpdate) this.blockchainState = blockchainUpdate
 
     await Promise.resolve()
-    const walletInfo: WalletInfo = this.getWalletInfo()
+
+    const walletInfo = await (this.getWalletInfo() || this._getAsyncWalletInfo())
```
anxolin: I don't understand why add this kind of complications. I prefer if the Api interface return a single type, in this case, a `Promise<WalletInfo>`

This logic seems very strange

src/api/wallet/subscriptionHelpers.ts
```diff
@@ -0,0 +1,138 @@
...
+}: SubscribeParams<T>): Promise<() => void> => {
+  return new Promise<Subscription<Event2Data[T]>>((resolve, reject) => {
+    const detectValidSubCb = (e: Error): void => {
+      if (e) {
+        reject(e)
+        return
+      }
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const params: [any, any] = (logOptions && event === 'logs'
```
anxolin: The `createWeb3Subscription` implementation is difficult to follow. Comments of intent would help.

Velenir:
> @Velenir This PR includes a `console-feed` integration. I didn't see any reference in the description. I assume it was to debug the mobile browsers, right?
> 
> How do we do use it? Does it add any weight to the overall weight, or is disabled for production?

It's from this sub-PR #418
doesn't add any weight in production

src/api/wallet/WalletApi.ts
Velenir: I didn't want to drop the very useful case of sychronous WalletInfo
Most of the time it's synchronous, only with some providers (like TrustWallet) we can't count on synch  props on provider itself

src/api/wallet/WalletApi.ts
Velenir: I meant to delete it but forgot

src/api/wallet/WalletApi.ts
anxolin: Yep, but as soon as 1 provider does that async, we need to make the interface async to avoid confusion.

All parts using this, just need to do a `await` if they need synchronous code. I don't see the benefit in the "dual signature". Maybe I'm not seeing sth...

Velenir: The benefit comes from using it in component render

Velenir: If you will strongly about it, can of course change to always Promise

anxolin: But the issue I see, is that, you cannot use it in render anymore, since you need to handle both cases. So I don't understand the gain

Velenir:
```ts
export const useWalletConnection = (): (WalletInfo & { pending: false }) | typeof PendingState => {
  const [walletInfo, setWalletInfo] = useSafeState(() => walletApi.getWalletInfo())

  useEffect((): Command => {
    return walletApi.addOnChangeWalletInfo(setWalletInfo)
  }, [setWalletInfo])

  return useMemo(() => {
    return isPromise(walletInfo) ? PendingState : { ...walletInfo, pending: false }
  }, [walletInfo])
}
```

`isPromise()===false` we return walletInfo, true -- pending

Velenir: You do feel strongly about it, don't you
I'm happy to change

anxolin: Heheh, yes. But may only be cause I don't fully understand the benefits and I'm seeing the problems on an Api that don't return a pure type.

Happy to discuss too, I might be terribly wrong u might convince me

Issue #402, opened 2020-01-08
Pooling interface: Step 3 - Create strategy


anxolin:
Creates strategy.

It makes the actual call to BatchExchange contract with all the orders.

Details: https://docs.google.com/document/d/1yNddJk6eW1NluEh5B8IL-UghDv9tWnl0ve17SgZBf1k/edit

![image](https://user-images.githubusercontent.com/2352112/72000334-5c3e3600-3243-11ea-8577-47948ceb050f.png)

Related to: https://github.com/gnosis/dex-react/issues/400

PR #420, opened 2020-01-15
Adding implementation for placeValidFromOrders to ExchangeApi

alfetopito:
Adding implementation for `placeValidFromOrders` to ExchangeApi

Requirement for #402

### DIFF
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -201,8 +201,8 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
 
     const contract = await this._getContract(networkId)
 
-    const buyAmountsStr = buyAmounts.map(amountBn => amountBn.toString())
-    const sellAmountsStr = sellAmounts.map(amountBn => amountBn.toString())
+    const buyAmountsStr = buyAmounts.map(String)
+    const sellAmountsStr = sellAmounts.map(String)
 
     const tx = contract.methods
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr420--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApi.ts
```diff
@@ -169,6 +181,50 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
+    assert(
+      [sellTokens, validFroms, validUntils, buyAmounts, sellAmounts].every(el => el.length === length),
+      'Parameters length do not match',
+    )
+    assert(length > 0, 'At least one order required')
+
+    const contract = await this._getContract(networkId)
+
+    const buyAmountsStr = buyAmounts.map(amountBn => amountBn.toString())
+    const sellAmountsStr = sellAmounts.map(amountBn => amountBn.toString())
```
Velenir: Can do `sellAmounts.map(String)`, not that it matters

alfetopito:
> I like it!
> 
> One question. Although the interface in solidity has many arrays, don't you think that if we create our API for it it makes more sense to define it as an array of orders with buyToken, sellToken, etc.
> It feels more natural imo.

We definitely can do that and it's somehow what has been done on #425 but in a hook.

The reason I didn't do that in the API, is mainly to keep the same standard, as most (all?) other API methods follow very closely the contract interface.

anxolin:
Okay, fair enough. It's true that if we get too far from the implementation signature, we end up abstracting too much.
If we just work with the hook in the UI and the hook is simpler to use, I think it's good enough. Thx!

Issue #403, opened 2020-01-08
Pooling interface: Step 4 - Add funding


anxolin:
Final step.

Details: https://docs.google.com/document/d/1yNddJk6eW1NluEh5B8IL-UghDv9tWnl0ve17SgZBf1k/edit

![image](https://user-images.githubusercontent.com/2352112/72000415-914a8880-3243-11ea-8a23-f2408ea44215.png)

Related to: https://github.com/gnosis/dex-react/issues/400

Issue #401, opened 2020-01-08
Pooling interface: Step 2 - Select spread


anxolin:
Select the spread

Details: https://docs.google.com/document/d/1yNddJk6eW1NluEh5B8IL-UghDv9tWnl0ve17SgZBf1k/edit

![image](https://user-images.githubusercontent.com/2352112/72000299-44ff4880-3243-11ea-968e-bc16b75cc7ae.png)

Related to: https://github.com/gnosis/dex-react/issues/400

PR #421, opened 2020-01-16
400/pooling interface step 1 + ProgressBar

### LINKED ISSUE #400: Pooling interface: Step 1 - Select tokens
Main component that handle the state: `spread`, `tokens`, `newStrategyTx`

Depending on the step, it would show one component or another: 4 steps

The step 1 is the selection of tokens

Details: https://docs.google.com/document/d/1yNddJk6eW1NluEh5B8IL-UghDv9tWnl0ve17SgZBf1k/edit

image

### LINKED ISSUE #404: Pooling interface: Stepper component
(lesser prio than other task in the sprint)

Details: https://docs.google.com/document/d/1yNddJk6eW1NluEh5B8IL-UghDv9tWnl0ve17SgZBf1k/edit

image

Related to: https://github.com/gnosis/dex-react/issues/400

W3stside:
Closes #400 and closes #404 

Creates: 

1. ProgressBar (tied to step state in `PoolingWidget/index.tsx`)
2. SubComponents which renders issues #400, #401, #402, #403 based on steps (<-- comps still need to be implemented just render them in there though when you finish building them)
3. TokenSelector
4. Various styled components that are hopefully resusable


gnosis-info:
Travis automatic deployment:
 https://pr421--dexreact.review.gnosisdev.com

src/components/PoolingWidget/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+    if (item === newItem) return acc
+
+    const newAcc = acc.concat(item)
+    return newAcc
+  }, [])
+}
+
+const PoolingInterface: React.FC = () => {
+  const [step, setStep] = useSafeState(1)
+  const [selectedTokens, setSelectedTokens] = useSafeState<number[]>([])
```
alfetopito: What about using a `Set`? Don't know if you noticed, but I do like Sets :D

alfetopito: Actually, a `Map<number, TokenDetails>` might be better, because on the next step (https://github.com/gnosis/dex-react/issues/402) we'll need to show token symbols, and I would prefer to keep the objects around instead of having to find them again by id.

Anyway, doesn't have to be a Map, just thinking of an easier interface to check what's selected for Step 1, while still keeping TokenDetails objects for Step 3.

What do you think?

src/components/PoolingWidget/SubComponents.tsx
```diff
@@ -0,0 +1,27 @@
...
+  ) : step === 2 ? (
+    // STEP 2 - SELECT SPREAD
+    <div>Step 2</div>
+  ) : step === 3 ? (
+    // STEP 3 - CREATE STRATEGY
+    <div>Step 3</div>
+  ) : (
+    // STEP 4 - ADD FUNDS
+    <div>Step 4</div>
+  )
```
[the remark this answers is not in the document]
W3stside: haha, yeah ;)

src/components/PoolingWidget/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const { networkId } = useWalletConnection()
+  // Avoid displaying an empty list of tokens when the wallet is not connected
+  const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
+  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId).filter(({ symbol }) => symbol !== 'WETH'), [
+    fallBackNetworkId,
+  ])
```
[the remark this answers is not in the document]
W3stside: good idea

src/components/PoolingWidget/index.tsx
W3stside: @alfetopito I had it setup this way before using a `Map<tokenID, tokenDetails>` but i had it coded in a way i guess that was super inefficient so the state updates were horrendously slow. like we're talking a few seconds to see the selected tokens become highlighted etc

I'll play around with it some more

W3stside: Dmitry helped me debug my "slow" update problem, he da best

src/components/PoolingWidget/SubComponents.tsx
```diff
@@ -0,0 +1,31 @@
...
+    case 1:
+      return (
+        <TokenSelector handleTokenSelect={handleTokenSelect} tokens={tokens} selectedTokensMap={selectedTokensMap} />
+      )
+    case 2:
+      return <div>Step 2 </div>
+    case 3:
+      return <div>Step 3</div>
+    case 4:
+      return <div>Step 4</div>
```
Velenir: Do a _cool fall-through case_

```ts
case 2:
case 3:
case 4:
  return <div>Step {step}</div>
```

PR #425, opened 2020-01-16
403/use place order

alfetopito:
Part of #402 (shit, wrong branch name...)

Adding to `usePlaceOrder` hook the function `placeMultipleOrders`.

Also, adding `validFrom` and `validUntil` as parameters, where the former is only available for the new function.

Also, re-enabling success notification once tx is mined on `placeOrder` function.


gnosis-info:
Travis automatic deployment:
 https://pr425--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr425--dexreact.review.gnosisdev.com

anxolin:
Same as the other PR. In JS we don't have the limitations of solidity. We don't need to expose the signature with all these arrays of the same size. We can have an array of orders.

PR #427, opened 2020-01-17
402/price calculation

alfetopito:
Part of #402 
Depends on #420 and #425 

# Adding amount calculation for unlimited orders

For exchange contract to consider the order as `unlimited`, one of the amounts must be equal to max uint128 size: https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L35

To have the price as user selected spread, we derive the opposite amount from MAX.

Given that tokens have varied decimal precision, we need to find a way to pick which amount (buy/sell) should be set to MAX, and derive the opposite from there.


gnosis-info:
Travis automatic deployment:
 https://pr427--dexreact.review.gnosisdev.com

src/const.ts
```diff
@@ -17,6 +17,8 @@ export const ALLOWANCE_MAX_VALUE = TWO.pow(new BN(256)).sub(ONE) // 115792089237
 export const ALLOWANCE_FOR_ENABLED_TOKEN = TWO.pow(new BN(128)) // 340282366920938463463374607431768211456
 // How much a sell order must be selling to be considered `unlimited`
 export const MIN_UNLIMITED_SELL_ORDER = ALLOWANCE_MAX_VALUE.div(TEN).mul(new BN(7)) // Currently set to 70% of max allowance
+// Amount for an order to be considered unlimited, from contract's point of view: https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L35
+export const UNLIMITED_ORDER_AMOUNT = TWO.pow(new BN(128))
```
W3stside: Out of curiousity, does this not need ot be `TWO.pow(new BN(128)).sub(ONE)`? or overflow errors happen?

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+  // 1.2345, precision 2, amount == 12345 / 10 ^ (4 - 2) => 12345 / 100 => 123
+  // 1.5550, precision 1, amount == 15550 / 10 ^ (4 - 1) => 15550 / 1000 => 16
+  return amount.divRound(TEN.pow(difference))
+}
+
+function bigNumberToBN(n: BigNumber | BN): BN {
+  if (n instanceof BN) {
+    return n
+  }
+  return new BN(n.integerValue().toString(10))
```
W3stside: `.toString(10)` is default i believe, e.g `toString() === toString(10)`

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+interface Amounts {
+  buyAmount: BN
+  sellAmount: BN
+}
+
+const MAX = new BigNumber(UNLIMITED_ORDER_AMOUNT.toString())
```
W3stside: inside `/const` :smile:

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Calculates the max amounts within given `spread` for unlimited orders
+ *
+ * Uses BigNumber internally to keep track of decimals
+ * Returns BN for compatibility with ExchangeApi
+ *
+ * @param spread Value between 0 and 100, not inclusive
+ * @param buyTokenPrecision Decimals of buy token
+ * @param sellTokenPrecision Decimals of sell token
+ */
+export function maxAmountsForSpread(spread: number, buyTokenPrecision: number, sellTokenPrecision: number): Amounts {
```
W3stside: nice work with this fn, easy to get tripped up here with the switch between `buyAmount` and `sellAmount` and the tricky BN syntax

src/const.ts
Velenir: Good catch, even excellent catch :+1: 
`uint128(-1)` in the contract is converted into max value for uint128 (or at least I expect it to work like that in solidity ), which is `340282366920938463463374607431768211455`, or in JS `2n ** 128n  - 1`

alfetopito: So I need to subtract 1 from this amount?

src/utils/price.ts
alfetopito: Yes, they are equivalent, BUT, if base is not set, it defaults to the exponent notation representation: [Bignumber.js docs](https://mikemcl.github.io/bignumber.js/#toS)

> If a base is not specified, and this BigNumber has a positive exponent that is equal to or greater than the positive component of the current EXPONENTIAL_AT setting, or a negative exponent equal to or less than the negative component of the setting, then exponential notation is returned. 

Thus, I set base to 10 to force NOT returning a string with exponent.

src/utils/price.ts
alfetopito: Well, that amount is already there as BN. I 'duplicated' it here for local usage only.
But you are right, will move it there.

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (higherPrecision === lowerPrecision) {
+    // no adjustment required
+    return amount
+  }
+
+  assert(higherPrecision > lowerPrecision, 'higherPrecision must be > lowerPrecision')
+
+  const difference = new BN(higherPrecision - lowerPrecision)
+
+  // divides amount by difference in precision, rounding. E.g:
```
anxolin: I love that you add these examples. Keep doing that. Code is easier to follow

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
+import BN from 'bn.js'
+import { assert } from './miscellaneous'
+import { TEN, UNLIMITED_ORDER_AMOUNT_BIGNUMBER } from 'const'
+import BigNumber from 'bignumber.js'
+
+export function adjustAmountToLowerPrecision(amount: BN, higherPrecision: number, lowerPrecision: number): BN {
```
anxolin: Should the params be an object instead of simple params? We had issues in the past with the params not being clear and having not obvious mistakes by entering the params in the wrong order

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (higherPrecision === lowerPrecision) {
+    // no adjustment required
+    return amount
+  }
+
+  assert(higherPrecision > lowerPrecision, 'higherPrecision must be > lowerPrecision')
```
anxolin: If we have this assert, means this util need to be called with the precisions in the right order. Wouldn't it be better to allow this util to provide the two precisions without any assumptions?

We would check which one is bigger inside the util (once) instead of doing on every call to the util.

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Uses BigNumber internally to keep track of decimals
+ * Returns BN for compatibility with ExchangeApi
+ *
+ * @param spread Value between 0 and 100, not inclusive
+ * @param buyTokenPrecision Decimals of buy token
+ * @param sellTokenPrecision Decimals of sell token
+ */
+export function maxAmountsForSpread(spread: number, buyTokenPrecision: number, sellTokenPrecision: number): Amounts {
+  // Enforcing positive spreads: 0 < spread < 100
+  assert(spread > 0 && spread < 100, 'Invalid spread amount')
```
anxolin: I like that you add explicit asserts, also the JSdoc :)

src/utils/price.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const MAX = UNLIMITED_ORDER_AMOUNT_BIGNUMBER
+  const ONE = new BigNumber(1)
+
+  const spreadPercentage = new BigNumber(spread).dividedBy(new BigNumber(100))
+
+  let buyAmount
+  let sellAmount
+
+  if (buyTokenPrecision === sellTokenPrecision) {
```
anxolin: Very neat way to handle the complexity of this, and hide it inside this function

src/utils/price.ts
alfetopito: > If we have this assert, means this util need to be called with the precisions in the right order. Wouldn't it be better to allow this util to provide the two precisions without any assumptions?

That is something I considered afterwards. Instead of the assert, the function would select the higher precision and adjust from there.
My concern with this approach is that this can lead to a silent failure.

Let's say we have token `A` precision 12, token `B` precision 2.
If due to some bug we selected `B` with the higher precision and called the function with (amount, 2, 12).
In the current implementation, it throws and error.
In the new approach, it simply adjusts the precision assuming 12 is the higher amount.

Another way of looking at this is that it isn't this function responsibility to deal with bug somewhere else, and the function should do what it's supposed to do: adjust to lower precision.

/Conflicted.

> We would check which one is bigger inside the util (once) instead of doing on every call to the util.

I don't get that. For every precision we need to either do the assert or check which is bigger, so every call still needs to do at least one check. No?

src/utils/price.ts
alfetopito: I'm selfish and do them for my self to understand what's going on :D

src/utils/price.ts
alfetopito: Sure, I'll do that. Hang on for the PR...

src/utils/price.ts
anxolin: Understood, it's fine then

src/utils/price.ts
alfetopito: https://github.com/gnosis/dex-react/pull/438

PR #433, opened 2020-01-20
Detect <doman>.eth.link  as IPFS

Velenir:
Currently when `window.location` is detected as belonging to IPFS we use HashRouter instead of BrowserRouter

Detection relies on matching `Qm<hash>` with `location.pathname`

If the app is hosted on EthDNS at `<domain>.eth.link` that detection fails and we can't start browsing at a non-index route, also can't refresh the page at a subroute

This PR adds detection for `location.hostname`

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -5,20 +5,20 @@
   <%if (htmlWebpackPlugin.options.ipfsHack) { %>
   <script>
     (function () {
-      const { pathname } = window.location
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr433--dexreact.review.gnosisdev.com

anxolin:
Cool, thanks for this

PR #435, opened 2020-01-21
402/step 3

### LINKED ISSUE #402: Pooling interface: Step 3 - Create strategy

alfetopito:
Closes #402 

Implemented the interface defined on #402 and also placing the orders.

### DIFF
--- a/src/components/PoolingWidget/index.tsx
+++ b/src/components/PoolingWidget/index.tsx
@@ -142,16 +142,16 @@ const PoolingInterface: React.FC = () => {
     fallBackNetworkId,
   ])
 
-  const prevStep = (): void => {
+  const prevStep = useCallback((): void => {
     if (step == 1) return
 
     return setStep(step - 1)
-  }
-  const nextStep = (): void => {
+  }, [setStep, step])
+  const nextStep = useCallback((): void => {
     if (step == 4) return
 
     return setStep(step + 1)
-  }
+  }, [setStep, step])
 
   const handleTokenSelect = useCallback(
     (token: TokenDetails): void => {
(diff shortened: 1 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr435--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr435--dexreact.review.gnosisdev.com

src/components/PoolingWidget/index.tsx
```diff
@@ -128,9 +132,46 @@ function addRemoveMapItem(map: Map<number, TokenDetails>, newToken: TokenDetails
   return copyMap
 }
 
+// TODO: Decide the best place to put this. This file is too long already, but feels to specific for utils
```
anxolin: I like the comments below, also agree that the "combinations" should be done in a util

alfetopito:
> Amazing!
> 
> Can we delete the link? Underlined is enough.

Right! I thought it was link to somewhere but didn't know where to point at yet. Will fix.

> image
> 
> Also, I saw that we wait for the TX to be mined until we go to next step.
> In the proposal, I was suggesting to go to next step as soon as the tx is sent, not mined. However, maybe you are right and it's better to wait for it to be mined, just in case we need a re-attempt. In that case, maybe we need to show something like the message of step 4 in step 3:
> image
> 
> What do you think is best?

Hmmm, yeah. I kind of remembered about this step but did not implement it properly.

If we are to show the message of step 4, does it make sense to still have the step 4?

Otherwise, I can fix and move straight to step for before tx is mined.

One issue that I have with this though is, once tx is sent, there's no going back. How is best to translate this into the interface?

> 
> One question. How much gas are you using? Are you estimating somehow?

I'm using Dima's gas estimation

PR #441, opened 2020-01-22
Comments for subscriptions helper

Velenir:
More comments for subscriptions helpers, as they are reportedly hard to read

In a sentence what happens is:

1. We want to sub to `newBlockHeaders`
2. We try to do it through `web3.eth.subscribe('newBlockHeaders')`
3. Works -- great
4. Doesn't -- start checking if block number changed on interval (supposed to be efficient, MMask does exactly that)
5. When changed, fetch new block

Also removes `interval`, which isn't used anyway, forgot to delete


gnosis-info:
Travis automatic deployment:
 https://pr441--dexreact.review.gnosisdev.com

PR #444, opened 2020-01-23
Add balance for an unapproved token

### LINKED ISSUE #407: Allow to withdraw, even if the token is not enabled
Estimation: 2

If we see a balance in the token, we should allow to withdraw it, even in the case the token is not enabled.

Enable token in reality only affects the deposit. So I would enable the deposit button if the there's a balacen, regardless of the state of the "abilitation" of the token.

anxolin:
Closes #407 

* Allows to withdraw even if the token was not enabled.
* The "Withdraw" button is only shown if the user has a balance (no other condition is required)


gnosis-info:
Travis automatic deployment:
 https://pr444--dexreact.review.gnosisdev.com

PR #447, opened 2020-01-23
Fix linking dex js

anxolin:
Fix dex-js dependency usage

This is from the session with @alfetopito where we linked dex-react and dex-js and fixed the issues


gnosis-info:
Travis automatic deployment:
 https://pr447--dexreact.review.gnosisdev.com

Issue #436, opened 2020-01-21
Handle unlimited orders on orders page


alfetopito:
With the new way unlimited orders are created (see https://github.com/gnosis/dex-react/pull/427), we need to adapt the orders page to show `Unlimited` amount

![screenshot_2020-01-21_07-37-53](https://user-images.githubusercontent.com/43217/72856190-acb78980-3c6e-11ea-85de-065d49a68067.png)

Also, apply the same to the telegram-bot.

Thus, this will have to be done on `dex-js` package and exported as a utils function.

PR #448, opened 2020-01-23
Using isOrderUnlimited from dex-js

alfetopito:
Part of #436 
Depends on #447 
Requires https://github.com/gnosis/dex-js/pull/67 to be published


gnosis-info:
Travis automatic deployment:
 https://pr448--dexreact.review.gnosisdev.com

src/const.ts
```diff
@@ -16,10 +17,7 @@ export const TEN = new BN(10)
 export const ALLOWANCE_MAX_VALUE = TWO.pow(new BN(256)).sub(ONE) // 115792089237316195423570985008687907853269984665640564039457584007913129639935
 // Arbitrarily big number for checking if the token is enabled
 export const ALLOWANCE_FOR_ENABLED_TOKEN = TWO.pow(new BN(128)) // 340282366920938463463374607431768211456
```
W3stside: should this also be subtracted by one? I don't think you can approve that high a value

PR #457, opened 2020-01-27
Fornawesome bundle size workaround

Velenir:
Workaround from an issue in [react-forntawesome repo](https://github.com/FortAwesome/react-fontawesome/issues/70#issuecomment-486626761)

This effectively changes 
```ts
import { iconName } from '@fortawesome/free-solid-svg-icons';
// to
import { iconName } from '@fortawesome/free-solid-svg-icons/iconName';
```

BEFORE:
Screenshot from 2020-01-27 18-40-19

AFTER:
Screenshot from 2020-01-27 18-40-11

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,6 +11,23 @@
         "pure": true
       }
     ],
+    [
+      "transform-imports",
+      {
+        "@fortawesome/free-regular-svg-icons": {
+          "transform": "@fortawesome/free-regular-svg-icons/${member}",
+          "skipDefaultConversion": true
+        },
+        "@fortawesome/free-solid-svg-icons": {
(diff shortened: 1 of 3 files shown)


Velenir:
> One question though: don't we need to also update the fontawesome imports to make this work?

NO, we import a component from `react-fontawesome`, which in turn imports a utility function from some engine core behind the scenes, which finally imports the icon from `free-solid-svg-icons`.
That last one import isn't very, or at all, treeshakable. Babel rewrites it to point to exact icon file, or whatever, instead of importing that icon as a variable (`import {iconName}`) from an index file.

W3stside:
niiiiiiice

anxolin:
Great findings on this!!!
nice job

PR #462, opened 2020-01-28
Fix auto dark mode

Velenir:
localStorage stick around across page refresh, as a result dark mode stays even when auto is chosen

To reproduce:

1. Select dark mode
2. Select auto
3. Refresh page


gnosis-info:
Travis automatic deployment:
 https://pr462--dexreact.review.gnosisdev.com

PR #463, opened 2020-01-28
Misc design fixes

W3stside:
Fixes some padding/margin on smaller screens, fonts, hover issues as discovered while testing

### DIFF
--- a/src/components/PoolingWidget/DefineSpread.tsx
+++ b/src/components/PoolingWidget/DefineSpread.tsx
@@ -43,6 +43,10 @@ export const SpreadInformation: React.FC<SpreadInformationProps> = ({ selectedTo
             &::before {
               content: none;
             }
+          }
+          > td {
+            padding: 0.7rem 0 !important;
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr463--dexreact.review.gnosisdev.com

src/components/PoolingWidget/DefineSpread.tsx
```diff
@@ -43,6 +43,10 @@ export const SpreadInformation: React.FC<SpreadInformationProps> = ({ selectedTo
             &::before {
               content: none;
             }
+          }
+          > td {
+            padding: 0.7rem 0 !important;
+            margin: 0 0.6rem !important;
```
Velenir: I generally don't like `!important`, but I trust you.
Maybe raise specificity with something like a class

W3stside: i agree here

PR #466, opened 2020-01-28
ValidFrom orders have now a wait period when no value is set

### LINKED ISSUE #459: The liquidity orders are failing if there's no enough time to be mined
Mining can take some time. The contracts ensures you don't provide a past batch number as the start. In order to leave some time for the tx to mine, we can use `batchNumber + 3` when submitting the orders.

Review also, how we do with normal orders, and if there's sth that it would be good to do in the smart contract in adition.

alfetopito:
Closes #459 

## Steps to reproduce:
1. Create a new liquidity
2. Send the transaction
3. Using metamask, keep the tx confirmation window open
4. After 5 min, confirm the tx

- On a version without the fix, tx failed https://rinkeby.etherscan.io/tx/0x4df9101bf5dc15acfaaa76c8fc831820b76564c0d677dfe09aa65837b2cb1c13
- On version with the fix, tx succeeded https://rinkeby.etherscan.io/tx/0xa8258c29e1e7371f2fc8b4c1501c563bc8ac7e46beea4e8593d209eb88785e54

### DIFF
--- a/src/hooks/usePlaceOrder.ts
+++ b/src/hooks/usePlaceOrder.ts
@@ -10,7 +10,7 @@ import { PlaceOrderParams as ExchangeApiPlaceOrderParams } from 'api/exchange/Ex
 import { log } from 'utils'
 import { txOptionalParams as defaultTxOptionalParams } from 'utils/transaction'
 import { useWalletConnection } from './useWalletConnection'
-import { DEFAULT_ORDER_DURATION } from 'const'
+import { DEFAULT_ORDER_DURATION, BATCHES_TO_WAIT } from 'const'
 import useSafeState from './useSafeState'
 
 interface PlaceOrderParams<T> {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr466--dexreact.review.gnosisdev.com

src/hooks/usePlaceOrder.ts
```diff
@@ -147,8 +147,8 @@ export const usePlaceOrder = (): Result => {
           buyTokens.push(order.buyToken)
           sellTokens.push(order.sellToken)
 
-          // if not set, order is valid from placement
-          validFroms.push(order.validFrom || currentBatchId)
+          // if not set, order is valid from placement + wait period
+          validFroms.push(order.validFrom || currentBatchId + BATCHES_TO_WAIT)
```
Velenir: Wanna use the new fancy null coalescence here?
```suggestion
          validFroms.push(order.validFrom ?? currentBatchId + BATCHES_TO_WAIT)
```

It would allow for `validFrom` to be 0, not that it will ever be 0

W3stside: :+1:

alfetopito: Well, why not?

Issue #455, opened 2020-01-27
Orders are not show


anxolin:
Reported by @koeppelmann:

I created a bunch of orders directly via the “placeValidFrom” command. Potentially used a token ID that did not exist on rinkeby

![image](https://user-images.githubusercontent.com/2352112/73164660-a43dd380-40f2-11ea-9bae-e56a3f181aef.png)

--------
anxo: Probably is one of these 2:
- Use pagination for loading the orders of the user: The new smart contract version allow to paginate the orders, so you can overcome the limits of gaslimit in a call
- Related to support any token in the interface: #376, #378, #379

W3stside:
@anxolin I don't get this one, can you elaborate? Is this Martin messing around with the contracts directly and then checking if the FE supported this in the orders page?

anxolin:
>  Is this Martin messing around with the contracts directly and then checking if the FE supported this in the orders page?

Probably :) 

I've got this directly from his feedback in https://docs.google.com/document/d/1WCP3YNnJ6wagRL5ZhdUJ9wcmxoNXCp55jsQCeuiVDqI/edit# (I just made the issue, and added what 2 things I thought could be the cause)

@koeppelmann , is this still happening to you? Can you help us reproduce it?

W3stside:
@anxolin i don't see Martin's report in the bug doc

alfetopito:
Not sure it's related to the issue or not, but I can already reproduce a use case where pagination is handy.
~I have a couple ~ Actually I deleted my orders because the page was too slow.

Anyway, that's my point. With over 50 expired orders, switching between active/expired takes quite awhile to load.

anxolin:
> Not sure it's related to the issue or not

Yep. I also think it can be that @alfetopito!

> @anxolin i don't see Martin's report in the bug doc

David, it was, but I moved here. With all the info that he wrote. Ping him if you need help reproducing, although I believe it's what Leandro suggests. This method was not paginated in the past, now we can paginate to avoid this issues
![image](https://user-images.githubusercontent.com/2352112/73380112-bfac0880-42c3-11ea-9929-ecf149254195.png)

alfetopito:
I'll start working on the pagination. Can re-prioritize during next sprint planning but besides the UI (how the pages should look like), the `backend` (interacting with the contract, hooks and so on) logic won't change.

anxolin:
Cool! thanks leandro

alfetopito:
[...]

Issue #159, opened 2019-10-31
Add Curly rule for linter


anxolin:
https://eslint.org/docs/rules/curly

discussed in this thread: https://github.com/gnosis/dex-react/pull/136#discussion_r340536562

alfetopito:
Don't close it

alfetopito:
Keep it

PR #473, opened 2020-01-29
Adding getOrdersPaginated method to ExchangeApi

alfetopito:
Part of #455 

Implemented `getOrdersPaginated` corresponding to the contract's https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L479

- `pageSize` is optional defaulting to `10`
- returns an object with `orders` and -- when there might be a next page -- `nextIndex`

### DIFF
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -145,9 +145,11 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
       // no more pages left, indicate by not returning `nextIndex`
       return { orders }
     } else {
-      // there is at least 1 item in the next page,
-      // pop the extra element, get its id as nextIndex
-      const nextIndex = Number((orders.pop() as AuctionElement).id)
+      // there is at least 1 item in the next page
+      // pop the extra element
+      const nextPageOrder = orders.pop() as AuctionElement
+      // get its id as nextIndex
+      const nextIndex = Number(nextPageOrder.id)
       return { orders, nextIndex }
     }
   }
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -133,23 +133,23 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
       `[ExchangeApiImpl] Getting Orders Paginated for account ${userAddress} with offset ${offset} and pageSize ${pageSize}`,
     )
 
-    const encodedOrders = await contract.methods.getEncodedUserOrdersPaginated(userAddress, offset, pageSize).call()
+    // query 1 more than required to check whether there's a next page
+    const encodedOrders = await contract.methods.getEncodedUserOrdersPaginated(userAddress, offset, pageSize + 1).call()
 
     // is null if Contract returns empty bytes
     if (!encodedOrders) return { orders: [] }
 
     const orders = decodeAuctionElements(encodedOrders, offset)
 
-    let nextIndex: number | undefined
-    if (orders.length < pageSize) {
+    if (orders.length <= pageSize) {
       // no more pages left, indicate by not returning `nextIndex`
-      nextIndex = undefined
+      return { orders }
     } else {
-      // this page was full, keep going
-      nextIndex = Number(orders[orders.length - 1].id) + 1
+      // there is at least 1 item in the next page,
+      // pop the extra element, get its id as nextIndex
+      const nextIndex = Number((orders.pop() as AuctionElement).id)
+      return { orders, nextIndex }
     }
-
-    return { orders, nextIndex }
   }
 
   public async getNumTokens(networkId: number): Promise<number> {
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr473--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApi.ts
```diff
@@ -109,6 +121,38 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
   }
 
+  public async getOrdersPaginated({
+    userAddress,
+    networkId,
+    offset,
+    pageSize,
+  }: GetOrdersPaginatedParams): Promise<GetOrdersPaginatedResult> {
+    const contract = await this._getContract(networkId)
+    const _pageSize = pageSize || DEFAULT_ORDERS_PAGE_SIZE
```
W3stside: could also set the default in the params above as `pageSize = DEFAULT_ORDERS_PAGE_SIZE`

src/api/exchange/ExchangeApi.ts
```diff
@@ -109,6 +121,38 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
+
+    log(
+      `[ExchangeApiImpl] Getting Orders Paginated for account ${userAddress} with offset ${offset} and pageSize ${_pageSize}`,
+    )
+
+    const encodedOrders = await contract.methods.getEncodedUserOrdersPaginated(userAddress, offset, _pageSize).call()
+
+    // is null if Contract returns empty bytes
+    if (!encodedOrders) return { orders: [] }
```
anxolin: we never ended up adding the curly bracket rule?

src/api/exchange/ExchangeApi.ts
```diff
@@ -109,6 +121,38 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
+
+    const orders = decodeAuctionElements(encodedOrders, offset)
+
+    let nextIndex: number | undefined
+    if (orders.length < _pageSize) {
+      // no more pages left, indicate by not returning `nextIndex`
+      nextIndex = undefined
+    } else {
+      // this page was full, keep going
+      nextIndex = Number(orders[orders.length - 1].id) + 1
```
anxolin: I like this part where we return the nextIndex, although I would follow a slightly different pattern.

Instead of querying `pageSize results` (the ones you need),  query `pageSize + 1` then. 
For the result, you'll discard the element `pageSize + 1` and return it's id as `nextIndex`. 

This way, you know for sure if you have more pages or not. With the current approach, if for example you query 10 results and there's 10 items, you'll return `nextIndex=11` instead of `nextIndex =undefined`

src/const.ts
```diff
@@ -32,6 +32,9 @@ export const BATCHES_TO_WAIT = 3
 // Furtherst batch id possible (uint32), must be a js Number
 export const MAX_BATCH_ID = 2 ** 32 - 1
 
+// How many orders should we query per call, when invoking https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L479
+export const DEFAULT_ORDERS_PAGE_SIZE = 10
```
anxolin: Why 10. I find it a bit small. We should maybe see a good number that return the maximum results but is still performant.

alfetopito: Arbitrary number, no particular reason. What do you suggest? 20? 50? 100?

src/api/exchange/ExchangeApi.ts
alfetopito: I like that, will do it.

src/api/exchange/ExchangeApi.ts
alfetopito: Nope. Still in the backlog https://github.com/gnosis/dex-react/issues/159

alfetopito:
> What do you think we'll be the strategy for getting the orders? Are you thinking about paginating also in the UI, or maintain the current functionality (load all orders, in bathes of N-elements).

As we discussed in the call, I would lazy load the orders in multiple calls until all of them are fetched on first load and display them without changing the interface at first.

src/const.ts
anxolin: Yes, I would try to see how it behaves with a bigger number
Maybe we can do this when we do the part of fetching all the pages (finding the right number that is a good balance for batch fetching)

alfetopito: This is what I'm considering. I'll adjust the number based on the next PRs where I actually use this method.
Will merge like it is for now.

src/api/exchange/ExchangeApi.ts
```diff
@@ -109,6 +121,37 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
+
+    const orders = decodeAuctionElements(encodedOrders, offset)
+
+    if (orders.length <= pageSize) {
+      // no more pages left, indicate by not returning `nextIndex`
+      return { orders }
+    } else {
+      // there is at least 1 item in the next page,
+      // pop the extra element, get its id as nextIndex
+      const nextIndex = Number((orders.pop() as AuctionElement).id)
```
anxolin: nit: pls consider do just one thing per line for clarity. In that line, you are doing a pop (mutate the array to discard the last element) and then getting it's id.

alfetopito: Sure

PR #460, opened 2020-01-28
Consolidate bn.js dependency

Velenir:
Use one version of `bn.js` everywhere. Shaves off 3.5 Mb. Doesn't seem to break anything, though potentially could.

BEFORE:
Screen Shot 2020-01-28 at 13 46 19

AFTER:
Screen Shot 2020-01-28 at 13 45 50

PR #474, opened 2020-01-30
Update dependencies

### LINKED ISSUE #157: Update dependencies
Update dependencies in preparation for #97
Since many dependencies will have improved integration for techniques for reducing bundle size (react-router and lazy loading for example), better update them first

Velenir:
Updating dependencies, so it's easier to implement lazy loading afterwards
Part of the big picture of Bundle size optimizations.

Most updates were non-breaking
`react-hook-form` broke types. `web3connect` no longer exposes individual connectors, and broke types

We will be fully up-to-date with dependencies, except for `bn.js`, because of #460 

Closes #157


gnosis-info:
Travis automatic deployment:
 https://pr474--dexreact.review.gnosisdev.com

src/utils/format.ts
```diff
@@ -38,6 +38,18 @@ function _decomposeBn(amount: BN, amountPrecision: number, decimals: number): {
   return { integerPart, decimalPart }
 }
 
+export function formatAmount(
```
anxolin: why so many `formatAmount` declarations?

Velenir: That comes from `TokenRow.tsx::L152`:
```ts
setValue(inputId, formatAmountFull(balance.totalExchangeBalance, balance.decimals, false), true)
```
Now that `react-form-hooks` is more strict with its types it only accepts `string` in `setValue(inputId, string here)`
But `formatAmountFull` returns `string | null`. We know from how that function is written that it would return `string` for input of `amount: BN` and `null` for `amount: null | undefined`, we just need to write types of shape `BN => string; null | undefined => null`

That's where function overloading comes in. We create two distinct declarations:
```ts
function formatAmountFull(amount: BN, amountPrecision?: number, thousandSeparator?: boolean): string
// which allows for
formatFullAmount(amountBN): string

//and
function formatAmountFull(amount?: undefined): null
/// which allows for
fotmatFullAmount(): null
formatFullAmount(undefined): null
```
Same for `formatAmount`

And now in case of
```ts
setValue(inputId, formatAmountFull(balance.totalExchangeBalance, balance.decimals, false), true)
```
as long as (and there it always is) `balance.totalExchangeBalance: BN (not undefined)` we always get a `string` in return (previously was `string | null`)

https://www.typescriptlang.org/docs/handbook/functions.html#overloads

Don't you love Typescript? :smile:

anxolin: eheh, yes. It looks like this is "whitebox type definition", where you add the semantics because you know the actual implementation.

Thanks a lot for the detailed explanation!

PR #489, opened 2020-01-31
455/paginating orders page

alfetopito:
Part of #455 

Querying orders in pages.

- Increased page size to `50`
- Orders are queried on load until done.
- On new block, address/network change, query only from last order
- After confirming the cancellation/deletion of an order, refresh all orders from 0

## Warning

This is still a bit dumb. When switching pages, the orders are fetched from 0 again.
I'll address that in the next PR.

## Fetching from start, then updating only the top on new block

screenshot_2020-01-31_12-47-56

## Deleting order and updating everything from start

screenshot_2020-01-31_12-49-03

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -37,7 +37,7 @@ const ShowOrdersButton: React.FC<ShowOrdersButtonProps> = ({ type, isActive, sho
 }
 
 const OrdersWidget: React.FC = () => {
-  const allOrders = useOrders()
+  const { orders: allOrders, forceOrdersRefresh } = useOrders()
 
   const [orders, setOrders] = useSafeState<AuctionElement[]>(allOrders)
 
@@ -99,6 +99,9 @@ const OrdersWidget: React.FC = () => {
         // reset selections
         setOrders(orders.filter(order => !markedForDeletion.has(order.id)))
         setMarkedForDeletion(new Set<string>())
+
+        // update the list of orders
+        forceOrdersRefresh()
       }
     },
     [deleteOrders, markedForDeletion, orders, setMarkedForDeletion, setOrders],
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -104,7 +104,7 @@ const OrdersWidget: React.FC = () => {
         forceOrdersRefresh()
       }
     },
-    [deleteOrders, markedForDeletion, orders, setMarkedForDeletion, setOrders],
+    [deleteOrders, forceOrdersRefresh, markedForDeletion, orders, setMarkedForDeletion, setOrders],
   )
 
   return (
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -70,7 +70,7 @@ export function useOrders(): Result {
             setOrders(filteredOrders)
           } else if (filteredOrders.length) {
             // incremental update: append
-            setOrders(curr => [...curr, ...filteredOrders])
+            setOrders(curr => curr.concat(filteredOrders))
           }
 
           // Save the last seen order
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -27,19 +27,62 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
   )
 }
 
-export function useOrders(): AuctionElement[] {
+interface Result {
+  orders: AuctionElement[]
+  forceOrdersRefresh: () => void
+}
+
+export function useOrders(): Result {
   const { userAddress, networkId, blockNumber } = useWalletConnection()
   const [orders, setOrders] = useSafeState<AuctionElement[]>([])
+  const [lastSeenOrderId, setLastSeenOrderId] = useSafeState<number>(-1)
+  const [offset, setOffset] = useSafeState<number | undefined>(undefined)
+
+  /**
+   * Forces the immediate refresh of orders
+   */
+  function forceOrdersRefresh(): void {
+    // Setting `offset` to 0 triggers a new query from the start
+    setOffset(0)
+  }
+
+  useEffect(() => {
+    // whenever there's a state change (new block, network /address change),
+    // set `offset` to trigger a new query
+    setOffset(lastSeenOrderId + 1)
+  }, [blockNumber, networkId, userAddress])
 
   useEffect(() => {
     userAddress &&
-      networkId &&
-      exchangeApi
-        .getOrders({ userAddress, networkId })
-        .then(filterDeletedOrders)
-        .then(setOrders)
-    // updating list of orders on every block by listening on `blockNumber`
-  }, [networkId, setOrders, userAddress, blockNumber])
+    networkId &&
+    offset !== undefined && // stop querying for new orders when there are no more pages
+      exchangeApi.getOrdersPaginated({ userAddress, networkId, offset }).then(({ orders, nextIndex }) => {
+        if (orders.length) {
+          // Save the id of last order before filtering
+          const orderId = +orders[orders.length - 1].id
 
-  return orders
+          // Apply filters (remove deleted orders)
+          const filteredOrders = filterDeletedOrders(orders)
+
+          // Store new orders, if any
+          if (offset === 0) {
+            // fresh start/refresh: replace whatever is stored
+            setOrders(filteredOrders)
+          } else if (filteredOrders.length) {
+            // incremental update: append
+            setOrders(curr => [...curr, ...filteredOrders])
+          }
+
+          // Save the last seen order
+          setLastSeenOrderId(orderId)
+        }
+
+        // `nextIndex` can be `undefined`, which means there are no more pages
+        // and causes the querying to stop
+        setOffset(nextIndex)
+      })
+    // Update ONLY when `offset` changes
+  }, [offset])
+
+  return { orders, forceOrdersRefresh }
 }
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -94,7 +94,7 @@ export function useOrders(): Result {
       }
 
       await exchangeApi.getOrdersPaginated({ userAddress, networkId, offset }).then(({ orders, nextIndex }) => {
-        if (orders.length) {
+        if (orders.length > 0) {
           // Apply filters (remove deleted orders)
           const filteredOrders = filterDeletedOrders(orders)
 
@@ -102,7 +102,7 @@ export function useOrders(): Result {
           if (offset === 0) {
             // fresh start/refresh: replace whatever is stored
             setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, false))
-          } else if (filteredOrders.length) {
+          } else if (filteredOrders.length > 0) {
             // incremental update: append
             setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, true))
           }
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -6,6 +6,12 @@ import { exchangeApi } from 'api'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
 import { ZERO } from 'const'
 
+interface Result {
+  orders: AuctionElement[]
+  forceOrdersRefresh: () => void
+  isLoading: boolean
+}
+
 /**
  * Filter out deleted orders.
  *
@@ -26,12 +32,6 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
         order.validUntil === 0
       ),
   )
-}
-
-interface Result {
-  orders: AuctionElement[]
-  forceOrdersRefresh: () => void
-  isLoading: boolean
 }
 
 export function useOrders(): Result {
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -1,7 +1,8 @@
+import { useEffect, useCallback } from 'react'
+
 import { useWalletConnection } from './useWalletConnection'
 import useSafeState from './useSafeState'
 import { exchangeApi } from 'api'
-import { useEffect } from 'react'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
 import { ZERO } from 'const'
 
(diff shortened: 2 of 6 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr489--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
```diff
@@ -27,19 +27,62 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+          // Apply filters (remove deleted orders)
+          const filteredOrders = filterDeletedOrders(orders)
+
+          // Store new orders, if any
+          if (offset === 0) {
+            // fresh start/refresh: replace whatever is stored
+            setOrders(filteredOrders)
+          } else if (filteredOrders.length) {
+            // incremental update: append
+            setOrders(curr => [...curr, ...filteredOrders])
```
Velenir:
```suggestion
            setOrders(curr => curr.concat(filteredOrders))
```

a little more optimized

src/hooks/useOrders.ts
```diff
@@ -27,19 +27,62 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+  function forceOrdersRefresh(): void {
+    // Setting `offset` to 0 triggers a new query from the start
+    setOffset(0)
+  }
+
+  useEffect(() => {
+    // whenever there's a state change (new block, network /address change),
+    // set `offset` to trigger a new query
+    setOffset(lastSeenOrderId + 1)
+  }, [blockNumber, networkId, userAddress])
```
Velenir: Change in `networkId, userAddress` should refetch all orders as all orders are network|user specific

src/hooks/useOrders.ts
W3stside: we should pin this optimised array logic that keeps getting picked up in reviews (by dima) ;)

src/components/OrdersWidget/index.tsx
```diff
@@ -99,6 +99,9 @@ const OrdersWidget: React.FC = () => {
         // reset selections
         setOrders(orders.filter(order => !markedForDeletion.has(order.id)))
         setMarkedForDeletion(new Set<string>())
+
+        // update the list of orders
+        forceOrdersRefresh()
```
W3stside: do we need this if we subscribe instead to userAccount changes via subs?

W3stside: or networkId or whatever or block

W3stside: or wait is this what we talked about on the last call that we don't want to do that, don't remember now :thinking:

Velenir: this is to refetch everything on order deletion, regardless network or account or whatever

src/hooks/useOrders.ts
alfetopito: You are 100% correct. After submitting this PR I started working on this. Unfortunately, I haven't been able to make it work it.
It behaves weird when switching network/account for some reason. Still trying to figure it out.

src/hooks/useOrders.ts
```diff
@@ -27,19 +27,62 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
-        .getOrders({ userAddress, networkId })
-        .then(filterDeletedOrders)
-        .then(setOrders)
-    // updating list of orders on every block by listening on `blockNumber`
-  }, [networkId, setOrders, userAddress, blockNumber])
-
-  return orders
+    networkId &&
+    offset !== undefined && // stop querying for new orders when there are no more pages
+      exchangeApi.getOrdersPaginated({ userAddress, networkId, offset }).then(({ orders, nextIndex }) => {
```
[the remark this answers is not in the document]
alfetopito: True, will add it.

src/hooks/useOrders.ts
```diff
@@ -27,19 +27,62 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+  }
+
+  useEffect(() => {
+    // whenever there's a state change (new block, network /address change),
+    // set `offset` to trigger a new query
+    setOffset(lastSeenOrderId + 1)
+  }, [blockNumber, networkId, userAddress])
 
   useEffect(() => {
     userAddress &&
```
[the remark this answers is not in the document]
alfetopito: What do you mean by if statement?

If you mean, only do the update if one of those changed, I think I would need to store these values in a local state to verify whether any has changed and make the useEffect hook be invoked on every change.

src/hooks/useOrders.ts
alfetopito: Got to a somewhat working solution with a bunch of `useEffect`s and duplicated block, network and address. Ugly.

Any suggestions to make it better?

src/components/OrdersWidget/index.tsx
alfetopito: All of the above.

src/hooks/useOrders.ts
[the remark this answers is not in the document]
alfetopito: Ended up doing as I understood your suggestion to overcome the mess that useEffect created.
Let me know if this is what you had in mind.

alfetopito:
> Great!
> 
> I would have some open questions/ideas:
> 
>     * I'm a bit confused by `setOffset(lastSeenOrderId + 1)` logic. Also I have an issue figuring out what would happen if you are in the middle of getting all the pages, and a new block is mined
> 
>     * I believe we should have a flag for when we are loading the orders, so we can represent it in the UI. As I understood, you update the UI on every page until there's no more pages. If we have this loading flag we can add at the bottom of the order list a row with a spinner that shows that we are not done.
> 
>     * There's sth I find not super natural about `setOffset(nextIndex)` for iterating on the pages.  My current thinking is in the direction that we should define a recursive function instead and only update the offset when we don't have more pages at the end (with the last index we have). Then it's just calling fetchAll logic on every block. Since the `offset` is the latest successful order we fetched, we can always easily check if there's new orders
> 
> 
> One idea would be:
> 
>     * Defining the recursive function `getOrdersPage(userAddress, networkId, offset)`
> 
>     * We call this function, at the beginning, with`offset == 0`. Also on `forceOrdersRefresh` is `0` what loads all the orders again. `getOrdersPage(userAddress, networkId, 0)`. I think this is done automatically when we mutate the offset to 0 and for the first load too.
> 
>     * This recursive function would have a step in it's implementation, where we do the recursive call `getOrdersPage(userAddress, networkId, nextIndex)`. Of course inside the implementation we also have the concat logic you implemented here.
> 
>     * If `blockNumber` changes, we can call `getOrdersPage(userAddress, networkId, offset)` that would get any new page if there is new orders only. Or do nothing if there's no new orders. We only do this IF we are not `loading` (so we ensure no re-fetch is done while we are getting all the orders)

Did a refactor based on your suggestions and fixed the issue pointed out by @Velenir 

The code feels weird though. (All of the useEffect for updating the copy of block, network and address).

Try it out and let me know what you think.

src/hooks/useOrders.ts
```diff
@@ -27,19 +28,130 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+      networkId: number | undefined,
+      offset: number | undefined,
+      areThereOrders?: boolean,
+    ): Promise<void> => {
+      if (userAddress === undefined || networkId === undefined || offset === undefined) {
+        return
+      }
+
+      await exchangeApi.getOrdersPaginated({ userAddress, networkId, offset }).then(({ orders, nextIndex }) => {
+        if (orders.length) {
```
anxolin: this is `orders.length !== 0` right? or there's a case where orders.length is `undefined`?

src/hooks/useOrders.ts
```diff
@@ -27,19 +28,130 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+            // incremental update: append
+            setOrders(curr => curr.concat(filteredOrders))
+          }
+        } else if (offset === 0 && areThereOrders) {
+          // There were orders. We fetched again from the beginning, and now there are none (in the first page). Set to []
+          // Example: switching network or address.
+          setOrders([])
+        }
+
+        // `nextIndex` can be `undefined`, which means there are no more pages
```
anxolin: I like very much that you add these kind of comments. Keep doing it pls!

src/hooks/useOrders.ts
```diff
@@ -27,19 +28,130 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
+
+          // Store new orders, if any
+          if (offset === 0) {
+            // fresh start/refresh: replace whatever is stored
+            setOrders(filteredOrders)
+          } else if (filteredOrders.length) {
+            // incremental update: append
+            setOrders(curr => curr.concat(filteredOrders))
+          }
+        } else if (offset === 0 && areThereOrders) {
```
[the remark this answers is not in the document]
alfetopito: Sadly, no.
If I use `orders` from the outer scope, there are two possibilities:
1) The value of `orders` will always be the same -- if using no deps on useCallback -- making it useless
2) The callback will be updated on every change of `orders`, and since the function is recursive, that is not a good thing.

The first version was just the way you describe and took me awhile to understand why it was failing

src/hooks/useOrders.ts
alfetopito: It should always return an array, even when empty: https://github.com/gnosis/dex-react/blob/develop/src/api/exchange/ExchangeApi.ts#L140

src/hooks/useOrders.ts
```diff
@@ -27,19 +28,130 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
-  }, [networkId, setOrders, userAddress, blockNumber])
-
-  return orders
+    setLastUserAddress(userAddress)
+  }, [userAddress, setLastUserAddress])
+
+  useEffect(() => {
+    // whenever there's a state change (new block, network /address change)
+
+    if (!isLoading && networkId && userAddress) {
```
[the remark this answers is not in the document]
alfetopito: Yeah, I think so. At least the first call to update will fail until the current fetch chain is complete.

Hmm, how best to abort a recursion in progress?

Because I think if I just ignore it when changing network/address it might add both set of orders to the same list...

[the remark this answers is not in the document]
alfetopito: Well, even without changing anything in the code this is already a problem.

To reproduce:
1. Using metamask, create two accounts and fund each with some ETH for gas, create several liquidity orders to force pagination.
1. Click on orders page and before it finishes loading, switch the account.

Result:
Orders from both accounts will be loaded and a bunch of errors will be printed on the console from react due to duplicate keys:

```
react-dom.development.js?7f13:531 Warning: Encountered two children with the same key, `45`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

Trying to fix it...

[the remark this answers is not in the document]
alfetopito: Idea: Instead of storing a flat list of orders, I'll use a mapping of network and address to avoid overwriting the list under these race conditions. Something like:

```typescript
interface OrdersState {
  orders: AuctionElement[]
  offset: number
}

interface OrdersPerNetwork {
  [networkId: number]: { [address: string]: OrdersState }
}
```

This is already sort of inline with what I was thinking for the global state. I'll try to get this working today...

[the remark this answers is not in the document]
alfetopito: So, managed to get this idea working, slightly different that my suggestion above on https://github.com/gnosis/dex-react/pull/489/commits/d3e94327153abb985b4f695b7df8560487b84925

Works as expected now when switching addresses in the middle of an update, although every time the orders page is loaded something is throwing this error:

```
Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in OrdersWidget (created by Context.Consumer)
```

Couldn't figure out which hook is causing that though. 
And still works as intended, just has this error in the middle. Partial updates (from new blocks) do not raise the error.

src/hooks/useOrders.ts
anxolin: so maybe would be slightly better to do `orders.length > 0` or `orders.length !== 0`?

src/hooks/useOrders.ts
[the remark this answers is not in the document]
anxolin: okok

src/hooks/useOrders.ts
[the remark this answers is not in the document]
anxolin: Great. Sound good enough. 
It keeps adding complexity, although I like that we don't expose out that complexity. 

I think that if we need to do something similar to this again, we could explore other options like https://rxjs-dev.firebaseapp.com to be able to "produce" the orders, stop fetching on certain events and so on in a more natural way

src/hooks/useOrders.ts
```diff
@@ -27,19 +34,165 @@ function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
...
-        .then(setOrders)
-    // updating list of orders on every block by listening on `blockNumber`
-  }, [networkId, setOrders, userAddress, blockNumber])
-
-  return orders
+    // whenever there's a state change (new block, network /address change)
+
+    if (!isLoading && networkId && userAddress) {
+      // should try to update orders
+      if (networkId !== lastNetworkId || userAddress !== lastUserAddress) {
```
Velenir: I don't quite get the reason for keeping track of `lastNetworkId, lastUserAddress`
How about separate useEffects
```
[networkId, userAddress] => setOffset(0)
[blockNumber] => fetch from offset
```

alfetopito: This is complicated.

I was using two useEffect. 1 for blockNumber one for network/address.
This caused on the initial load to trigger two recursions in parallel, because useEffect is always run on the first render (right?), which made the list of orders contain duplicates.

And this is sort of why tracking I'm tracking lastNetworkId, lastUserAddress.

Unless you know how I could have 2 useEffect in a way that one doesn't clash with the other?

src/hooks/useOrders.ts
alfetopito: Alright

alfetopito:
> This and line 104 can be refactorized? just the isAppend is changing
> 
> ```js
> const isAppend = offset === 0
> setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, isAppend))
> ```

Well, not quite the same thing as the current:

```typescript
// Store new orders, if any
if (offset === 0) {
  // fresh start/refresh: replace whatever is stored
  setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, false))
} else if (filteredOrders.length) {
  // incremental update: append
  setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, true))
}
```

Because there are 3 possible branches, and I'm doing an optimization by not updating the state on the third case: not the first page (offset !== 0) and there are no active orders (filteredOrders.length === 0)

anxolin:
> Because there are 3 possible branches... 

Ok, good for me then. Anyways, maybe possible with sth in this direction. I'm not sure about the `!` and conditions, I'm writing while I'm leaving... :P 

```js
const isAppend = offset !== 0
if (!isAppend || filteredOrders.length === 0) {
    setOrdersState(curr => updateOrdersState(networkId, userAddress, curr, filteredOrders, !isAppend))
}
```

Issue #495, opened 2020-02-04
Add OWL, SNX and sETH to the interface


anxolin:
SNX
https://etherscan.io/token/0xc011a72400e58ecd99ee497cf89e3775d4bd732f
and sETH
https://etherscan.io/token/0x5e74c9036fb86bd7ecdcb084a0673efc32ea31cb?a=0x9a4935749dbdfaf786a19df1c61a6d28b7a6cf94

alfetopito:
For this we'll likely need the token filtering task implemented. (Can't find the issue. Was there one?)

anxolin:
Filtering was for adding OWL for example in the telegram bot and not the web. I understand they want these tokens in the interface

Adding arbitrary tokens should be part of our next sprint, so people can make experiments easier without re-deploying the apps

alfetopito:
> Filtering was for adding OWL for example in the telegram bot and not the web. I understand they want these tokens in the interface

Yes, but keep in mind that the same list is used on 3 places on the web: trade, wallet and liquidity.
As far as I am aware, we won't be displaying non stablecoins on liquidity page, thus the need for filtering.

David left us a TODO indicating where the filtering is currently done to prevent WETH from being listed https://github.com/gnosis/dex-react/blob/develop/src/components/PoolingWidget/index.tsx#L185

Of course, we can filter it out there as well.

PR #497, opened 2020-02-05
Filter tokens not in the liquidity list

anxolin:
Part of this taks https://github.com/gnosis/dex-react/issues/495

I took the simpler approach, I think :) 
We just keep a list of tokens for the liquidity page. The list is maintained there. It looks straightforward and good enough

If in the future we need more fancy stuff we can reevaluate

### DIFF
--- a/src/components/PoolingWidget/index.tsx
+++ b/src/components/PoolingWidget/index.tsx
@@ -30,6 +30,8 @@ import { maxAmountsForSpread, log } from 'utils'
 import { DEFAULT_PRECISION } from 'const'
 import { Link } from 'react-router-dom'
 
+const LIQUIDITY_TOKEN_LIST = ['USDT', 'TUSD', 'USDC', 'PAX', 'GUSD', 'DAI']
+
 interface ProgressBarProps {
   step: number
   stepArray: string[]
@@ -182,10 +184,15 @@ const PoolingInterface: React.FC = () => {
   // Avoid displaying an empty list of tokens when the wallet is not connected
   const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
 
-  // TODO: switched to tagged tokens @anxo @leandro
-  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId).filter(({ symbol }) => symbol !== 'WETH'), [
-    fallBackNetworkId,
-  ])
+  const tokens = useMemo(() => {
+    return (
+      // Get all the tokens for the current network
+      tokenListApi
+        .getTokens(fallBackNetworkId)
+        // Filter out the tokens not in the list
+        .filter(({ symbol }) => symbol && LIQUIDITY_TOKEN_LIST.includes(symbol))
+    )
+  }, [fallBackNetworkId])
 
   const prevStep = useCallback((): void => {
     if (step == 1) return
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr497--dexreact.review.gnosisdev.com

src/components/PoolingWidget/index.tsx
```diff
@@ -182,10 +184,15 @@ const PoolingInterface: React.FC = () => {
...
-  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId).filter(({ symbol }) => symbol !== 'WETH'), [
-    fallBackNetworkId,
-  ])
+  const tokens = useMemo(() => {
+    return (
+      // Get all the tokens for the current network
+      tokenListApi
+        .getTokens(fallBackNetworkId)
+        // Filter out the tokens not in the list
+        .filter(({ symbol }) => symbol && LIQUIDITY_TOKEN_LIST.includes(symbol))
```
Velenir: ~Unnecessary~ small optimization, but `LIQUIDITY_TOKEN_LIST` would be better as a Set

W3stside: i agree!

anxolin: Done! 14a6190d2dc603d5b421ff22b4e6c6f32128030f

src/components/PoolingWidget/index.tsx
```diff
@@ -30,6 +30,8 @@ import { maxAmountsForSpread, log } from 'utils'
 import { DEFAULT_PRECISION } from 'const'
 import { Link } from 'react-router-dom'
 
+const LIQUIDITY_TOKEN_LIST = new Set(['USDT', 'TUSD', 'USDC', 'PAX', 'GUSD', 'DAI'])
```
alfetopito: Shouldn't sUSD be in this list?

src/components/PoolingWidget/index.tsx
```diff
@@ -182,10 +184,15 @@ const PoolingInterface: React.FC = () => {
...
-  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId).filter(({ symbol }) => symbol !== 'WETH'), [
-    fallBackNetworkId,
-  ])
+  const tokens = useMemo(() => {
+    return (
+      // Get all the tokens for the current network
+      tokenListApi
+        .getTokens(fallBackNetworkId)
+        // Filter out the tokens not in the list
+        .filter(({ symbol }) => symbol && LIQUIDITY_TOKEN_LIST.has(symbol))
```
alfetopito:
```suggestion
        .filter(({ symbol }) => LIQUIDITY_TOKEN_LIST.has(symbol))
```

Checking for `undefined` in the set should work, right?

src/components/PoolingWidget/index.tsx
anxolin: doing in the release

src/components/PoolingWidget/index.tsx
anxolin: I'm afraid it's expecting a String:
image

alfetopito: Ah right, typescript, typescripting :)

PR #498, opened 2020-02-05
478/optimistic orders

### LINKED ISSUE #478: The order is not showing while the transaction is not mined
No order is shown in `/orders` page until transaction with that order has been mined

Optimistically show that order upon transaction sent:

1. get hash, order data
2. persist in memory or localStorage
3. list it in orders

**Estimation: 3**

W3stside:
Closes #478 

Curious to get feedback on this...

uses GlobalState for pending oders, and adding to localStorage on exit

Still, the behaviour is strange in the Orders page, the tables flicker as sth is happening with the conditional render of the tables, e.g `pendingOrders.length && renderTable` is probaby what is happening. Not sure why the `useEffect` inside `useOrders` would affect this but I am looking to you for help.

Still a draft at this point

To test - simply make some orders in trade and dont confirm, just switch from Trade to some other page then make trades, approve them all and then go directly to Orders page.

For now the way it's rendering 1 table on top of another is weird but it's just to get a feel for rendering the data


gnosis-info:
Travis automatic deployment:
 https://pr498--dexreact.review.gnosisdev.com

src/reducers-actions/index.ts
```diff
@@ -14,8 +20,12 @@ export interface GlobalState {
  * make sure the name of the state key(s) is/are the same as the reducer key(s) below
  */
 
-export const INITIAL_STATE: GlobalState = {
-  tokens,
+export const INITIAL_STATE: () => GlobalState = () => {
+  const initiatedPendingOrders = pendingOrders()
+  return {
+    tokens,
+    pendingOrders: initiatedPendingOrders,
```
alfetopito: nitpick: why not use `PendingOrdersInitialState` directly?

src/reducers-actions/pendingOrders.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const enum ActionTypes {
+  SAVE_PENDING_ORDERS = 'SAVE_PENDING_ORDERS',
+  REMOVE_PENDING_ORDERS = 'REMOVE_PENDING_ORDERS',
+}
+
+interface Actions<T> {
+  type: ActionTypes
+  payload: T
+}
```
alfetopito: Seems very similar to tokenRow actions.

It it worth making a generic Actions interface?

src/reducers-actions/pendingOrders.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const savePendingOrdersAction = (payload: {
+  orders: PendingTxObj
+  networkId: number
+}): Actions<{
+  orders: PendingTxObj
+  networkId: number
+}> => ({
```
alfetopito: As much as I don't like several interfaces scattered about (looking at you, ExchangeApi!) it makes the function signature more readable. Could you do that here as well?

src/reducers-actions/pendingOrders.ts
```diff
@@ -0,0 +1,79 @@
...
+  type: ActionTypes.REMOVE_PENDING_ORDERS,
+  payload,
+})
+
+export type PendingOrdersState = typeof GP_ORDER_TX_HASHES
+
+export const EMPTY_PENDING_ORDERS_STATE = {
+  1: [],
+  4: [],
+}
```
alfetopito: One thing that I realized when testing the pagination is that address is another dimension to be considered.
In this case, when changing addresses the pending txs will still be the same, no?

src/const.ts
```diff
@@ -72,3 +72,8 @@ export const INFURA_ID = '8b4d9b4306294d2e92e0775ff1075066'
 export const INITIAL_INFURA_ENDPOINT = `wss://rinkeby.infura.io/ws/v3/${INFURA_ID}`
 
 export const STORAGE_KEY_LAST_PROVIDER = 'lastProvider'
+
+export const GP_ORDER_TX_HASHES = {
```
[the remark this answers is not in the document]
W3stside: Gnosis Protocol

src/reducers-actions/pendingOrders.ts
W3stside: :+1:

src/reducers-actions/pendingOrders.ts
W3stside: you mean making `PendingOrderInterface` instead of 

```js
{
  orders: PendingTxObj
  networkId: number
}
``` 
?

src/reducers-actions/pendingOrders.ts
W3stside: ah FUCKKKKK

W3stside: true

src/components/OrdersWidget/index.tsx
```diff
@@ -37,10 +37,11 @@ const ShowOrdersButton: React.FC<ShowOrdersButtonProps> = ({ type, isActive, sho
 }
 
 const OrdersWidget: React.FC = () => {
-  const allOrders = useOrders()
+  const { orders: allOrders, pendingOrders } = useOrders()
```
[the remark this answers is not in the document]
W3stside: bleh

src/components/TradeWidget/index.tsx
```diff
@@ -160,18 +164,44 @@ const TradeWidget: React.FC = () => {
...
       const { success } = await placeOrder({
         buyAmount,
         buyToken: cachedBuyToken,
         sellAmount,
         sellToken: cachedSellToken,
+        txOptionalParams: {
+          onSentTransaction: (txHash: string): void => {
+            pendingTxHash = txHash
+
+            if (!networkId) return
```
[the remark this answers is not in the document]
W3stside: ah yep, thanks

src/reducers-actions/pendingOrders.ts
W3stside: ok switched it, good catch

src/reducers-actions/pendingOrders.ts
alfetopito: Yep!

src/hooks/useOrders.ts
```diff
@@ -188,7 +188,7 @@ export function useOrders(): Result {
    */
   const forceOrdersRefresh = useCallback((): void => {
     fetchOrdersAndHandleErrors(userAddress, networkId, 0, areThereOrders(networkId, userAddress, ordersState))
-  }, [fetchOrdersAndHandleErrors, userAddress, networkId, ordersState, areThereOrders])
+  }, [fetchOrdersAndHandleErrors, userAddress, networkId, ordersState])
```
alfetopito: Why remove `areThereOrders`?

src/hooks/usePendingOrders.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function usePendingOrders(): PendingTxArray {
+  const { userAddress, networkId } = useWalletConnection()
+
+  const [{ pendingOrders: pendingOrdersGlobal }] = useGlobalState()
+  const [pendingOrders, setPendingOrders] = useSafeState<PendingTxArray>([])
+
+  useEffect(() => {
+    userAddress && networkId && setPendingOrders(pendingOrdersGlobal[networkId][userAddress] || [])
+    // updating list of orders on every block by listening on `blockNumber`
```
alfetopito: You are not listening to `blockNumber`.
I know this is from `useOrders` hook, but should you be updating pending orders on every block?

src/components/OrdersWidget/index.tsx
```diff
@@ -153,6 +153,50 @@ const OrdersWidget: React.FC = () => {
...
+                >
+                  <thead>
+                    <tr>
+                      <th className="checked">
+                        <input
+                          type="checkbox"
+                          onChange={toggleSelectAll}
+                          checked={orders.length === markedForDeletion.size}
+                          disabled={deleting}
+                        />
```
alfetopito: I think the checkbox should either be disabled or not present for pending orders, because we can't cancel.
Also, when checking the box it select only active orders.

Yes, this will be replaced soon, so might not be that relevant.

src/hooks/useOrders.ts
W3stside: I think my hooks linter did that :scream:

src/components/OrdersWidget/index.tsx
W3stside: yeah thats what I was thinking, leaving out any design stuff

src/hooks/usePendingOrders.ts
W3stside: pending orders are only detected via the Trade page on submission - so checking via block update wouldn't do anything as they are auto filtered out on mining via the async place order function - unless I'm missing sth

alfetopito: Then it's fine, just remove the comment :)

W3stside: :+1:

src/hooks/useOrders.ts
W3stside: getting:
```ts
React Hook useCallback has an unnecessary dependency: 'areThereOrders'. Either exclude it or remove the dependency array. Outer scope values like 'areThereOrders' aren't valid dependencies because mutating them doesn't re-render the component.eslint(react-hooks/exhaustive-deps)
```

alfetopito: Nothing against removing it, just wanted to understand why.

BTW, something is not working on my vscode, because the hooks deps are not being updated anymore.

W3stside: try reinstalling the addon?

PR #510, opened 2020-02-10
useOrders v2

Velenir:
minor refactor for `useOrders` hook


src/hooks/useOrders.ts
```diff
@@ -71,6 +73,92 @@ function areThereOrders(networkId: number | undefined, userAddress: string | und
...
+        // check cancelled bool from parent scope
+        if (cancelled) return
+
+        // Apply filters (remove deleted orders)
+        const filteredOrders = filterDeletedOrders(orders)
+
+        // ensures we don't have multiple reruns for each update
+        // i.e. offset change -> render
+        //      isLoading change -> another render
+        unstable_batchedUpdates(() => {
```
alfetopito: This is nice! Probably why I was having so many re-renders on my first try.

src/hooks/useOrders.ts
```diff
@@ -71,6 +73,92 @@ function areThereOrders(networkId: number | undefined, userAddress: string | und
...
+  const forceOrdersRefresh = useCallback((): void => {
+    setOffset(0)
+    setOrders([])
+  }, [setOffset, setOrders])
+
+  useEffect(() => {
+    // fresh start/refresh: replace whatever is stored
+    forceOrdersRefresh()
+    // whenever userAddress or networkId changes
+  }, [userAddress, networkId, setIsLoading, setOffset, setOrders, forceOrdersRefresh])
```
alfetopito: Do you need to depend on these 3 functions? `setIsLoading, setOffset, setOrders`

src/hooks/useOrders.ts
```diff
@@ -71,6 +73,92 @@ function areThereOrders(networkId: number | undefined, userAddress: string | und
...
+      }
+    }
+
+    fetchOrders(offset)
+
+    // if offset changed mid-fetch
+    // which happens when forceOrdersRefresh si called
+    // ignore current fetch results
+    return (): void => {
+      cancelled = true
```
alfetopito: In this case `useRef` would be better (some react warning/linter told me so)?

src/hooks/useOrders.ts
Velenir: blame linter

src/hooks/useOrders.ts
Velenir: blame linter but also don't trust it too much

alfetopito: Not saying it told me now. I saw this is the past as a best practice for having 'instance variables' like behaviour on hooks/FC is to use `useRef` rather than plain variables for reasons.
Not really needed?

src/hooks/useOrders.ts
```diff
@@ -71,6 +73,92 @@ function areThereOrders(networkId: number | undefined, userAddress: string | und
...
+          if (!nextIndex) {
+            // no more orders left
+            // done fetching for now
+            setIsLoading(false)
+          }
+          // move offset forward, fetch new batch
+          setOffset(offset + orders.length)
+        })
+      } catch (error) {
+        console.log('Failed to fetch orders', e)
```
alfetopito: Build failure:

```
ERROR in /home/travis/build/gnosis/dex-react/src/hooks/useOrders.ts(123,47):

TS2304: Cannot find name 'e'.
```

```suggestion
        console.log('Failed to fetch orders', error)
```

src/hooks/useOrders.ts
Velenir: we need a variable in function scope as close as possible
so inside `useEffect` here
`useRef` would be outer scope (relatively) and would apply to all `fetchOrders` calls
whereas we need only one cancelled per `fetchOrders` call

gnosis-info:
Travis automatic deployment:
 https://pr510--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
Velenir: I don't get that linter error for some reason

src/hooks/useOrders.ts
Velenir: can't believe I missed that

W3stside: tsk tsk

W3stside:
i would say run with this version so i can merge into usePendingOrders

PR #513, opened 2020-02-11
useOrders optimization

Velenir:
small change to eleminate one rerender

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -69,7 +69,8 @@ export function useOrders(): Result {
           if (offset === 0) {
             // fresh start/refresh: replace whatever is stored
             setOrders(filteredOrders)
-          } else {
+            // only if we have something to add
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr513--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
```diff
@@ -69,7 +69,8 @@ export function useOrders(): Result {
           if (offset === 0) {
             // fresh start/refresh: replace whatever is stored
             setOrders(filteredOrders)
-          } else {
+            // only if we have something to add
+          } else if (filterDeletedOrders.length > 0) {
```
alfetopito: I had that before on my super not complicated at all previous version :D

Velenir: And you were right to have it :+1: 
I somehow deleted it during refactor

PR #514, opened 2020-02-11
When a token is not registered in the exchange, do not allow the order to be placed

### LINKED ISSUE #511: Trades with 1 unregistered token does not fail and use OWL instead
### Steps to reproduce

1. On Rinkeby network, create a trade with any Synthetics token and DAI.
1. Once the order is confirmed, check the Orders page.
1. Your order will be DAI <-> OWL instead of the selected token

### Reasoning

- OWL is token id `0` by convention on the contract.
- Synthetics tokens have been added recently and only to Mainnet, not yet to Rinkeby
- Before placing an order, we check for the registered token id on the contract.
- When a token is not registered, it returns `0`.

### Why this is a problem

Currently, this is not a big deal on Mainnet since all tokens are properly registered and listed tokens are always registered.
But, once we allow any tokens to be traded and some happen to not be registered, we are not identifying that on the interface and users might be trading against OWL by accident.

alfetopito:
Closes #511 

No longer able to place orders using tokens that are not registered in the exchange:

screenshot_2020-02-11_13-17-54

Works when 1 or both tokens are not registered.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -33,6 +33,9 @@ export const getDefaultProvider = (): string | null =>
 function createWeb3Api(): Web3 {
   // TODO: Create an `EthereumApi` https://github.com/gnosis/dex-react/issues/331
   const web3 = new Web3(getDefaultProvider())
+  // `handleRevert = true` makes `require` failures to throw
+  // For more details see https://github.com/gnosis/dex-react/issues/511
+  web3.eth['handleRevert'] = true
 
   if (process.env.MOCK_WEB3 === 'true') {
     // Only function that needs to be mocked so far. We can add more and add extra logic as needed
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr514--dexreact.review.gnosisdev.com

src/api/index.ts
```diff
@@ -33,6 +33,9 @@ export const getDefaultProvider = (): string | null =>
 function createWeb3Api(): Web3 {
   // TODO: Create an `EthereumApi` https://github.com/gnosis/dex-react/issues/331
   const web3 = new Web3(getDefaultProvider())
+  // `handleRevert = true` makes `require` failures to throw
+  // For more details see https://github.com/gnosis/dex-react/issues/511
+  web3.eth['handleRevert'] = true
```
Velenir: Is it a new addition, seeing as it's not even typed?

W3stside: never seen this

alfetopito: Not sure, it's on the docs though https://web3js.readthedocs.io/en/v1.2.5/web3-eth.html#handlerevert

PR #515, opened 2020-02-12
fix useQuery updates

Velenir:
`useQuery` doesn't update when `location.search` changes
Luckily it doesn't break anything as the source of truth after the first load is the form, but it potentially could

See reproduction gif where `sellAmount` stays the initial value
useQuery


gnosis-info:
Travis automatic deployment:
 https://pr515--dexreact.review.gnosisdev.com

PR #519, opened 2020-02-12
Fix refetch from 0 on mount

Velenir:
`useEffect` is run when deps changes
That includes on mount/unmount
So 
```
useEffect(() => {
    forceOrdersRefresh()
    dispatch(overwriteOrders({ orders: [] }))
  }, [userAddress, networkId, forceOrdersRefresh, dispatch])

```

Would run and `forceRefresh` whenever we mount `OrdersRow`

One solution is to skip first run.
Another could be to centralize forceRefresh somewhere in global state near Provider, but then we would fetch even when we don't need to display orders

aside: setState is batched by default inside `useEffect` and in sync event handlers. some time in the future, with concurrent mode, will be batched everywhere

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -117,18 +117,14 @@ export function useOrders(): Result {
 
   // allow to fresh start/refresh on demand
   const forceOrdersRefresh = useCallback((): void => {
-    unstable_batchedUpdates(() => {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr519--dexreact.review.gnosisdev.com

PR #520, opened 2020-02-12
455/global orders

### LINKED ISSUE #480: Paginate when getting orders for the user
Simple 1st version:

1. get all
2. keep updating on block
3. on cancel we fetch everything again

Drawbacks:

- frontend only aware of order state change (cancellation) through frontend
- big refetch on cancel

Related #455
Initial work done in #473

**Estimation: 4**

alfetopito:
Closes #480

Moving orders to global state.

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -76,10 +76,10 @@ export function useOrders(): Result {
         unstable_batchedUpdates(() => {
           if (offset === 0) {
             // fresh start/refresh: replace whatever is stored
-            dispatch(overwriteOrders({ orders: filteredOrders }))
+            dispatch(overwriteOrders(filteredOrders))
           } else if (filteredOrders.length > 0) {
             // incremental update: append
-            dispatch(appendOrders({ orders: filteredOrders }))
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr520--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
```diff
@@ -110,14 +115,21 @@ export function useOrders(): Result {
...
-    setOffset(0)
+    dispatch(updateOffset(0))
     setIsLoading(true)
-  }, [setIsLoading, setOffset])
+  }, [dispatch, setIsLoading])
+
+  const runEffect = useRef(false)
 
   useEffect(() => {
+    if (!runEffect.current) return
```
Velenir: I thought about it and a simpler approach would be
```ts
if (!runEffect.current) {
  runEffect.current = true
  return
}
```

instead of an extra `useEffect`
Sorry about that

alfetopito: Unforgivable!

alfetopito: Done

PR #521, opened 2020-02-13
Mini improv for bundle compilation

Velenir:
Some improvements after I finally noticed that `parsed size > stats size` in bundle report :flushed: :see_no_evil:


gnosis-info:
Travis automatic deployment:
 https://pr521--dexreact.review.gnosisdev.com

PR #523, opened 2020-02-13
482.5/fix form reset

W3stside:
```js
const DEFAULT_FORM_STATE = {
  sellToken: '0',
  receiveToken: '0',
  validUntil: '30',
}
```

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -69,6 +69,12 @@ export type TradeFormData = {
   [K in keyof typeof TradeFormTokenId]: string
 }
 
+const DEFAULT_FORM_STATE = {
+  sellToken: '0',
+  receiveToken: '0',
+  validUntil: '30',
+}
+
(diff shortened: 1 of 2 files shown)


src/components/TradeWidget/index.tsx
```diff
@@ -184,7 +190,7 @@ const TradeWidget: React.FC = () => {
       })
       if (success) {
         // reset form on successful order placing
-        reset()
+        reset(DEFAULT_FORM_STATE)
```
Velenir: Should this also go into `L102::defaultValues`

alfetopito: Don't think so, because those default values are used to set the form values from the url.
Resetting it here should be enough.

Velenir: Still, something is wrong with resetting. Unlimited checkbox gets stuck.

gnosis-info:
Travis automatic deployment:
 https://pr523--dexreact.review.gnosisdev.com

Issue #534, opened 2020-02-13
Review-refactor Michel re-designed parts


alfetopito:
* Remove pagewrapper parent from ORDERS overview and fully make it functional on the new Trade page

### Estimation:
*3*

anxolin:
This doesn't need an issue

PR #541, opened 2020-02-14
526/time remaining in batch

### LINKED ISSUE #526: Display time remaining in batch
There was a task for that before https://github.com/gnosis/dex-react/issues/235

We decided to add it and use local time instead of blockchain time.

### Notes:

Also in RF.1.2 Timer of the current batch

using local time of the browser

### Estimation:
*1.5*

alfetopito:
Closes #526 

- Uses local time to calculate time remaining in batch
- Code on `feature/ui-redesign` branch since it has the time already on the UI
- NOT passing tests/build due to linter issues. Will be addressed on https://github.com/gnosis/dex-react/issues/534
- Able to run locally with `yarn mock` and see the clock ticking :)

### Note:
Not sure there's a smarter way to format seconds into XXm XXs from seconds rather than from a Date or timestamp. Open to suggestions.

### DIFF
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -35,6 +35,32 @@ export function formatDateFromBatchId(batchId: number): string {
   return formatDistanceToNow(date, { addSuffix: true })
 }
 
+/**
+ * Calculates seconds remaining in current batch
+ * Assumes local time is accurate and can be used as source of truth
+ */
+export function getSecondsRemainingInBatch(): number {
+  return BATCH_TIME - (getEpoch() % BATCH_TIME)
+}
+
+export function formatSecondsToMinutes(seconds: number): string {
+  const minutes = Math.floor(seconds / 60)
+  const remainderSeconds = seconds % 60
+  let s = ''
+
+  if (minutes > 0) {
+    s += `${minutes}m `
+  }
+  if (remainderSeconds > 0) {
+    s += `${remainderSeconds}s`
+  }
+  if (minutes === 0 && remainderSeconds === 0) {
+    s = '0s'
+  }
+
+  return s
+}
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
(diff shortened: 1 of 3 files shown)


src/utils/time.ts
```diff
@@ -35,6 +35,32 @@ export function formatDateFromBatchId(batchId: number): string {
...
 
+/**
+ * Calculates seconds remaining in current batch
+ * Assumes local time is accurate and can be used as source of truth
+ */
+export function getSecondsRemainingInBatch(): number {
+  return BATCH_TIME - (getEpoch() % BATCH_TIME)
+}
+
+export function formatSecondsToMinutes(seconds: number): string {
```
anxolin: I would call this just `formatSeconds` since it can be missleasing to think that `30` --> `30s` and not `0.5m`

Velenir: Cool but wouldn't `date-fns` have a function for that

alfetopito: That's what I though, but as far as I can see, their functions are Date/timestamp based. I would need to add those seconds to now() and compare against it with something like https://date-fns.org/v2.9.0/docs/formatDistanceToNow

Or is there a simpler way?

alfetopito: See https://github.com/gnosis/dex-react/pull/542

Velenir:
```js
const date = new Date(seconds * 1000)
const minutes = date.getMinutes()
const seconds = date.getSeconds()
```
Is it easier? Not really

As for `date-fn` the best I found
```sh
> d.formatDistance(17000, 0, { includeSeconds: true})
'less than 20 seconds'
> d.formatDistance(97000, 0, { includeSeconds: true})
'2 minutes'
> d.formatDistance(77000, 0, { includeSeconds: true})
'1 minute'
```

which is not ideal

alfetopito: Not much better in my opinion. I'd like to keep the current version in this case.

Issue #481, opened 2020-01-31
Define the order by sell amount and price


Velenir:
1. Add inputs for price
2. Make buy amount input readonly
3. Make buy amount function of `(sellAmount, price)` and update in real time

**Estimation: ~~2.5~~ 2**

PR #551, opened 2020-02-17
481/add price

anxolin:
Closes: #481 

Main task:
* Adds the two prices A/B and B/A to define an order
* Make `receive amount` read only
* Make `receive amount` calculated field: `sell amount x price`
* If you enter price A/B, the B/A is populated automatically
* If you enter price B/A , the A/B is populated automatically

image

**Also Includes**

The query params has been modified:
* Now `buy` amount is not allowed any more as a param --> requires also a change in the telegram bot (note that this will also solve the issue with the unlimited orders)
* Added `price` param
* Default price is 0

Some refactors:
* Moves the `PriceWrapper` to it's own component, where the two prices logic/UI are handled
* Refactors the utils for `onChange` and `onBlur`  logic in the token row so it can be used also in the prices too 
* `WalletDetail` is the component used for displaying the error and warning messages.  It has been moved to its own component so it can be reused also for the prices and to make the TradeWidget smaller. This component is called now `FormMessage`

Added some utils
* parse util for big numbers
* NOTE: The utils need to be moved to dex-js at some point. I've added a TODO, but we need to address this in a separate one

*ISSUES*
* The tooltip wrapper doesn't like functional components. Added a TODO for now since it's not part of this PR. You might see a warning in the console
* The tooltip wrapper breaks the form. I think it doesn't work the magic of `useForm` hook for some reason. This is why is commented out


gnosis-info:
Travis automatic deployment:
 https://pr551--dexreact.review.gnosisdev.com

Velenir:
> The tooltip wrapper breaks the form. I think it doesn't work the magic of useForm hook for some reason. This is why is commented out

As we trigger tooltips on hover and focus, if useForm uses any of those (I bet it does focus/blur), they conflict
Need to try without focus or composing handlers without TooltipWrapper

anxolin:
@Velenir  thanks for the hint for forwarding ref to solve the issue. The warning is gone with this commit:
b98281e2222e6027fd2047f47bf7f9a138108e99

src/components/TradeWidget/index.tsx
```diff
@@ -275,25 +217,45 @@ const TradeWidget: React.FC = () => {
...
+
+  // Update receive amount
+  useEffect(() => {
+    let receiveAmount: string | undefined
+    if (priceValue && sellValue) {
+      const sellAmount = parseBigNumber(sellValue)
+      const price = parseBigNumber(priceValue)
+
+      if (sellAmount && price) {
+        receiveAmount = sellAmount.multipliedBy(price).toString()
```
alfetopito:
```suggestion
        receiveAmount = sellAmount.multipliedBy(price).toString(10)
```

Default BigNumber toString() uses exponential notation. Setting the base (10) forces it not to.

Unless we do want exponential notation in the price?

Examples:

Screenshot from 2020-02-20 09-12-24
Screenshot from 2020-02-20 09-11-19

src/hooks/useQuery.ts
```diff
@@ -3,15 +3,15 @@ import { useMemo } from 'react'
 import { sanitizeInput, sanitizeNegativeAndMakeMultipleOf } from 'utils'
 import { DEFAULT_FORM_STATE } from 'components/TradeWidget'
 
-export function useQuery(): { sellAmount: string; buyAmount: string; validUntil?: string } {
+export function useQuery(): { sellAmount: string; price: string; validUntil?: string } {
```
alfetopito: Did you update the default route on App.tsx as well?

alfetopito:
Oh yes, I think this was a feedback from the user try outs. Store selections (tokens, values, prices) on global state.
Not on this PR though, it's quite big already.

Velenir:
I have to input both DAI/USDC and USDC/DAI prices for Submit button to become enabled, even though it autocalcs second price based on first

ezgif-7-0dd35f652d80

src/components/TradeWidget/Price.tsx
```diff
@@ -0,0 +1,190 @@
+import React from 'react'
+import styled from 'styled-components'
+import { TokenDetails } from 'types'
+import { useFormContext } from 'react-hook-form'
+import { TradeFormData } from '.'
+import BigNumber from 'bignumber.js'
+import { parseBigNumber, validatePositive, validInputPattern } from 'utils'
+import FormMessage from './FormMessage'
+import { useNumberInput } from './useNumberInput'
+import { DEFAULT_PRECISION } from '../../../test/data'
```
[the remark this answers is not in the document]
anxolin: it should! thanks

src/components/TradeWidget/Price.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <PriceInputBox>
+        <label>
+          <input
+            className={isError ? 'error' : ''}
+            name={priceInputId}
+            type="text"
+            onChange={(e: React.ChangeEvent<HTMLInputElement>): void => {
+              enforcePrecisionPrice()
+              onChangePrice(priceInverseInputId, e)
+            }}
```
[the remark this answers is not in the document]
anxolin: 2cf39a6

src/components/TradeWidget/index.tsx
```diff
@@ -263,10 +200,15 @@ const TradeWidget: React.FC = () => {
   const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
 
   const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
+  const sellInputId = TradeFormTokenId.sellToken
+  const receiveInputId = TradeFormTokenId.receiveToken
+  const priceInputId = TradeFormTokenId.price
+  const priceInverseInputId = TradeFormTokenId.priceInverse
+  const validUntilId = TradeFormTokenId.validUntil
```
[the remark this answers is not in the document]
anxolin: It's a good suggestion, how ever, as far as I know:
```
# this don't work, because TradeFormTokenId is a enum, not an object
const sellInputId = const {
    sellToken: sellInputId,
    receiveToken: receiveInputId,
    price: priceInputId,
    priceInverse: priceInverseInputId,
    validUntil: validUntilId,
  } = TradeFormTokenId
```

Or is there a way? @alfetopito

src/components/TradeWidget/index.tsx
```diff
@@ -275,25 +217,45 @@ const TradeWidget: React.FC = () => {
     () =>
       getToken('symbol', receiveTokenSymbol, tokens) || (getToken('symbol', 'USDC', tokens) as Required<TokenDetails>),
   )
-  const [unlimited, setUnlimited] = useState(!validUntil || !Number(validUntil))
-  const sellInputId = TradeFormTokenId.sellToken
-  const receiveInputId = TradeFormTokenId.receiveToken
-  const validUntilId = TradeFormTokenId.validUntil
+  const [unlimited, setUnlimited] = useState(!validUntilParam || !Number(validUntilParam))
```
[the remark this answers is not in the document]
anxolin: I thought about this, and my thinking is that we don't need the additional checks of the "useSafeState" for this case.

This is used only to init the form

src/components/TradeWidget/index.tsx
anxolin: Nop, decimal is better I think. Thanks for the tip!

src/components/TradeWidget/index.tsx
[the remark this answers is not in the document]
alfetopito: You are doing `const` twice

```
const sellInputId = const {...} = TradeFormTokenId
```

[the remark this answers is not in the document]
alfetopito: But then again, not important, just a nit pick.

alfetopito:
Damn, still happening for some reason:

Screenshot from 2020-02-21 11-18-49


Also (for another PR), tab order is crazy! From sell amount input it goes to Order tab link. We might need to add tabIndex again.

anxolin:
> Also (for another PR), tab order is crazy! From sell amount input it goes to Order tab link. We might need to add tabIndex again.

yes, I saw that too. We need a lot to polish :)

src/components/TradeWidget/index.tsx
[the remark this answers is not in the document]
anxolin: yep, the code I added had the double const. However, I think the issue is that enum is a type, cannot be destructured as far as I know.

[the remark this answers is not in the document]
alfetopito: Got it, please disregard.

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -265,42 +211,13 @@ const TokenRow: React.FC<Props> = ({
...
-    [inputId, setValue],
-  )
-
-  const onKeyPress = useCallback(
-    (event: React.KeyboardEvent<HTMLInputElement>): void =>
-      event.key === 'Enter' ? removeExcessZeros(event) : preventInvalidChars(event),
-    [removeExcessZeros],
-  )
+  // Form validation
+  const inputRef = !readOnly
```
Velenir: Can `readOnly` change when the component is mounted? Because I don't think it will re-register as react-hook-form does it only once.
Totally ok if `readOnly` is the same throughout the life of the component

src/components/TradeWidget/useNumberInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const enforcePrecision = useCallback(() => {
+    const newValue = adjustPrecision(inputValue, precision)
+    if (inputValue !== newValue) {
+      setValue(inputId, newValue, true)
+    }
+  }, [inputId, inputValue, precision, setValue])
+
+  useEffect(() => {
+    enforcePrecision()
+  }, [enforcePrecision])
```
Velenir: An almost empty hook, funny :smile:

src/hooks/useQuery.ts
anxolin: Now i do. Thanks! b3528e04376687c98fed3c10d35a4a77ba908e49

src/components/TradeWidget/TokenRow.tsx
anxolin: Readonly don't change, cause it's only used to define the receivingAmount as readOnly. That will remain true forever.

Hoever, as a component, no-one warranties the use you will give to that property, so I agree it can be an issue. It's a bad practice to depend on "how you use the props"

anxolin: Do you see a simple solution?

src/components/TradeWidget/useNumberInput.tsx
anxolin: Does it have an issue? 

I think we can change it to:
`useEffect(enforcePrecision, [enforcePrecision])`

Doing the change now. Let me know if it's fine :)

Velenir: :+1:

src/components/TradeWidget/TokenRow.tsx
Velenir:
```ts
const inputRef = useMemo(() => !readOnly
   ? register({
        pattern: { value: validInputPattern, message: 'Invalid amount' },
        validate: { positive: validatePositive },
      })
    : register,
// eslint-disable don't-remember-what
[register])
```

Velenir: No, don't include readOnly in deps, the whole point is to not depend on it
And don't copy my eslint comment, lol

anxolin: ahahahah

PR #563, opened 2020-02-18
block form only until txHash (dont wait mine)

### LINKED ISSUE #561: Trade page blocked until mined tx - unblock
Currently users arent able to make orders quikcly in succession as the trade page form is blocked until tx is mined

with pending txs in place this makes zero sense - unblock the form to allow users to make consecutive txs

for 0.8.2

W3stside:
Closes #561 

Blocks on initial `onSubmit`
Unblocks on `txHash` via the `onSentTransaction` callback

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -4,11 +4,13 @@ import { faExchangeAlt, faPaperPlane, faSpinner } from '@fortawesome/free-solid-
 import { FieldValues } from 'react-hook-form/dist/types'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { useHistory } from 'react-router'
+import { toast } from 'toastify'
 
 import TokenRow from './TokenRow'
 import OrderDetails from './OrderDetails'
 import OrderValidity from './OrderValidity'
 import Widget from 'components/Layout/Widget'
+import { TxNotification } from 'components/TxNotification'
 
 import { useForm, FormContext } from 'react-hook-form'
 import { useParams } from 'react-router'
@@ -111,7 +113,7 @@ const TradeWidget: React.FC = () => {
       [validUntilId]: validUntil,
     },
   })
-  const { handleSubmit, watch, reset } = methods
+  const { handleSubmit, watch } = methods
 
   const searchQuery = buildSearchQuery({
     sell: watch(sellInputId),
@@ -147,7 +149,7 @@ const TradeWidget: React.FC = () => {
     [NULL_BALANCE_TOKEN, balances, receiveToken],
   )
 
-  const { isSubmitting, placeOrder } = usePlaceOrder()
+  const { placeOrder, isSubmitting, setIsSubmitting } = usePlaceOrder()
   const history = useHistory()
 
   const swapTokens = useCallback((): void => {
@@ -188,6 +190,9 @@ const TradeWidget: React.FC = () => {
 
     if (isConnected && userAddress) {
       let pendingTxHash: string | undefined = undefined
+      // block form
+      setIsSubmitting(true)
+
       const { success } = await placeOrder({
         buyAmount,
         buyToken: cachedBuyToken,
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr563--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -217,9 +226,11 @@ const TradeWidget: React.FC = () => {
         },
       })
       if (success && pendingTxHash) {
+        // Following 2 fns: Commented out for testing consecutive txs
```
alfetopito: Is it intentional to push this test?

src/hooks/usePlaceOrder.ts
```diff
@@ -60,7 +61,6 @@ export const usePlaceOrder = (): Result => {
         return { success: false }
       }
 
-      setIsSubmitting(true)
```
alfetopito: Should the behaviour also change for liquidity?

Even though the answer might be `no`, I think the other function (`placeMultipleOrders`) should match the change on this, because we'll soon be using that to place orders (when setting custom `from` date)

Thus I'd say to also update that function and liquidity code.

W3stside: ok, makes sense

src/components/TradeWidget/index.tsx
W3stside: ill remove

src/components/TradeWidget/index.tsx
```diff
@@ -217,10 +226,6 @@ const TradeWidget: React.FC = () => {
         },
       })
       if (success && pendingTxHash) {
-        // reset form on successful order placing
-        reset(DEFAULT_FORM_STATE)
-        setUnlimited(false)
-        // remove pending tx
```
alfetopito: What I meant in the previous comment was: do you need push this test?
In other words, will the form no longer reset and `setUnlimit false` ?

W3stside: it shouldnt - if a user is in the middle of making a new order and the old tx mines, then the form resets... sucks as UX

W3stside: yeah probably best to move this directly after tx hash, not remove entirely

src/components/TradeWidget/index.tsx
```diff
@@ -196,7 +201,14 @@ const TradeWidget: React.FC = () => {
         validUntil,
         txOptionalParams: {
           onSentTransaction: (txHash: string): void => {
+            // reset form on successful order placing
+            reset(DEFAULT_FORM_STATE)
+            setUnlimited(false)
+            // unblock form
+            setIsSubmitting(false)
```
Velenir: That's a lot of setState, wrap in unstable_batch

W3stside:
> Nice!
> 
> One question though. Why `setIsSubmitting` is exposed in the hook?
> Wouldn't this be responsibility of the hook implementation to know, so you just expose the isSubmitting?

it's exposed as I needed access to it via the `onSubmitTransaction` callback inside trade - not sure how else to get the txHash inside the hook

src/components/TradeWidget/index.tsx
W3stside: it's just 2.. also does unstable batch do anything for state hooks spread across different hooks?

W3stside: done! worked like a charm

anxolin:
> it's exposed as I needed access to it via the onSubmitTransaction callback inside trade - not sure how else to get the txHash inside the hook

Well the hook exposes `placeOrder`, this means, that it can abstract any user of the hook of the part of notifying when the transaction was sent. This way you don't need to do it in every place you use the hook. The hook would be more selfcontained.

W3stside:
happy to make this change in the future as you're right. As of now it works and my change was relatively minor. Let's circle back on this yeah?

PR #566, opened 2020-02-18
Tooltips sample integration

Velenir:
Adds

- Hook for Popper.js
- Hook with reasonable defaults ready to plug-n-play
- TooltipComponent for easy plug-n-play tooltips
- Sample tooltips on `/trade` page (style pending)

localhost_8080_trade_DAI-USDC_sell=0 buy=0 expires=2880

Tooltips don't have to look like that or like anything really, can be whole components inside

### DIFF
--- a/src/components/Portal.tsx
+++ b/src/components/Portal.tsx
@@ -1,0 +1,12 @@
+import React from 'react'
+import { createPortal } from 'react-dom'
+
+// mounts children in the body
+// which isolates them from the overall App styles
+// imported for inherited styles
+// usual use cases -- tooltips, popups, modals
+const Portal: React.FC = ({ children }) => {
+  return createPortal(children, document.body)
+}
+
+export default Portal
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -109,7 +109,10 @@ WrapperPropsAll<C> & { children?: ReactNode }): React.ReactElement => {
   // can attach ref to element
   // if children is a single element, not just text, and `as` tag|Component not specified
   if (chilrenNumber === 1 && typeof as === 'undefined' && isElement(children) && !isFragment(children)) {
-    let finalTargetProps
+    let finalTargetProps:
+      | typeof targetProps
+      | Omit<typeof targetProps, 'onMouseEnter' | 'onMouseLeave'>
+      | Omit<typeof targetProps, 'onFocus' | 'onBlur'>
     if (focus && hover) {
       // pass all targetProps by default
       finalTargetProps = targetProps
@@ -118,6 +121,7 @@ WrapperPropsAll<C> & { children?: ReactNode }): React.ReactElement => {
       finalTargetProps = {
         onFocus: targetProps.onFocus,
         onBlur: targetProps.onBlur,
+        ref: targetProps.ref,
       }
     } else if (hover) {
       // <TooltipWrapper focus={false}> --> only show tooltip on hover
@@ -126,6 +130,7 @@ WrapperPropsAll<C> & { children?: ReactNode }): React.ReactElement => {
       finalTargetProps = {
         onMouseEnter: targetProps.onMouseEnter,
         onMouseLeave: targetProps.onMouseLeave,
+        ref: targetProps.ref,
       }
     }
     const TargetComponent = React.cloneElement(children, finalTargetProps)
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -3,6 +3,7 @@ import Portal from './Portal'
 import { usePopperDefault, TOOLTIP_OFFSET } from 'hooks/usePopper'
 import { State, Placement } from '@popperjs/core'
 import styled from 'styled-components'
+import { isElement, isFragment } from 'react-is'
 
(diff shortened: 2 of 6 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr566--dexreact.review.gnosisdev.com

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -305,36 +307,42 @@ const TokenRow: React.FC<Props> = ({
       <div>
         <strong>{selectLabel}</strong>
         <span>
-          <button>+ Deposit</button>
+          <TooltipWrapper as="button" tooltip="Deposit">
+            + Deposit
+          </TooltipWrapper>
+          {/* <button>+ Deposit</button> */}
```
alfetopito: dead code?

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { State, Placement } from '@popperjs/core'
+import styled from 'styled-components'
+import { isElement, isFragment } from 'react-is'
+
+// visibility necessary for correct boundingRect calculation by popper
+const TooltipOuter = styled.div<Pick<TooltipBaseProps, 'isShown'>>`
+  visibility: ${(props): 'hidden' | false => !props.isShown && 'hidden'};
+`
+// can style anything but TOOLTIP_OFFSET fields, position and transform: rotate
+const TooltipArrow = styled.div`
```
W3stside: why all this styling? I thought the library handled styling more gracefully via props? or am i wrong

src/components/Tooltip.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const Wrapper: React.FC<WrapperProps> = ({ children, tooltip, placement, as }) => {
+  const { targetProps, tooltipProps } = usePopperDefault<HTMLDivElement>(placement)
+
+  const chilrenNumber = React.Children.count(children)
+
+  // can attach ref to element
+  // if children is a single element, not just text, and `as` tag|Component not specified
+  if (chilrenNumber === 1 && typeof as === 'undefined' && isElement(children) && !isFragment(children)) {
```
W3stside: what is this for? seems like crazy amount of adding of code

W3stside:
not saying it's a bad thing, but there's a lot seemingly going on in this PR that seems to be adding to the library. The hook looks very complicated - what was the reasoning behind adding it vs using the library directly?

src/components/TradeWidget/TokenRow.tsx
Velenir: Not dead code. Didn't delete as `<TooltipWrapper>` is just an example. Maybe we don't need a tooltip for that button

src/components/Portal.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { createPortal } from 'react-dom'
+
+// mounts children in the body
```
anxolin: Great that you added this comments

src/hooks/usePopper.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const defaultConfig: Partial<Options> & Pick<Options, 'modifiers'> = {
+  // default tooltip placement
+  placement: 'top',
+  modifiers: [
+    {
+      // move slightly to make space for the arrow element
+      name: 'offset',
+      options: {
+        offset: [0, 8],
```
anxolin: Is this 8 supposed to be the constant above?

src/components/Tooltip.tsx
Velenir: It handle styling via attributes, like vanilla JS would. And even then only positioning through transforms, not actual style styling

src/components/Tooltip.tsx
Velenir: To be able to clone Element later and pass it a `ref`


> not saying it's a bad thing, but there's a lot seemingly going on in this PR that seems to be adding to the library. The hook looks very complicated - what was the reasoning behind adding it vs using the library directly?

Library directly do anything with React.

src/hooks/usePopper.ts
```diff
@@ -0,0 +1,136 @@
...
+    fn: ({ state }) => {
+      // popper.state is mutable, need a new instance
+      setState({ ...state })
+    },
+  })
+
+  return finalConfig
+}
+
+interface PopperHookResult<T extends HTMLElement, U extends HTMLElement = HTMLDivElement> {
```
anxolin: Do we have a convention  for this interfaces? 
Personally I think it's fine to call it just Result, specially if you don't export it

Velenir:
> not saying it's a bad thing, but there's a lot seemingly going on in this PR that seems to be adding to the library. The hook looks very complicated - what was the reasoning behind adding it vs using the library directly?

Library directly doesn't do anything with React.

src/hooks/usePopper.ts
Velenir: Forgot to inckude, it's `TOOLTIP_OFFSET`

src/hooks/usePopper.ts
Velenir: Renamed

PR #571, opened 2020-02-19
483/liquidity orders tab

### LINKED ISSUE #483: Separate liquidity orders from normal orders
Visually separate normal orders and liquidity (unlimited) orders
In a different Tab or somewhere else

**Estimation: 1.5**

alfetopito:
Closes #483 

 - Using a state for tracking all orders per tab, splitting orders/markedForDeletion
 - Showing amount of orders selected to be deleted
 - Every tab tracks independently which orders were selected
 - Order deletion now works again

Screenshot from 2020-02-19 15-09-58

### Notes: 
- Did not do any styling at all. Would very much like to have the tabs like Michel's design (https://projects.invisionapp.com/share/5XVR9OE6FPJ#/screens/403151253_GnosisProtocol_-_V2-2_-_1_Order)
- ~~Oh my, it's soooo slow. What am I doing wrong?~~ It seems only local is painfully slow. Deployed version is much better, although a bit slow with over 100 orders.
- Still don't know why you need to first switch to a different tab upon refresh to load orders. Related to the `networkId == 0` issue maybe?

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -31,8 +31,7 @@ interface ShowOrdersButtonProps {
 
 const ShowOrdersButton: React.FC<ShowOrdersButtonProps> = ({ type, isActive, count, onClick }) => (
   <button className={type} disabled={isActive} onClick={onClick}>
-    {!isActive ? <Highlight>{count}</Highlight> : <>{count}</>}
-    <> {type}</>
+    {!isActive ? <Highlight>{count}</Highlight> : { count }} {type}
   </button>
 )
 
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -3,6 +3,7 @@ import React, { useMemo, useCallback, useEffect } from 'react'
 import { unstable_batchedUpdates } from 'react-dom'
 import { faTrashAlt } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { isOrderUnlimited } from '@gnosis.pm/dex-js'
 
 import { useOrders } from 'hooks/useOrders'
 import useSafeState from 'hooks/useSafeState'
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr571--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -19,23 +20,32 @@ import Highlight from 'components/Highlight'
...
+  count: number
+  onClick: (event: React.SyntheticEvent<HTMLButtonElement | HTMLFormElement>) => void
 }
 
-const ShowOrdersButton: React.FC<ShowOrdersButtonProps> = ({ type, isActive, shownCount, hiddenCount, onClick }) => {
-  const count = isActive ? shownCount : hiddenCount
+const ShowOrdersButton: React.FC<ShowOrdersButtonProps> = ({ type, isActive, count, onClick }) => (
+  <button className={type} disabled={isActive} onClick={onClick}>
+    {!isActive ? <Highlight>{count}</Highlight> : <>{count}</>}
+    <> {type}</>
```
Velenir: Why like that?
```jsx
{!isActive ? <Highlight>{count}</Highlight> : count}
     {type}
```

alfetopito: Dunno, don't remember doing that. Will fix.

src/components/OrdersWidget/index.tsx
```diff
@@ -44,49 +54,104 @@ const OrdersWidget: React.FC = () => {
...
+  // Update filteredOrders state whenever there's a change to allOrders
+  // splitting orders into respective tabs
   useEffect(() => {
     const now = new Date()
-    const filtered = allOrders.filter(order => showActive === isOrderActive(order, now))
-    setOrders(filtered)
-  }, [allOrders, setOrders, showActive])
 
-  const shownOrdersCount = orders.length
+    const filteredOrders = emptyState()
```
anxolin: Leave it like this. I think it's good. But an alternative would be to use `reduce` starting on an empty state

alfetopito:
> Kinda a lot of code for just keeping checked/unchecked orders across tabs. But if it works, it works
> +1

Indeed.
One improvement I have in mind (for another task) is to create a hook for the filtered orders, and on top of that store it on global state as user's might expect their selections to be kept, maybe?

Both might be overkill though. Thoughts?

gnosis-info:
Travis automatic deployment:
 https://pr571--dexreact.review.gnosisdev.com

PR #574, opened 2020-02-20
Fix stale networkId: 0 issue

Velenir:
Added logic for discarding Promises with stale (as in there's been new since init) WalletInfo

### DIFF
--- a/src/api/wallet/WalletApi.ts
+++ b/src/api/wallet/WalletApi.ts
@@ -211,6 +211,7 @@ export class WalletApiImpl implements WalletApi {
   }
 
   public async connect(givenProvider?: Provider): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr574--dexreact.review.gnosisdev.com

Velenir:
> - maybe the wallet should be in the global state
it already kinda is, its on the provider, which is global

> - Only one "get wallet" is required on load, we are doing multiple with at least 2 providers
but wallet can change, and wallet state does change on account change and other
> - I don't think useWalletConnection need to init the state by doing a call to the API. addOnChangeWalletInfo with trigger=true can do that
But if we don't request a new fresh WalletInfo, we can't be sure until new block if it maybe changed

Anyway happy to revisit all the points later

PR #577, opened 2020-02-20
smart disclaimer - persist close

### LINKED ISSUE #535: Allow to close the disclamer
Persist the closing of the disclaimer

### Estimation:
*0.5*

W3stside:
Closes #535 

As title states - very simple implementation

I think in the future, with more persisted state, we should consider using [`localForage`](https://github.com/localForage/localForage) which allows for more than just strings

Using JSON parse/stringify is a pain

### DIFF
--- a/src/components/LegalBanner.tsx
+++ b/src/components/LegalBanner.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import styled from 'styled-components'
 import { LEGALDOCUMENT } from 'const'
 
@@ -33,8 +33,18 @@ const Text = styled.p`
   }
 `
 
+const BannerCloser = styled.span`
+  &:before {
+    content: 'X';
+  }
+  position: absolute;
+  top: 1rem;
+  right: 1rem;
+
+  cursor: pointer;
+`
+
 interface LegalBannerProps {
-  startOpen: boolean
   title: string
   useFull: boolean
 }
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr577--dexreact.review.gnosisdev.com

alfetopito:
Dumb question: I don't see the banner. Is it because it's not part of the new design yet?

src/components/LegalBanner.tsx
```diff
@@ -72,17 +82,24 @@ const LegalText: React.FC = () => (
   </Text>
 )
 
-const LegalBanner: React.FC<LegalBannerProps> = ({ startOpen, useFull, title }) => {
-  const [open, setOpen] = useState(startOpen)
+const DISCLAIMER = 'GP_DISCLAIMER'
+
+const LegalBanner: React.FC<LegalBannerProps> = ({ useFull, title }) => {
+  const [open, setOpen] = useState(JSON.parse(localStorage.getItem(DISCLAIMER) as string))
```
Velenir: It starts closed, as `localStorage.getItem(DISCLAIMER) === null` when nothing is yet in localStorage

W3stside: true
good point
oops

W3stside: i'll rock with strings

W3stside:
thanks @Velenir good catch

src/components/LegalBanner.tsx
```diff
@@ -72,17 +82,24 @@ const LegalText: React.FC = () => (
   </Text>
 )
 
-const LegalBanner: React.FC<LegalBannerProps> = ({ startOpen, useFull, title }) => {
-  const [open, setOpen] = useState(startOpen)
+const DISCLAIMER = 'GP_DISCLAIMER'
```
anxolin: I think `DISCLAIMER` is too general. 

nit: Maybe `IS_CLOSED_PARAM` or sth like that.

src/components/LegalBanner.tsx
```diff
@@ -72,17 +82,24 @@ const LegalText: React.FC = () => (
...
+
+const LegalBanner: React.FC<LegalBannerProps> = ({ useFull, title }) => {
+  const [open, setOpen] = useState(localStorage.getItem(DISCLAIMER) || 'OPEN')
+
+  useEffect(() => {
+    open === 'OPEN' ? localStorage.setItem(DISCLAIMER, 'OPEN') : localStorage.setItem(DISCLAIMER, 'CLOSED')
```
anxolin: Being two the values, why not a boolean as a string?

```js
localStorage.setItem(IS_CLOSED_PARAM, 'true')
...
const [isClosed, setIsClosed] = useState(localStorage.getItem(IS_CLOSED_PARAM) === 'true')
...

```

src/components/LegalBanner.tsx
W3stside: it's the key name

src/components/LegalBanner.tsx
W3stside: good rec, will work it in

W3stside: actually there are issues with the null type and the way it's stringified so for simplicities sake and the fact its working now I'll leave it as is :D if it's super annoying we can change it later

anxolin: But `localStorage.getItem(IS_CLOSED_PARAM) === 'true'` 

As I understand, `localStorage.getItem(IS_CLOSED_PARAM)` can be:
* `null`: `null == 'true'`   ---> `false` (no closed)
* `'true'`: `'true' == 'true'`   ---> `true` (closed)

src/components/LegalBanner.tsx
anxolin: That's my point. `DISCLAIMER` var should make a reference of what it is. I think it should have then the prefix/suffix `_KEY`

PR #578, opened 2020-02-20
Tooltips for form input + extra

Velenir:
Simplifies FormMessage component
Enables to use tooltips with <inputs> already hooked to form-hook


gnosis-info:
Travis automatic deployment:
 https://pr578--dexreact.review.gnosisdev.com

PR #584, opened 2020-02-21
528/show from date on orders

### LINKED ISSUE #528: Show "from date" in orders page
If the order is not tradable yet, we should show it

Similar to how we do with the telegram bots

### Estimation:
*1*

Related to https://github.com/gnosis/dex-react/issues/527

alfetopito:
Closes #528

Went a bit over the task and also implemented all statuses seen in https://projects.invisionapp.com/share/5XVR9OE6FPJ#/screens/403151253_GnosisProtocol_-_V2-2_-_1_Order:

- Pending (well, pending...)
- Scheduled (`validFrom` in the future)
- Active (`validUntil` in the future AND (liquidity order OR matched amount == 0))
- Partial fill (NOT liquidity AND matched amount > 0)
- Filled (matched amount >= 99% order amount)
- Expired (`validUntil` in the past)
- Low balance warning displayed for *Active* and *Partial fill* orders

Screenshot from 2020-02-20 17-45-16
Screenshot from 2020-02-20 17-45-01
Screenshot from 2020-02-20 17-44-48

### Notes:
- Trade form currently not working. ~~Guess something from last merge. Will check and submit a new PR for that.~~ `data` value on `onSubmit` is not being filled. I'll leave this up to https://github.com/gnosis/dex-react/pull/551 to address.

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -150,32 +150,6 @@ const Amounts: React.FC<AmountsProps> = ({ sellToken, order, isUnlimited }) => {
   )
 }
 
-//TODO: no longer needed? remove
-// interface AccountBalanceProps extends Pick<Props, 'order' | 'isOverBalance'> {
-//   sellToken: TokenDetails
-//   isUnlimited: boolean
-// }
-
-// const AccountBalance: React.FC<AccountBalanceProps> = ({ sellToken, order, isOverBalance, isUnlimited }) => {
-//   const accountBalance = useMemo(() => formatAmount(order.sellTokenBalance, sellToken.decimals) || '0', [
-//     order.sellTokenBalance,
-//     sellToken.decimals,
-//   ])
-//   const isActive = isOrderActive(order, new Date())
-
-//   return (
-//     <td data-label="Account Balance" className="sub-columns three-columns">
-//       <div>{accountBalance}</div>
-//       <strong>{displayTokenSymbolOrLink(sellToken)}</strong>
-//       {isOverBalance && isActive && !isUnlimited && (
-//         <div className="warning">
-//           <FontAwesomeIcon icon={faExclamationTriangle} />
-//         </div>
-//       )}
-//     </td>
-//   )
-// }
-
 const Expires: React.FC<Pick<Props, 'order' | 'pending'>> = ({ order }) => {
   const { isNeverExpires, expiresOn } = useMemo(() => {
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr584--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -150,32 +158,6 @@ const Amounts: React.FC<AmountsProps> = ({ sellToken, order, isUnlimited }) => {
   )
 }
 
-//TODO: no longer needed? remove
```
anxolin: good call

src/utils/miscellaneous.ts
```diff
@@ -51,3 +52,9 @@ export function getImageUrl(tokenAddress?: string): string | undefined {
 }
 
 export const isOrderActive = (order: AuctionElement, now: Date): boolean => batchIdToDate(order.validUntil) >= now
+
+export function isOrderFilled(order: AuctionElement): boolean {
+  // consider an oder filled when 99% or more is matched
```
anxolin: why 99%?

Can we make it a bit bigger? Like with that. If we sell 1M DAI we can have 10K DAI still pending.
Maybe I would sth like `99.99999999999%` :) 

Or an alternative is sth in the direction of: `99.99%` and remaining amount `< 0.01` whatever is more restrictive

alfetopito: > why 99%?
So I got feedback on what a better value would be :)

Will do 0.01%

alfetopito: Updated. Oh, just noticed you also mentioned a fixed value. Updating...

alfetopito: Actually, that makes it much more complicated. I need the token decimals to calculate what's the amount left, and that's not part of the order object.
Can we stick to just the 0.01% for now?

anxolin: Although I would make it even smaller `0.0001%`
`0.0001%` of `100 Ether` --> 0.0001 ETH --> `0.026612 $` at current price

Because the ` 0.01%` would be:
`0.01%` of `100 Ether` --> 0.01 ETH --> `2.6612 $` at current price

alfetopito: I could argue that relatively speaking, $ 2.6612 out of $ 26,612.00 is negligible, but I won't :D

Instead I'll lower down from 0.01% to 0.0001%

PR #586, opened 2020-02-21
Balances overview styles update.

biocom:
Styled the balances overview:
- Supports longer digits by wrapping inside the cell.
- For now on mobile the table uses overflow-x scrolling. Seems the most convenient approach for now:
Screenshot 2020-02-21 at 17 38 54
Screenshot 2020-02-21 at 17 56 54Screenshot 2020-02-21 at 17 57 00

**TODO after merging:**
- Make the inline search input field functional.
- Hide zero balances -> filter the overview.
- Manage Tokens > Should open a modal. Perhaps we should call it 'Add Token' instead and simply call the TokenSelector. Only show tokens on that list, which aren't added yet?

--- 
**Also included:**
- The trade view now doesn't directly load in a wide layout. Only when necessary it will expand. I feel that keeps the overall aesthetique of the interface while supporting larger digits without wrapping/hidden overflow:
Screenshot 2020-02-21 at 17 53 42
Screenshot 2020-02-21 at 17 53 52
Screenshot 2020-02-21 at 17 55 08


gnosis-info:
Travis automatic deployment:
 https://pr586--dexreact.review.gnosisdev.com

biocom:
Brought up to speed with the latest `develop` branch / resolved a conflict.

biocom:
> Everything looks great.
> 
> I'm just not a fan of the scrolling on mobile. My personal preference is the card style David created, that consolidates all info for a given token in the screen.
> 
> Still, approving, I'm not the holder of UI decisions and I trust you know better :)

Do you have an example of that view? I agree that it can be optimised. I think on a 2nd iteration I'll try to have it wrap (which I think is what you refer to) so you have longer cards.

alfetopito:
> > Everything looks great.
> > I'm just not a fan of the scrolling on mobile. My personal preference is the card style David created, that consolidates all info for a given token in the screen.
> > Still, approving, I'm not the holder of UI decisions and I trust you know better :)
> 
> Do you have an example of that view? I agree that it can be optimised. I think on a 2nd iteration I'll try to have it wrap (which I think is what you refer to) so you have longer cards.

You can see that in the current prod version https://dex.gnosis.io/wallet

Screenshot from 2020-02-21 20-03-57

biocom:
@alfetopito Pushed an update for an optimised mobile experience:
Screenshot 2020-02-24 at 11 34 54
Screenshot 2020-02-24 at 11 35 04

PR #590, opened 2020-02-21
588/wallet not connected orders page

### LINKED ISSUE #588: Improve message when we load the orders
When we load orders we show, instead of a spinner this messge:

image

Additionally, we should handle specifically the case of an unconnected wallet and show a specific message for that:

image

Probably the solution would be just to show this message:

image

alfetopito:
Closes #588 

Did not touch the styling, simply re-used the same component as in ConnectWallet page.

### DIFF
--- a/src/components/ConnectWalletBanner.tsx
+++ b/src/components/ConnectWalletBanner.tsx
@@ -1,0 +1,56 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import iconWallet from 'assets/img/wallet.svg'
+import { MEDIA } from 'const'
+
+const Wrapper = styled.div`
+  background: #ffffff;
+  box-shadow: 0 -1rem 4rem 0 rgba(0, 0, 0, 0.05), rgba(0, 0, 0, 0.02) 0 0.276726rem 0.221381rem 0,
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr590--dexreact.review.gnosisdev.com

biocom:
Thank you Leandro.

Right now because of the way I restructured the use of the Pagewrapper styles, the widgets themselves have this white background/box shadow attached.

This causes it to be shown in particular on the orders panel:
Screenshot 2020-02-21 at 21 09 09

I am aware, that the original setup of using a single Pagewrapper component, is in principal the correct approach to prevent situations like this (box-shadow styles on child widgets).

I would say that for now it's fine and I'd opt for sorting this in a separate PR.

alfetopito:
> Great!
> 
> Only one question. Since now we have it in the widget page, maybe we need to re-think about the use case where the user:
> 
>     1. Is not connected. Fills the trade form
> 
>     2. Hit submit --> navigate to connect wallet
> 
> 
> Maybe this is fine for now, although we could see how we can improve the workflow. Maybe even opening automatically the "connect wallet" modal or something?

Hmmm yes, makes sense. Will do that in another PR.

PR #598, opened 2020-02-24
Triggering wallet connection from trade page on submission

alfetopito:
Follow up to #588 

When user is logged and tries to submit an order, the wallet dropdown is shown over the trade page.

### Notes
- ~~There's a bug on `onSubmit` where the form data is not being passed down onto `data` object. To test I manually set the values. I guess this will be fixed by https://github.com/gnosis/dex-react/pull/551 so I didn't spend time trying to fix it. For this reason, you'll have to trust me it works or test locally.~~
- ~~It seems like the wallet widget is re-rendered every second, guess due to the batch time remaining thingy. Still have to check that.~~ Yep, that's the case.
- ~~For a reason that I yet haven't figured out, the wallet dropdown is expanded after logging in.~~

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -488,6 +488,8 @@ const TradeWidget: React.FC = () => {
       if (validFrom === 0) {
         // ; for destructure reassign format
         ;({ success } = await placeOrder({
+          networkId,
+          userAddress,
           buyAmount,
           buyToken,
           sellAmount,
@@ -510,6 +512,8 @@ const TradeWidget: React.FC = () => {
       } else {
         // ; for destructure reassign format
         ;({ success } = await placeMultipleOrders({
+          networkId,
+          userAddress,
           orders: [
             {
               buyAmount,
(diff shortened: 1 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr598--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -461,8 +461,8 @@ const TradeWidget: React.FC = () => {
         dispatch(removePendingOrdersAction({ networkId, pendingTxHash, userAddress }))
       }
     } else {
-      const from = history.location.pathname + history.location.search
-      history.push('/connect-wallet', { from })
+      // Not connected. Prompt user to connect his wallet
+      await connectWallet()
```
Velenir: So it only asks to connect, but doesn't submit the transaction after connection
Kinda not intuitive

W3stside: this is where redux middleware would be very useful

Velenir: can just send the tx if connection is successful

alfetopito: Would you rather do that after connecting?

Velenir:
```js
if (!isConnected) //send tx
else {
  // connect
  if (success) // send tx
}
```

would be my preference
@anxolin , @W3stside, @biocom  ?

W3stside: yes i think it makes the most sense :100:

PR #597, opened 2020-02-24
Improve tokenSelector styles.

biocom:
Improved the styles for the tokenSelector:
Screenshot 2020-02-24 at 13 47 46
Screenshot 2020-02-24 at 13 45 38

I need assistance with adding some extra elements around the token selector, as per design:
Screenshot 2020-02-24 at 13 50 07

- Show balance for each token
- Search input field
- Title + X close button

PR #599, opened 2020-02-24
Display balance on trade page token selector

alfetopito:
Addressing point: **Show balance for each token**
From https://github.com/gnosis/dex-react/pull/597

### Update: March 2nd

Finally did it, managed to align the balance

YES!

Screenshot from 2020-03-02 14-18-52

### DIFF
--- a/src/components/TokenSelector.tsx
+++ b/src/components/TokenSelector.tsx
@@ -34,10 +34,19 @@ const Wrapper = styled.div`
       margin: 0;
     }
 
-    > div {
+    .tokenDetails {
       display: flex;
-      flex-direction: column;
-      margin-left: 1rem;
+      justify-content: space-between;
+
+      .tokenName {
+        display: flex;
+        flex-direction: column;
+        margin-left: 1rem;
+      }
+
+      .tokenBalance {
+        font-weight: bold;
+      }
     }
 
     > div > div {
@@ -110,12 +119,16 @@ function renderOptionLabel(token: TokenDetails | TokenBalanceDetails): React.Rea
   return (
     <div className="optionItem">
       <TokenImgWrapper src={token.image} alt={token.name} />
-      <div>
-        <div>
-          <strong>{token.symbol}</strong>
+      <div className="tokenDetails">
+        <div className="tokenName">
+          <div>
+            <strong>{token.symbol}</strong>
+          </div>
+          <div>{token.name}</div>
         </div>
-        <div>{token.name}</div>
-        {'totalExchangeBalance' in token && <div>{formatAmount(token.totalExchangeBalance, token.decimals)}</div>}
+        {'totalExchangeBalance' in token && (
+          <div className="tokenBalance">{formatAmount(token.totalExchangeBalance, token.decimals)}</div>
+        )}
       </div>
     </div>
   )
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr599--dexreact.review.gnosisdev.com

src/components/TokenSelector.tsx
```diff
@@ -88,7 +90,7 @@ const SelectedTokenWrapper = styled.span`
   }
 `
 
-function renderOptionLabel(token: TokenDetails): React.ReactNode {
+function renderOptionLabel(token: TokenDetails | TokenBalanceDetails): React.ReactNode {
```
Velenir: I don't particularly like how we don't know at this point if token is one type or another
Maybe we should always deal with `TokenBalanceDetails`
The problem stems from `TradeWidget/index.tsx::L367`

biocom:
Just checked it out, this is how it renders for me:
Screenshot 2020-02-25 at 14 39 28
 
Is it possible to have the balances right align? That would require the balances and the Token img/name/symbol to be in its own elements and share 1 direct parent. So we can use display: flex.

alfetopito:
> Just checked it out, this is how it renders for me:
> Screenshot 2020-02-25 at 14 39 28
> 
> Is it possible to have the balances right align? That would require the balances and the Token img/name/symbol to be in its own elements and share 1 direct parent. So we can use display: flex.

Ah yes, I just added it to the list, without caring for styling. I assumed you would style it later anyway, so I didn't bother.
I can make it right aligned, though.

src/components/TokenSelector.tsx
alfetopito: Yes, I thought about using throughout `TokenBalanceDetails`, but this type requires things related to user balance, which are not available when logged out.
Since we do want to display the token selector even when not logged in, I opted for the current version.

src/components/TradeWidget/index.tsx
```diff
@@ -361,7 +361,13 @@ const TradeWidget: React.FC = () => {
   // Avoid displaying an empty list of tokens when the wallet is not connected
   const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
 
-  const tokens = useMemo(() => tokenListApi.getTokens(fallBackNetworkId), [fallBackNetworkId])
+  const { balances } = useTokenBalances()
+
+  // If user is connected, use balances, otherwise get the default list
+  const tokens = useMemo(
+    () => (isConnected && balances.length > 0 ? balances : tokenListApi.getTokens(fallBackNetworkId)),
```
anxolin: I find strange how we return either balances or the list of tokens
Wouldn't it be better to have it as two separate state properties, each one with it's own type, and then depending on `isConnected && balances.length` we render one or the other one?

biocom:
> Ah yes, I just added it to the list, without caring for styling. I assumed you would style it later anyway, so I didn't bother.
> I can make it right aligned, though.

Sure, I think the only thing I need there, is for those elements to have the right structure as mentioned. Otherwise would have to resort to 'position:absolute' to manually set the balance to the right. So feel free to do whatever and I can review styling after. 👍

src/components/TradeWidget/index.tsx
alfetopito: We could yes, and we had it before. I'm changing the behaviour because we'll only need the list of tokens when logged out.
I could argue that we are saving one API call to tokenListApi, although that's a local call.

Still, if you feel strongly about this I can revert.

PR #600, opened 2020-02-25
hotfix/0.8.4

Velenir:
Fixes:

- not being able to connect with Mobile wallets
- orders not being updated on initial fetch on Mobile wallets

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -56,7 +56,11 @@ export function useOrders(): Result {
     const fetchOrders = async (offset: number): Promise<void> => {
       // isLoading is the important one
       // controls ongoing fetching chain
-      if (!userAddress || !networkId || !isLoading) return
+      if (!userAddress || !networkId || !isLoading) {
+        // next isLoading = true will be when userAddress and networkId are valid
+        setIsLoading(false)
+        return
+      }
 
       // contract call
       try {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr600--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
```diff
@@ -50,13 +50,24 @@ export function useOrders(): Result {
...
   useEffect(() => {
     let cancelled = false
 
     const fetchOrders = async (offset: number): Promise<void> => {
       // isLoading is the important one
       // controls ongoing fetching chain
-      if (!userAddress || !networkId || !isLoading) return
+      if (!userAddress || !networkId || !isLoading) {
+        // next isLoading = true will be when userAddress and networkId are valid
+        setIsLoading(false)
```
alfetopito: I solved this a little differently https://github.com/gnosis/dex-react/commit/30ae106415f125232afbff3d915f277f99876f53
Regardless, to avoid a re-render, can you make this state change conditional?
Or does React take care of that already (by not re-rendering when the state is set to the same value)?

Velenir: yes,` setState(oldValue)` is noop

Velenir: Also for the complete fix this `setIsLoading` should come after the other one in `useEffect` 34b33cd

PR #608, opened 2020-02-26
527/set order start time

### LINKED ISSUE #527: Schedule an order: Set order start date
Similar to custom expiration date.

It should have the option to create orders ASAP or with a custom time in the future, in 5 min intervals.

- If ASAP --> we use the current method for placing orders
- If a date is selected --> We use the one with arrays (place orders from)
- 15 min is the minimum time in the future (to avoid issues with transaction mining)

### Estimation:
*3*

Related to https://github.com/gnosis/dex-react/issues/528

W3stside:
Closes #527 

Allows user to set valid from date for their orders (simple implementation)

Sets from time in minutes and has an ASAP toggle (needs tooltip probz)


gnosis-info:
Travis automatic deployment:
 https://pr608--dexreact.review.gnosisdev.com

src/components/TradeWidget/OrderValidity.tsx
```diff
@@ -58,137 +82,211 @@ const Wrapper = styled.div`
...
+  const validFromInputValue = watch(validFromInputId)
+  const validUntilInputValue = watch(validUntilInputId)
+  const overMax = ZERO
+  const validFromClassName = validFromError ? 'error' : overMax.gt(ZERO) ? 'warning' : ''
+  const validUntilClassName = validUntilError ? 'error' : overMax.gt(ZERO) ? 'warning' : ''
+
+  // const validFromErrorComp = validFromError && <WalletDetail className="error">{validFromError.message}</WalletDetail>
+  // const validUntilErrorComp = validUntilError && (
+  //   <WalletDetail className="error">{validUntilError.message}</WalletDetail>
+  // )
```
alfetopito: Dead code?

src/components/TradeWidget/OrderValidity.tsx
```diff
@@ -58,137 +82,211 @@ const Wrapper = styled.div`
...
-  // function handleUnlimitedClick(): void {
-  //   setUnlimited(!isUnlimited)
-  //   !isUnlimited ? setValue(inputId, '', true) : setValue(inputId, '30', true)
-  // }
+  useEffect(() => {
+    handleValidUntilChange()
+  }, [handleValidUntilChange])
+
+  function handleUnlimitedClick(): void {
+    setUnlimited(!isUnlimited)
```
alfetopito: Shouldn't it be `setUnlimited(unlimited => !unlimited)` ? same for `setAsap`

src/components/TradeWidget/index.tsx
```diff
@@ -414,41 +426,88 @@ const TradeWidget: React.FC = () => {
...
+                priceNumerator: buyAmount,
+                priceDenominator: sellAmount,
+                user: userAddress,
+                remainingAmount: ZERO,
+                sellTokenBalance: ZERO,
+                validFrom: 0,
+                validUntil: 0,
+              }
+
+              return dispatch(savePendingOrdersAction({ orders: newTxState, networkId, userAddress }))
```
alfetopito: Could this be moved into a function for reuse?

src/components/TradeWidget/index.tsx
```diff
@@ -525,7 +587,7 @@ const TradeWidget: React.FC = () => {
       </FormContext>
       <OrdersPanel>
         {/* Toggle panel visibility (arrow) */}
-        <button onClick={(): void => setOrdersVisible(!ordersVisible)} />
+        <button onClick={(e): void => (e.preventDefault(), setOrdersVisible(!ordersVisible))} />
```
alfetopito: Do @Velenir 's `type='button'` magic and forget about `e.preventDefault()`

src/hooks/usePlaceOrder.ts
```diff
@@ -158,9 +163,9 @@ export const usePlaceOrder = (): Result => {
           sellTokens.push(order.sellToken)
 
           // if not set, order is valid from placement + wait period
-          validFroms.push(order.validFrom ?? currentBatchId + BATCHES_TO_WAIT)
+          validFroms.push(order.validFrom ? currentBatchId + order.validFrom : currentBatchId + BATCHES_TO_WAIT)
```
alfetopito:
```suggestion
          validFroms.push(currentBatchId + order.validFrom ? order.validFrom : BATCHES_TO_WAIT)
```

src/components/TradeWidget/OrderValidity.tsx
```diff
@@ -58,137 +82,211 @@ const Wrapper = styled.div`
...
-//   &.warning {
-//     color: orange;
-//   }
-// `
-
-// function validatePositive(value: string): true | string {
-//   return Number(value) == 0 || Number(value) >= 5 || 'Invalid expiration time'
-// }
+const OrderValidityInputsWrapper = styled.div<{ $visible: boolean }>`
+  opacity: ${({ $visible }): string => ($visible ? '1' : '0')};
```
Velenir: If you want to hide completely while preserving layout use `visibility: hidden`

src/hooks/usePlaceOrder.ts
Velenir: wait, wait, these aren't equivalent
```suggestion
          validFroms.push(currentBatchId + (order.validFrom ? order.validFrom : BATCHES_TO_WAIT))

```

src/components/TradeWidget/OrderValidity.tsx
W3stside: i think both are fine but ill go with this for safety @Velenir

src/components/TradeWidget/OrderValidity.tsx
Velenir: Also `opacity: 0` doesn't prevent clicks, so you can't click everything behind this fully opaque element (can't click Token selector)

Velenir: Why do you render that modal when not in use for that matter?

src/utils/format.ts
```diff
@@ -56,17 +56,17 @@ export function sanitizeNegativeAndMakeMultipleOf(value?: string | null, default
...
 }
 export const validInputPattern = new RegExp(/^\d+\.?\d*$/) // allows leading and trailing zeros
 export const leadingAndTrailingZeros = new RegExp(/(^0*(?=\d)|\.0*$)/, 'g') // removes leading zeros and trailing '.' followed by zeros
 export const trailingZerosAfterDot = new RegExp(/(.*\.\d+?)0*$/) // selects valid input without leading zeros after '.'
 
-export const formatValidity = (validTime: string | number): string =>
-  +validTime == 0
-    ? 'Unlimited'
-    : +validTime < 0
+export const formatValidity = (validTime: string | number, zeroValueText: string, constraint = 0): string =>
```
anxolin: ubernit: shouldn't you be coherent with the naming `constraint` and `zeroValueText `. 
U understand you created `constraint` to allow to use something different than `zero`, but then the label is has `zero` in it.

Also `validTime < constraint` I think there you mean `validTime < 0` for several reasons, one being that you want to be coherent with the message `Invalid time - time cannot be negative`. I think that's unrelated to the `constraint`

Then, `constraint` I don't see you use that anywere. So I think that is using always the default. So for now I would just remove it, otherwise it only makes the code more complicated.

src/components/TradeWidget/OrderValidity.tsx
```diff
@@ -58,137 +82,211 @@ const Wrapper = styled.div`
...
-          />
-          <div className="radio-container">
-            <input type="checkbox" disabled={isDisabled} defaultChecked={isUnlimited} onClick={handleUnlimitedClick} />
-            <small>Never</small>
-          </div>
-        </div>
-        {errorOrWarning}
-      </InputBox> */}
+
+      <OrderValidityInputsWrapper $visible={showOrderConfig}>
```
Velenir: Would be better to conditionally render here instead of switching visibility
If it breaks form-hooks somehow, try putting the whole thing + useFormContext into its own component and conditionally rendering that

src/utils/format.ts
anxolin: I think you are making this `formatValidity` more general. So you cannot call it `formatValidity` any more.

I'm not good with names, maybe `formatTimeMinutes`

src/components/TradeWidget/OrderValidity.tsx
W3stside: no i dont, i like the fade effect

src/components/TradeWidget/OrderValidity.tsx
W3stside: why would it break hooks?

src/components/TradeWidget/OrderValidity.tsx
Velenir: Both are fine here
function setState is used when we either don't have access to current state or to account for multiclicks or similar. As in
`setState(false)  x 10 times` will end up false, `setState(flip => !flip)  x 10 times` will end up the opposite

src/hooks/usePlaceOrder.ts
```diff
@@ -107,7 +107,12 @@ export const usePlaceOrder = (): Result => {
           //  In reality, no need to ceil cause we know batch time is 300, but it was done to avoid relying on knowing
           //  the actual value of the constant
           const validityInMinutes = Math.ceil((validUntil * BATCH_TIME) / 60)
-          toast.success(`Transaction mined! Succesfully placed order valid for ${formatValidity(validityInMinutes)}`)
+          toast.success(
+            `Transaction mined! Succesfully placed order valid for ${formatValidity(
```
anxolin: You might want to add when the order will become tradable if it's not ASAP.

src/components/TradeWidget/OrderValidity.tsx
Velenir: no idea
just thinking

W3stside: @anxolin @alfetopito thoughts on this? - I'm fine with it using the opactiy as its not a huge component but i get the sentiment from Dima

src/components/TradeWidget/OrderValidity.tsx
Velenir: You can't just use opacity
you can't click on what's behind right now
Try clicking USDC token selector

src/components/TradeWidget/OrderValidity.tsx
Velenir: It's ok to just hide, but you can't just use opacity
If you really want to, at least add `pointer-events: none`, otherwise your modal steals clicks

src/hooks/usePlaceOrder.ts
alfetopito: oops, pretend you didn't see that

src/components/TradeWidget/OrderValidity.tsx
W3stside: ah true, didn't think of that. ok I'll work it in

alfetopito: No opinion, so I'll just back @Velenir

anxolin: I'm ok with hidding too. What I'm confused about is that this part looks like is the implementation of a modal (I might be wrong), while we had a modal component already. So theoretically we didn't need to re-style this again and handle the visibility thing. It's handled in the library

I think this probably has a relation to the nit-issue reported here: https://github.com/gnosis/dex-react/issues/629

anxolin:
@W3stside do you think we can review/merge this tomorrow for the version?

PR #616, opened 2020-02-26
Fix form submission

Velenir:
`<button>` without explicit type is `type="submit"` by default
So any such `<button>` inside a form submits that form

Added explicit type to every button I found, let's do it explicitly from now on


gnosis-info:
Travis automatic deployment:
 https://pr616--dexreact.review.gnosisdev.com

PR #617, opened 2020-02-26
Hide functionalities not included in ethcc version

anxolin:
Just removes some functionalities we likely won't include in EthCC.

Created issues for all of them.

For doing it, i created a global `not-implemented` css style, so it's more declarative. The idea is later tackle those issues and remove that class.

Deposit and enable token:
image

Filter, hide 0 balances, and manage tokens:
image

Some broken links:
image


gnosis-info:
Travis automatic deployment:
 https://pr617--dexreact.review.gnosisdev.com

src/components/DepositWidget/index.tsx
```diff
@@ -346,14 +346,14 @@ const DepositWidget: React.FC = () => {
     <section>
       <BalancesWidget>
         <BalanceTools>
```
biocom: Actually @anxolin we could also use a single 'not-implemented' class for the whole BalanceTools element. So that it doesn't output at all.

anxolin: Yes, I was going to do it like that, but since they have independent issues. I thought it was better to do it one by one, so we can implement the issues independently with less conflicts

gnosis-info:
Travis automatic deployment:
 https://pr617--dexreact.review.gnosisdev.com

PR #632, opened 2020-02-27
Fix balance always on error state

anxolin:
Fix error message in the balance always showing up


gnosis-info:
Travis automatic deployment:
 https://pr632--dexreact.review.gnosisdev.com

PR #636, opened 2020-02-27
Create pending API and mocks

anxolin:
Adds a new `API` ---> `PendingAPI`

* This API has for now, ONLY getOrders. So it has the pending orders
* I created it for mocking the pending orders, this is why the REAL implementation returns an empty array
* In the future, we could use this API for persisting in the browser the pending orders, so it's refresh resistant

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -64,7 +64,8 @@ export function useOrders(): Result {
     const fetchOrders = async (offset: number): Promise<void> => {
       // isLoading is the important one
       // controls ongoing fetching chain
-      if (!userAddress || !networkId || !isLoading) {
+      if (!userAddress || !networkId /*|| !isLoading*/) {
+        // TODO: Do not merge with this, just for doing a quick test!!
         // next isLoading = true will be when userAddress and networkId are valid
         setIsLoading(false)
         return
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -94,12 +94,6 @@ export interface Order {
   remainingAmount: BN
 }
 
-export interface PendingTxObj extends AuctionElement {
-  txHash: string
-}
-
-export type PendingTxArray = PendingTxObj[]
-
 export interface GetOrdersPaginatedResult {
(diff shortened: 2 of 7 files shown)


src/hooks/useOrders.ts
```diff
@@ -64,7 +64,8 @@ export function useOrders(): Result {
     const fetchOrders = async (offset: number): Promise<void> => {
       // isLoading is the important one
       // controls ongoing fetching chain
-      if (!userAddress || !networkId || !isLoading) {
+      if (!userAddress || !networkId /*|| !isLoading*/) {
+        // TODO: Do not merge with this, just for doing a quick test!!
```
alfetopito: ?

anxolin: Yes! :P 

This is so this don't get merged until that issue is fixed.
I commented that so Michel can see the orders. The problem is that there was a race condicion loading the wallet, and the fix broke the mocks for the orders. 

This is a temporary fix. You can still review the other parts of this PR.

Velenir:
> * In the future, we could use this API for persisting in the browser the pending orders, so it's refresh resistant

But that would be just an extra abstraction over what we use for example for orders in general -- global state. We can just persist a slice of global state if we need to.
API should be used to interface with something external, like ExchangeApi is. PendingApi as it is is just another global state by a different name

Velenir:
Also, travis build broke

W3stside:
Yes what @Velenir said - we simply need to persist into localStorage via a hook. We can easily persist a slice, i.e pendingOrders in this case - in fact it should already be doing that

W3stside:
wait I don't understand why we need this PR, can someone expand?

anxolin:
Okay, I'll try to use global state instead then.
Let me see if I can re-shape in time for the version....

Velenir:
> Okay, I'll try to use global state instead then.
> Let me see if I can re-shape in time for the version....

It's ok to merge this if only for this version, revert later
Don't bother with global state on such short notice

src/hooks/useOrders.ts
alfetopito: @anxolin we still have that. Guess this should be reverted?

PR #637, opened 2020-02-27
Close form on Escape

### LINKED ISSUE #629: nit: Close modals using ESC
It's an app convention that when you have a modal open, and you press cancel, the modal closes

Doesn't happen at least in the Balances page:

image

Velenir:
Adds a hook to call callback on keydown
Uses it to close Deposit/Withdraw from on /wallet page

Modali already has that functionality, so no need there

Closes #629


gnosis-info:
Travis automatic deployment:
 https://pr637--dexreact.review.gnosisdev.com

alfetopito:
Why not use modali on these forms then?

Velenir:
> Why not use modali on these forms then?

Don't know. In the old design these Forms weren't Modals. So it changed somewhere in the transition.

PR #638, opened 2020-02-27
Autofocus sell input

Velenir:
Adds automatic focus on the sell amount input

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -270,7 +270,7 @@ const TokenRow: React.FC<Props> = ({
       </div>
       <InputBox>
         <InputWithTooltip
-          autoFocus={autoFocus}
+          autoFocus={!readOnly && autoFocus}
           className={className}
           tooltip={tooltipText}
           placeholder="0"
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -159,6 +159,7 @@ interface Props {
   tabIndex: number
   readOnly: boolean
   tooltipText: string
+  autoFocus?: boolean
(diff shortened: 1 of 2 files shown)


anxolin:
Nice Dima. I tried to use that, but I was writing it in lowercase (as in the `HTML` field), and I thought it was not supported. So I went for the other approach.

Strange. Why did travis not build for this PR?

anxolin:
Nice Dima. I tried to use that, but I was writing it in lowercase (as in the `HTML` field), and I thought it was not supported. So I went for the other approach.

Strange. Why did travis not build for this PR?

alfetopito:
> Strange. Why did travis not build for this PR?

Too many PRs!

anxolin:
Nice Dima. I tried to use that, but I was writing it in lowercase (as in the `HTML` field), and I thought it was not supported. So I went for the other approach.

Strange. Why did travis not build for this PR?

alfetopito:
> Strange. Why did travis not build for this PR?

Too many PRs!

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -268,6 +270,7 @@ const TokenRow: React.FC<Props> = ({
       </div>
       <InputBox>
         <InputWithTooltip
+          autoFocus={!readOnly && autoFocus}
```
anxolin: I did it as this first, but then I thought it would be more declarative with an optional field "focus". 

It's a bit strange at a purist level to "focus" in every "TokenRow" that's editable. We can do this because we have knowledge that there's ONLY two rows.

However, it doesn't look there's going to be more than two in the future.

gnosis-info:
Travis automatic deployment:
 https://pr638--dexreact.review.gnosisdev.com

PR #639, opened 2020-02-27
Join orders in single table and handle pending

anxolin:
* Unifies the pending and normal table --> it gets simplified too :) 
* Takes the count of pending orders into account. So when you send a transaction it increments already the count
* Handles the separation between liquidity/active/expired ALSO for pending orders

image


src/components/OrdersWidget/index.tsx
```diff
@@ -96,16 +102,26 @@ const OrdersWidget: React.FC = () => {
       }
     })
 
+    allPendingOrders.forEach(order => {
```
alfetopito: Feels like you could extract a function to split the orders.
Something like: 

```
function splitOrders(sourceOrders:AuctionElement[], ordersState: state, stateKey: string):void {
  ...
}
```

src/components/OrdersWidget/index.tsx
```diff
@@ -238,85 +248,57 @@ const OrdersWidget: React.FC = () => {
...
+                      <th>Limit price</th>
+                      <th className="filled">Filled / Total</th>
+                      <th>Expires</th>
+                      <th className="status">Status</th>
+                    </tr>
+                  </thead>
+                  <tbody>
+                    {displayedPendingOrders.map((order: PendingTxObj) => (
+                      <OrderRow
+                        key={Math.random()}
```
Velenir: Please, don't use random keys. OrderRow will remount each time.
Why not `order.id` or `order.txHash`. `order.id` here should not interfere with same key on L289, they are different arrays, if you are worried about that

anxolin: Haven't done this, it was there already. This PR was unification.

I didn't see it, I agree it's super bad practice. I'll address it in separate PR

W3stside: let's um, not use git blame here

W3stside: think u just discovered the flashing pending orders bug :joy: :joy: god daaamn

src/components/OrdersWidget/index.tsx
```diff
@@ -96,16 +102,26 @@ const OrdersWidget: React.FC = () => {
       }
     })
 
+    allPendingOrders.forEach(order => {
+      if (!isOrderActive(order, now)) {
+        filteredOrders.expired.pendingOrders.push(order)
```
W3stside: would `expiredOrders` ever have pending orders though?

src/components/OrdersWidget/index.tsx
```diff
@@ -238,85 +248,57 @@ const OrdersWidget: React.FC = () => {
...
-                        <OrderRow
-                          key={Math.random()}
-                          order={order}
-                          networkId={networkId}
-                          isOverBalance={false}
-                          pending
+            {ordersCount > 0 ? (
+              <div className="ordersContainer">
+                <CardTable
+                  // $columns="minmax(2rem, min-content) minmax(13.625rem, 1fr) repeat(2, minmax(6.2rem, 0.6fr)) minmax(5.5rem, 1fr)"
```
W3stside: dead code?

src/components/OrdersWidget/index.tsx
```diff
@@ -238,85 +248,57 @@ const OrdersWidget: React.FC = () => {
...
+                  $columns="minmax(2rem,.4fr)  minmax(11rem,1fr)  minmax(11rem,1.3fr)  minmax(5rem,.9fr)  minmax(auto,1.4fr)"
+                  // $cellSeparation="0 .5rem;"
```
W3stside: dead code?

src/components/OrdersWidget/index.tsx
alfetopito: I though the same, but since we are at it, why not? :D

src/components/OrdersWidget/index.tsx
anxolin: Gilt is shared among reviewers too :) 
Nice that Dima spotted it now!

src/components/OrdersWidget/index.tsx
W3stside: seems like UNOPTIMISED CODE

/nooooooooooooooooooooo

anxolin: I also thought the same :) 

It can happen, it's not likely, but if you say it expires in `15min` and has a bad gas price can happen

src/components/OrdersWidget/index.tsx
anxolin: Not from this PR, but I'll delete

src/components/OrdersWidget/index.tsx
anxolin: Not from this PR, but I'll delete

src/components/OrdersWidget/index.tsx
anxolin: https://github.com/gnosis/dex-react/issues/657

gnosis-info:
Travis automatic deployment:
 https://pr639--dexreact.review.gnosisdev.com

PR #641, opened 2020-02-27
Delete unused column

anxolin:
There was a hidden column in the orders that was not used :) 

Removed from the code

That's what it used to do:

image


gnosis-info:
Travis automatic deployment:
 https://pr641--dexreact.review.gnosisdev.com

PR #642, opened 2020-02-27
Do not disable the send button

### LINKED ISSUE #634: Price empty on trade page triggers no error
### Steps to reproduce:
1. On trade page, try to submit the form.
2. You'll see an error about invalid value for the `sell amount`
3. You should see errors about the `price` inputs being empty, but there are none

Screenshot from 2020-02-27 08-32-55

### Requirement:
Show error messages when price inputs are empty and the user tries to submit the form.

anxolin:
Close #634  

This is a proposal to solve the issue above describe. It's simple, do not disable the submit button.

use-form already handle the validation, also it's nice that it focuses in the input field that is incorrect.

i think this approach is more UX friendly

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -522,7 +522,7 @@ const TradeWidget: React.FC = () => {
           <SubmitButton
             data-text="This order might be partially filled."
             type="submit"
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr642--dexreact.review.gnosisdev.com

PR #643, opened 2020-02-27
627/liquidity spread input

### LINKED ISSUE #627: nit: The spread input in liquidity doesn't behave properly
image

It has the same behaviour as the old inputs in the deposit widget. It has leading zeros, and the behaviour is strange.

Maybe is a matter of using the hook `useNumberInput` or finding a more general solution for all our inputs.

It's feels overcomplicated what we do in the deposit widget. Maybe we can find a simpler approach and extrapolate it to all the app.

alfetopito:
Closes #627 

Adding extra input cleaning and tooltip to `spread` input on liquidity page.
Also moving to next step on `enter`.


gnosis-info:
Travis automatic deployment:
 https://pr643--dexreact.review.gnosisdev.com

src/components/PoolingWidget/index.tsx
```diff
@@ -296,8 +296,9 @@ const PoolingInterface: React.FC = () => {
       txHash,
       txReceipt,
       txError,
+      nextStep,
     }),
-    [handleTokenSelect, selectedTokensMap, setSpread, spread, tokens, txError, txHash, txReceipt],
+    [handleTokenSelect, nextStep, selectedTokensMap, setSpread, spread, tokens, txError, txHash, txReceipt],
```
Velenir: While you are at it can you please change `prevStep, nextStep` to use `setStep(step => step +- 1)` and remove `[setStep, ~step~]` from deps

src/utils/validation.ts
```diff
@@ -0,0 +1,15 @@
+export const validInputPattern = new RegExp(/^\d+\.?\d*$/) // allows leading and trailing zeros
+
+export function preventInvalidChars(event: React.KeyboardEvent<HTMLInputElement>): void {
+  if (
+    event.key !== 'Backspace' &&
+    event.key !== 'Delete' &&
```
Velenir: I wonder if these kind of guards can be tricked with `Ctrl+X` - cut text :thinking:

W3stside:
> It doesn't quite work like you would expect
> ezgif-7-5e3517811f35

Dima the gifboi

src/utils/validation.ts
```diff
@@ -0,0 +1,15 @@
+export const validInputPattern = new RegExp(/^\d+\.?\d*$/) // allows leading and trailing zeros
```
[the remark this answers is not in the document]
alfetopito: Yes, because there's another validation preventing numbers >= 100 that kicks in before this one.

src/components/PoolingWidget/DefineSpread.tsx
```diff
@@ -35,17 +40,53 @@ export const SpreadInformation: React.FC<SpreadInformationProps> = ({ selectedTo
   )
 }
 
-const DefineSpread: React.FC<DefineSpreadProps> = ({ selectedTokensMap, spread, setSpread }) => {
-  const handleSpreadChange = ({ target: { value } }: React.ChangeEvent<HTMLInputElement>): void => {
-    if (+value < 0 || +value >= 100) return
+const DefineSpread: React.FC<DefineSpreadProps> = ({ selectedTokensMap, spread, setSpread, nextStep }) => {
+  const onChange = useCallback(
+    ({ currentTarget: { value } }: React.SyntheticEvent<HTMLInputElement>): void => {
+      if (+value === NaN || +value <= 0 || +value >= 100) {
```
W3stside: use `isNaN`

src/components/PoolingWidget/SubComponents.tsx
```diff
@@ -16,10 +16,22 @@ interface SubComponentProps extends TokenSelectorProps {
   txHash: string
   txReceipt?: Receipt
   txError?: Error
+  nextStep: () => void
 }
 
 const SubComponents: React.FC<SubComponentProps> = props => {
```
W3stside: why not destructure in signature?

src/utils/validation.ts
alfetopito: Maybe, but in the end, `onBlur` should clear out that stuff

src/components/PoolingWidget/DefineSpread.tsx
Velenir: second that, nobody equals NaN, even `NaN === NaN -> false`

src/utils/validation.ts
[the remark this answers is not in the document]
alfetopito: Now that is fixed onChange

src/components/PoolingWidget/DefineSpread.tsx
```diff
@@ -35,17 +39,69 @@ export const SpreadInformation: React.FC<SpreadInformationProps> = ({ selectedTo
...
-  const handleSpreadChange = ({ target: { value } }: React.ChangeEvent<HTMLInputElement>): void => {
-    if (+value < 0 || +value >= 100) return
+const DefineSpread: React.FC<DefineSpreadProps> = ({ selectedTokensMap, spread, setSpread, nextStep }) => {
+  const onChange = useCallback(
+    ({ currentTarget: { value } }: React.SyntheticEvent<HTMLInputElement>): void => {
+      const newValue = value
+        // put a `0` at the start, if starting with `.`
+        .replace(/^(\..*)$/, '0$1')
+        // remove extra decimal places, if present and more than 1
+        .replace(/^(.*\.\d{0,1}).*$/, '$1')
```
Velenir: Small simplification suggestion
```suggestion
        .replace(/^\./, '0$&')
        // remove extra decimal places, if present and more than 1
        .replace(/(\.\d)\d+$/, '$1')
```

The most apparent benefit is that `/^(.*\.\d{0,1}).*$/` always matches even `0.4` and therefore runs the replacements function, even though it literally replaces `0.4 -> 0.4`
`(\.\d)\d+$` doesn't match `0.4` and doesn't run replacement

src/components/PoolingWidget/DefineSpread.tsx
```diff
@@ -35,17 +39,69 @@ export const SpreadInformation: React.FC<SpreadInformationProps> = ({ selectedTo
...
+  const onKeyPress = useCallback(
+    (event: React.KeyboardEvent<HTMLInputElement>): void => {
+      if (event.key === 'Enter') {
+        onBlur(event)
+        nextStep()
+      } else if (
+        // allow to remove chars and add numbers and a dot
+        !event.key.match(/(Backspace|Delete|\.|[0-9])/) ||
+        // do not allow a second dot
+        (event.key === '.' && event.currentTarget.value.match(/\./))
```
Velenir: It is generally recommended to use `regexp.test` if you are only interested in `true|false`. And `string.includes` to check for 1 char only
```suggestion
        !/(Backspace|Delete|\.|[0-9])/.test(event.key) ||
        // do not allow a second dot
        (event.key === '.' && event.currentTarget.value.includes('.')
```

src/utils/format.ts
```diff
@@ -56,13 +56,19 @@ export function sanitizeNegativeAndMakeMultipleOf(value?: string | null, default
...
 }
 
-export function validatePositive(value: string, constraint = 0): true | string {
-  return Number(value) > constraint || 'Invalid amount'
-}
-export const validInputPattern = new RegExp(/^\d+\.?\d*$/) // allows leading and trailing zeros
 export const leadingAndTrailingZeros = new RegExp(/(^0*(?=\d)|\.0*$)/, 'g') // removes leading zeros and trailing '.' followed by zeros
 export const trailingZerosAfterDot = new RegExp(/(.*\.\d+?)0*$/) // selects valid input without leading zeros after '.'
 
+/**
```
Velenir: Why are we using RegExp constructor above? How did it get here
Please change to `/(^0*(?=\d)|\.0*$)/g, /(.*\.\d+?)0*$/`

src/components/PoolingWidget/DefineSpread.tsx
Velenir: I'm gonna commit this, because it solves a problem with trying to input `0.25`, as per @anxolin experience

Velenir:
Superseded by #683 ?

W3stside:
> Superseded by #683 ?

Would like that yes, makes it easier for merging etc

alfetopito:
https://github.com/gnosis/dex-react/pull/643#issuecomment-596973483

PR #647, opened 2020-02-28
Render optimization

Velenir:
Some optimizations I made when trying to address _flashing screen when opening Modali_ issue
Doesn't solve that problem but a nice to have nevertheless


gnosis-info:
Travis automatic deployment:
 https://pr647--dexreact.review.gnosisdev.com

PR #650, opened 2020-02-28
Fix for modali screen flash

Velenir:
When opening Modali modal from Withdraw modal (when overriding withdraw request) screen briefly flashes
ezgif-7-c2fda6feb21c

Happens because Withdraw modal has it's own overlay (not really overlay, but box shadow), so what actually goes on is:

1. Withdraw overlay on -> overlay background darkish
2. Withdraw off -> no overlay -> white background
3. Modali on -> overlay background darkish

A proper fix would be to change not modals but content of one modal: Form -> Modali contents,; without changing overlay in-between

Candidate for refactor? Candidate for refactor!

### DIFF
--- a/src/styles/global.ts
+++ b/src/styles/global.ts
@@ -947,6 +947,23 @@ const GlobalStyles = createGlobalStyle`
     color: #476481;
   }
   // End WalletConnect Modal
+
+  /* Hack to fix Modali screen flash */
+  .modali-overlay {
+    display: none
+  }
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr650--dexreact.review.gnosisdev.com

Velenir:
> Should we add a issue for revisiting this?

Done

PR #652, opened 2020-02-28
Fix stale wallet info

Velenir:
Get new WalletInfo right away

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -570,8 +570,10 @@ const TradeWidget: React.FC = () => {
       })
     } else {
       // Not connected. Prompt user to connect his wallet
-      const success = await connectWallet()
-      if (success) {
+      const walletInfo = await connectWallet()
+      if (walletInfo && walletInfo.userAddress && walletInfo.networkId) {
+        // pass new WalletInfo data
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr652--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -570,7 +570,21 @@ const TradeWidget: React.FC = () => {
       })
     } else {
       // Not connected. Prompt user to connect his wallet
-      await connectWallet()
+      const walletInfo = await connectWallet()
+      if (walletInfo && walletInfo.userAddress && walletInfo.networkId) {
```
W3stside: `if (walletInfo?.userAddress && walletInfo?.networkId) { ...`

Velenir:
@alfetopito already did good!

src/components/TradeWidget/index.tsx
alfetopito: Hmmm nah. TS warning:

```
Property 'networkId' does not exist on type 'false | WalletInfo'.
  Property 'networkId' does not exist on type 'false'.ts(2339)
```

PR #661, opened 2020-02-28
OrderValidity modal sanitization logic

W3stside:
when manually setting `0` values for both `validFrom` and `validUntil` the app wasn't setting the toggle on the checkboxes - WELL now it does

### DIFF
--- a/src/components/TradeWidget/OrderValidity.tsx
+++ b/src/components/TradeWidget/OrderValidity.tsx
@@ -184,6 +184,8 @@ const OrderValidity: React.FC<Props> = ({
       const sanitizedUntilValue = makeMultipleOf(5, validUntilInputValue).toString()
 
       batchedUpdates(() => {
+        if (sanitizedFromValue === '0') setAsap(true)
+        if (sanitizedUntilValue === '0') setUnlimited(true)
         setValue(validFromInputId, sanitizedFromValue, true)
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr661--dexreact.review.gnosisdev.com

src/components/TradeWidget/OrderValidity.tsx
```diff
@@ -184,14 +184,18 @@ const OrderValidity: React.FC<Props> = ({
       const sanitizedUntilValue = makeMultipleOf(5, validUntilInputValue).toString()
 
       batchedUpdates(() => {
+        if (sanitizedFromValue === '0') setAsap(true)
```
anxolin: I find strange that we need two vars for the from and the until. I mean, For example the `asap`, couldn't just be derived from the value of `validFrom` instead of having a hook for it's state?

W3stside: this is true, i'll look into it for after the merge

W3stside: but it is also used for the `checkbox`es next to the inputs, you can set the `validFrom` by checking the box so it needs a toggle

W3stside: or you mean setting the form to zero on checkbox change? :thinking: that also makes sense

PR #677, opened 2020-03-02
Non-blocking Deposit/Withdraw modals

### LINKED ISSUE #621: Improve deposit/withdrawals so it's not blocking
The current version makes the user wait until the transaction is mined, then it closed the modal.

Implement an async workflow.

image

Velenir:
Deposit/Withdraw modals now close `onSentTransaction`. That is when user confirms transaction in the wallet.
To show tx in progress as modals are closed now there are spinners near values that will be updated
That's why there are individual sets to track claiming/withdrawing/depositing txs in progress

Closes #621

### DIFF
--- a/src/components/DepositWidget/Styled.tsx
+++ b/src/components/DepositWidget/Styled.tsx
@@ -95,27 +95,19 @@ export const RowClaimButton = styled.button`
   }
 `
 
-export const RowClaimLink = styled.a`
-  text-decoration: none;
-  cursor: pointer;
+export const RowClaimSpan = styled.span`
   font-size: 1.2rem;
   line-height: 1;
+  display: block;
+  border: 0.1rem solid #63ab52;
+  border-radius: 2rem;
+  background: transparent;
+  color: #63ab52;
+  padding: 0.4rem 1rem;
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr677--dexreact.review.gnosisdev.com

src/components/DepositWidget/Styled.tsx
```diff
@@ -95,27 +95,19 @@ export const RowClaimButton = styled.button`
   }
 `
 
-export const RowClaimLink = styled.a`
-  text-decoration: none;
-  cursor: pointer;
+export const RowClaimSpan = styled.span`
```
W3stside: why not make this a `div` if using display block? (curious)

Velenir: div is generally not to be used inside button
only phrasing content
or so says the spec
for example `<button><div/></button>` would fail w3c validation, not that really matters

Velenir: But you are right that it doesn't have to be `block`, anything inside a flex is already a `block`. Will remove

PR #682, opened 2020-03-04
Fix Firefox Escape keydown

Velenir:
So this is a quick fix and essentially a workaround

When Deposit/Withdraw modal opens in Chrome focus returns from the button just clicked to body
In Firefox it doesn't

That button swallows all onClick/KeyDown and doesn't let it bubble because react catches it, normally wouldn't prevent anything, or I think that's what happens

If we explicitly move focus bu clicking somewhere it works. Also it makes sense to just autofocus input

### DIFF
--- a/src/components/DepositWidget/Form.tsx
+++ b/src/components/DepositWidget/Form.tsx
@@ -126,6 +126,7 @@ export const Form: React.FC<FormProps> = (props: FormProps) => {
                 value={amountInput}
                 onChange={(e: ChangeEvent<HTMLInputElement>): void => setAmountInput(e.target.value)}
                 placeholder="0"
+                autoFocus
               />
             </div>
             {/* Error Message */}


gnosis-info:
Travis automatic deployment:
 https://pr682--dexreact.review.gnosisdev.com

Velenir:
> And also, can you point this to the hotfix/0.9.1 branch instead of develop?

bad idea, I tried and got a thousand commits to merge
this isn't an essential fix, can be merged from develop to hotfix or even master whenever

alfetopito:
> > And also, can you point this to the hotfix/0.9.1 branch instead of develop?
> 
> bad idea, I tried and got a thousand commits to merge
> this isn't an essential fix, can be merged from develop to hotfix or even master whenever

You could rebase, but agree, not a priority fix.

PR #701, opened 2020-03-11
Pooling Input Error

W3stside:
Added formatted error under inputs - need to rework other inputs next in a separate PR

### DIFF
--- a/src/components/TradeWidget/FormMessage.tsx
+++ b/src/components/TradeWidget/FormMessage.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react'
+import React, { useMemo } from 'react'
 import styled from 'styled-components'
 import { formatSchemaErrorMessage } from 'utils'
 
@@ -64,7 +64,7 @@ export const FormInputError: React.FC<FormInputErrorProps> = ({
   visibilityHidden = true,
   wrapperMargin = '0.5rem 0',
 }) => {
-  const message = useCallback(() => {
+  const message = useMemo(() => {
     if (errorMessage) {
       if (schemaError) {
         return formatSchemaErrorMessage(errorMessage)
@@ -82,7 +82,7 @@ export const FormInputError: React.FC<FormInputErrorProps> = ({
 
   return (
     <FormMessageWrapper $margin={wrapperMargin}>
-      <FormMessage className={errorMessage ? 'error' : 'hidden'}>{message()}</FormMessage>
+      <FormMessage className={errorMessage ? 'error' : 'hidden'}>{message}</FormMessage>
     </FormMessageWrapper>
   )
 }
--- a/src/components/TradeWidget/FormMessage.tsx
+++ b/src/components/TradeWidget/FormMessage.tsx
@@ -1,4 +1,11 @@
+import React, { useCallback } from 'react'
 import styled from 'styled-components'
+import { formatSchemaErrorMessage } from 'utils'
+
+export const FormMessageWrapper = styled.div<{ $margin?: string; $padding?: string }>`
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr701--dexreact.review.gnosisdev.com

src/utils/validation.ts
```diff
@@ -58,3 +58,9 @@ export const stringOrNumberResolverFactory = <FormData, CustomContext = {}>(
       : {},
   }
 }
+
+export function formatSchemaErrorMessage(errorString?: string): string | undefined {
+  if (!errorString) return undefined
+  const cleanedString = errorString.replace(/"/g, '')
```
Velenir: Why? How does it look without replace?

src/components/TradeWidget/FormMessage.tsx
```diff
@@ -39,6 +46,45 @@ const FormMessage = styled.div.attrs<{ className?: string }>(props => ({
...
+  wrapperMargin?: string
+}
+
+export const FormInputError: React.FC<FormInputErrorProps> = ({
+  errorMessage,
+  schemaError = true,
+  visibilityHidden = true,
+  wrapperMargin = '0.5rem 0',
+}) => {
+  const message = useCallback(() => {
```
Velenir: Since you are calling `message()` every time anyway, no need for it to be a function, do `useMemo` and construct message inside

src/components/TradeWidget/FormMessage.tsx
```diff
@@ -39,6 +46,45 @@ const FormMessage = styled.div.attrs<{ className?: string }>(props => ({
     box-sizing: border-box;
     margin: 0.3rem 0 1rem;
   }
+  &.hidden {
+    visibility: hidden;
```
Velenir: Why switching between using classes for css and props.
Stick to one. Like `visibility: ${props => props.visible ? 'hidden' : 'visible'}`

Or don't. Depends on semantics I guess

What's your opinion?

src/utils/validation.ts
W3stside: `"spread" is dumb, change that shit` - there are some settings inside the API that allow me to change it but i couldnt figure it out

src/components/TradeWidget/FormMessage.tsx
W3stside: A+

src/components/TradeWidget/FormMessage.tsx
W3stside: i was having issues with the `Styled` syntax of `styled.div.attrs<{ ... }>` and adding my own props - it was throwing in TS. Any advice? An example would be useful

Velenir: [example](https://github.com/gnosis/dex-react/blob/67f578ce5f0187641297b158e86cca1e2010779d/src/components/TradeWidget/OrderValidity.tsx#L87)

W3stside: No I mean when it's a styled component using `.attrs`

W3stside:
> I like it!
> 
> Also, with my new locale, I get localized error messages. And a combo of messages as well :D
> 
> Captura de tela_2020-03-11_08-03-11

ole'!

src/components/TradeWidget/FormMessage.tsx
W3stside: will not change this for now, thanks tho

anxolin:
Please, leave more time for review, cause I was off yesterday and it takes me a bit of time to catch up some times. Let's always try to leave a fair amount of time

image

PR #705, opened 2020-03-11
Fix overflow undulating at 886px window width

Velenir:
In general not a good idea to use `height: 100vh` as it doesn't account for scrollbar size.

overflow


gnosis-info:
Travis automatic deployment:
 https://pr705--dexreact.review.gnosisdev.com

PR #707, opened 2020-03-12
Pooling interface input precision check

W3stside:
1. remove trailing zeros on SpreadInformation data
2. precision on liquidity spread input set to 6


gnosis-info:
Travis automatic deployment:
 https://pr707--dexreact.review.gnosisdev.com

PR #710, opened 2020-03-12
Pooling interface back button rework

W3stside:
There were some remaining issues with back button addressed here:

1. disappearing back button when it shouldn't have been
2. formatting
3. not disabled when it should have been


gnosis-info:
Travis automatic deployment:
 https://pr710--dexreact.review.gnosisdev.com

anxolin:
> There were some remaining issues with back button addressed here
Can you develop a bit the description?

anxolin:
The positioning of the button looks off. I see now it's disabled when you send the transaction. I think it's fine, although maybe better to hide it, also would solve the positioning thing too :) 

image

Velenir:
I don't see that wrong `< BACK` positioning

W3stside:
Positioning was fixed in last commit

PR #711, opened 2020-03-12
696/trade page formatting

### LINKED ISSUE #696: Review some design issues in trade page
Some users report:
* Not clear to me which input fields I can amend and which ones I cannot. 
* Also, I think the formatting of the balances (here the 99.52 and the 0 are off).

image

W3stside:
Closes #696 

1. Read only inputs now have a darker bg via datatype prop
2. fixes zindex overlap issue with table header in orders and tooltip (not sure why the theader was set to zindex 5 in the first place?? may have been an artifact from earlier card css before michel)
3. fixes placeholder issue

### DIFF
--- a/src/components/Layout/Card/Card.tsx
+++ b/src/components/Layout/Card/Card.tsx
@@ -128,8 +128,6 @@ export const CardTable = styled.table<{
     position: sticky;
     background: #ecf2f7;
     top: 0;
-    // TODO: why is this 5? see no problems in testing
-    // z-index: 5;
     font-size: 1.1rem;
     color: #2F3E4E;
     letter-spacing: 0;
--- a/src/components/Layout/Card/Card.tsx
+++ b/src/components/Layout/Card/Card.tsx
@@ -128,7 +128,8 @@ export const CardTable = styled.table<{
     position: sticky;
     background: #ecf2f7;
     top: 0;
-    z-index: 5;
+    // TODO: why is this 5? see no problems in testing
+    // z-index: 5;
     font-size: 1.1rem;
     color: #2F3E4E;
     letter-spacing: 0;
--- a/src/styles/global.ts
+++ b/src/styles/global.ts
@@ -123,7 +123,6 @@ const GlobalStyles = createGlobalStyle`
     &::placeholder {
       color: inherit;
       font-size: inherit;
-      line-height: inherit;
     }
     
     &:focus::placeholder {
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -124,8 +124,8 @@ const InputBox = styled.div`
     }
 
     &[datatype='readOnlyInput'] {
-      background: #334e69;
-      color: white;
+      background-color: var(--color-background-pageWrapper);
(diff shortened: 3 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr711--dexreact.review.gnosisdev.com

src/components/Layout/Card/Card.tsx
```diff
@@ -128,7 +128,8 @@ export const CardTable = styled.table<{
     position: sticky;
     background: #ecf2f7;
     top: 0;
-    z-index: 5;
+    // TODO: why is this 5? see no problems in testing
+    // z-index: 5;
```
Velenir: Who knows, was probably below something otherwise

src/styles/global.ts
```diff
@@ -123,7 +123,6 @@ const GlobalStyles = createGlobalStyle`
     &::placeholder {
       color: inherit;
       font-size: inherit;
-      line-height: inherit;
```
Velenir: ? was unnecessary ?

W3stside: yeah

anxolin:
mm, not sure the new color is the right choice.

I think is still odd and not super clear:
image

I did this attempt making the editable fields white (with a border) and leaving the gray for read-only, but still feels not clear.

image

I would ask for advice to @biocom and @leafilipo about this.

biocom:
As a quick reply, I'd just refer to https://material.io/components/text-fields/#filled-text-field where the current input fields are inspired from:

mio-components_assets_17_fSR2ls9Rp3kERxc2adlMDcn1YGuIeY_textfields-filled-prefix-states

anxolin:
I see the point @biocom, but we still have people claiming that it's difficult to know that the input is not editable.

What about making the buy amount not an input? Just a text/label

alfetopito:
Placeholder :+1: 

Token selector :-1: 

Something broke, not sure in this PR, though. Notice the rounded grey area:

Captura de tela_2020-03-12_09-36-50

(and that is before my new weird/broken token selector styles, just merged that)

Velenir:
> As a quick reply, I'd just refer to https://material.io/components/text-fields/#filled-text-field where the current input fields are inspired from

I have to say Material Design looks too subtle for me

biocom:
You can also look at the previous version of the disabled state:
https://storage.googleapis.com/spec-host-backup/mio-components%2Fassets%2F1oQcwFHZmAd-JYB-x46AUuEH7uzkrCKCB%2Ftextfields-outlined-suffix-states.png

Otherwise, seems like we tried most of options by now.

biocom:
Screenshot 2020-03-13 at 11 38 33

One suggestion ^

leafilipo:
Hi just stepping in the conversation. I am not so familiar with the interface but this last version seems like a better choice in comparison to the first dark grey version for a read only button.

W3stside:
@anxolin @alfetopito

biocom:
> Like the new input
> 
> BTW, the token selector still has the half of the page rounded dark background. Is that coming from somewhere else?

In my opinion, given that is a button where user interactivity is required, I'd opt for keeping it as is. Curious on your thoughts there.

PR #721, opened 2020-03-16
691/use max locale

### LINKED ISSUE #691: Issue with locale and `use max` button
I understand from this comment we have an issue with the locale:
https://github.com/gnosis/dex-react/issues/500#issuecomment-594522648

`use max` uses an English/America format for numbers, while the inputs and validation expects the local user's locale, this is causing the form validation to fail, so the user has to modify the symbols manually.

alfetopito:
Fixes #691 

Captura de tela de 2020-03-16 07-53-25


gnosis-info:
Travis automatic deployment:
 https://pr721--dexreact.review.gnosisdev.com

PR #723, opened 2020-03-17
Fix duplicate orders

### LINKED ISSUE #718: Pending orders and normal orders are duplicated at one point
1. didn't have orders

2. Created an order. It was showing as pending:

IMG_0505

3. When the tx got confirmed:
IMG_0506

4. Then, after 30s or so, the pending disappeared:
IMG_0507

Velenir:
So I thought about filtering `pendingOrders` against `orders`
But as @alfetopito  mentioned in #718 issue we can't really
Bu then again we can pretty much guess what id will be when assigning pendingOrders as ids are incremental.
And there you have it

Anticipating questions:
Q: Why not use `offset` for `lastId`?
A: While it almost always matches, there are cases when we reset offset to refetch every order (on order cancellation/deletion), so it's less reliable.

Q: But why not ...
A: No,no trust me on this.

Also, I can't really reproduce duplicate orders, for me it's always mined first, new block with orders later. So would be cool if someone who can reproduce tested this PR.

Closes #718

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -56,11 +56,18 @@ function classifyOrders(
   orders: AuctionElement[],
   state: FilteredOrdersState,
   ordersType: 'orders' | 'pendingOrders',
-): void {
+  alreadyCounted?: Set<string>,
+): Set<string> {
   const now = new Date()
   const isOrderActiveFn = ordersType === 'pendingOrders' ? isPendingOrderActive : isOrderActive
+  const counted = new Set<string>()
 
   orders.forEach(order => {
+    // don't process already processed orders
+    if (alreadyCounted && alreadyCounted.has(order.id)) return
+
+    counted.add(order.id)
+
     if (!isOrderActiveFn(order, now)) {
       state.closed[ordersType].push(order)
     } else if (isOrderUnlimited(order.priceDenominator, order.priceNumerator)) {
@@ -69,6 +76,8 @@ function classifyOrders(
       state.active[ordersType].push(order)
     }
   })
+
+  return counted
 }
 
 const OrdersWidget: React.FC = () => {
@@ -108,8 +117,12 @@ const OrdersWidget: React.FC = () => {
   useEffect(() => {
     const filteredOrders = emptyState()
 
-    classifyOrders(allOrders, filteredOrders, 'orders')
-    classifyOrders(allPendingOrders, filteredOrders, 'pendingOrders')
+    const countedOrderIds = classifyOrders(allOrders, filteredOrders, 'orders')
+
+    // don't include orders fetched from blockchain among pending orders
+    // consider them as finished pending
+    // even if pending tx hasn't been reported as mined yet
+    classifyOrders(allPendingOrders, filteredOrders, 'pendingOrders', countedOrderIds)
 
     setFilteredOrders(curr => {
       // copy markedForDeletion
--- a/src/components/PoolingWidget/index.tsx
+++ b/src/components/PoolingWidget/index.tsx
@@ -103,9 +103,11 @@ const validationSchema = joi.object({
 const numberResolver = stringOrNumberResolverFactory<PoolingFormData>(validationSchema, 'number')
 
 const PoolingInterface: React.FC = () => {
+  const { networkId, userAddress } = useWalletConnection()
   const [
     {
       orders: { orders },
+      pendingOrders,
     },
     dispatch,
   ] = useGlobalState()
@@ -117,9 +119,11 @@ const PoolingInterface: React.FC = () => {
   const [txReceipt, setTxReceipt] = useSafeState<Receipt | undefined>(undefined)
   const [txError, setTxError] = useSafeState(undefined)
 
-  const lastOrderIdRef = useLastOrderIdRef(orders)
+  const lastOrderIdRef = useLastOrderIdRef(
+    orders,
+    networkId && userAddress ? pendingOrders[networkId][userAddress] : [],
+  )
 
-  const { networkId, userAddress } = useWalletConnection()
   // Avoid displaying an empty list of tokens when the wallet is not connected
   const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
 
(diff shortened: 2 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr723--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -56,11 +56,18 @@ function classifyOrders(
...
-): void {
+  alreadyCounted?: Set<string>,
+): Set<string> {
   const now = new Date()
   const isOrderActiveFn = ordersType === 'pendingOrders' ? isPendingOrderActive : isOrderActive
+  const counted = new Set<string>()
 
   orders.forEach(order => {
+    // don't process already processed orders
+    if (alreadyCounted && alreadyCounted.has(order.id)) return
```
anxolin: I didn't understand why an order would be two times.

It's pending, but once we get the confirmation the changes into a not pending one.

anxolin: If we do this only in the widget, wouldn't we have this problem in any part using the global state?

Would it make sense to just delete the pending orders once we have the confirmation?

src/components/PoolingWidget/index.tsx
```diff
@@ -166,10 +173,11 @@ const PoolingInterface: React.FC = () => {
             setTxHash(txHash)
 
             batchedUpdates(() => {
+              let orderId = lastOrderIdRef.current + 1
```
anxolin: I see why you do this, but would this work the second time I send an order?
Are the pending orders  also consider in the `useLastOrderIdRef(orders)`? If not, Is it possible that together with the change of not showing trades with the same id would make it to not display the second pending orders.

src/components/PoolingWidget/index.tsx
```diff
@@ -166,10 +173,11 @@ const PoolingInterface: React.FC = () => {
...
               orders.forEach(({ buyToken: buyTokenId, sellToken: sellTokenId, buyAmount, sellAmount }) => {
                 const pendingOrder = {
                   txHash,
-                  id: 'PENDING ORDER',
+                  id: (orderId++).toString(10),
```
anxolin: I think it was wrong to have `'PENDING ORDER'` as the ID an I see the point on guessing it from our current orders, but in general I think it would be fine to create a unique temporal ID (i.e. using a timestamp) and then DELETE the pending order when it's confirmed.

I remember working in this part of the code, and it's asking for a bit of refactor. Looks a bit complex right now, and we keep adding more complexity.

anxolin:
Also, this is relying on "assuming" optimistic mining and no other dapp or script creating orders in parallel.

i.e. 
0. User has 0 orders
1. User sends an order. OrderId = 1
2. it's still pending when he refresh the browser
3. He sends a second order. Order = 1

Theoretical final result with the changes as I understand in this PR. Only the first order is rendered

src/components/OrdersWidget/index.tsx
Velenir: We delete it on confirmation of mined tx already. I don't understand what you mean.

@alfetopito described exactly why we have duplicate orders in the issue

src/components/PoolingWidget/index.tsx
Velenir: Prending orders are being deleted anyway. Once tx is mined

src/components/PoolingWidget/index.tsx
Velenir: You are right
I address it in 0270223

Velenir:
> Also, this is relying on "assuming" optimistic mining and no other dapp or script creating orders in parallel.

Well yes, that's why I say "guess" in the description.
But guess is enough, in my opinion. All that guessing clears up as soon as pending orders are mined and we only have normal orders in state.

@alfetopito, as you were already considering this issue, care to weigh in on this approach

gnosis-info:
Travis automatic deployment:
 https://pr723--dexreact.review.gnosisdev.com

src/components/PoolingWidget/index.tsx
anxolin: I think we would still have the issue of the refresh making the new pending order to dissapear if we get confirmation on an old pending order that was lost because of the refresh https://github.com/gnosis/dex-react/pull/723#issuecomment-600049389

anxolin:
I don't think the guessing is that bad. I would be comfortable with the concurrency. But I think an issue that concerns me more would be that in the situation described above, that is not that uncommon, the user will see a pending tx be replaced to a one that was pending and not in the state of the app (cause of the refresh)

alfetopito:
> > Also, this is relying on "assuming" optimistic mining and no other dapp or script creating orders in parallel.
> 
> Well yes, that's why I say "guess" in the description.
> But guess is enough, in my opinion. All that guessing clears up as soon as pending orders are mined and we only have normal orders in state.
> 
> @alfetopito, as you were already considering this issue, care to weigh in on this approach

Yes, this indeed looks a bit more deterministic, although only if the interface is the only source of truth.
I have the same concern as Anxo that this might lead to incorrect that when placing orders outside of the contract.
Although this also might have a low chance of happening.

Is this something easy to reproduce for you Anxo?

I'm considering that not taking any action might be better than having any work around which will not work 100% of the time.

Velenir:
But blockchain is the source of truth
Orders are always updated from chain. Only pending orders are from ui and they don't take precedence.

Pending txs would just disappear quicker in case of order outside
 of ui
Which only happens if user refreshes between order submission and mining, and makes another order immediately

Velenir:
> PROPOSAL: Use timestamps for pending orders. Delete pending order once it's confirmed. Cause it's gonna be added to the "normal orders" in that same block.

That's what we are doing prior to this PR, and why we have the duplicate issue

Velenir:
It seems nobody is happy with this approach
I'm fine with closing this PR
The initial issue isn't a serious one anyway.

alfetopito:
> It seems nobody is happy with this approach
> I'm fine with closing this PR
> The initial issue isn't a serious one anyway.

I'm happy to give it a try, but as you said, might not be that big of a deal.

@anxo or is it?

src/hooks/useGlobalState.tsx
```diff
@@ -23,3 +24,18 @@ export function withGlobalContext<P>(
 const useGlobalState = (): [GlobalState, Function] => useContext(GlobalStateContext) as [GlobalState, Function]
 
 export default useGlobalState
+
```
W3stside: can you move this out? 

In regards tot he functionality, I'm on the fence about adding it as both methods have their respective pros and cons. May be better not to add it for now.

Velenir:
Outdated and unnecessary

Issue #658, opened 2020-02-28
Pre-fill the price


anxolin:
Feedback from Martin
-------------

I think we should prefill the sell price. Even if it is something completely random like 1 or better 0.995 - it just makes it clear that when you type in “sell amount” the other receive amount is calculated.

0.995 might be a good number because for the stable/stable coins case that is sensible. For the stable-ETH usecase it does of course make no sense but we will get there in a future version.

W3stside:
this would also solve the initial 0 validation problem reported by Marco, I am ok with it

anxolin:
We might consider using the graph prices here. Probably it's easier and good enough estimate.

anxolin:
Other alternative could be to get the latest available price from the contract. In any case, let's make a simple concrete proposal so we can validate it with Martin and implement it.

This initial criteria can be evolved and in the future we could have better price estimation, but let's get an initial version rolling :)

alfetopito:
As suggested during the sprint planning, we could use a naive approach and get the value from the contract https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L63

Although that price is with reference to OWL... might not be what we are looking for.

Velenir:
So 
```js
Token1InOwl = contract.currentPrices(Token1)
Token2InOwl = contract.currentPrices(Token2)
price = Token1InOwl/Token2InOwl
```

or something?

anxolin:
Yes, although let me double confirm today

alfetopito:
Maybe this won't work as we expect.
Just queried the mainnet contract manually for the prices for tokens 0 to 20, and I got prices != 0 only for 3 (technically 4, but OWL will always be set): 1, 7 and 15.

@fleupold @bh2smith @josojo do we expect to have the prices set under what circumstances?

If they are only set when they have been traded in the last batch, it might not be useful directly from the blockchain. We'll need another layer like the graph.

For context, we know @fleupold is working on a service for providing the price estimation. This is just intended as a quick and naive temporary solution.

fleupold:
> do we expect to have the prices set under what circumstances?

The smart contract only stores the prices of tokens that were traded in the most recent settlement. Thus, for a more complete picture (e.g. the last price any token was traded) we would either need to query `instance.contract.getPastEvents("SolutionSubmission", {fromBlock: 0})` (which is still fully decentralized and quite performant for now) or the Graph.

>  @fleupold is working on a service for providing the price estimation

[...]

PR #726, opened 2020-03-17
658/get current prices

alfetopito:
Part of #658 

Implementing contract read method `currentPrices`


gnosis-info:
Travis automatic deployment:
 https://pr726--dexreact.review.gnosisdev.com

alfetopito:
> Are we still going forward with this?

Well, not sure we are using it, but adding the method won't hurt (had already coded it before realizing wouldn't be that useful...)

anxolin:
I think this one is not needed, and probably won't ever be. It's a pity it's working and we don't add it, but if we don't need it maybe we should "not merge"?

alfetopito:
> I think this one is not needed, and probably won't ever be. It's a pity it's working and we don't add it, but if we don't need it maybe we should "not merge"?

Agreed

PR #733, opened 2020-03-18
TheGraphApi and getPriceEstimation service

alfetopito:
Implementation of TheGraphApi for now only querying the prices.

Implementation of `getPriceEstimation` service.

Using Dima's suggestion, `getPrices` query for multiple prices at once, and `getPriceEstimation` service query for all known tokens at once as well for caching all prices locally.

Open for feedbacks and suggestions regarding this approach.

In another PR, will implement `usePriceEstimation` hook and add it to the interface somehow.

### DIFF
--- a/src/api/thegraph/TheGraphApi.ts
+++ b/src/api/thegraph/TheGraphApi.ts
@@ -1,10 +1,13 @@
-import BigNumber from 'bignumber.js'
 import BN from 'bn.js'
 import { assert, ZERO } from '@gnosis.pm/dex-js'
 
 export interface TheGraphApi {
-  getPrice(params: GetPriceParams): Promise<BigNumber>
-  getPriceInWei(params: GetPriceParams): Promise<BN>
+  getPrice(params: GetPriceParams): Promise<BN>
+  getPrices(params: GetPricesParams): Promise<GetPricesResult>
+}
+
+interface GetPricesResult {
+  [tokenId: number]: BN
 }
 
 export interface GetPriceParams {
@@ -12,7 +15,12 @@ export interface GetPriceParams {
   tokenId: number
 }
 
-export interface QueryParams {
+export interface GetPricesParams {
+  networkId: number
+  tokenIds: number[]
+}
+
+interface QueryParams {
   networkId: number
   queryString: string
 }
@@ -23,8 +31,7 @@ interface UrlByNetwork {
 
 interface PricesData {
   data: {
-    prices: {
-      batchId: string
+    [tokenKey: string]: {
       priceInOwl: string
     }[]
   }
@@ -51,31 +58,23 @@ export class TheGraphApiImpl {
     this.urlByNetwork = params.urls
   }
 
-  public async getPrice({ networkId, tokenId }: GetPriceParams): Promise<BigNumber> {
-    const queryString = `{ 
-        prices(
-          first: 1, 
-          orderBy: batchId
-          orderDirection: desc
-          where: { token: "${tokenId}" }
-        ) {
-            batchId
-            priceInOwl
-        }}`
-
-    const gqlResult = await this.query<GqlResult>({ networkId, queryString })
-
-    if (isGqlError(gqlResult)) {
-      throw new Error(gqlResult.errors[0].message)
-    }
-
-    const price = gqlResult.data.prices[0].priceInOwl
-    return new BigNumber(price)
+  public async getPrice({ networkId, tokenId }: GetPriceParams): Promise<BN> {
+    const prices = this.getPrices({ networkId, tokenIds: [tokenId] })
+    return prices[tokenId]
   }
 
-  public async getPriceInWei({}: GetPriceParams): Promise<BN> {
-    // TODO: implement this, maybe?
-    return ZERO
+  public async getPrices({ networkId, tokenIds }: GetPricesParams): Promise<GetPricesResult> {
+    const tokens = tokenIds.map(tokenId => this.buildPriceQuery(tokenId)).join('\n')
+    const queryString = `{${tokens}}`
+
+    try {
+      const response = await this.query<PricesData>({ networkId, queryString })
+
+      return this.parsePricesData(response)
+    } catch (e) {
+      console.error(e)
+      throw new Error(`Failed to query prices: ${e.message}`)
+    }
   }
 
   private async query<T>({ networkId, queryString }: QueryParams): Promise<T> {
@@ -84,7 +83,6 @@ export class TheGraphApiImpl {
 
     const data = {
       query: queryString,
-      variables: null,
     }
 
     const response = await fetch(url, {
@@ -101,6 +99,34 @@ export class TheGraphApiImpl {
       throw new Error(`Request failed: [${response.status}] ${response.body}`)
     }
 
-    return response.json()
+    const gqlResult = await response.json()
+
+    if (isGqlError(gqlResult)) {
+      throw new Error(`Query failed: ${gqlResult.errors[0].message}`)
+    }
+
+    return gqlResult
+  }
+
+  private buildPriceQuery(tokenId: number): string {
+    return `Token${tokenId}: prices(
+  first: 1, 
+  orderBy: batchId, 
+  orderDirection: desc, 
+  where: {token: "${tokenId}"}
+) {
+  priceInOwl
+}`
+  }
+
+  private parsePricesData({ data }: PricesData): GetPricesResult {
+    return Object.keys(data).reduce((acc, tokenKey) => {
+      // not possible to have a key with only integers, thus `Token` prefix was added
+      const tokenId = +tokenKey.replace(/Token/, '')
+
+      // When there's no data for a token, return 0
+      acc[tokenId] = new BN(data[tokenKey].length > 0 ? data[tokenKey][0].priceInOwl : ZERO)
+      return acc
+    }, {})
   }
 }
--- a/src/api/thegraph/TheGraphApi.ts
+++ b/src/api/thegraph/TheGraphApi.ts
@@ -59,6 +59,8 @@ export class TheGraphApiImpl {
 
     const response = await fetch(url, {
       method: 'POST',
+      mode: 'cors',
+      credentials: 'omit',
       headers: {
         'Content-Type': 'application/json',
       },
--- a/src/api/thegraph/TheGraphApi.ts
+++ b/src/api/thegraph/TheGraphApi.ts
@@ -55,10 +55,14 @@ type GqlResult = PricesResponse | GqlError
 
 const isGqlError = (gqlResult: GqlResult): gqlResult is GqlError => 'errors' in gqlResult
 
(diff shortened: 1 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr733--dexreact.review.gnosisdev.com

Velenir:
Depending on how you connect GraphApi to TradeWidget and when you fetch, it may be a  good idea to fetch for every token together. For GraphQL it's only a single request (viva gql), and we can just pull prices and not bother again

[Example](https://thegraph.com/explorer/subgraph/gnosis/dfusion-staging?query=multiple%20prices)

gnosis-info:
Travis automatic deployment:
 https://pr733--dexreact.review.gnosisdev.com

Velenir:
You didn't squash :smile:

alfetopito:
> You didn't squash smile

Right, because I'll squash this PR when merging to develop, so double squash doesn't matter :)

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,59 @@
...
+    // expect to have tokens ids not often outside of our list.
+    // Thus, we can actually make a request once, and reuse the response for all token pairs falling into the same list
+
+    // all known token ids
+    const tokenIdsSet = new Set(tokenListApi.getTokens(networkId).map(token => token.id))
+    // just in case the tokens are not in our initial list:
+    tokenIdsSet.add(numeratorTokenId)
+    tokenIdsSet.add(denominatorTokenId)
+    // make it nice and sorted to help with caching
+    const tokenIds = [...tokenIdsSet].sort()
```
Velenir: _For an 'optimization' of sorts :wink:_ can presort in parent scope  if numeratorTokenId and denominatorTokenId are in the set, like they will usually be, and will certainly be included on repeated calls
Like:
```ts
// L20:
const tokenIdsSet = new Set(tokenListApi.getTokens(networkId).map(token => token.id))
const tokenIds = Array.from(tokenIdsSet).sort() // just my preference over ...spread for single iterables, fell free to ignore
...
// L33:
let needToResort = false
if (needToResort = needToResort || !tokenIdsSet.has(numeratorTokenId)) {
  tokenIdsSet.add(numeratorTokenId)
  
  tokenIds.push(numeratorTokenId)
}
if (needToResort = needToResort || !tokenIdsSet.has(denominatorTokenId)) {
  tokenIdsSet.add(denominatorTokenId)
  
  tokenIds.push(denominatorTokenId)
}
if (needToResort) {
  tokenIds.sort() // it sorts in place
}
```

_optimization intensifies_ :feelsgood:

src/api/thegraph/TheGraphApi.ts
```diff
@@ -0,0 +1,157 @@
...
+  token {
+    decimals
+  }
+}`
+  }
+
+  private parsePricesResponse({ data }: PricesResponse, inWei: boolean): GetPricesResult {
+    return Object.keys(data).reduce((acc, tokenKey) => {
+      // not possible to have a key with only integers, thus `Token` prefix was added
+      const tokenId = +tokenKey.replace(/Token/, '')
```
Velenir:
```suggestion
      const tokenId = +tokenKey.replace('Token', '')
```

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,59 @@
...
+    assert(numeratorTokenId !== denominatorTokenId, 'Token ids cannot be the equal')
+
+    // ---
+    // For an 'optimization' of sorts, we always query all known tokens plus the requested token ids.
+    // Assuming the caching is on for the TheGraphApi layer (it was last time I checked) and we don't
+    // expect to have tokens ids not often outside of our list.
+    // Thus, we can actually make a request once, and reuse the response for all token pairs falling into the same list
+
+    // all known token ids
+    const tokenIdsSet = new Set(tokenListApi.getTokens(networkId).map(token => token.id))
```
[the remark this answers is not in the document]
alfetopito: ~Well, I'm working under the assumption that the tokenList might be dynamic in the near future when we allow users to list any tokens.~

~For example, consider we have 4 tokens in the initial list and user adds 1.~

~When he selects 2 tokens in the list, we'll query for 4 tokens.
When he selects 1 from the list and 1 he added, we'll query for 5 tokens.~

~That gives us 2 different queries we make to the graph API, making the cache not very useful...~

You know what, writing my original thinking made me realize that doesn't make much sense. I'll do as you suggest.

src/api/thegraph/TheGraphApi.ts
alfetopito: but but but, I LIKE regexes :(

src/services/factories/getPriceEstimation.ts
alfetopito: 1 little problem with your logic, but it's a problem that I'm willing to be authoritarian about anyway.

It requires the `networkId` on factory creation time. 

But if we really think about it, doesn't make sense to have prices for any network other than Mainnet.

1. That's where we have trades and actual prices can be derived
2. For now only the mainnet endpoint (staging) can be queried for prices :D

So, I'll enforce the networkId and we are good :)

alfetopito: https://github.com/gnosis/dex-react/pull/733/commits/eb3c84047e656bfa13ad4d1c9ee332b82a55c4f1

gnosis-info:
Travis automatic deployment:
 https://pr733--dexreact.review.gnosisdev.com

src/services/factories/getPriceEstimation.ts
Velenir: Props to you for switching statement order in `needToResort || !tokenIdsSet.has(denominatorTokenId) -> addTokenId(denominatorTokenId) || needsToSort`

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return async (params: GetPriceParams): Promise<BigNumber> => {
+    const { numeratorTokenId, denominatorTokenId } = params
+
+    assert(numeratorTokenId !== denominatorTokenId, 'Token ids cannot be the equal')
+
+    // As much as it'd be visually pleasing, cannot simplify it to a single expression
+    // We need to execute both functions.
+    let needsToSort = addTokenId(numeratorTokenId)
```
anxolin: > As much as it'd be visually pleasing, cannot simplify it to a single expression. We need to execute both functions

I didn't fully get it :) 
I guess you mean this is not possible for some reason:
`const needsToSort = addTokenId(numeratorTokenId) || addTokenId(denominatorTokenId)`

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (denominatorTokenId === 0) {
+      // Optimization more for a matter of principle than performance:
+      // OWL token is always tokenId == 0 on the contract
+      // All prices are given in terms of OWL
+      // OWL price is always 1
+      // Anything divided by 1...
+      return prices[numeratorTokenId]
+    } else if (prices[numeratorTokenId].isZero() || prices[denominatorTokenId].isZero()) {
+      // there was never a trade for one of these tokens
+      return ZERO_BIG_NUMBER
```
anxolin: Should we return `undefined` so the UI decides what to do? probably adding a zero ;)

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const prices = await theGraphApi.getPrices({ networkId, tokenIds })
+
+    if (denominatorTokenId === 0) {
+      // Optimization more for a matter of principle than performance:
+      // OWL token is always tokenId == 0 on the contract
+      // All prices are given in terms of OWL
+      // OWL price is always 1
+      // Anything divided by 1...
+      return prices[numeratorTokenId]
```
anxolin: I see the point, but why doing this optimization, and not the case where numerator is 0, so the result is 1/price

I would remove this optimization and just make the:
```js
numeratorPriceInOwl = numeratorTokenId === 0 ? ONE : prices[numeratorTokenId]
denominatorPriceInOwl = denominatorTokenId === 0 ? ONE : prices[denominatorTokenId]
return numeratorPriceInOwl.dividedBy(denominatorPriceInOwl)
```

alfetopito:
> Hey, would that be equivalent thanks to boolean algebra?
> 
> ```js
> needsToSort = addTokenId(numeratorTokenId) || addTokenId(denominatorTokenId) // but not necessarily executed
> and
> needsToSort = !(!addTokenId(numeratorTokenId) && !addTokenId(denominatorTokenId)) // executes
> ```
> 
> Single expression yay! But nigh unreadable, don't use

My eyes, it hurts, please staaaap

src/services/factories/getPriceEstimation.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function getPriceEstimationFactory(factoryParams: {
+  theGraphApi: TheGraphApi
+  tokenListApi: TokenList
+}): (params: GetPriceParams) => Promise<BigNumber> {
+  const { theGraphApi, tokenListApi } = factoryParams
+
+  // Only make sense to fetch prices for mainnet, thus we won't accept networkId parameter on this service
+  const networkId = Network.Mainnet
```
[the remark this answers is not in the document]
alfetopito: For these reasons https://github.com/gnosis/dex-react/pull/733#discussion_r395695421

Plus, we cannot rely on Rinkeby token ids. We have that long pending issue with ids being hardcoded to Mainnet ids, and some tokens in the list are not registered on Rinkeby.

src/services/factories/getPriceEstimation.ts
alfetopito: Yes!
@Velenir proved me wrong right away of course, but I rather not do that.
I'll update the comment to be clearer.

src/services/factories/getPriceEstimation.ts
alfetopito: Hmmm :/

I don't really like that. Then we have to type guard and and so on...
Price 0 is enough to tell there's no price, no?

src/services/factories/getPriceEstimation.ts
alfetopito: Done, no more implicit or needing comments.

gnosis-info:
Travis automatic deployment:
 https://pr733--dexreact.review.gnosisdev.com

src/services/factories/getPriceEstimation.ts
alfetopito: Done

src/services/factories/getPriceEstimation.ts
alfetopito: Maybe that's really better.
Ok, stopped being a whining kid and did so.

gnosis-info:
Travis automatic deployment:
 https://pr733--dexreact.review.gnosisdev.com

PR #734, opened 2020-03-19
658/fix fetching

Velenir:
Main problem was wrong graph endpoints

Another problem is -- Price query field is only available on graph-staging right now
So I put staging url for mainnet, rinkeby won't work
Need to update the graph for them to work (gnosis/dex-subgraph#25)

Try
```js
await theGraphApi.getPrice({networkId: 1, tokenId 1)
```

### DIFF
--- a/src/api/thegraph/TheGraphApi.ts
+++ b/src/api/thegraph/TheGraphApi.ts
@@ -59,6 +59,8 @@ export class TheGraphApiImpl {
 
     const response = await fetch(url, {
       method: 'POST',
+      mode: 'cors',
+      credentials: 'omit',
       headers: {
         'Content-Type': 'application/json',
       },
--- a/src/api/thegraph/TheGraphApi.ts
+++ b/src/api/thegraph/TheGraphApi.ts
@@ -21,6 +21,26 @@ interface UrlByNetwork {
   [network: number]: string
 }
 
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr734--dexreact.review.gnosisdev.com

anxolin:
The issue with CORS i've also experienced it building the CLI, and I believe it was the same reason, we need to use https://api.thegraph.com instead of https://thegraph.com

Thanks Dima for the change. Today I'll prepare something so we can take the latest changes in staging to production.

Velenir:
> The issue with CORS i've also experienced it building the CLI

Well, it isn't an issue with cors per se. `curl https://thegraph.com` returns html, you can't POST there only GET, it reports cors error probably because of middleware order

src/api/thegraph/TheGraphApi.ts
```diff
@@ -59,6 +89,8 @@ export class TheGraphApiImpl {
 
     const response = await fetch(url, {
```
W3stside: let's make the fetch options a const somewhere?

Velenir: I'll leave that up to @alfetopito 
GraphApi will change anyway

PR #709, opened 2020-03-12
Add links to devportal in liquidity page

anxolin:
Closes #615 
Part of #673

Issue #615, opened 2020-02-26
Include link for "learn more about the spread and liquidity"


anxolin:
![image](https://user-images.githubusercontent.com/2352112/75346920-fadc3000-589f-11ea-97f7-7953f871122a.png)

![image](https://user-images.githubusercontent.com/2352112/75346955-0a5b7900-58a0-11ea-9f34-3fc65f5d0622.png)

anxolin:
@c3rnst when do you think we will have a link for here?

After or before EthCC?

PR #737, opened 2020-03-19
Restore commits

Velenir:
PRs #615 and #709 were reverted or just overwritten at some point
I suspect when StepDescriptor component was abstracted away in its own file


gnosis-info:
Travis automatic deployment:
 https://pr737--dexreact.review.gnosisdev.com

PR #739, opened 2020-03-20
Fix focus stealing in production

Velenir:
Original check was for a specific `displayName` of a child component.
That name as all the others was mangled and minified in production

Check proposed here looks at children of that child component, which comes from
```ts
// TokenSelector.tsx::L255
const noOptionsMessage = (): string => 'No results'
```

Still ugly but more robust


gnosis-info:
Travis automatic deployment:
 https://pr739--dexreact.review.gnosisdev.com

PR #747, opened 2020-03-23
Trade page global state

alfetopito:
Depends on #658 

Adding global state management to trade parameters.

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -26,7 +26,6 @@ import { usePlaceOrder } from 'hooks/usePlaceOrder'
 import { useQuery, buildSearchQuery } from 'hooks/useQuery'
 import useGlobalState from 'hooks/useGlobalState'
 import { savePendingOrdersAction, removePendingOrdersAction } from 'reducers-actions/pendingOrders'
-import { updateTradeState } from 'reducers-actions/trade'
 import { MEDIA } from 'const'
 
 import { tokenListApi } from 'api'
@@ -39,6 +38,14 @@ import Price, { invertPriceFromString } from './Price'
 import { useConnectWallet } from 'hooks/useConnectWallet'
 import { PendingTxObj } from 'api/exchange/ExchangeApi'
 import { usePriceEstimation } from 'hooks/usePriceEstimation'
+import {
+  updatePrice,
+  updateSellAmount,
+  updateSellToken,
+  updateBuyToken,
+  updateValidFrom,
+  updateValidUntil,
+} from 'reducers-actions/trade'
 
 const WrappedWidget = styled(Widget)`
   overflow-x: visible;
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr747--dexreact.review.gnosisdev.com

gnosis-info:
Travis automatic deployment:
 https://pr747--dexreact.review.gnosisdev.com

src/reducers-actions/trade.ts
```diff
@@ -0,0 +1,30 @@
...
+export const INITIAL_TRADE_STATE: TradeState = {
+  price: null,
+  sellAmount: null,
+  sellToken: null,
+  buyToken: null,
+  validFrom: '0',
+  validUntil: '2880',
+}
+
+export function reducer(state: TradeState, subset: TradeState): TradeState {
```
Velenir: `subset` should be `Partial<TradeState>` as it is a subset, why TS let you do not partial...

src/reducers-actions/trade.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface TradeState {
+  price: string | null
+  sellAmount: string | null
+  sellToken: Required<TokenDetails> | null
+  buyToken: Required<TokenDetails> | null
+  validFrom: string
+  validUntil: string
+}
+
+export function updateTradeState<K extends keyof TradeState>(
```
Velenir: What is this function, this should be an action constructor, something like:
```ts
const updateTradeState = (payload:Partial<TradeState>): YourAction => ({type: YOUR_TYPE, payload})
```

But if you go without actions in reducer (which is a completely valid approach), essentially making it the same as `setState`, then you don't even need `updateTradeState`. Just `dispatch(newTradeState_partial)`

Velenir: Now about that _completely valid approach_.
Won't work nicely every time. Why? Because the dispatch you use in react components is a `GlobalState.dispatch`. Whatever you give it goes through the `combinedReducer` before it gets to `trade::reducer`. It only works here because every other reducer checks for `action.type` and `updateTradeState` returns without `type`.

If you had any other reducer that operated without types, that other reducer would react to TradeState and spread it over whatever it spreads over, like butter :custard:  over bread :bread: , but wrong butter :fish_cake:  over wrong bread :baguette_bread: 

So do proper actions instead.

src/components/TradeWidget/index.tsx
```diff
@@ -403,7 +416,27 @@ const TradeWidget: React.FC = () => {
...
+  }, [dispatch, sellToken])
+  useEffect(() => {
+    dispatch(updateTradeState('buyToken', receiveToken as Required<TokenDetails>))
+  }, [dispatch, receiveToken])
+  useEffect(() => {
+    dispatch(updateTradeState('validFrom', validFromValue))
+  }, [dispatch, validFromValue])
+  useEffect(() => {
+    dispatch(updateTradeState('validUntil', validUntilValue))
+  }, [dispatch, validUntilValue])
```
Velenir:
```suggestion
  useEffect(() => {
    dispatch(updateTradeState({
      price: priceValue,
      sellAmount: sellValue,
      sellToken,
      buyToken: receiveToken,
      validFrom: validFromValue,
      validUntil: validUntilValue
    }))
  }, [dispatch, priceValue, sellValue, sellToken, receiveToken, validFromValue, validUntilValue])
```
if you change `updateTradeState` to accept `Partial<TradeState>`

src/reducers-actions/trade.ts
alfetopito: Hmmm, so that's why. If you check the commit history, I started cutting it off bit by bit to see how far the action boilerplate was needed. Turns out it wasn't, but now you explained why it's there in the first place :)
Will roll back

src/components/TradeWidget/index.tsx
```diff
@@ -403,7 +416,27 @@ const TradeWidget: React.FC = () => {
   const validFromValue = watch(validFromId)
   const validUntilValue = watch(validUntilId)
 
-  const initialPrice = useRef(priceParam)
+  // Updating global state, on change of each respective prop
+  useEffect(() => {
```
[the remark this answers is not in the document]
alfetopito: Yes, just wanted to be precise and update only what changed. Guess this optimization is not worth it then?

PR #748, opened 2020-03-24
Select with arrows

### LINKED ISSUE #741: Allow to use the arrows keys in the token selector
Normally, SELECTS allow to use the arrows keys (up/down key), to iterate over the options.

For example, in this example, as I user I would love to be able to select `TUSD` by:
- Input `us`. A filter is applied:
image

- Press down one time --> `USDT` is hightlighted:
image

- Press down another time --> `TUSD` is hightlighted
image

- Press INTRO --> `TUSD` is selected
image

Velenir:
`react-select` was already changing active option on `ArrowUp|Down`, it just wasn't visibly styled

This PR also takes care of the recurring issue of `scrollTop` which happens when `menuList` overflows and selected element changes from the last to first.
Can check the behavior if you press `ArrowDown` long enough in develop.

overflow

As I understand when `menuList` determines how much to scroll it takes measurement  of the whole `MenuList` components and count from Top, but as we added our search component there top got moved.

By moving our custom elements above the `boundingRect` of `MenuList` , default `MenuList` is once again flush with custom `MenuList` dimensions..

Closes #741

### DIFF
--- a/src/components/TokenSelector.tsx
+++ b/src/components/TokenSelector.tsx
@@ -136,6 +136,10 @@ const Wrapper = styled.div`
       @media ${MEDIA.mobile} {
         height: 5.6rem;
       }
+
+      &--is-focused {
+        background: rgba(33, 141, 255, 0.1);
+      }
     }
   }
 `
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr748--dexreact.review.gnosisdev.com

PR #751, opened 2020-03-24
Use last balance on error

### LINKED ISSUE #749: Issue with disappearing tokens
tokens is in the balance tab are suddenly disappearing.

image

image

image

NOTE:  Looks like a network error getting the balance that is poorly handled

Velenir:
When error happens during balance fetching, use last balance for that token

About refetching failed balance instead -- there'll be another balance fetch on newBlock in a few seconds anyway, so I don't think there's much need to spam requests.

Error can be simulated by blocking Metamask requests in devtools, only timing is difficult, need to block after `eth_blockNumber` request, during `eth_call` request for balance.

Screenshot from 2020-03-24 14-24-43
 Closes #749

### DIFF
--- a/src/hooks/useTokenBalances.ts
+++ b/src/hooks/useTokenBalances.ts
@@ -62,6 +62,17 @@ async function fetchBalancesForToken(
   }
 }
 
+const balanceCache: { [K: string]: TokenBalanceDetails } = {}
+interface CacheKeyParams {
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr751--dexreact.review.gnosisdev.com

PR #756, opened 2020-03-25
Update FAQ.md

c3rnst:
slight changes in wording.

### DIFF
--- a/src/assets/md/FAQ.md
+++ b/src/assets/md/FAQ.md
@@ -16,20 +16,23 @@ The Gnosis Protocol smart contracts have been audited externally. Check this sec
 
 **How do I trade on Mesa?**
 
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr756--dexreact.review.gnosisdev.com

Velenir:
Interestingly markdown-loader seems to break on `<br>`
So need to make space between paragraphs:
```
// NO
**Users must also have at least one of the tokens of their liquidity provision deposited in their Exchange Wallet to enable trades.**    
Users do not pay gas for the execution of trades. After the initial setup, no maintenance is required, other than cancellation. 

//YES
**Users must also have at least one of the tokens of their liquidity provision deposited in their Exchange Wallet to enable trades.**

Users do not pay gas for the execution of trades. After the initial setup, no maintenance is required, other than cancellation.

PR #760, opened 2020-03-25
Add any token

Velenir:
Preliminary work for adding any token
For now sets up logic to add token to list or exchange contract
~Then encounters a circular dependency and crashes. :rofl:~


gnosis-info:
Travis automatic deployment:
 https://pr760--dexreact.review.gnosisdev.com

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
+
     networkIds.forEach(networkId => {
       // initial value
-      this._tokensByNetwork[networkId] = getTokensByNetwork(networkId)
+      const tokenList = getTokensByNetwork(networkId).concat(this.loadUserTokenList(networkId))
+      this._tokensByNetwork[networkId] = tokenList
+
+      tokenList.forEach(({ address }) => {
+        this._tokenAddressNetworkSet.add(
+          TokenListApiImpl.constructAddressNetworkKey({ tokenAddress: address, networkId }),
```
alfetopito: Can't you use this for static methods?

```suggestion
          this.constructAddressNetworkKey({ tokenAddress: address, networkId }),
```

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
   public getTokens(networkId: number): TokenDetails[] {
     return this._tokensByNetwork[networkId] || []
   }
+
+  private static constructAddressNetworkKey({ tokenAddress, networkId }: HasTokenParams): string {
+    return tokenAddress.toLowerCase() + '|' + networkId
+  }
+
+  private static getUserTokenListName(networkId: number): string {
+    return 'USER_TOKEN_LIST_' + networkId
```
alfetopito: Does it have to be static?

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
+  }
+  public addToken({ networkId, token }: AddTokenParams): void {
+    logDebug('Added new Token to userlist', token)
```
alfetopito:
```suggestion
    logDebug('[TokenListApi]: Added new Token to userlist', token)
```

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
     this.networkIds = networkIds
 
     // Init the tokens by network
     this._tokensByNetwork = {}
+    this._tokenAddressNetworkSet = new Set<string>()
+
     networkIds.forEach(networkId => {
       // initial value
-      this._tokensByNetwork[networkId] = getTokensByNetwork(networkId)
+      const tokenList = getTokensByNetwork(networkId).concat(this.loadUserTokenList(networkId))
```
alfetopito: This is something I'm dealing with in my dynamic tokens PR, so you may not need to change anything.
Just raising the point that, we can have duplicates here. If a token added locally is later added to list of tokens.

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
+    return 'USER_TOKEN_LIST_' + networkId
+  }
+  public addToken({ networkId, token }: AddTokenParams): void {
+    logDebug('Added new Token to userlist', token)
+
+    this._tokensByNetwork[networkId].push(token)
+    this._tokenAddressNetworkSet.add(
+      TokenListApiImpl.constructAddressNetworkKey({ tokenAddress: token.address, networkId }),
+    )
+    this.persistUserTokenList(token, networkId)
```
alfetopito: nit: The function name gives the impression that it would take a list as parameter

src/api/tokenList/TokenListApi.ts
anxolin: nit: why the word `USER`? since this is for the local storage, it's implied. Why not just `TOKENS_` + networkId?

src/api/tokenList/TokenListApi.ts
```diff
@@ -13,21 +26,67 @@ export interface TokenList {
...
+
   public getTokens(networkId: number): TokenDetails[] {
     return this._tokensByNetwork[networkId] || []
   }
+
+  private static constructAddressNetworkKey({ tokenAddress, networkId }: HasTokenParams): string {
+    return tokenAddress.toLowerCase() + '|' + networkId
+  }
+
+  private static getUserTokenListName(networkId: number): string {
```
anxolin: ubernit: Maybe `getLocalStorageKey`?

src/services/index.ts
```diff
@@ -19,6 +21,54 @@ export const getTokenFromExchangeByAddress = getTokenFromExchangeByAddressFactor
...
+
+export const addTokenToList = async ({ networkId, tokenAddress }: AddTokenToListParams): Promise<AddTokenResult> => {
+  const checkSumAddress = web3.utils.toChecksumAddress(tokenAddress)
+  const token = await getTokenFromExchangeByAddress({ networkId, tokenAddress: checkSumAddress })
+  if (token) {
+    logDebug('Added new Token to userlist', token)
+
+    tokenListApi.addToken({ token, networkId })
+  } else {
+    logDebug('Token at address', tokenAddress, 'not available in Exchange contract')
```
alfetopito:
```suggestion
    logDebug(`[services:addTokenToList] Token at address ${tokenAddress} not available in Exchange contract`)
```

src/services/index.ts
```diff
@@ -19,6 +21,54 @@ export const getTokenFromExchangeByAddress = getTokenFromExchangeByAddressFactor
...
+    tokenList: tokenListApi.getTokens(networkId),
+  }
+}
+
+export const addTokenToExchange = async ({
+  userAddress,
+  networkId,
+  tokenAddress,
+}: AddTokenToExchangeParams): Promise<AddTokenResult> => {
+  const checkSumAddress = web3.utils.toChecksumAddress(tokenAddress)
```
anxolin: Good you remembered to do this :)

src/api/tokenList/TokenListApi.ts
W3stside: what does it return if nothing is instantiated? would probably be `undefined` or would it error?

src/api/tokenList/TokenListApi.ts
W3stside: i think because he's using a `lambda` function (`=>`) this will be applied to the function itself not the calling entity, static or not

W3stside: the lamba being inside the static method being the problem

Velenir: can't
`this` is instance reference, static lives on the class, i.e. constructor function
```js
this.constructor === TokenListApiImpl
this.constructor.constructAddressNetworkKey === TokenListApiImpl.constructAddressNetworkKey
```

src/api/tokenList/TokenListApi.ts
Velenir: why not static? I don't use static enough, let me use static

Velenir: what instantiated? `networkId`? TS takes care that we only pass in `number`

src/api/tokenList/TokenListApi.ts
Velenir: that is an interesting point
not gonna do anything about it, though :stuck_out_tongue: 
not here

src/api/tokenList/TokenListApi.ts
Velenir: but we will have our own `TOKEN_LIST` later when we start fetching ids and persisting in localStorage
User should have its own list

src/api/tokenList/TokenListApi.ts
Velenir: We are inside the constructor, `this` is class instance, lambdas have `this` from lexical (parent) scope -- same classs instance
I don't understand the problem

Issue #757, opened 2020-03-25
Issues with the price


anxolin:
I think we have some issue with the price suggestion:

**Sometimes we get NaN as the amount**
![image](https://user-images.githubusercontent.com/2352112/77537396-50efc380-6e9e-11ea-89ba-bc6dd4c78443.png)
If the price is not good, maybe we just default to `0 / Infinity` and show `0` in the receive amount

**Sometimes the price is not updated**

1. Select DAI-WETH. You'll get sth like `0.00746 / 134.0482` 
2. Change to DAI-GUSD. The amount don't change at all

alfetopito:
> I think we have some issue with the price suggestion:
> 
> **Sometimes we get NaN as the amount**
> ![image](https://user-images.githubusercontent.com/2352112/77537396-50efc380-6e9e-11ea-89ba-bc6dd4c78443.png)
> If the price is not good, maybe we just default to `0 / Infinity` and show `0` in the receive amount
> 
> **Sometimes the price is not updated**
> 
>     1. Select DAI-WETH. You'll get sth like `0.00746 / 134.0482`
> 
>     2. Change to DAI-GUSD. The amount don't change at all

The issue is that GUSD doesn't have a value. It was never traded in the exchange it seems.

PR #762, opened 2020-03-25
Reseting prices when switching tokens even if estimation is null

alfetopito:
Part of #757 

Prices are reset even when a token pair has no price estimation.

### Note

This is no longer an issue. Leaving for future reference.

~One caveat is that one use case the price won't be updated, when it meets this conditions:~
- ~Price is set on the URL~
- ~One of the tokens in the URL doesn't have a price~
- ~The user picks different tokens, without changing the one without price.~

~Example:~
- ~Starting URL: `/trade/GUSD-SNX?price=10`~
- ~Token without a price: `GUSD`~

~1. Load the page~
~2. Switch `SNX` by any other token~

~Price will remain `10`~

~In that case, the price won't change UNTIL the _priceless_ token is swapped out. First time this happens price will work as expected.~

~3. Switch `GUSD` by any other token~

~Price will be updated~

~4. Switch back to `GUSD`~

~Price will be reset to `0`.~

---

~This happens because of the way the price is loaded and how we deal with prices coming from the URL.~

~We'll only start inserting the price estimation once we have a valid price, because on first load the price is always `null`.~

~And we want to avoid overwriting URL prices when they are present.~

~Any better ideas to get this working?~

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -26,7 +26,7 @@ import { usePlaceOrder } from 'hooks/usePlaceOrder'
 import { useQuery, buildSearchQuery } from 'hooks/useQuery'
 import useGlobalState from 'hooks/useGlobalState'
 import { savePendingOrdersAction, removePendingOrdersAction } from 'reducers-actions/pendingOrders'
-import { MEDIA } from 'const'
+import { MEDIA, PRICE_ESTIMATION_PRECISION } from 'const'
 
 import { tokenListApi } from 'api'
 
@@ -450,7 +450,7 @@ const TradeWidget: React.FC = () => {
 
       if (shouldUsePriceEstimation) {
         // Price estimation can be null. In that case, set the input to 0
-        const newPrice = priceEstimation ? priceEstimation.toFixed(5) : '0'
+        const newPrice = priceEstimation ? priceEstimation.toFixed(PRICE_ESTIMATION_PRECISION) : '0'
 
         setValue(priceInputId, newPrice)
         setValue(priceInverseInputId, invertPriceFromString(newPrice))
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -446,17 +446,17 @@ const TradeWidget: React.FC = () => {
       // Keep in mind that next time is when token selection is changed.
       // In that case, initial price doesn't matter anymore
       initialPrice.current = ''
+    }
 
-      logDebug('[TradeWidget] priceEstimation', priceEstimation.toString(10))
+    logDebug(`[TradeWidget] priceEstimation ${priceEstimation}`)
 
-      if (shouldUsePriceEstimation) {
-        const newPrice = priceEstimation.toFixed(5)
+    if (shouldUsePriceEstimation) {
+      const newPrice = priceEstimation ? priceEstimation.toFixed(5) : '0'
 
-        setValue(priceInputId, newPrice)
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr762--dexreact.review.gnosisdev.com

src/hooks/usePriceEstimation.ts
```diff
@@ -11,13 +11,24 @@ interface Params {
...
+  const [priceEstimation, setPriceEstimation] = useSafeState<BigNumber | null>(null)
 
   useEffect(() => {
-    getPriceEstimation({ baseTokenId, quoteTokenId }).then(setPrice, logDebug)
-  }, [quoteTokenId, baseTokenId, setPrice])
+    setIsPriceLoading(true)
+
+    getPriceEstimation({ baseTokenId, quoteTokenId }).then(price => {
+      setPriceEstimation(price)
+      setIsPriceLoading(false)
```
anxolin: No catch? you don't need to do a `setIsPriceLoading(false)` also if you fail. Also handle the error somehow?

src/components/TradeWidget/index.tsx
```diff
@@ -438,19 +438,19 @@ const TradeWidget: React.FC = () => {
...
+        initialPrice.current = ''
+      }
 
-      logDebug('[TradeWidget] priceEstimation', priceEstimation.toString(10))
+      logDebug(`[TradeWidget] priceEstimation ${priceEstimation}`)
 
       if (shouldUsePriceEstimation) {
-        const newPrice = priceEstimation.toFixed(5)
+        // Price estimation can be null. In that case, set the input to 0
+        const newPrice = priceEstimation ? priceEstimation.toFixed(5) : '0'
```
anxolin: Why `5` ?

alfetopito: Because reasons.

Seriously, how far do we want to go precision wise? Does it make sense to input a 18 decimals price?

src/hooks/usePriceEstimation.ts
alfetopito: Check the next line https://github.com/gnosis/dex-react/pull/762/files#diff-c47f9093788861eb688116377dd46937R30

But I get your point.
Will address

src/components/TradeWidget/index.tsx
anxolin: No, I think it's fine for the estimation, but maybe a const would be good. `5` look like a magic number. Just a nit

src/hooks/usePriceEstimation.ts
alfetopito: Done undoing @Velenir 's improvement suggestion

src/components/TradeWidget/index.tsx
alfetopito: Makes sense, done.

src/hooks/usePriceEstimation.ts
```diff
@@ -22,12 +22,18 @@ export function usePriceEstimation(params: Params): Result {
...
-    getPriceEstimation({ baseTokenId, quoteTokenId }).then(price => {
-      setPriceEstimation(price)
-      setIsPriceLoading(false)
-    }, logDebug)
+    async function estimatePrice(): Promise<void> {
+      setIsPriceLoading(true)
+      try {
+        setPriceEstimation(await getPriceEstimation({ baseTokenId, quoteTokenId }))
+      } catch (e) {
+        logDebug(e)
```
anxolin:
```suggestion
        console.error('Error getting price estimation', e)
```

PR #779, opened 2020-03-27
Track token list updates

Velenir:
Adds subscription to TokenListApi so that updates propagate to everywhere we need tokenList


gnosis-info:
Travis automatic deployment:
 https://pr779--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -356,6 +356,8 @@ const TradeWidget: React.FC = () => {
 
   // If user is connected, use balances, otherwise get the default list
   const tokens = useMemo(
+    // it's okay to tokenListApi.getTokens() here without subscribing to updates
+    // because balances from useTokenBalances is already subscribed
     () => (isConnected && balances.length > 0 ? balances : tokenListApi.getTokens(fallBackNetworkId)),
```
alfetopito: But, balances is only used/updated if wallet is connected.
Won't we allow tokens to be added on trade page without having connected the wallet?

I guess that is fine for adding tokens, but can be a problem due to the way I'll load the tokens. (can probably address on my PR, though)

src/hooks/useTokeList.ts
```diff
@@ -0,0 +1,19 @@
...
+// for stable reference
+// to avoid updates on setState([])
+const emptyArray: TokenDetails[] = []
+
+export const useTokenList = (networkId?: number): TokenDetails[] => {
+  const [tokens, setTokens] = useSafeState(networkId === undefined ? emptyArray : tokenListApi.getTokens(networkId))
+
+  useEffect(() => {
+    if (networkId === undefined) return setTokens(emptyArray)
+    return tokenListApi.subscribe(setTokens)
```
alfetopito: :thinking: 
When the hook umounts, shouldn't it return `tokenListApi.unsubscribe`?

src/services/index.ts
```diff
@@ -7,7 +7,7 @@ import {
   getPriceEstimationFactory,
 } from './factories'
 import { logDebug } from 'utils'
-import { TokenDetails } from '@gnosis.pm/dex-js'
+import { TokenDetails } from 'types'
```
alfetopito: Are they different?

src/api/tokenList/Subscriptions.ts
```diff
@@ -0,0 +1,32 @@
...
+type SubscriptionCallback<T> = (newState: T) => void
+
+export interface SubscriptionsInterface<State> {
+  subscribe(callback: SubscriptionCallback<State>): Command
+  unsubscribe(callback: SubscriptionCallback<State>): void
+  unsubscribeAll(): void
+  triggerSubscriptions(newState: State): void
+}
+
+export default class GenericSubscriptions<State> implements SubscriptionsInterface<State> {
```
anxolin: One idea. Why don't you create a new interface `Subscriptions<State>` and then do `GenericSubscriptions` implement this interface plus add the public `triggerSubscriptions` method.

This way TokenList would extend Subscriptions, and the Impl/Mock would extend GenericSubscriptions

src/hooks/useTokenBalances.ts
```diff
@@ -12,6 +12,7 @@ import { ALLOWANCE_FOR_ENABLED_TOKEN } from 'const'
 import { TokenBalanceDetails, TokenDetails } from 'types'
 import { WalletInfo } from 'api/wallet/WalletApi'
 import { PendingFlux } from 'api/deposit/DepositApi'
+import { useTokenList } from './useTokeList'
```
W3stside: `useTokeList` typo error

W3stside: unless it's a list of you smoking doobies idk

src/components/TradeWidget/index.tsx
Velenir: ugh, you are right
having default network-less tokenList is a problem

src/hooks/useTokeList.ts
Velenir:
```ts
public subscribe(callback: SubscriptionCallback<State>): Command {
    this._subscriptions.push(callback)

    return (): void => this.unsubscribe(callback)
  }
```

src/services/index.ts
Velenir:
```ts
//dex-js
export interface TokenDetails {
    id: number;
    name?: string;
    symbol?: string;
    decimals?: number;
    address: string;
    addressMainnet?: string;
    image?: string;
}

//dex-react
export interface MinimalTokenDetails {
  address: string
  symbol?: string
  name?: string
  decimals: number
}

export interface TokenDetails extends MinimalTokenDetails {
  id: number
  addressMainnet?: string
  image?: string
}
```

`decimals` is required in dex-react
which probably doesn't matter much, but TS was complaining

src/api/tokenList/Subscriptions.ts
Velenir: Can't extend interface
Can only extend class `GenericSubscriptions`, so `triggerSubscriptions` becomes public, so we remove it from interface with 
```ts
export interface TokenList extends Omit<SubscriptionsInterface<TokenDetails[]>, 'triggerSubscriptions'>
```

But you gave me an idea, I'll make it `protected`, 6abef37b3728105c8fe616875be650b913c81729

Velenir:
I'll merge this and #760 to unblock @alfetopito

src/hooks/useTokeList.ts
alfetopito: :brainfreeze:

Issue #708, opened 2020-03-12
Control better the error where a token is not listed


anxolin:
If an user try to use a token that is not listed, right now we show this error:

![image](https://user-images.githubusercontent.com/2352112/76507207-acf12b80-644c-11ea-8672-bb1d95dbead4.png)

Ideally we would present a modal or something inviting the user to list it himself.
This modal should contain a link to the devportal:
https://docs.gnosis.io/dfusion/docs/addtoken1/

cc/ @c3rnst

PR #780, opened 2020-03-27
Add token in TokenSelector

Velenir:
Example of adding a new token through TokenSelector with minimal implementation just to see if it works
Try adding `0x45804880de22913dafe09f4980848ece6ecbaf78` on mainnet.
Not rinkeby because of id overlap

1. Open TokenSelector
2. Input token address
3. Press enter
4. Accept dialog
5. _Presto_, new token

ezgif-2-3f0bfaa8a1ee

Honestly, I don't want to deal with react-select/Creatable.


gnosis-info:
Travis automatic deployment:
 https://pr780--dexreact.review.gnosisdev.com

src/components/TokenSelector.tsx
```diff
@@ -331,17 +366,33 @@ const TokenSelector: React.FC<Props> = ({ isDisabled, tokens, selected, onChange
...
+      } else if (e.key === 'Enter' && networkId && isAddress(e.target.value.toLowerCase())) {
+        const tokenAddress = e.target.value
+        if (tokenListApi.hasToken({ tokenAddress, networkId })) return
+
+        // prevents double-catch of this event
+        // double because it's captured from onKeyDown in MenuList
+        // and in general on Select, I guess
+        e.stopPropagation()
+        console.log('tokenAddress', tokenAddress)
+        if (window.confirm(`Do you want to add ${tokenAddress} to your local tokneList`)) {
```
alfetopito:
```suggestion
        if (window.confirm(`Do you want to add ${tokenAddress} to your local list of tokens?`)) {
```

src/components/TokenSelector.tsx
```diff
@@ -331,17 +366,33 @@ const TokenSelector: React.FC<Props> = ({ isDisabled, tokens, selected, onChange
...
-    }
-  }, [])
+  const onKeyDown = useCallback(
+    (e: React.KeyboardEvent<HTMLInputElement> & { target: HTMLInputElement }): void => {
+      if (e.key === 'Escape') {
+        // Close menu on `Escape`
+        e.stopPropagation()
+        setIsFocused(false)
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      } else if (!e.target.value! && e.key === ' ') {
```
alfetopito: Why the `!` on `.value!`?

alfetopito:
Can't we use Modali for the confirmation here?

src/components/TokenSelector.tsx
W3stside: forces TS to accept it as not `falsy`

src/components/TokenSelector.tsx
```diff
@@ -331,17 +366,33 @@ const TokenSelector: React.FC<Props> = ({ isDisabled, tokens, selected, onChange
...
+        // That closes the menu. (/shrug)
+        e.preventDefault()
+      } else if (e.key === 'Enter' && networkId && isAddress(e.target.value.toLowerCase())) {
+        const tokenAddress = e.target.value
+        if (tokenListApi.hasToken({ tokenAddress, networkId })) return
+
+        // prevents double-catch of this event
+        // double because it's captured from onKeyDown in MenuList
+        // and in general on Select, I guess
+        e.stopPropagation()
```
W3stside: oof all these stopPropagations etc are getting messy. This is a `react-select` problem I'm assuming?

Velenir: yes, it handles event like a blackbox, so these issues pop up from time to time

src/components/TokenSelector.tsx
Velenir: `window.confirm` is going away anyway

src/components/TokenSelector.tsx
Velenir: TS complained at some point, but not anymore, removing

src/components/TokenSelector.tsx
```diff
@@ -298,6 +304,26 @@ const components = { MenuList }
 
 const noOptionsMessage = (): string => 'No results'
 
+const addTokenFromInput = async ({ networkId, tokenAddress }: AddTokenToListParams): Promise<boolean> => {
+  try {
+    const tokenAddedResult = await addTokenToList({ networkId, tokenAddress })
+
+    if (!isTokenAddedSuccess(tokenAddedResult)) {
+      toast.warn(tokenAddedResult.error)
```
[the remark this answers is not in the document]
anxolin: Related: https://github.com/gnosis/dex-react/issues/708#issue-579806745

src/utils/deferred.ts
```diff
@@ -0,0 +1,24 @@
...
+export type PromiseReject = (reason?: unknown) => void
+
+export interface Deferred<T> {
+  promise: Promise<T>
+  resolve: PromiseResolve<T>
+  reject: PromiseReject
+}
+
+export const createDeferredPromise = <T>(): Deferred<T> => {
+  let resolve: PromiseResolve<T> = () => void 0
```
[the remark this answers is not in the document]
Velenir: `void <anything> === undefined` `void 0` is just convention, TS was complaining about empty arrow function `() => {}`

Velenir:
> 3\. When not connected, I'm not able to add the example token. I get `token not in the exchange` message.

#812

Velenir:
> Probably separate PR, but would it be very difficult to show a preview the token before you hit enter, as Uniswap does?

That's rather hard because the `noMessage` that is rendered there is not very controlled and supposed to be a string

Velenir:
> 2\. As a user, I have no way to know that when I put an address in the token selector I can add it when I hit enter. Should we update the `No results` message to make that clear, or something else?

Can do, probably, would be a terrible hack though. `noOptionsMessage` is supposed to be a string. And it doesn't get when you would expect

Velenir:
I'll merge this and other related PRs into #780 to have everything in one place
It's becoming unmanageable otherwise

PR #782, opened 2020-03-28
695/add schema validation trade page

### LINKED ISSUE #672: validUntil & validFrom Modal - should we disable button on invalid amounts?
e.g `validFrom` must be `>= 15` -- should we disable button on incorrect values?

### LINKED ISSUE #695: Do not allow negative numbers in some inputs
Do not allow negative numbers in some inputs: i.e. Start date in the trade

image

### LINKED ISSUE #750: Order Starts In (min)
First issue is that if you set a value that is smaller than 15 it should show a validation error. It doesn't do that, and  it allows to press the "Set order parameter" in that state:

image

Then, if you try to send the transaction, it won't, and it won't give the user any feedback.

I also find extrange how the input doesn't tell you anything if you for example set `14`, and the close the modal:
image

Then, if you  check the value (open the modal again), you'll see that the value was magically changed into `15`. 

One idea for the message, would be something like:
`Select a start time multiple of 5 since batches last 5min. Quick fix: Set 15min`

I'm sure we can come up with a better message anyways :)

W3stside:
Closes #750 and closes #672  AND closes #695  [THREE-FER @Velenir]

# Add schema validation into trade page
<bl />

### Minor, quasi-unrelated changes:
1. TokenRow - uses schema errors
2. validFrom/Until uses `null` instead of `0` to indicate ASAP/Never
3. `resolverFactory` no longer casts values - validator handles this. Converts `""` to `undefined` (to handle ASAP/Never cases)


gnosis-info:
Travis automatic deployment:
 https://pr782--dexreact.review.gnosisdev.com

src/utils/validation.ts
```diff
@@ -61,6 +61,9 @@ export const stringOrNumberResolverFactory = <FormData, CustomContext = {}>(
 
 export function formatSchemaErrorMessage(errorString?: string): string | undefined {
   if (!errorString) return undefined
-  const cleanedString = errorString.replace(/"/g, '')
-  return cleanedString[0].toUpperCase() + cleanedString.slice(1)
+  const cleanedString = errorString
+    .split(' ')
+    .slice(1)
+    .join(' ')
```
Velenir: This looks strange. What `errorString` do you get for input and what do you expect as output?

src/utils/validation.ts
```diff
@@ -27,23 +27,23 @@ interface FormDataAsNumbers {
...
+export const resolverFactory = <FormData, CustomContext = {}>(validationSchema: ObjectSchema<unknown>) => (
+  data: FormData,
+): ReturnType<ValidationResolver<FormData, CustomContext>> => {
+  const castedData: FormDataAsNumbers | FormData = Object.keys(data).reduce<FormDataAsNumbers>((acc, key) => {
+    const oldValue = data[key]
+    const castedValue = oldValue ? oldValue : undefined
+    const castedObj = { ...acc, [key]: castedValue }
+
+    return castedObj
+  }, {})
```
Velenir: What happens here?
```
data = {a: '', b: 0, c: false, e: Truthy}
castedData = {a: undefined, b: undefined, c: undefined, : Truthy}
```
Is that what you want? How does validation treat it? Skips undefined keys?

src/utils/validation.ts
```diff
@@ -27,23 +27,23 @@ interface FormDataAsNumbers {
...
-      : data
+export const resolverFactory = <FormData, CustomContext = {}>(validationSchema: ObjectSchema<unknown>) => (
+  data: FormData,
+): ReturnType<ValidationResolver<FormData, CustomContext>> => {
+  const castedData: FormDataAsNumbers | FormData = Object.keys(data).reduce<FormDataAsNumbers>((acc, key) => {
+    const oldValue = data[key]
+    const castedValue = oldValue ? oldValue : undefined
+    const castedObj = { ...acc, [key]: castedValue }
+
+    return castedObj
```
Velenir:
```suggestion
    acc[key] = data[key] || undefined
    return acc
```

src/utils/validation.ts
W3stside: falsy = explicit undefined, was some issue with `FormData` taking undefined values and setting as `""` which `schemaValidation` wasn't treating as `undefined`

src/utils/validation.ts
W3stside: It is exactly: `"{{keyName}}" {{restOfErrorMessage}}`

e.g `"sellToken value must be positive"`

which is ugly

Velenir: then
```suggestion
  const cleanedString = errorString.replace(/^"\w+"\s+/, '')
```

W3stside: no like i dont want the key name at all

Velenir:
```
`"sellToken" value must be positive`.replace(/^"\w+"\s+/, '') === "value must be positive"

W3stside:
> Approved with same comments as Leandro.
> 
> On top of that, a small nit, I feel it's strange to display the name of the field like this:
> image
> 
> It should be plain english, either:
> 
>     * The price must be greater than 0
> 
>     * .. or probaly "must be greater than 0"
> 
>     * .. or maybe "Enter a valid price" since if you get that, and you have a 0, it's not difficult to understand and it sounds more natural
> 
> 
> Same applies to this messages:
> image

I addresses the keyName and quotes part in the comments but I will make further changes in another PR - for now `Value` replaces all key names

alfetopito:
The prices are no longer in sync. I'm able to insert anything and it never adjusts the opposite price:

screenshot_2020-04-01_07-39-28

W3stside:
> The prices are no longer in sync. I'm able to insert anything and it never adjusts the opposite price:
> 
> screenshot_2020-04-01_07-39-28

fixed in #810

PR #786, opened 2020-03-30
Tooltip text change

Rafanator:
- Changed text to reflect more clearly that users need to deposit. 

@anxolin Is this the correct file to do this? 
cc/ @c3rnst

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -181,7 +181,7 @@ interface Props {
 
 const BalanceTooltip = (
   <HelpTooltipContainer>
-    This balance reflects the amount deposited in the Exchange Wallet on Mesa, not the overall amounts in your wallet.
+    This balance reflects the amount deposited in the Exchange Wallet on Mesa. Only orders with a balance will be considered for matching.
   </HelpTooltipContainer>
 )
 


src/components/TradeWidget/TokenRow.tsx
```diff
@@ -181,7 +181,7 @@ interface Props {
 
 const BalanceTooltip = (
   <HelpTooltipContainer>
-    This balance reflects the amount deposited in the Exchange Wallet on Mesa, not the overall amounts in your wallet.
+    This balance reflects the amount deposited in the Exchange Wallet on Mesa. Only orders with a balance will be considered for matching.
```
c3rnst: I think now we have "matching" on top of words like settlement etc. In the balance page, we also call it Exchange Balance (and not wallet).
How about 
"This balance reflects your amount available in the Exchange. You can deposit on the Balances page. Only orders with a balance can be considered for settlement."

Rafanator: I would keep matching. There is a difference between matching and settlement. 

**- Matching:** two orders being matched, or a trade being executed  
**- Settlement:** is a part of post-trade processing, what happens after a trade is executed. (Change records of ownership, transfer of assets, etc)

Reference: https://www.investopedia.com/terms/p/post-trade-processing.asp 

I would say we go for: `"This balance reflects your amount available in the Exchange. Only orders with an Exchange balance can be considered for matching."`

I would go against adding `...You can deposit on the Balances page.` As I expect this to be apparent  and makes it too long.

c3rnst: I agree with this. But before you said that the status of an order in a batch should be titled "waiting for settlement". I don't think there should be different words used. We can change it to "waiting for match".

Rafanator: Sure. Waiting for match is more accurate.

c3rnst: I would still think that "You can deposit on the Balances page" - in light of all the user feedback we got - is important?

PR #787, opened 2020-03-30
Add dune link, fix a broken link, and include smart contract addresses

anxolin:
image

Adds a link to dune analytics
Also, link to the smart contracts

As a bonus I had to fix the Ethescan link component, cause the `networkId` was being used as a default, instead of being set as the network you want. So you could not use a different network than the one you are connected with your wallet. 
I reviewed and we were not using the `networkId` param anywere, so I think it's safe to change the behaviour. I added an aditional optional parameter for `networkIdDefault` in case we need the old behaviour

### DIFF
--- a/src/components/EtherscanLink.tsx
+++ b/src/components/EtherscanLink.tsx
@@ -9,6 +9,7 @@ export interface EtherscanLinkProps {
   type: EtherscanLinkType
   identifier: string
   networkId?: number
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr787--dexreact.review.gnosisdev.com

PR #793, opened 2020-03-31
Merging both lists of tokens to avoid duplicates

alfetopito:
List of tokens were being concatenated with previously saved list of tokens from localStorage, making it quite large...

Here I do a simple 'merge' by only using unique addresses and ignoring repeated entries.

### DIFF
--- a/src/api/tokenList/TokenListApi.ts
+++ b/src/api/tokenList/TokenListApi.ts
@@ -40,7 +40,7 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
 
     networkIds.forEach(networkId => {
       // initial value
-      const tokenList = getTokensByNetwork(networkId).concat(this.loadUserTokenList(networkId))
+      const tokenList = this.mergeTokenLists(getTokensByNetwork(networkId), this.loadUserTokenList(networkId))
       this._tokensByNetwork[networkId] = tokenList
 
       tokenList.forEach(({ address }) => {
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr793--dexreact.review.gnosisdev.com

src/api/tokenList/TokenListApi.ts
```diff
@@ -59,6 +59,20 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
...
   }
 
+
+  private mergeTokenLists(baseList: TokenDetails[], newList: TokenDetails[]): TokenDetails[] {
+    const seenAddresses = new Set<string>()
+    const result: TokenDetails[] = []
+
+    baseList.concat(newList).forEach(token => {
+      if (!seenAddresses.has(token.address.toLocaleLowerCase())) {
+        seenAddresses.add(token.address.toLocaleLowerCase())
```
Velenir: Why `Locale`? No need, `toLowerCase` is enough

src/api/tokenList/TokenListApi.ts
```diff
@@ -59,6 +59,20 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
     return this._tokensByNetwork[networkId] || []
   }
 
+
+  private mergeTokenLists(baseList: TokenDetails[], newList: TokenDetails[]): TokenDetails[] {
```
Velenir: This could be a `static` method. Not that it need to be, but maybe should. Does it have to? No. Ought to? Eh....
And other modal verbs :expressionless:

src/api/tokenList/TokenListApi.ts
alfetopito: Oops, auto-complete late in the evening sleepy coding. Thanks for catching it :)

src/api/tokenList/TokenListApi.ts
alfetopito: ok ok, I know you want static. I'll give you `static`.

Issue #791, opened 2020-03-30
Good to have: Order book


anxolin:
We'll incorporate an order book as a pop up in the trade widget.

@fleupold will move the code to
https://6jxh4.codesandbox.io/

![image](https://user-images.githubusercontent.com/2352112/77945677-4e87d200-72c1-11ea-83b6-f3185808a6f7.png)

The original idea was to do a standalone app for this, but Felix prefers to add it as a page 

The original idea was:
* Make an independent project
* Have it’s own URL
* Expose a page with a minimal UI so anyone can use it with iframes. It would take the tokens as parameters
* Expose a normal UI, with the token selector: entry point for the web. Just minimal design, with a title, the selector, etc..
* We could try to make a popup with an iframe to it in Mesa, so it opens up the minimal UI for the selected token pair

I'm fine with Felix proposal too. We can make the placeholder for him: A blank page, with a blank widget where he'll encapsulate this graph. This component will have as parameter the 2 tokens. 
We could add the token selector in this page, to facilitate his development. So he would start with a blank component that has preselected 2 tokens.

Does this work?

anxolin:
cc: @alfetopito @Velenir @W3stside

fleupold:
Works for me. Let me know where to put the code. It can be found here: https://codesandbox.io/s/orderbook-centralized-czii9

anxolin:
I've made this PR with the placeholders so you can add it there.
https://github.com/gnosis/dex-react/pull/800

PR #800, opened 2020-03-31
Add order book page

anxolin:
Part of #791

I would like to create as mentioned in #791  a placeholder so Felix can work on integrating his order book.

his PR includes:
- A new page with the selectors logic
- A new component for the Otherbook
- Add a wrapper with some example CSS --> just to show how to add CSS, but Felix pls remove the CSS in the next PR

image

Test it by navigating to `/book` address:
* http://localhost:8080/book
* https://pr800--dexreact.review.gnosisdev.com/book

### DIFF
--- a/src/pages/OrderBook.tsx
+++ b/src/pages/OrderBook.tsx
@@ -1,0 +1,45 @@
+import React, { useEffect } from 'react'
+import { ContentPage } from 'components/Layout/PageWrapper'
+import OrderBookWidget from 'components/OrderBookWidget'
+import TokenSelector from 'components/TokenSelector'
+import { useTokenList } from 'hooks/useTokenList'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import { Network } from 'types'
+import useSafeState from 'hooks/useSafeState'
+import { TokenDetails } from 'types'
+
+const OrderBook: React.FC = () => {
+  const { networkId } = useWalletConnection()
+  const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
+  const tokenList = useTokenList(fallBackNetworkId)
+  const [baseToken, setBaseToken] = useSafeState<TokenDetails | null>(null)
+  const [quoteToken, setQuoteToken] = useSafeState<TokenDetails | null>(null)
+
+  const tokensLoaded = tokenList.length !== 0
+  useEffect(() => {
+    if (tokensLoaded) {
+      if (baseToken === null) {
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr800--dexreact.review.gnosisdev.com

src/pages/OrderBook.tsx
```diff
@@ -0,0 +1,45 @@
...
+import TokenSelector from 'components/TokenSelector'
+import { useTokenList } from 'hooks/useTokenList'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import { Network } from 'types'
+import useSafeState from 'hooks/useSafeState'
+import { TokenDetails } from 'types'
+
+const OrderBook: React.FC = () => {
+  const { networkId } = useWalletConnection()
+  const fallBackNetworkId = networkId ? networkId : Network.Mainnet // fallback to mainnet
```
alfetopito: Do we need to show prices, or even, does the tool have prices for anything that's not Mainnet?

fleupold:
Should I add the code to this PR or create a new one on top?

src/pages/OrderBook.tsx
anxolin: First version Felix did was using pure web3, so only that was needed (that, and the batchexchange contract and the viewer)

I think he is using now his own service, and I believe he also has two environments (rinkeby and mainnet), so probably yes, I think we should show the current network, and mainnet as a default

anxolin:
> Should I add the code to this PR or create a new one on top?

One on top is better

src/pages/OrderBook.tsx
fleupold: Yes, I'm querying the backend service and need network info to form the correct URL

c3rnst:
cool! from where will the order book be accessible on the interface? I have some concerns that it will increase confusion on the site: first we say it's not one order book per pair and then we present it :) but in any case, good to have the info.

anxolin:
> cool! from where will the order book be accessible on the interface? I have some concerns that it will increase confusion on the site: first we say it's not one order book per pair and then we present it :) but in any case, good to have the info.

@c3rnst  This page won't be visible probably (navigable). This is just for isolating a component Felix will work on. 

This component might be shown for example, as a popup in the trading widget if the user click in one suggested price or a link we make for showing the order book. 
Would that work?

c3rnst:
> Would that work?

anxolin:
@fleupold I've merged this one. Point yours to develop if you want

PR #801, opened 2020-03-31
First attempt at a rate limiter api

alfetopito:
First feeble attempt at a rate limiter implementation.

Failing miserably so far.

The interval and the actual method calls seem to be in a different context.

For example, when the interval clears (1s), the value is always `0`. (`Resetting count. Current 0` log entries)
When the method is called, the counter keeps incrementing, and is never reset (`x < 5` log entries):

screenshot_2020-03-31_14-07-39

- I added the rate limiter to some contract read only methods. Not extensive and through yet.
- Still need some work on the interface typing and so on.
- Build probably failing due to the use of `any`.

Looking for:
1. Opinions on this interface
2. Suggestions on the apparently different context for instance variables

### DIFF
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -138,7 +138,8 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
     )
 
     // query 1 more than required to check whether there's a next page
-    const encodedOrders = await contract.methods.getEncodedUserOrdersPaginated(userAddress, offset, pageSize + 1).call()
+    const promise = contract.methods.getEncodedUserOrdersPaginated(userAddress, offset, pageSize + 1).call()
+    const encodedOrders = await this.rateLimiterApi.call(promise)
 
     // is null if Contract returns empty bytes
     if (!encodedOrders) return { orders: [] }
@@ -176,12 +177,16 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
 
   public async getTokenAddressById({ tokenId, networkId }: GetTokenAddressByIdParams): Promise<string> {
     const contract = await this._getContract(networkId)
-    return contract.methods.tokenIdToAddressMap(tokenId).call()
+
+    const promise = contract.methods.tokenIdToAddressMap(tokenId).call()
+    return this.rateLimiterApi.call(promise)
   }
 
   public async getTokenIdByAddress({ tokenAddress, networkId }: GetTokenIdByAddressParams): Promise<number> {
     const contract = await this._getContract(networkId)
-    const tokenId = await contract.methods.tokenAddressToIdMap(tokenAddress).call()
+
+    const promise = contract.methods.tokenAddressToIdMap(tokenAddress).call()
+    const tokenId = await this.rateLimiterApi.call(promise)
     return +tokenId
   }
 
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr801--dexreact.review.gnosisdev.com

src/api/exchange/ExchangeApi.ts
```diff
@@ -176,12 +177,16 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
 
   public async getTokenAddressById({ tokenId, networkId }: GetTokenAddressByIdParams): Promise<string> {
     const contract = await this._getContract(networkId)
-    return contract.methods.tokenIdToAddressMap(tokenId).call()
+
+    const promise = contract.methods.tokenIdToAddressMap(tokenId).call()
```
anxolin, line 181: You are actually doing the call here, you don't rate limit if you execute first the call and get the promise already. Right?

Typically the concepts related to this is called debounce or throttling. In this case we would need the later. I think we are overcomplicating it by creating an API and a custom implementation for it.
I believe, it would be just to define a util, and this util may just delegate in a well tested npm library that just do that.

Making a quick search I found this https://github.com/sindresorhus/p-throttle simple, but I'm sure there will be many options.

A simple interface of this function, would just accept a function and some basic params, and would return the throttled version of this function. So if you call it, it will delay slightly the call.

Having done this function, maybe it's worth it to add a throttle in a more central place, like close to your web3 provider that constrict you. This way, all calls will be throttling using the same timers, no matter which API is using them

Additionally, another idea worth thinking about that might come handy is some kind of subprovider of composition of web3. Something like this https://www.npmjs.com/package/web3-provider-engine uses the "chain of responsibility pattern" that might come handy

Here are some, random thought :)

Velenir: Use https://lodash.com/docs/4.17.15#throttle or something
No need to make a whole class for this

anxolin: @Velenir that's my suggestion too, to define just a function. But regarding lodash, do you think we can use JUST the throttle version or you are suggesting to add all the lib for just that function

Velenir:
```js
import throttle from 'lodash/throttle'
```
should work just fine

alfetopito: > You are actually doing the call here, you don't rate limit if you execute first the call and get the promise already. Right?

No, the contract `call()` returns a promise, and I'm not executing it there.

I looked around for a lib, but given that we are lib averse recently decided to write a simple api (that clearly did not work :D), to avoid adding tons of dependencies.

Will give lodash throttle a try then.

> Having done this function, maybe it's worth it to add a throttle in a more central place, like close to your web3 provider that constrict you. This way, all calls will be throttling using the same timers, no matter which API is using them

That is my thinking as well, but I guess you mean to add it like a wrapper to any web3 call?
I didn't want to have the rate limiter for user facing calls like placing orders, and have it for background fetching calls.

Will give this all a try.

alfetopito: Maybe I'm trying to solve something that isn't the problem. See https://github.com/gnosis/dex-react/pull/815

alfetopito:
Turns out not needed as of yet.

PR #813, opened 2020-04-01
[Orderbook] Rendering the transitive orderbook

fleupold:
This PR pastes the code from https://czii9.csb.app/ into the Orderbook widget. The orderbook is based on https://www.amcharts.com/docs/v4/ and using the gnosis hosted price-estimation service to get data.

## Test Plan

compare http://localhost:8080/book to https://czii9.csb.app/ and make sure they look a like,

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -1,6 +1,10 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import styled from 'styled-components'
 import { TokenDetails } from 'types'
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import * as am4core from '@amcharts/amcharts4/core'
+import * as am4charts from '@amcharts/amcharts4/charts'
+import am4themesSpiritedaway from '@amcharts/amcharts4/themes/spiritedaway'
 
 interface OrderBookProps {
   baseToken: TokenDetails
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr813--dexreact.review.gnosisdev.com

Velenir:
Some suggestions #819

src/components/OrderBookWidget.tsx
```diff
@@ -1,31 +1,180 @@
...
+          //    Default            Bids          Asks
+          //            |      |                        |
+          //   ---------       ---------       ---------
+          //  |                         |      |
+          //       x                    x      x
+          //
+          // For asks, we can offset the "startLocation" by 0.5. However, Amcharts does not support a "startLocation" of -0.5.
+          // For bids, we therefore offset the curve by -1 (expose the previous total volume) and use an offset of 0.5.
+          // Otherwise our steps would be off by one.
+          askValueY: type == Offer.Bid ? previousTotalVolume : null,
```
anxolin, line 93: This can be moved up to a `isBid` const so you don't reevaluate for every order
the `isAsk` can be defined as `!isBid` right?

src/components/OrderBookWidget.tsx
```diff
@@ -1,31 +1,180 @@
...
+  chart.cursor = new am4charts.XYCursor()
+  return chart
+}
+
 const OrderBookWidget: React.FC<OrderBookProps> = props => {
-  const { baseToken, quoteToken } = props
+  const { baseToken, quoteToken, networkId } = props
+  const mountPoint = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
```
anxolin, line 168: well done! I see you go the hang of it fast :) 
also, very well the `useRef`

alfetopito:
I'm getting this now:

screenshot_2020-04-03_09-02-36

fleupold:
I'm in the process of updating the endpoint. Will ping you here when the PR is based on the new URL

src/components/OrderBookWidget.tsx
fleupold: @Velenir did that for me 😄

src/components/OrderBookWidget.tsx
```diff
@@ -9,23 +13,162 @@ interface OrderBookProps {
...
+}
+
+const draw = (baseToken: TokenDetails, quoteToken: TokenDetails, dataSource: string): am4charts.XYChart => {
+  am4core.useTheme(am4themesSpiritedaway)
+  const chart = am4core.create('chartdiv', am4charts.XYChart)
+
+  // Add data
+  chart.dataSource.url = dataSource
+  chart.dataSource.adapter.add('parsedData', function(data) {
+    data = JSON.parse(data)
```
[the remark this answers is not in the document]
fleupold: It did allow me, but this should no longer be necessary.

src/components/OrderBookWidget.tsx
```diff
@@ -9,23 +13,162 @@ interface OrderBookProps {
...
+  type: Offer,
+): ProcessedItem[] => {
+  let totalVolume = 0
+  return (
+    list
+      // Account fo decimals
+      .map(element => {
+        return {
+          price: element.price / 10 ** (quoteToken.decimals - baseToken.decimals),
+          volume: element.volume / 10 ** baseToken.decimals,
```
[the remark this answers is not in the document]
fleupold: I think it should be big enough (JS float double precision, right?) There might be some precision loss due to rounding but I think that's ok. Order amounts can only have 128 bits.

fleupold:
Ok, it should be displayable again now. There is still a bug that when you reload the page multiple times the graph changes. @alfetopito this is why you saw the scales not match when comparing to the code sandbox.

Waiting on review of https://github.com/gnosis/dex-price-estimator/pull/11 and then it should just work™️ 🤞

alfetopito:
> Ok, it should be displayable again now. 

Still the same:

screenshot_2020-04-03_12-43-03

Could it be related to the query? `?base=1"e=2`

fleupold:
It didn’t rebuild the deployed version I pushed my changes. Is this screenshot from your local checkout?

alfetopito:
> It didn’t rebuild the deployed version after I pushed my changes. Is this screenshot from your local checkout?

Nope, ran from the deployed version in this PR: https://pr813--dexreact.review.gnosisdev.com/book

fleupold:
The URL changed in the meantime so the original build is no longer working.

alfetopito:
> The URL changed in the meantime so the original build is no longer working.

:thinking: 

Don't you need to push the new url to this repo?

fleupold:
Oups, I forgot to do the push command yesterday 🙈🙉🙊 Should be there now...

fleupold:
The backend is updated and I think the weird behavior is now gone. That is the images of this build and https://czii9.csb.app/ match and there is no longer a weird overlap (cf images below)

Screen Shot 2020-04-06 at 12 57 11
Screen Shot 2020-04-06 at 12 57 36
Screen Shot 2020-04-06 at 12 57 47


The small discrepency in volumes between DAI<->USDC and USDC<->DAI stems from the fact that one side is denominated in DAI and the other one in USDC (thus the volume is not expected to match 1:1 but instead according to the offered price).

Let me know if there are any concerns, otherwise I'm going to merge this towards the end of today.

alfetopito:
This gap does not look right:

screenshot_2020-04-06_08-42-31

Or is it ok?

fleupold:
Great catch @alfetopito. The problem is a sorting error in the underlying data when prices are really close together. We are relying on the orderbooks coming in ordered by price, but that ordering is done based on their [BN fraction representation](https://github.com/gnosis/dex-contracts/blob/97852ec0c992f30847a4dbb85b9e1c45c9bd1f16/src/orderbook.ts#L373)

However we then express them as floats where the ordering doesn't seem to preserve. The following test case fails: 

```js
const smaller = new Fraction(
        new BN("2cd42711c557ec02d09d6f7e0e0", "hex"),
        new BN("2c8f7d89a459955fcc182dc00e3", "hex")
); // 1.0060190481092344

const larger = new Fraction(
        new BN("190ccccccccccccccccccccccccccd9bf0", "hex"),
        new BN("18e66e96b6904a657edfd9efcafe985789", "hex")
); // 1.0060190481092346

assert.isTrue(smaller.toNumber() < larger.toNumber());
assert.isTrue(smaller.lt(larger)); // This fails
```

However, this is a backend issue (we should sort the json by float and not by fraction and thus can be fixed server side).

PR #812, opened 2020-04-01
Add token when disconnected

Velenir:
Allows to add token in TokenSelector even when no Wallet connected
Assumes mainnet

PR #815, opened 2020-04-01
Why doesnt this work??

alfetopito:
Simple test to assert I'm crazy. Or not.

Head out to /about page (https://pr815--dexreact.review.gnosisdev.com/about)

1. Click on the nice buttons: Get id and Get address.
2. Fail :(

screenshot_2020-04-01_15-28-09

Why aren't they working?
It seems to work fine when placing new orders: https://github.com/gnosis/dex-react/blob/develop/src/hooks/usePlaceOrder.ts#L78

### DIFF
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -176,12 +176,17 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
 
   public async getTokenAddressById({ tokenId, networkId }: GetTokenAddressByIdParams): Promise<string> {
     const contract = await this._getContract(networkId)
-    return contract.methods.tokenIdToAddressMap(tokenId).call()
+    console.log(`[${networkId}][${tokenId}] Trying to get address for id`)
+    const address = await contract.methods.tokenIdToAddressMap(tokenId).call()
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr815--dexreact.review.gnosisdev.com

alfetopito:
Ok ok, I think I see what might be wrong.
Switching for network `4` instead of `1` I do get a valid response.

Assumption: default provider connection is Rinkeby (networkId == 4), thus it's only possible to query the contract for this network.

alfetopito:
It works in other places because we always deal with the currently selected network.

Hmm, what about places where we default to Mainnet when wallet is not connected?

alfetopito:
Understood what is going on and why it was failing.

Currently, when not connected to a wallet, we default to using an Infura endpoint for Rinkeby network: https://github.com/gnosis/dex-react/blob/develop/src/const.ts#L100

That is changing though to Mainnet in this PR: https://github.com/gnosis/dex-react/pull/812/files#diff-437c799b591f0165059d5b0ff3a3a359R100

That's why this simple experiment fails. Querying for Mainnet when connected to Rinkeby will not work.

anxolin:
Arrived late to the party but happy to see you solved it, that we are addressing it, and that you are not crazy either :)

PR #818, opened 2020-04-02
Add token form url

Velenir:
Allows to add token from url

try https://pr818--dexreact.review.gnosisdev.com/trade/0x45804880de22913dafe09f4980848ece6ecbaf78-USDC?sell=0&price=0&from=0&expires=2880

### DIFF
--- a/src/api/proxy/CacheMixin.ts
+++ b/src/api/proxy/CacheMixin.ts
@@ -31,7 +31,7 @@ export class CacheMixin {
       const methodName = method.toString()
 
       const fnToCache = instance[methodName].bind(instance)
-
+      \
       const params: CacheMethodParams<T, ReturnType<typeof fnToCache>> = {
         ...cacheConfig,
         fnToCache,
@@ -59,7 +59,11 @@ export class CacheMixin {
       value = fnToCache(...params)
 
       // save it for next round
-      this.store(hash, value, ttl)
+      Promise.resolve(value)
+        .then(val => {
+          this.store(hash, val, ttl)
+        })
+        // don't cache errors
 
       return value as R
     }
--- a/src/api/proxy/CacheMixin.ts
+++ b/src/api/proxy/CacheMixin.ts
@@ -31,7 +31,7 @@ export class CacheMixin {
       const methodName = method.toString()
 
       const fnToCache = instance[methodName].bind(instance)
-      \
+
       const params: CacheMethodParams<T, ReturnType<typeof fnToCache>> = {
         ...cacheConfig,
         fnToCache,
@@ -59,11 +59,10 @@ export class CacheMixin {
       value = fnToCache(...params)
 
       // save it for next round
-      Promise.resolve(value)
-        .then(val => {
-          this.store(hash, val, ttl)
-        })
-        // don't cache errors
+      Promise.resolve(value).then(val => {
+        this.store(hash, val, ttl)
+      })
+      // don't cache errors
 
       return value as R
     }
--- a/src/api/proxy/CacheMixin.ts
+++ b/src/api/proxy/CacheMixin.ts
@@ -59,10 +59,18 @@ export class CacheMixin {
       value = fnToCache(...params)
 
       // save it for next round
-      Promise.resolve(value).then(val => {
-        this.store(hash, val, ttl)
-      })
-      // don't cache errors
+      // store possibly a promise
+      this.store(hash, value, ttl)
+      Promise.resolve(value)
+        .then(val => {
+          // store a value of resolved promise
+          this.store(hash, val, ttl)
+        })
+        // don't cache errors
+        .catch(() => {
+          // remove failed promise from store
+          this.remove(hash)
+        })
 
       return value as R
     }
@@ -76,6 +84,10 @@ export class CacheMixin {
     }
 
     return obj
+  }
+
+  private remove(hash: string): void {
+    this.cache.del(hash)
   }
 
   private store<R>(hash: string, obj: R, ttl?: number): void {
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -7,6 +7,8 @@ import { FieldValues } from 'react-hook-form/dist/types'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { toast } from 'toastify'
 import BN from 'bn.js'
+import Modali from 'modali'
+import { isAddress } from 'web3-utils'
 
 import TokenRow from './TokenRow'
 import OrderValidity from './OrderValidity'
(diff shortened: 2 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr818--dexreact.review.gnosisdev.com

src/api/proxy/CacheMixin.ts
```diff
@@ -59,7 +59,18 @@ export class CacheMixin {
...
+      Promise.resolve(value)
+        .then(val => {
+          // store a value of resolved promise
+          this.store(hash, val, ttl)
+        })
+        // don't cache errors
+        .catch(() => {
+          // remove failed promise from store
+          this.remove(hash)
+        })
```
alfetopito, line 73: Thank you! I was thinking about this yesterday :)

src/hooks/useAddTokenModal.tsx
```diff
@@ -152,16 +158,18 @@ export const useAddTokenModal = (): UseAddTokenModalResult => {
...
   // using deferred promise that will be resolved separately
-  const result = useRef<Deferred<boolean>>()
+  const result = useRef<Deferred<TokenDetails | null>>()
   // to faster show the Token on Confirm, prefetch sooner
-  const prefetchToken = useRef<Promise<TokenDetails | null>>(Promise.resolve(null))
+  const prefetchToken = useRef<Promise<TokenFromExchangeResult | null>>(Promise.resolve(null))
   const [token, setToken] = useState<TokenDetails | null>(null)
   const [error, setError] = useState<Error | null>(null)
 
+  const [isProcessing, setIsProcessing] = useState(false)
```
alfetopito, line 167: useSafeState?

src/components/TradeWidget/index.tsx
```diff
@@ -336,6 +340,25 @@ function calculateReceiveAmount(priceValue: string, sellValue: string): string {
   return receiveAmount
 }
 
+interface TokenAdderProps {
+  tokenAddress: string
+  networkId: number
+  onSelectChange: (selected: TokenDetails) => void
+}
+
+const TokenAdder: React.FC<TokenAdderProps> = ({ tokenAddress, networkId, onSelectChange }: TokenAdderProps) => {
```
anxolin, line 349: consider moving this to it's own file

anxolin: Does it make sense that instead of defining the hook in it's own file, we define the component TokenAdderModal with all the logic there?

src/components/TradeWidget/index.tsx
```diff
@@ -370,13 +393,17 @@ const TradeWidget: React.FC = () => {
   const [sellToken, setSellToken] = useState(
     () =>
       trade.sellToken ||
-      getToken('symbol', sellTokenSymbol, tokens) ||
+      (sellTokenSymbol && isAddress(sellTokenSymbol?.toLowerCase())
```
anxolin, line 396: Can we refactor this a bit, also it grow the oneline statement a bit. I think it's the moment to add some more lines/statements/intermediate-vars or whatever it takes to read it simpler

src/components/TradeWidget/index.tsx
```diff
@@ -690,8 +717,46 @@ const TradeWidget: React.FC = () => {
     receiveValue,
   ])
 
+  const fallBackNetworkId = networkId || Network.Mainnet // fallback to mainnet
```
anxolin, line 720: we have this logic in many places, since we get the networkId always from `useWalletConnection` should we move the logic there and expose `networkIdOrDefault` for example?

src/components/TradeWidget/index.tsx
```diff
@@ -690,8 +717,46 @@ const TradeWidget: React.FC = () => {
...
+
+  const { needToAddSellToken, needToAddReceiveToken } = useMemo(() => {
+    const needToAddSellToken =
```
anxolin, line 723: A bit repetitive. Is it possible to return the array of addresses to add, so we also remove the repetition in the JSX part?

src/services/factories/getTokenFromExchange.ts
```diff
@@ -49,23 +49,38 @@ interface GetByAddressParams {
   tokenAddress: string
 }
 
+export const enum TokenFromExchange {
```
anxolin, line 52: I can't guess what this enum models reading it's name.
Looks like it's used for documenting a "reason" of the value in token? 

I mean, shouldn't it return the data, unless there's a problem like "Unregistered" and in that case you reject providing that reason? why do we want to fulfill the promise in that case?

Velenir:
> 1. After clicking the button in modal to confirm adding the token, validation was triggered (empty price) and nothing happened. I had to click again. Hopefully resolved by David's PR.

This happens because focus is somewhere else, may or may not be solved by David's PR
Will revisit later

src/services/factories/getTokenFromExchange.ts
Velenir: I added reason to be able to distinguish why token couln't be fetched.
As when fetching to early, which happens when opening modal right away with token from URL, we may be between providers (switching to lastUsedProvider). And it fails to fetch fro ExchangeApi
Then we retry. But if it failed for any other reason then no use retrying.

Velenir: Can of course throw an error, but then those can be different errors and we'll still need to differentiate the reason

src/components/TradeWidget/index.tsx
Velenir: Please elaborate

src/components/TradeWidget/index.tsx
Velenir: Maybe in a different PR, too many places to change

Velenir:
> 2\. After confirming the second modal (where the token img is displayed) causing it to go away, I could see for a moment the modal with a different styling, similar to what I got in the other PR. Might be FF issues, was just for a fraction of a sec, not a big deal. But it was there!

Can you make a video/gif or something. I'm not sure how to reproduce

Velenir:
> * Can we join the two modals into one? So we have one. Something like:

I'm reusing same modal as for TokenSelector add, so I was hoping to avoid making a separate component

Let's discuss what kind of UX we want wholistically.

Both for TokenSelector, from URL and then later for Manage token in /wallet
@anxolin @alfetopito @W3stside

src/services/factories/getTokenFromExchange.ts
anxolin: if it's easier for you go ahead, although in principle you can throw an error with a type field, that would allow to catch and retry depending on the type

src/components/TradeWidget/index.tsx
anxolin: instead of returning:
```
{
      needToAddSellToken,
      needToAddReceiveToken,
    }
```

you return an array of addresses. So this array would be:
* empty
* one address
* two addresses

Then I say is repetitive, cause line 723 and 728 expressions looks very ilike. 
Then in https://github.com/gnosis/dex-react/blob/43b68b7f503ad7a413512c45cd54dcc0951e9d94/src/components/TradeWidget/index.tsx#L746 , lines 746 and 753 are also repetitive, and if you have the array I mentioned above, you wouldn't have that repetition.

anxolin:
> I'm reusing same modal as for TokenSelector add, so I was hoping to avoid making a separate component

Yes let's discuss it. I'm ok using the same modal. But ideally a single modal, and with the question, warning frase and preview of the token info. Does it make sense?

PR #819, opened 2020-04-02
Reactify charts

Velenir:
Some suggestions to make charts more react-like

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useEffect, useRef } from 'react'
 import styled from 'styled-components'
 import { TokenDetails } from 'types'
 import * as am4core from '@amcharts/amcharts4/core'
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr819--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -157,18 +162,19 @@ const draw = (baseToken: TokenDetails, quoteToken: TokenDetails, dataSource: str
...
   useEffect(() => {
-    if (chart) {
-      chart.dispose()
-    }
-    setChart(draw(baseToken, quoteToken, orderbookUrl(baseToken, quoteToken, networkId)))
+    if (!mountPoint.current) return
+
+    const chart = draw(mountPoint.current, baseToken, quoteToken, orderbookUrl(baseToken, quoteToken, networkId))
+
+    return (): void => chart.dispose()
```
fleupold, line 173: I think we still need to dispose on every redraw (e.g. when tokens change)

Velenir: And that's exactly how  `useEffect` works
```ts
useEffect(() => {
...
    return (): void => chart.dispose() //runs on clean-up stage
  }, [baseToken, quoteToken, networkId]) //deps
```
deps change -> cleanup function runs -> actual function runs -> deps change -> clean-up ...

Velenir:
> Awesome, thanks a lot! Can I just merge into my branch?

Go for it :+1:

Issue #467, opened 2020-01-28
Dynamically load token ids


alfetopito:
With the addition of `sUSD` on https://github.com/gnosis/dex-react/issues/461 -- which has been manually added to the contract with id `9`, but not immediately after the previous in the list (DAI with id `7`) -- the ids on https://github.com/gnosis/dex-js/blob/master/src/tokenList.json will only be valid for the current contract deployment on Mainline.

As soon as there's another Mainline deployment, or even for Rinkeby, we cannot guarantee that sUSD will be added to the contract in the same position (10th == id 9).

For this reason, we shouldn't rely on the hardcoded ids, and instead fetch the them from the contract.

anxolin:
But, one of the data fields in the token json file is `id` wouldn't it be enough to add the right value there?

Velenir:
Different ids on different networks

anxolin:
> Different ids on different networks

That's right. Maybe defining the id as an object like `addressByNetwork`?

alfetopito:
> > Different ids on different networks
> 
> That's right. Maybe defining the id as an object like `addressByNetwork`?

Well, yes. In that case we could still have a hard coded list, with the `id` bound to the network.

* Upside is that there's no client side id fetching, faster loading for the user
* Downside is that every deployment (be it mainnet or rinkeby) requires:
  1. manual changes to the file, 
  2. new dex-js npm release and 
  3. bump version on dex-react and dex-telegram. 

Not that we expect often deployments, so this might not be a big deal.

anxolin:
Yup, it's not difficult, maybe fetching is better so adding tokens in the list is easier and we also prepare the ground for using TCRs.

Besides, thanks to the wonderful client-side catching it shouldn't be a performance issue to load the token ids, only for the first load and I don't think it's that bad (should be fast anyways)

alfetopito:
Thoughts:

1. Ignore completely list of tokens from `dex-js`
2. Whenever not connected, default to `mainnet` list of tokens
3. Store list of tokens for each network on local storage
4. On `TokenList` constructor
    1. Load list of tokens from local storage, if any
    2. Async; Fetch [number of tokens registered in each network](https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L54)
    3. If number of tokens == local list length, stop (per network)
    4. Else, fetch token details from token id = localList.length -1 until numTokensInExchange
        1. Fetch address from exchange contract
[...]

PR #824, opened 2020-04-02
Dynamically loading tokens ids, per network, first time getTokens is called

alfetopito:
Part of #467 

See title.

Also, removing tokens which don't exist for given network. Example:

screenshot_2020-04-02_15-03-02_231436198

Querying sequentially, nice and easy, nothing fancy parallel extra optimized etc.

And, no need for throttle after all...

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -96,14 +96,14 @@ function createExchangeApi(erc20Api: Erc20Api, injectedDependencies: DepositApiD
   return exchangeApi
 }
 
-function createTokenListApi(exchangeApi: ExchangeApi): TokenList {
+function createTokenListApi(): TokenList {
   const networkIds = [Network.Mainnet, Network.Rinkeby]
 
   let tokenListApi: TokenList
   if (process.env.MOCK_TOKEN_LIST === 'true') {
     tokenListApi = new TokenListApiMock(tokenList)
   } else {
-    tokenListApi = new TokenListApiImpl({ networkIds, exchangeApi })
+    tokenListApi = new TokenListApiImpl({ networkIds })
   }
 
   window['tokenListApi'] = tokenListApi // register for convenience
@@ -135,5 +135,5 @@ const injectedDependencies = {
 export const erc20Api: Erc20Api = createErc20Api(injectedDependencies)
 export const depositApi: DepositApi = createDepositApi(erc20Api, injectedDependencies)
 export const exchangeApi: ExchangeApi = createExchangeApi(erc20Api, injectedDependencies)
-export const tokenListApi: TokenList = createTokenListApi(exchangeApi)
+export const tokenListApi: TokenList = createTokenListApi()
 export const theGraphApi: TheGraphApi = createTheGraphApi()
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -96,14 +96,14 @@ function createExchangeApi(erc20Api: Erc20Api, injectedDependencies: DepositApiD
   return exchangeApi
 }
 
-function createTokenListApi(): TokenList {
-  const networks = [Network.Mainnet, Network.Rinkeby]
+function createTokenListApi(exchangeApi: ExchangeApi): TokenList {
+  const networkIds = [Network.Mainnet, Network.Rinkeby]
 
   let tokenListApi: TokenList
   if (process.env.MOCK_TOKEN_LIST === 'true') {
     tokenListApi = new TokenListApiMock(tokenList)
   } else {
-    tokenListApi = new TokenListApiImpl(networks)
+    tokenListApi = new TokenListApiImpl({ networkIds, exchangeApi })
   }
 
   window['tokenListApi'] = tokenListApi // register for convenience
@@ -135,5 +135,5 @@ const injectedDependencies = {
 export const erc20Api: Erc20Api = createErc20Api(injectedDependencies)
 export const depositApi: DepositApi = createDepositApi(erc20Api, injectedDependencies)
 export const exchangeApi: ExchangeApi = createExchangeApi(erc20Api, injectedDependencies)
-export const tokenListApi: TokenList = createTokenListApi()
+export const tokenListApi: TokenList = createTokenListApi(exchangeApi)
 export const theGraphApi: TheGraphApi = createTheGraphApi()
--- a/src/api/tokenList/TokenListApi.ts
+++ b/src/api/tokenList/TokenListApi.ts
@@ -2,11 +2,17 @@ import { TokenDetails } from 'types'
 import { getTokensByNetwork } from './tokenList'
 import { logDebug } from 'utils'
 import GenericSubscriptions, { SubscriptionsInterface } from './Subscriptions'
+import { ExchangeApi } from 'api/exchange/ExchangeApi'
 
 export interface TokenList extends SubscriptionsInterface<TokenDetails[]> {
   getTokens: (networkId: number) => TokenDetails[]
   addToken: (params: AddTokenParams) => void
   hasToken: (params: HasTokenParams) => boolean
+}
+
+export interface TokenListApiParams {
+  exchangeApi: ExchangeApi
+  networkIds: number[]
 }
 
 export interface AddTokenParams {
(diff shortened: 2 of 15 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr824--dexreact.review.gnosisdev.com

src/api/tokenList/TokenListApi.ts
```diff
@@ -108,6 +122,55 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
...
+
+    // Sequentially fetch updated ids
+    const updatedTokenList = []
+    let failedToUpdate = false
+    for (let i = 0; i < this._tokensByNetwork[networkId].length; i++) {
+      const token = this._tokensByNetwork[networkId][i]
+      const tokenAddress = token.address
+
+      try {
+        token.id = await this.exchangeApi.getTokenIdByAddress({ networkId, tokenAddress })
```
Velenir, line 143: Aggregate tokenIds and do `getTokenIdByAddress` calls in `Promise.all` otherwise it will add up quickly

alfetopito: That's what I was doing at first. Then, with all the failures I was getting (due to wrong network) I thought throttling was necessary... you know the story.

I decided to do it one by one to maybe avoid that? Not sure makes any difference though.

Alright, will do the bundled way :)

src/api/tokenList/TokenListApi.ts
```diff
@@ -108,6 +122,55 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
...
+  }
+
+  private async updateTokenIds(networkId: number): Promise<void> {
+    // Set as updated on start to prevent concurrent queries
+    this.updatedIdsForNetwork.add(networkId)
+
+    // Sequentially fetch updated ids
+    const updatedTokenList = []
+    let failedToUpdate = false
+    for (let i = 0; i < this._tokensByNetwork[networkId].length; i++) {
```
anxolin, line 138: ubermeganit: Can we do a foreach instead?

src/api/tokenList/TokenListApi.ts
```diff
@@ -108,6 +122,55 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
...
+      const token = this._tokensByNetwork[networkId][i]
+      const tokenAddress = token.address
+
+      try {
+        token.id = await this.exchangeApi.getTokenIdByAddress({ networkId, tokenAddress })
+        updatedTokenList.push(token)
+
+        logDebug(`[network:${networkId}][address:${tokenAddress}] Token id updated: ${token.id}`)
+      } catch (e) {
+        if (e.message.match(/Must have Address to get ID/)) {
```
anxolin, line 148: Shouldn't we check by using `tokenAddressToIdMap`, sounds a bit strange this check

src/api/tokenList/TokenListApi.ts
alfetopito: Can't (find out how to) `await` inside a forEach/map/reduce etc.

But I'll update it like Dima asked for anyway, this won't be a problem anymore.

src/api/tokenList/TokenListApi.ts
alfetopito: Not really. In my view error handling is up to the caller.

src/api/tokenList/TokenListApi.ts
anxolin: Yep, it's not supported. Usually it's a map+promiseAll

src/api/index.ts
```diff
@@ -96,14 +96,14 @@ function createExchangeApi(erc20Api: Erc20Api, injectedDependencies: DepositApiD
...
-  const networks = [Network.Mainnet, Network.Rinkeby]
+function createTokenListApi(exchangeApi: ExchangeApi): TokenList {
+  const networkIds = [Network.Mainnet, Network.Rinkeby]
 
   let tokenListApi: TokenList
   if (process.env.MOCK_TOKEN_LIST === 'true') {
     tokenListApi = new TokenListApiMock(tokenList)
   } else {
-    tokenListApi = new TokenListApiImpl(networks)
+    tokenListApi = new TokenListApiImpl({ networkIds, exchangeApi })
```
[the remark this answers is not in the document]
alfetopito: > mmm, can we avoid making an API depend on another API directly?

Yeah, that is a problem I thought about.

> maybe depend on a pure function? getTokenIdByAddress? 

That would still make it dependent on exchangeApi. The difference is that instead of depending on the whole api we depend on only 1 function.

> Otherwise, I think we created the service layer for this

I have an idea, not sure it'll work or whether it makes sense: Make `updateIdsPerNetwork` a service. 
For that we'd need:
1. expose a way to update the local token list (easy)
2. when tokens are updated, update subscribers (easy I guess?)
3. figure out a way to call `updateIdsPerNetwork` once per network, per page load, without circular dependencies (hard?)

`3` is hard because the current implementation invokes it when `getTokens` is called, and tracks it within the class. I think I might need to expose another service to encapsulate `getTokens` instead of allowing direct access to `tokenListApi`.
But then, how do I deal with the subscriptions?

src/api/tokenList/TokenListApi.ts
anxolin: Yes, I know, but the strange part is how we check which kind of error it is, right?

wait wait, I misunderstood there, I thought it was handling the case where the address is not in the exchange. But it's just checking the implementation I see:
```js
public async getTokenIdByAddress({ tokenAddress }: GetTokenIdByAddressParams): Promise<number> {
    assert(typeof this.tokenAddressToId[tokenAddress] === 'number', 'Must have Address to get ID')
    return this.tokenAddressToId[tokenAddress]
  }
```

So I still think it's a bid odd. Cause it's relying on a message of an exception of an specific implementation of the API. If someone changes this message, the Api implementation or whatever, this piece won't realise. Or maybe I miss sth.

alfetopito: > Cause it's relying on a message of an exception of an specific implementation of the API. If someone changes this message, the Api implementation or whatever, this piece won't realise

That is true.

What we could do is to have error codes, like what Dima did for the service response here https://github.com/gnosis/dex-react/pull/818/files#diff-e89dfdf83d68cb15aaac942477290d0bR52

But, this won't change the fact that the implementation can change and we are at the same problem as you point out.

Although, the contract implementation requires a new deployment, which doesn't happen every day. So I'd say we are safe on that front.

src/api/index.ts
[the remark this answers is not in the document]
anxolin: > That would still make it dependent on exchangeApi. The difference is that instead of depending on the whole api we depend on only 1 function.

Yes, this is not my prefered option. And yes, that's so true, however it would depend on a pure function that would be injected, so it will be blind to what someone injects there

>> Otherwise, I think we created the service layer for this
> I have an idea, not sure it'll work or whether it makes sense: Make updateIdsPerNetwork a service.

If I understand correctly, I like the proposal. By the look of `updateTokenIds` looks more like a service to me than an API. Coordinate other services. Actually, I don't think updateTokenIds should be a service, at least "public" I think it should be getTokens the service, right?

With that, you don't have circular dependencies, do you?

src/api/tokenList/TokenListApi.ts
anxolin: Aha, I was not 100% sure where that error message came from. I see is in the IdToAddressBiMap. Yes, we don't expect this to change easily, won't happen unless we change the contract implementation. Hoewever, I find that if this "service logic" is relying 
on handling that case, we are better of using the funcion with a pink arrow, this way, we are type safe. If we change the smart contract implementation, that would give us a compile error. Also, it's more declarative.

image

src/api/tokenList/TokenListApi.ts
alfetopito: https://github.com/gnosis/dex-react/pull/824/commits/e6c39f6217e0f46f98b0749edc3d0a4bab12425c

src/api/tokenList/TokenListApi.ts
alfetopito: :point_right:  https://github.com/gnosis/dex-react/pull/824/commits/e6c39f6217e0f46f98b0749edc3d0a4bab12425c

Honestly though, old school for loop was more efficient, looping through list only once. Now I'm looping it three times:
1. Prepare list of promises
2. Resolve promises
3. Use results

Or I am just really bad at this.

src/api/index.ts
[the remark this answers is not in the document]
alfetopito: Done what we discussed over slack.

Exposing 2 services for getting the tokens and for subscribing to updates.
TokenListApi should no longer be directly used as an api, but only from the service.
Except by other services. Don't want to create dependencies between services either.

src/api/tokenList/TokenListApi.ts
alfetopito: Aha, this function then: https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L417

Will do

alfetopito: https://github.com/gnosis/dex-react/pull/824/commits/ef569fce6b8a0f4bf9b920194a4e1983edb61bbd

PR #825, opened 2020-04-03
No options message change

Velenir:
We can change NoOptionsMessage dynamically to give user hints that a token can be added

ezgif-4-7a67f43d3fd3

Can even do 
localhost_8080_trade_DAI-USDC_sell=0 price=1 01955 from=0 expires=2880
but will be hacky and will have to rewrite `useAddTokenModal` logic a lot
Let's first discuss UX across all ways to add a token


gnosis-info:
Travis automatic deployment:
 https://pr825--dexreact.review.gnosisdev.com

PR #834, opened 2020-04-06
790/fetch prices from dex price estimator

### LINKED ISSUE #790: Use new price endpoint
@fleupold will provide the endpoint URL

This price, will provide a better estimate. Instead of suggesting the last price, it suggest one that takes into account the user balances and open orders, so the trade will be more likely to happen at this price

We'll replace the implementation of out price suggestion to use the new endpoint

alfetopito:
Closes #790 

~Still working on the service factory to expose this new API.~

~In the mean time, you can try it out in the console with something like:~

```
window.dexPriceEstimatorApi.getPrice({networkId:1, baseToken:{id:5}, quoteToken:{id:7}}).then(p=>console.log(p.toString()))
```
Where tokenId `7` == `DAI`; tokenId `5` == `PAX`

### Edit 1:
Api is ready. Adding it to trade widget.

~That is the weird part. For some reason, the price fetching never stops.~
~I'm using a craze high debounce value (5s) to test it more easily.~
~It seems that, even when no values are changed, the `userEffect` to update the price (https://github.com/gnosis/dex-react/pull/834/commits/a65fc2c622296c15eb0f53f5d3d66264101df38d#diff-c47f9093788861eb688116377dd46937R64) is triggered).~
~I thought it was maybe due to the amount being refreshed too often, but even when I set a static value (https://github.com/gnosis/dex-react/pull/834/commits/a65fc2c622296c15eb0f53f5d3d66264101df38d#diff-d69cd25170f3907dcfdd742b5b057099R446) the loop continues.~

~Ideas?~

### Edit 2:
Thanks to @W3stside , managed to stop the loop.
Contrary to what I expected, React is not smart enough to detect updates in nested objects on hook deps.

Price suggestion updated on
1. token change or
2. debounced by 200ms after updating sell price (to avoid sending a request for price on every keystroke).

:warning: One thing to note is that, when inserting a sufficiently high sell amount, the price will turn 0.

That's because currently we can't tell from price api's response whether given token doesn't exist or there's no volume for given amount.

### TODO:
- [x] Handle failure when token in unknown
- [x] ~Service layer~ turns out this api is simple enough, no need for a service layer
- [x] Use it!
- [x] Remove debugging logs


gnosis-info:
Travis automatic deployment:
 https://pr834--dexreact.review.gnosisdev.com

anxolin:
> One thing to note is that, when inserting a sufficiently high sell amount, the price will turn 0.

Probably a good thing imo 👍 
Gives the user a hint of the depth, and he is free to use the price he gets for low volume with high volume, but at least if he don't get the order matched (or if it take more batches), he would know why

anxolin:
Amending my last comment, I see something a bit weird the following case:

If I sell 1000 DAI for OWL I get:
image

But If I sell 10000 DAI for OWL, I get 1 OWL:
image

Ideally, if there's a big drop in the price, we could show the price for low volume, with a pre-fixed slippage and a message of low liquidity, warning the user that it may take several batches.

anxolin:
@fleupold, I'm reviewing the case exposed above, and it looks like the endpoint is actually returning:

For 1000 (`/markets/0-7/estimated-buy-amount/1000000000000000000000?atoms=true`) ---> `1,329.3 OWL`

```json
{"baseTokenId":"0","quoteTokenId":"7","buyAmountInBase":1.329337332e+21,"sellAmountInQuote":1e+21}
```

For 10000 (`/markets/0-7/estimated-buy-amount/10000000000000000000000?atoms=true`) ---> `0.9795495633 OWL`
```json
{"baseTokenId":"0","quoteTokenId":"7","buyAmountInBase":979549563336939400,"sellAmountInQuote":1e+22}
```

How can it be?

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    const url = baseUrl + queryString
+
+    const response = await fetch(url)
+
+    if (!response.ok) {
+      throw new Error(`Request failed: [${response.status}] ${response.body}`)
+    }
+
+    const body = await response.text()
```
Velenir, line 115: Why `.text()` and then `JSON.parse`? Why not `response.json()`?

fleupold:
@anxolin the more you sell the worse your price gets due to slippage. While intuitively you might expect to get at least as much OWL for 10k DAI as for 1k DAI, our "uniform clearing price" constraint enforces that you get the worst price for the entirety of your volume.

So if the worst counter-offer you have to accept to sell 1k DAI pays 1.3 OWL per DAI you get 1300 OWL for the whole lot, but if for 10k DAI people the worst count-offer only pays 0.01 per DAI you only get 100 OWL for the whole lot.

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
alfetopito: Because the response can be empty. That throws when doing `response.json()`

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -0,0 +1,123 @@
...
+  sellAmountInQuote: string
+}
+
+export interface Params {
+  networkIds: number[]
+}
+
+function getDexPriceEstimatorUrl(networkId: number): string {
+  const networkName = getNetworkFromId(networkId).toLowerCase()
+  return `https://price-estimate-${networkName}.dev.gnosisdev.com/api/v1/`
```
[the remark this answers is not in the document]
alfetopito: @fleupold Is there a production url?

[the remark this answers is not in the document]
alfetopito: Added a todo entry and created an issue to not forget about it https://github.com/gnosis/dex-react/issues/869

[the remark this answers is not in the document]
fleupold: Not yet, but we can ask DevOps to deploy to prod on Tuesday

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
Velenir: That sounds like a non-normalized api :disappointed:

alfetopito: Well, there are some rough edges, yes, but I'm sure we can ask @fleupold nicely or even do the change ourselves :)

PR #838, opened 2020-04-07
Fix path for to the fileMock in the tests

anxolin:
There's an issue with using CSS in the tests because the path to the mock JS is wrong.

I saw this in other branch, it only broke cause I included modali (which imports CSS) in the row component which that contains tests. I assume we never attempted a import of CSS before in a test :)


Velenir:
Good catch!
More proof that CSS doesn't get along with JS

PR #850, opened 2020-04-08
Order book btn

anxolin:
> DRAFT: 

Button for opening a modal with the order book.

The trade widget selection is used as the default for the tokens in the order book. But it allows the user to change the selection.
This selection won't change the trade widget selection on purpose.

image

image

------------------
<del>
@biocom is this enough for styling it?
Remaining things I see are:
- the window is not big enough
- beter layout for the token selectors
- Not sure why the close button is in the left. I'm using the defaults from modali too
- Not sure why the close button icon (times icon in the top right corner is hidden)
</del>

edit: Thanks for that PR Michel 😊


gnosis-info:
Travis automatic deployment:
 https://pr850--dexreact.review.gnosisdev.com

Velenir:
> * Not sure why the close button is in the left. I'm using the defaults from modali too

Because it's how Modali rolls. Need to pad with empty space
```js
buttons: [
   <>&nbsp;</>,
  <Modali.Button/>
]

File: src/components/OrderBookBtn.tsx [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+    large: true,
+    title: `${safeTokenName(baseToken)}-${safeTokenName(quoteToken)} Order book`,
+    message: (
+      <ModalWrapper>
+        <TokenSelector tokens={tokenList} selected={baseToken} onChange={setBaseToken} />
+        <TokenSelector tokens={tokenList} selected={quoteToken} onChange={setQuoteToken} />
+        <OrderBookWidget baseToken={baseToken} quoteToken={quoteToken} networkId={networkId} />
+      </ModalWrapper>
+    ),
+    buttons: [<Modali.Button label="Close" key="yes" isStyleDefault onClick={(): void => modalHook.hide()} />],
```
Velenir, line 40: Why are you using `modalHook.hide()` instead of `toggleModal`? Not that it matters as they are the same function. But we use toggle everywhere else.

anxolin: I just saw the hide function in their API and I thought it was more declarative than the toggle for the close button and accept button

U don't like it? :)

anxolin:
>Because it's how Modali rolls. Need to pad with empty space

A biiiit weird, but I'm happy If it set's the button in the right place (being right the both meanings in this case)

src/components/OrderBookBtn.tsx
Velenir: I honestly don't care. Same function, even if named hide, also toggles
:+1:

anxolin: > I honestly don't care. Same function, even if named hide, also toggles.

what?? hahaha. what a weird library
I'll use toggle from now on

PR #852, opened 2020-04-08
Weth api, hooks

anxolin:
**Done:**
- [x] Uses the new dex-js with WETH support
- [x] Adds Api and Mock for wrapping Ether
- [x] Adds a hook for wrapping/unwrapping
- [x] Refactor the load token balances so they also load the ETH balance
- [x] Load and displays the ETH and WETH in the modal for Wrapping and Unwrapping
- [x] Substitutes the alerts of the modal when you wrap/unwrap by the hooks that do the actual call to the API
- [x] Handle the loading? 
- [X] Make sure the window is closed at the right time. When the tx is sent
- [x] Make sure we have a success/error toast: On mining, on error
- [x] Make sure the user connects before opening the modal: otherwise, we don't know the balance
- [x] Finish the IMPL. Signature should be OK, and it's kind of done, need to make sure it works

**Out of this PR (it's quite big already):**
- Finish the MOCK. It's not done, probably I'll need to inject other mock APIs cause this modifies the ETH balance (wallet api) and the WETH balance (erc20 api) --> Expect a PR just for this. Please excuse the commented code in the mock

**Aditional notes on this PR:**
* I decided to keep the functionality of set max. It’s a bit weird for wrapping, but maybe it’s good to leave it, not sure. Glad to hear feedback
* I decided to remove the wrap ether from the trade widget for now, until we have deposit, otherwise, it don’t make sense to me ---> but it works if you uncomment it. So it’s ready to be added


gnosis-info:
Travis automatic deployment:
 https://pr852--dexreact.review.gnosisdev.com

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Basic implementation of WETH API
+ * Note that there's already another API for all ERC20, so only the deposit/withdraw is implemented in this API
+ */
+export class WethApiImpl implements WethApi {
+  protected _contractPrototype: WethContract
+  private web3: Web3
+  private fetchGasPrice: WethApiDependencies['fetchGasPrice']
+  protected static _contractsCache: { [network: number]: { [address: string]: WethContract } } = {}
```
alfetopito, line 48: I would drop the address part. Not like we'll ever have more than one WETH contract deployed per network, right?

```suggestion
  protected static _contractsCache: { [network: number]: WethContract } = {}
```

src/api/weth/WethApiMock.ts
```diff
@@ -0,0 +1,76 @@
+import { logDebug } from 'utils'
+import { Receipt } from 'types'
+import { waitAndSendReceipt } from 'utils/mock'
+import { WethApi, WrapUnwrapParams } from './WethApi'
+
+import { RECEIPT } from '../../../test/data'
+
+// interface WethApiMockParams {
```
alfetopito, line 8: ~Bunch of dead code~

I know I know, will wait for next PR

src/hooks/useWrapUnwrapEth.ts
```diff
@@ -0,0 +1,89 @@
...
+  wrappingEth: boolean
+  unwrappingWeth: boolean
+  wrapEth(amount: string, txOptionalParams?: TxOptionalParams): Promise<Receipt>
+  unwrapWeth(amount: string, txOptionalParams?: TxOptionalParams): Promise<Receipt>
+}
+
+interface WrapUnwrapAuxParams {
+  userAddress: string | undefined
+  networkId: number | undefined
+  isConnected: boolean | undefined
```
alfetopito, line 20: nitpick: just my personal preference, ignore if you prefer your version

```suggestion
  userAddress?: string
  networkId?: number
  isConnected?: boolean
```

webpack.config.babel.js
```diff
@@ -163,6 +163,7 @@ module.exports = ({ stats = false } = {}) => ({
       MOCK_WALLET: process.env.MOCK || 'false',
       MOCK_TOKEN_LIST: process.env.MOCK || 'false',
       MOCK_ERC20: process.env.MOCK || 'false',
+      MOCK_WETH: process.env.MOCK_WETH || 'false',
```
alfetopito, line 166: Nope, wrong var. It already compares against `MOCK_WETH`. You should compare against `MOCK`.

```suggestion
      MOCK_WETH: process.env.MOCK || 'false',
```

src/hooks/useWrapUnwrapEth.ts
W3stside: always better this way :+1:

src/api/index.ts
```diff
@@ -71,6 +73,18 @@ function createErc20Api(injectedDependencies: Erc20ApiDependencies): Erc20Api {
...
 }
 
+function createWethApi(injectedDependencies: WethApiDependencies): WethApi {
+  let wethApi
+  if (process.env.MOCK_WETH === 'true') {
+    wethApi = new WethApiMock()
+  } else {
+    wethApi = new WethApiImpl(injectedDependencies)
+  }
+  window['wethApi'] = wethApi // register for convenience
```
W3stside, line 83: I'm still unsure we even want this in production

src/api/weth/WethApi.ts
W3stside, line 48: can we move all the protected/`_` props to the top of the class?

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  public constructor(injectedDependencies: WethApiDependencies) {
+    Object.assign(this, injectedDependencies)
+    this._contractPrototype = new this.web3.eth.Contract(wethAbi) as WethContract
+  }
+
+  public async deposit(params: WrapUnwrapParams): Promise<Receipt> {
+    const { networkId, amount, userAddress, txOptionalParams } = params
+    const contract = await this._getContract(networkId)
```
W3stside, line 57: why not 
```ts
const [contract, gasPrice] = await Promise.all([
  this._getContract(networkId),
  this.fetchGasPrice(),
])
```

I think `await`s inside send tx calls are weird

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const tx = contract.methods
+      .deposit()
+      .send({ from: userAddress, value: amount, gasPrice: await this.fetchGasPrice() })
+
+    if (txOptionalParams && txOptionalParams.onSentTransaction) {
```
W3stside, line 63:
```suggestion
    if (txOptionalParams?.onSentTransaction) {
```

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    logDebug(`[WethApi] Wrapped ${amount} ETH for user ${userAddress} in network ${networkId}`)
+
+    return tx
+  }
+
+  public async withdraw(params: WrapUnwrapParams): Promise<Receipt> {
+    const { networkId, amount, userAddress, txOptionalParams } = params
+    const contract = await this._getContract(networkId)
```
W3stside, line 74: same suggestion about putting both `async` fn calls inside 1 `await Promise.all([ ... ])`

Velenir:
ezgif-4-aeada2ab52bb
I have to click `Confirm` twice to trigger tx, because of focus I imagine


ezgif-4-c70e5d8bf80d
If you open modal, start tx, close modal without confirming, confirm in MMask, -- modal opens again without styling.
Has to do with `toggleModal` timing I imagine

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const tx = contract.methods.withdraw(amount).send({ from: userAddress, gasPrice: await this.fetchGasPrice() })
+
+    if (txOptionalParams && txOptionalParams.onSentTransaction) {
```
W3stside, line 78:
```suggestion
    if (txOptionalParams?.onSentTransaction) {
```

src/api/weth/WethApi.ts
```diff
@@ -0,0 +1,116 @@
...
+    return this._getContractAtAddress(networkId, address)
+  }
+
+  protected _getContractAtAddress(networkId: number, address: string): WethContract {
+    let contract: WethContract | undefined = undefined
+
+    if (WethApiImpl._contractsCache[networkId]) {
+      contract = WethApiImpl._contractsCache[networkId][address]
+    } else {
+      WethApiImpl._contractsCache[networkId] = {}
```
W3stside, line 104: I don't get this line, if the contract doesn't exist on the `_contractsCache` why do we need to reinstantiate an empty object? is this a single item cache in which case it doesn't need to be an object?

src/components/DepositWidget/Row.tsx
```diff
@@ -116,13 +118,14 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
...
           )}
         </td>
-        <td data-label="Wallet" title={formatAmountFull({ amount: walletBalance, precision: decimals }) || ''}>
+        <td data-label="Wallet">
           {isWeth ? (
             <ul>
-              <li className="not-implemented">
-                0.1 ETH <WrapEtherBtn label="Wrap" className="wrapUnwrapEther" />
+              <li title={ethBalance ? formatAmountFull({ amount: ethBalance, precision: decimals }) : undefined}>
+                {ethBalance ? formatAmount(ethBalance, decimals) : ''} ETH{' '}
```
W3stside, line 125: empty string or `0` or maybe `-` ? seems weird as empty string

src/components/DepositWidget/Styled.tsx
```diff
@@ -84,6 +84,10 @@ export const TokenRow = styled.tr`
     margin: 0;
     padding: 0;
 
+    li {
+      white-space: nowrap;
```
W3stside, line 88: careful with this, if the WETH/ETH amounts are large this may cause formatting issues. But maybe we can just `formatSmart` with this later on

src/components/WrapEtherBtn.tsx
```diff
@@ -222,14 +244,38 @@ const WrapUnwrapEtherBtn: React.FC<WrapUnwrapEtherBtnProps> = (props: WrapUnwrap
...
+                      },
+                    },
                     required: 'The amount is required',
                     min: 0,
                   })}
                 />
               </InputBox>
             </div>
-            {amountError && <p className="error">Invalid amount</p>}
+            {amountError && <p className="error">{amountError.message}</p>}
```
W3stside, line 267: really not comfortable with wrap all, it's a recipe for disaster and has overwhelmingly worse implications than benefits if done incorrectly even once.

I would scrap it and see it as an extra security layer that users themselves have to consciously put in the amount max themselves to verify. it's like when u delete a repo and have to type in the name of it before being able to submit

src/components/WrapEtherBtn.tsx
```diff
@@ -241,15 +287,33 @@ const WrapUnwrapEtherBtn: React.FC<WrapUnwrapEtherBtnProps> = (props: WrapUnwrap
...
         isStyleDefault
         onClick={handleSubmit((data: WrapEtherFormData): void => {
-          const { wrapAmount } = data
+          const { wrapAmount: wrapAmountEther } = data
+          const wrapAmount = toWei(wrapAmountEther)
+
+          // Hide modal once the transaction is sent
+          const txOptionalParams = composeOptionalParams(() => modalHook.hide())
+
+          new BigNumber(wrapAmount)
```
W3stside, line 296: ? what is this

src/const.ts
```diff
@@ -139,5 +139,11 @@ export const INPUT_PRECISION_SIZE = 6
 export const VALID_UNTIL_DEFAULT = '2880'
 export const VALID_FROM_DEFAULT = '30'
 export const WETH_ADDRESS_MAINNET = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
+export const WETH_ADDRESS_RINKEBY = '0xc778417E063141139Fce010982780140Aa0cD5Ab'
```
W3stside, line 142: does this need to be toUpperCase or do we do this conversion elsewhere?

src/components/DepositWidget/Row.tsx
```diff
@@ -116,13 +118,14 @@ export const Row: React.FC<RowProps> = (props: RowProps) => {
             <>{withdrawing && spinner}0</>
           )}
         </td>
-        <td data-label="Wallet" title={formatAmountFull({ amount: walletBalance, precision: decimals }) || ''}>
+        <td data-label="Wallet">
           {isWeth ? (
             <ul>
-              <li className="not-implemented">
-                0.1 ETH <WrapEtherBtn label="Wrap" className="wrapUnwrapEther" />
+              <li title={ethBalance ? formatAmountFull({ amount: ethBalance, precision: decimals }) : undefined}>
```
Velenir, line 124: here `title={ethBalance ? ... : undefined}`
and below it's `title={... || ''}`
pick one, `'' or undefined`

src/hooks/useWrapUnwrapEth.ts
```diff
@@ -0,0 +1,89 @@
...
+  amount: string
+  txOptionalParams: TxOptionalParams
+  setLoadingFlag: React.Dispatch<React.SetStateAction<boolean>>
+  execute: (params: WrapUnwrapParams) => Promise<Receipt>
+}
+
+async function wrapUnwrapAux(params: WrapUnwrapAuxParams): Promise<Receipt> {
```
W3stside, line 27: what is `Aux`?

src/components/WrapEtherBtn.tsx
```diff
@@ -222,14 +244,38 @@ const WrapUnwrapEtherBtn: React.FC<WrapUnwrapEtherBtnProps> = (props: WrapUnwrap
...
+                    },
                     required: 'The amount is required',
                     min: 0,
                   })}
                 />
               </InputBox>
             </div>
-            {amountError && <p className="error">Invalid amount</p>}
+            {amountError && <p className="error">{amountError.message}</p>}
+            {wrapAllBalance && (
```
Velenir, line 268: I wouldn't care about this use-case at all, but whatever

src/components/WrapEtherBtn.tsx
Velenir, line 296: huh?

src/hooks/useBalances.ts
```diff
@@ -109,30 +110,49 @@ async function _getBalances(walletInfo: WalletInfo, tokens: TokenDetails[]): Pro
...
         })
+    }
   }, [setBalances, setError, walletInfo, tokens])
 
-  return { balances, error, tokens }
+  // Get ether balances
+  useEffect(() => {
+    if (walletInfo.isConnected) {
+      walletApi
+        .getBalance()
```
Velenir, line 144: Do we really need this call everywhere we use tokenBalances? Wouldn't it be better in its own hook that will be used only in /wallet, only for WETH, otherwise wasted extra request

webpack.config.babel.js
Velenir: You'll also need to add it to `test/globalSetup.ts`

src/hooks/useWrapUnwrapEth.ts
Velenir: Auxiliary?

src/hooks/useWrapUnwrapEth.ts
```diff
@@ -0,0 +1,89 @@
...
+  const { userAddress, isConnected, networkId } = useWalletConnection()
+  const [wrappingEth, setWrappingEth] = useSafeState(false)
+  const [unwrappingWeth, setUnwrappingWeth] = useSafeState(false)
+  const baseParams = {
+    userAddress,
+    networkId,
+    isConnected,
+  }
+
+  async function wrapEth(amount: string, txOptionalParams: TxOptionalParams): Promise<Receipt> {
```
Velenir, line 64: Memoize?

src/hooks/useWrapUnwrapEth.ts
```diff
@@ -0,0 +1,89 @@
...
+    return wrapUnwrapAux({
+      ...baseParams,
+      amount,
+      setLoadingFlag: setWrappingEth,
+      execute: params => wethApi.deposit(params),
+      txOptionalParams,
+    })
+  }
+
+  async function unwrapWeth(amount: string, txOptionalParams: TxOptionalParams): Promise<Receipt> {
```
Velenir, line 75: Memoize?

src/api/index.ts
anxolin: We do it for all the APIs, I think it's convenient. Happy to review, but should be reviewed all of our APIs (not this PR)

src/api/weth/WethApi.ts
anxolin: well, yes, getContract is cached, so after the first query is just sync code anyways. So I thought it's was slightly more readable than Promise.all while not big advantage

src/api/weth/WethApi.ts
anxolin: Totally right!

anxolin: 7bbb9a302e7ac90bc9ac386fb1ce910db9eefc7d

src/api/weth/WethApiMock.ts
anxolin: Yes, I should have cleanup before sending. But I wanted to have it for your review and it was pretty late already! thanks for understanding :)

src/hooks/useWrapUnwrapEth.ts
anxolin: Agree, changing...

webpack.config.babel.js
anxolin: Right! changing

src/api/weth/WethApi.ts
anxolin: Good!
I've change it in many other places we do it like that too: 7e3d076e9260e42812dea9de9014992bf552484d

src/api/weth/WethApi.ts
anxolin: We don't do it for the other APIs. I don't think we should do it. If you really really feel is an issue, is not a big deal. But this becomes sync code, so no big point for moving those 2 promise together

anxolin: Discussed better in one place ☝️

src/components/DepositWidget/Row.tsx
anxolin: Yep

src/components/DepositWidget/Styled.tsx
anxolin: That's was my thinking
Now we are sure of the currency here. We won't see huge values. These are the top ethereum accounts.
I don't see this accounts of Huobi, Binance as a user here, and with the smart format even would show sth decent

src/components/WrapEtherBtn.tsx
anxolin: I'm happy to remove the link to autofill the amount

I guess the warning it still make sense right? @W3stside , @Velenir , @alfetopito

anxolin: 8a477fe7bf86eaa043913875bb7260337fa1ec48

src/components/WrapEtherBtn.tsx
anxolin: U saw nothin! 🤪 22dc31a5df6892d005db977ab86764415b386111

src/const.ts
anxolin: doesn't matter, I think better to use the mixed case

src/hooks/useWrapUnwrapEth.ts
anxolin: Renamed wrapUnwrapAux to _wrapUnwrap

src/hooks/useBalances.ts
anxolin: Probably, I'll try to divide the hook

anxolin: Addressed here dfba6c26fb761b0ffecdeb4b85e276a040f7c3b1

src/hooks/useWrapUnwrapEth.ts
anxolin: Done: 5589167be40a47ba6bd9c7498c9d505aaafccdb8

src/hooks/useWrapUnwrapEth.ts
anxolin: Done: 5589167be40a47ba6bd9c7498c9d505aaafccdb8

src/components/WrapEtherBtn.tsx
anxolin: Closing, let reopen if u still have comments on the new approach

src/api/weth/WethApi.ts
anxolin: Now it's done differently --> by the way, is like that in the deposit.

Anyways. I think it to initialize the cache object for a network. Right after that, it adds a contract for a token.

src/components/WrapEtherBtn.tsx
anxolin: It's not the core part of this, but it's just taking care of a edge case. I think it's better now that I removed the link for setting the max amount in the WRAPPING

src/components/WrapEtherBtn.tsx
W3stside: WHAT IS THIIIIS

src/components/DepositWidget/index.tsx
```diff
@@ -465,7 +469,8 @@ const BalancesDisplay: React.FC<BalanceDisplayProps> = ({
 const BalancesDisplayMemoed = React.memo(BalancesDisplay)
 
 const DepositWidget: React.FC = () => {
-  const { balances, error } = useTokenBalances()
+  const { ethBalance } = useEthBalances()
+  const { balances, error } = useBalances()
```
alfetopito, line 473: In that case (splitting eth balance from token balances), I'd say keep the old name for the hook (`useTokenBalances`)

anxolin: Granted: 179949ada1f31b266c425a3928dd7092a882454a

PR #853, opened 2020-04-08
Reversing orders sorting. Latest on top

### LINKED ISSUE #820: Default sorting the orders: Newest on top
Invert the current orders. New orders should go on top by default

Related: https://github.com/gnosis/dex-react/issues/822

alfetopito:
Closes #820

Last minute quick fix?

### DIFF
--- a/src/reducers-actions/orders.ts
+++ b/src/reducers-actions/orders.ts
@@ -32,15 +32,31 @@ export const INITIAL_ORDERS_STATE = { orders: [], offset: 0 }
 export const reducer = (state: OrdersState, action: ReducerActionType): OrdersState => {
   switch (action.type) {
     case 'UPDATE_OFFSET':
-    case 'OVERWRITE_ORDERS': {
       return { ...state, ...action.payload }
-    }
-    case 'APPEND_ORDERS': {
+    case 'OVERWRITE_ORDERS': {
       const {
         payload: { orders },
       } = action
+
+      // default sorting order is ascending by order id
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr853--dexreact.review.gnosisdev.com

src/reducers-actions/orders.ts
```diff
@@ -32,15 +32,31 @@ export const INITIAL_ORDERS_STATE = { orders: [], offset: 0 }
...
-    case 'OVERWRITE_ORDERS': {
       return { ...state, ...action.payload }
+    case 'OVERWRITE_ORDERS': {
+      const {
+        payload: { orders },
+      } = action
+
+      // default sorting order is ascending by order id
+      // since we are starting from 0, reverse to have the latest on top
+      orders.reverse()
```
Velenir, line 43: Is it ok to mutate the array here? I guess it's not used anywhere else and only dispatched right after creation, but if it wasn't, mutating would be a bad idea.

W3stside: YOUVE ACTIVATED MY TRAP CARD

anxolin: I'll merge this for the user test, I think we can decide later what to do with this. My initial thought is that is fine, although if there's a risk, we can always create a new instance of the array before reversing.

If we do this, let's do it in another PR

alfetopito: https://github.com/gnosis/dex-react/pull/870

Issue #788, opened 2020-03-30
Make users more aware that they should do a deposit


anxolin:
![image](https://user-images.githubusercontent.com/2352112/77942628-2c3f8580-72bc-11ea-9920-6359d7092260.png)

Add a more clear message when the amount is bigger than the balance.
In the message, there should be a link to the balances page, so they can deposit.

In the future, we should add what it was already in the design, the deposit link. And, for the link in the message, instead of navigating to the balances, we also open the deposit modal right in the trade page:
![image](https://user-images.githubusercontent.com/2352112/77942828-85a7b480-72bc-11ea-8ba2-bf4648a04a94.png)

![image](https://user-images.githubusercontent.com/2352112/77942869-96582a80-72bc-11ea-9f4e-014e264686ac.png)

anxolin:
I think it would help also a lot to handle the specific case of BALANCE == 0

There's a high chance that if the balance is 0, the user is new, or don't know that this 0 means he don't have tokens in his **Exchange Wallet**, he might have them in his **Wallet**

Maybe a `Have you done the deposit?` link below the balance when it's 0 would just do the trick!

cc/ @c3rnst @alfetopito @biocom @Rafanator

PR #857, opened 2020-04-09
Add link to deposit to encourage to do the deposits

anxolin:
I think this PR can help #788 on top of what @biocom is working

What do you think?

It's just a plain link to the wallet, so user can deposit:

image


gnosis-info:
Travis automatic deployment:
 https://pr857--dexreact.review.gnosisdev.com

anxolin:
Okay, it's kind of an experiment I'm ok if it goes or not. Imo is better with this link.
 
I found that wrap ether was a bit weird without a deposit next to it. It was part of the original design, only difference is that is a link to the wallet instead of a pop up.

@biocom what would you suggest. I'm happy either way

biocom:
On one hand I agree it clutters things somewhat as @W3stside pointed out, but at the same time it's interesting to experiment with it. So fine for now with me.

PR #858, opened 2020-04-09
Fix wrong addresses on rinkeby

Velenir:
Can't use `networkIdOrDefault` everywhere after all
Fixes
```
[useTokenBalances] Error for ...
```
on rinkeby

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -391,7 +391,12 @@ const TradeWidget: React.FC = () => {
   const { balances, tokens: tokenList } = useTokenBalances()
 
   // If user is connected, use balances, otherwise get the default list
-  const tokens = isConnected && balances.length > 0 ? balances : tokenList
+  const tokens =
+    isConnected && balances.length > 0
+      ? balances
+      : tokenList.length > 0
+      ? tokenList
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr858--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -391,7 +391,12 @@ const TradeWidget: React.FC = () => {
   const { balances, tokens: tokenList } = useTokenBalances()
 
   // If user is connected, use balances, otherwise get the default list
-  const tokens = isConnected && balances.length > 0 ? balances : tokenList
+  const tokens =
```
W3stside, line 394: Would this logic make more sense inside the hook?

Velenir: no, this is specific to TradeWidget, because inside useTokenBalance we need to have empty array so we don't check balances for wrong addresses
But TradeWidget gets `[sell, receive] = tokenList` and then counts on them being defined

Unless you want a new specific hook only for that one line

W3stside: I DO NOT

PR #861, opened 2020-04-09
Order book: Improve logic of which tokens are selected

### LINKED ISSUE #876: Issues with Order book
* [X] It looks like now is ignoring the network (uses always mainnet): This is why I have this still as a draft https://github.com/gnosis/dex-react/pull/872
* [X] Also, we should prevent to select the same token (do a swap if the user selects the same token) #861
* [X] Still don't work as intenderd: See https://github.com/gnosis/dex-react/pull/861
* [X] Good to have: https://github.com/gnosis/dex-react/pull/868

anxolin:
This PR improves some logics in the orderbook that were causing some bad UX.

It solves this 3 points:
* [X] Do not allow to select the same token as quote and base. Now if a user picks the base token as quote token, or the other way around, the tokens will swap; i.e. If we have WETH-DAi and we select DAI instead of WETH, the order book will show DAI-WETH.
* [X] Restore the defaults when closing the modal: The modal opens with the default selection in the trade widget. But the user can change the selection in the order book. When an user do that, if he then closes the modal, and open it again, it should have the selection of the trading widget, not the last selection from the modal.
* [X] Any change in the trading widget should change the default in the modal

Closes #876


gnosis-info:
Travis automatic deployment:
 https://pr861--dexreact.review.gnosisdev.com

PR #870, opened 2020-04-09
Not reversing orders array in place because that's bad

alfetopito:
Follow up to https://github.com/gnosis/dex-react/pull/853#discussion_r406006081

### DIFF
--- a/src/reducers-actions/orders.ts
+++ b/src/reducers-actions/orders.ts
@@ -40,9 +40,9 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
 
       // default sorting order is ascending by order id
       // since we are starting from 0, reverse to have the latest on top
-      orders.reverse()
+      const reversed = orders.slice(0).reverse()
 
-      return { ...state, orders }
+      return { ...state, orders: reversed }
     }
     case 'APPEND_ORDERS': {
       const {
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr870--dexreact.review.gnosisdev.com

src/reducers-actions/orders.ts
```diff
@@ -51,10 +51,10 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
       const { orders: currentOrders } = state
 
       // reverse new orders
-      newOrders.reverse()
+      const reversed = newOrders.slice(0).reverse()
```
anxolin, line 54: ubernit: `newOrdersReversed`? so it's more clear when you read line 57

src/reducers-actions/orders.ts
```diff
@@ -40,9 +40,9 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
 
       // default sorting order is ascending by order id
       // since we are starting from 0, reverse to have the latest on top
-      orders.reverse()
+      const reversed = orders.slice(0).reverse()
```
Velenir, line 43: You don't need `0` in `array.slice()`, not that it matters

alfetopito: I read on [the internet this is faster](https://stackoverflow.com/a/21514254/1272513), so it must be true.

Velenir: Must be true then, no other way around it. Internet is the final authority

alfetopito: Exactly! You get me.

PR #872, opened 2020-04-09
Order Book: Show network name for testnets

anxolin:
I left this commit/branch unpushed, pushing now :) 

Mainnet:
image

Rinkeby:
image

If you are in a testnet (Rinkeby) and you are not very aware, the Order Book can be weird, because it uses the data from the smart contract in that network.

To make it more clear, I just show the name of the testnet if we are in a testnet, otherwise I don't show the name of the network (implying mainnet)


anxolin:
Weird! This used to work. It looks like now the order book uses always mainnet for some reason. Need to look at this in another moment!

gnosis-info:
Travis automatic deployment:
 https://pr872--dexreact.review.gnosisdev.com

PR #878, opened 2020-04-10
Fix 875 token address on rinkeby

Velenir:
Even after numerous workarounds by me and @alfetopito  the problem remained
Problem is stemming from the fact that oftentimes on network change `newBlockHeaders` fires first, before `onNetworkChanged`. Sync nature of `useTokenList` didn't help either.
So we get stale network in `WalletInfo`

This solution is hacky, but so is the problem

### DIFF
--- a/src/hooks/useTokenList.ts
+++ b/src/hooks/useTokenList.ts
@@ -16,7 +16,7 @@ export const useTokenList = (networkId?: number): TokenDetails[] => {
 
   useEffect(() => {
     return subscribeToTokenList(() => forceUpdate({}))
-  }, [networkId, forceUpdate])
+  }, [forceUpdate])
 
   return tokens
 }
--- a/src/hooks/useTokenList.ts
+++ b/src/hooks/useTokenList.ts
@@ -8,12 +8,16 @@ import { getTokens, subscribeToTokenList } from 'services'
 const emptyArray: TokenDetails[] = []
 
 export const useTokenList = (networkId?: number): TokenDetails[] => {
-  const [tokens, setTokens] = useSafeState(networkId === undefined ? emptyArray : getTokens(networkId))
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr878--dexreact.review.gnosisdev.com

src/hooks/useTokenList.ts
```diff
@@ -8,12 +8,15 @@ import { getTokens, subscribeToTokenList } from 'services'
...
+  const tokens = networkId === undefined ? emptyArray : getTokens(networkId)
+
+  // force update with a new value each time
+  const [, forceUpdate] = useSafeState({})
 
   useEffect(() => {
-    if (networkId === undefined) return setTokens(emptyArray)
-    return subscribeToTokenList(setTokens)
-  }, [networkId, setTokens])
+    return subscribeToTokenList(() => forceUpdate({}))
```
alfetopito, line 18: I don't get this.

The callback to the subscription is a function that updates an empty state of something that's ignored?

:confused:

Velenir: Just forces rerendet. seState(new value), and object is airways new, can use symbol too, or array, or counter +1

alfetopito: Complicated, but now I got it :)

anxolin: Sorry, I don't get it either 🤔

On every change of the token list, we call `forceUpdate` wich it looks doesn't have an effect, because no-one uses his value. The value is ignored.

That `useEffect ` is a bit strange to me, depends on `forceUpdate` function, and that doesn't change for a component using this hook. It made more sense to me, the previous hook, where were we had: `subscribeToTokenList(setTokens)`, so I'm a bit lost.

PR #883, opened 2020-04-13
Remove tooltip in trade widget

anxolin:
We agreed on removing the tooltips for the trade widget, they are too intrusive

This PR:
* [X]  Removes the tooltips for sell token and buy token in the trade widget
* [X] Removes the tooltip in the spread of the liquidity page
* [X] Extract the styled component `Input` from component `InputWithTooltip.tsx`. The idea is to have our own custom input instead of styling it everty time. In the future, maybe we want to implement some features in the input too, but for now is good enough to promote it to "Component

### DIFF
--- a/src/components/Input.tsx
+++ b/src/components/Input.tsx
@@ -1,0 +1,17 @@
+import styled from 'styled-components'
+
+export const Input = styled.input<{ $error?: boolean }>`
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr883--dexreact.review.gnosisdev.com

Issue #867, opened 2020-04-09
Text for deposit button 


Rafanator:
Hi @anxolin, Anna Chris and I agreed on the following text for the deposit button:

`The amount you deposit into your exchange wallet is your liquidity: at least deposit one stablecoin on ‘Balances’. `

Thank you!

anxolin:
@biocom this is the a quick fix feedback BD gave to improve the Liquidity page.
The complaint was that users are not aware that they need to deposit before creating the liquidity orders.

What we discuss is please @Rafanator double confirm me this, you want this texts in both STEP 1, and STEP 2, highlighted, and with a link to the Wallet.

 Is this correct?
cc/ @c3rnst

Rafanator:
Hi everyone! @c3rnst can you please confirm if @anxolin's message above is correct? 

Thanks!

c3rnst:
Hi @Rafanator  - this is what was the conclusion in the meeting with yourself, Anna and myself. I believe the suggestion was a framed text at the bottom left of both screens.

PR #885, opened 2020-04-14
Add liquidity message.

biocom:
For ticket #867

- Added message with a link to the Balances (wallet) page.
- Made the inputfield have a white background. In my opinion the grey inputfield inside the grey container wasn't clear.

Screenshot 2020-04-14 at 11 21 30 

Screenshot 2020-04-14 at 11 21 37

### DIFF
--- a/src/components/PoolingWidget/DefineSpread.styled.ts
+++ b/src/components/PoolingWidget/DefineSpread.styled.ts
@@ -15,7 +15,7 @@ export const DefineSpreadWrapper = styled(InputBox)`
     width: 100%;
     height: 5.6rem;
     padding: 0 1rem;
-    background: white;
+    background: var(--color-background-input-lighter);
 
     &.warning {
       border-color: orange;
--- a/src/components/PoolingWidget/DefineSpread.styled.ts
+++ b/src/components/PoolingWidget/DefineSpread.styled.ts
@@ -15,6 +15,7 @@ export const DefineSpreadWrapper = styled(InputBox)`
     width: 100%;
     height: 5.6rem;
     padding: 0 1rem;
+    background: white;
 
     &.warning {
       border-color: orange;
--- a/src/components/PoolingWidget/index.tsx
+++ b/src/components/PoolingWidget/index.tsx
@@ -294,7 +294,7 @@ const PoolingInterface: React.FC = () => {
             <LiquidityMessage>
               <p>
                 Your liquidity is equal to the amount you have deposited into your exchange wallet.
-                <br />{' '}
+                <br />
                 <b>
                   Be sure to deposit at least one stablecoin on the <a href="/wallet">Balances</a> tab.
                 </b>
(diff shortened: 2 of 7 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr885--dexreact.review.gnosisdev.com

W3stside:
> Great! Much clearer now.
> 
> Regarding the message, Kei gave us some feedback and changed the message slightly. Would you change it before merge to:
> 
> `Your liquidity is equal to the amount you have deposited into your exchange wallet. Be sure to deposit at least one stablecoin on the Balances tab.`

I would say:
`Your liquidity is equal to the amount you have deposited into your exchange wallet. Be sure to deposit at least one stablecoin via the Balances tab.`

src/components/PoolingWidget/DefineSpread.styled.ts
```diff
@@ -15,6 +15,7 @@ export const DefineSpreadWrapper = styled(InputBox)`
     width: 100%;
     height: 5.6rem;
     padding: 0 1rem;
+    background: white;
```
W3stside, line 18: can you make this `var(...)` for dark mode

biocom:
Updated message:
Screenshot 2020-04-14 at 12 00 50

src/components/PoolingWidget/DefineSpread.styled.ts
biocom: Ah, sorry forgot to change, will do now.

src/components/PoolingWidget/index.tsx
```diff
@@ -272,6 +291,16 @@ const PoolingInterface: React.FC = () => {
               <SubComponents step={step} {...restProps} />
             </ContentWrapper>
 
+            <LiquidityMessage>
+              <p>
+                Your liquidity is equal to the amount you have deposited into your exchange wallet.
+                <br />{' '}
```
Velenir, line 297: Why `{' '}`? It gets collapsed and you don't see it anyway

src/components/PoolingWidget/index.tsx
```diff
@@ -272,6 +291,16 @@ const PoolingInterface: React.FC = () => {
...
+                <b>
+                  Be sure to deposit at least one stablecoin on the <a href="/wallet">Balances</a> tab.
```
Velenir, line 299:
```suggestion
                  Be sure to deposit at least one stablecoin on the <Link to="/wallet">Balances</Link> tab.
```
use router, don't reload page

src/components/PoolingWidget/index.tsx
biocom: Think it was added by the yarn lint fix. Can remove.

src/components/PoolingWidget/index.tsx
biocom: Thanks, will update. I think we should also do it in the deposit message (in the trade form).

biocom:
Updated to use React router on two occasions (deposit and liquidity page).

biocom:
Thanks for the reviews!

PR #886, opened 2020-04-14
fixes saved trade params & second tx error

### LINKED ISSUE #874: Trade Widget broken on the second transaction
Maybe a `git bisect` would point us where we introduced this bug so we can fix it

Issue is than after we send one transaction, then the submit button don't work any more

W3stside:
closes #874 

Had to fix bug with resetting `validFrom` `validUntil` when setting `ASAP` or `Never`, respectively

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -677,8 +677,8 @@ const TradeWidget: React.FC = () => {
                   ...DEFAULT_FORM_STATE,
                   price,
                   priceInverse: invertPriceFromString(price),
-                  validFrom: formatTimeToFromBatch(validFrom, 'TIME').toString(),
-                  validUntil: formatTimeToFromBatch(validUntil, 'TIME').toString(),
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr886--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -719,7 +719,7 @@ const TradeWidget: React.FC = () => {
                   price,
                   priceInverse: invertPriceFromString(price),
                   validFrom: formatTimeToFromBatch(validFrom, 'TIME').toString(),
```
anxolin, line 721: Why this `validFrom` doesn't need the fix, but the one at 680 does?

Also, a nitpick would be to create a const with a name like `const isAsap = validFrom === 0` and `doesNotExpire = validUntil === 0` so it's more clear the semantics we are using and it's easier to read

W3stside: because the if block above - ok will change it

PR #890, opened 2020-04-14
Update infura key and set some env vars

### LINKED ISSUE #871: Make sure the Node URL is correct
@Velenir , pls Make sure we don't forget about this https://gnosisinc.slack.com/archives/C6Z2XNL5Q/p1586431893029700

anxolin:
Let's not use Dima's API key anymore, once this is in production is a good idea Dima to delete you leaked API key. Thanks for borrowing it to us until now :) 

* It uses the API key devops made for us
* It uses an optional env var for the API key and for the URL. Is not used, but it's just a convenient way to be able to change the URL without generating a new version

Closes #871

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -27,12 +27,11 @@ import {
   TOKEN_8,
 } from '../../test/data'
 import Web3 from 'web3'
-import { INITIAL_INFURA_ENDPOINT } from 'const'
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr890--dexreact.review.gnosisdev.com

PR #893, opened 2020-04-14
Volume: Leave only accumulated volume in the tooltip

anxolin:
The current order book looks like this:
image

This PR is a proposal to remove the "Volume" and just show the "Accumulated volume" like in this picture:
image

The "plain Volume" is the sum of orders at that price point (not the sum of that price point and all better price points). Imo, the "plain volume" is not super relevant and make the UX more confusing. It was very confusing for me that the "Volume" was not a monotonous function, so you could see that even if you go left in the bids, the volume can decrease. 

As a user, I'd be interested to get how much I can sell for a give price. So my understanding is that I only need the accumulated volume, that in this PR is renamed as just "Volume"

Additionally, as it can also be seen in the picture, this PR adds the currency for the volume. 
Units can always be confusing, because although normally the base token is used to measure the volume, in a market like ETHUSD, normally the volumes are given in USD, so not sure if an user might expect ETHUSDC to have the volumes expressed in USDC. In any case, adding units, makes it more clear.

Lastly, in this PR, I make use of the function `safeTokenName` since we cannot assume the token will have a symbol.

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -164,7 +164,7 @@ const draw = (
   bidCurve.fill = bidCurve.stroke
   bidCurve.startLocation = 0.5
   bidCurve.fillOpacity = 0.1
-  bidCurve.tooltipText = 'Bid: [bold]{categoryX}[/]\nTotal volume: [bold]{totalVolume}[/]\nVolume: [bold]{volume}[/]'
+  bidCurve.tooltipText = 'Bid: [bold]{categoryX}[/]\nVolume: [bold]{totalVolume}[/]'
 
   const askCurve = chart.series.push(new am4charts.StepLineSeries())
   askCurve.dataFields.categoryX = 'price'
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr893--dexreact.review.gnosisdev.com

biocom:
If the volume bars are all the same/are always full height, I'd question the need for having them at all. Better to remove?

E.g. Coinbase:
Screenshot 2020-04-14 at 14 16 20


Kraken:
Screenshot 2020-04-14 at 14 16 42

fleupold:
👍 

Agree with @biocom, you should be able to remove the ColumnSeries completely from the graph as they were just representing the volume at that price point.

Also agree with this PR that it's less confusing like this.

anxolin:
Good point, removing also the volume bars:

image

Rafanator:
Agree, leave only accumulated volume.

Rafanator:
Sorry accidentally closed, meant to only comment.

PR #896, opened 2020-04-14
Misc style fixes

Velenir:
Fixes invisible Close `X` button in Modali

Also fixes Order Settings button visibility transition (which happened previously, was fixed at some point and now broke again), by explicitly specifying which props to transition

ezgif-3-1ac391642c82


gnosis-info:
Travis automatic deployment:
 https://pr896--dexreact.review.gnosisdev.com

Issue #909, opened 2020-04-14
Order is filled but interface does not show it


alfetopito:
Once the page is refreshed though, the order is displayed as filled.

---

I suspect the reason is due to order caching.

Since on every new block we only query for new orders from the last point we checked, we'll never know the order has been matched.

See https://github.com/gnosis/dex-react/blob/develop/src/hooks/useOrders.ts#L91

We might need to either lose the optimization or maybe subscribe to order placement/matching events and update only the ones that changed.

Velenir:
Should we consider this fixed?

alfetopito:
Well, we did a temporary fix...
Should we have a long term fix updating orders by listening to trade events?

Velenir:
I'm not sure that kind of subscription will even work with some mobile wallets. Need to check

alfetopito:
Another improvement I thought about (considering my horrible first fix attempt) that might be useful to avoid flickering when we fetch orders from start is to have the option to merge the current list of orders.

I'm thinking about moving the filtering to the global state, and there (when merging) update the existing orders, removing ones that have potentially been cancelled.

This way, the list is always constant, while only modified orders are updated/removed.

What do ya think?

alfetopito:
Since this issue is fixed, I'll close it.

Opened a new one to address what I have in mind https://github.com/gnosis/dex-react/issues/948

PR #910, opened 2020-04-14
Refreshing orders on every new block: temporary fix

alfetopito:
Quick fix for #909 

Not an optimal fix as this is a bad UX, refreshing the list at every new block.

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -51,12 +51,23 @@ export function useOrders(): Result {
   // can only start loading when connection is ready. Keep it `false` until then
   const [isLoading, setIsLoading] = useSafeState<boolean>(false)
 
+  // allow to fresh start/refresh on demand
+  const forceOrdersRefresh = useCallback((): void => {
+    dispatch(updateOffset(0))
+    setIsLoading(true)
+  }, [dispatch, setIsLoading])
+
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr910--dexreact.review.gnosisdev.com

alfetopito:
> If I understand the problem/solution correct I'm not sure we should do this.
> 
> Fetching all orders can easily be lot's of requests, for example, when we have some liquidity orders. Fetching all orders over and over every 15s might be a bit overkilling.
> If this is using Metamask, or any infura RPC node, it can easily hit rate limits.
> 
> Maybe we can re-think this in the future, maybe using events?

Yes, definitely.
This is just a quick last minute fix as Martin raised this as a severe UX issue.

We for sure should deal with this a better way.

You can follow the thread on slack https://gnosisinc.slack.com/archives/CSGLUC1BN/p1586898945012800

anxolin:
What about having the same strategy as we have already, and **force re-fetch** once every 5min, precisely at minute 4 of every batch.  Still not ideal, but save 19 full refetches every 5min

Velenir:
> What about having the same strategy as we have already, and **force re-fetch** once every 5min, precisely at minute 4 of every batch. Still not ideal, but save 19 full refetches every 5min

I support this

anxolin:
superseeded by #916

PR #912, opened 2020-04-15
Fix meta viewport

biocom:
Fixed by modifying the meta viewport setting in the main html file.

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -42,7 +42,7 @@
   </script>
   <% } %>
   <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


biocom:
@anxolin Tested with Xcode iOS simulator. Please test on your device to verify the fix.

gnosis-info:
Travis automatic deployment:
 https://pr912--dexreact.review.gnosisdev.com

Velenir:
Does `maximum-scale=1` mean mobile users can't zoom in at all?
https://a11yproject.com/posts/never-use-maximum-scale/

I'm not sure it's a good idea

PR #914, opened 2020-04-15
908/lower threshold

### LINKED ISSUE #908: Reduce `Filled` order threshold to 0.01%
The order is considered filled when less than 0.0001% is left

Reduce the threshold to `0.01`%

See https://github.com/gnosis/dex-react/blob/release%2F0.11/src/utils/miscellaneous.ts#L79
and https://github.com/gnosis/dex-react/blob/release%2F0.11/src/const.ts#L22

W3stside:
Closes #908 

Threshold `0.0001` > `0.01`


gnosis-info:
Travis automatic deployment:
 https://pr914--dexreact.review.gnosisdev.com

PR #916, opened 2020-04-15
Force resfresh on batch time

### LINKED ISSUE #909: Order is filled but interface does not show it

Velenir:
Fixes #909 (more of a workaround than a fix)

Causes rerenders only when needed
Doesn't check on every second but rather being smart about initial timeout + intervals

To test revert a369fb5 with helpful logs still included

### DIFF
--- a/src/hooks/useTimeRemainingInBatch.ts
+++ b/src/hooks/useTimeRemainingInBatch.ts
@@ -49,7 +49,6 @@ export function useCheckWhenTimeRemainingInBatch(
     let id: number
 
     if (currentCheck.checkTime) {
-      // setCheckResult(currentCheck)
       callbackRef.current(currentCheck)
     }
 
--- a/src/hooks/useTimeRemainingInBatch.ts
+++ b/src/hooks/useTimeRemainingInBatch.ts
@@ -52,7 +52,7 @@ export function useCheckWhenTimeRemainingInBatch(
       callbackRef.current(currentCheck)
     }
 
-    // const nextCheckInSeconds = currentCheck.secondsRemaining + checkIntervalSeconds
+
     // first check on timeout not interval
     // as we start at random time in the BATCH
     const nextCheckInSeconds =
@@ -69,7 +69,7 @@ export function useCheckWhenTimeRemainingInBatch(
     }
 
     id = window.setTimeout(() => {
-      // not can go on interval
+      // now can go on interval
       id = window.setInterval(checkAndReport, BATCH_TIME * 1000)
 
       checkAndReport()
--- a/src/hooks/useTimeRemainingInBatch.ts
+++ b/src/hooks/useTimeRemainingInBatch.ts
@@ -70,7 +70,7 @@ export function useCheckWhenTimeRemainingInBatch(
 
     id = window.setTimeout(() => {
       // not can go on interval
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr916--dexreact.review.gnosisdev.com

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -13,3 +14,70 @@ export function useTimeRemainingInBatch(): number {
...
+  callback: (SResult: SecondsRemainingResult) => void,
+): void {
+  // allow for fluid callback change without resetting interval
+  const callbackRef = useRef(callback)
+  callbackRef.current = callback
+
+  useEffect(() => {
+    // how often to check
+    // when there are ${seconds} seconds left in the batch
+    const checkIntervalSeconds = BATCH_TIME - seconds
```
anxolin: nit: you are effectively operating 2 constants, I understand the idea behind doing this "interval calculation" (you want to minimize the checks). I get why you do the check in line 43, but I think that console.warn should be a throw or at least a more noisy error, since this case would be clearly a misconfiguration

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -13,3 +14,70 @@ export function useTimeRemainingInBatch(): number {
...
+
+    const currentCheck = checkIfTime(seconds)
+    let id: number
+
+    if (currentCheck.checkTime) {
+      // setCheckResult(currentCheck)
+      callbackRef.current(currentCheck)
+    }
+
+    // const nextCheckInSeconds = currentCheck.secondsRemaining + checkIntervalSeconds
```
anxolin: ubernit: It looks super effective ;) 
I was expecting sth more like check evey X secons for every batch. Your solution is much fancier :) 

Anyways, I would add an example in the comments, to make it more clear what you try to do. If the check is fired at minute 4:05, and if it's fired at minute 2:10 (i.e. at boot time)

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -13,3 +14,69 @@ export function useTimeRemainingInBatch(): number {
...
+        : // if not yet check time -- what's left to that time
+          currentCheck.secondsRemaining - seconds) + 1 // plus one second
+    // because condition in contract is >= 1 minutes and we check for <
+    // so we don't fall on the exact second and query blockchain
+    const checkAndReport = (): void => {
+      const currentCheck = checkIfTime(seconds)
+      callbackRef.current(currentCheck)
+    }
+
+    id = window.setTimeout(() => {
```
W3stside, line 71: why `window`? for global reference?

Velenir: TS complained about `NodeJS.Timeout` type from global `setTimeout`

anxolin, line 71: Timeout + set interval, I have to admit, it took me a bit to get my head around it

PR #924, opened 2020-04-15
918/url encode token symbols

### LINKED ISSUE #918: URL encode token symbols
Our assumptions about token symbols are incomplete.
For example, tokens can have `/` in their symbols:

> Just added a new token: 0x6c79f10543c7886c6946b8a996f824e474bac8f2
I am afraid it breakes the interface because it contains ” / ” in its name.

> just try it. It will forward to this link: https://mesa.eth.link/#/trade/oETH%20$100%20Put%2004/24/20-USDC?sell=0&price=1.01548&from=undefined&expires=2880
and this link will no longer load the side.

> Screen Shot 2020-04-15 at 01 17 02

alfetopito:
Closes #918 

Tested using `yarn mock` and selecting the new `Fake token`:

screenshot_2020-04-15_14-05-34

screenshot_2020-04-15_14-20-04

Certainly does not look good, but works now.

Not part of this PR: to make it look good.

### Edit 1:

Updated how we encode a possible dash (`-`) in the token symbol to avoid issues with the token symbol separator in the URL (`/trade/tokenA-tokenB/...`)

screenshot_2020-04-17_10-29-23

Inserting a circle dash (`⊝`) to remove ambiguity and keep somewhat the same semantics.

*Note*: this only affects the displayed URL, nothing else.

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -447,6 +447,26 @@ const chooseTokenWithFallback = ({
   )
 }
 
+function buildUrl(params: {
+  sell: string
+  price: string
+  from: string
+  expires: string
+  sellSymbol: string
+  buySymbol: string
+}): string {
+  const { sell, price, from, expires, sellSymbol, buySymbol } = params
+
+  const searchQuery = buildSearchQuery({
+    sell,
+    price,
+    from,
+    expires,
+  })
+
+  return `/trade/${encodeURIComponent(sellSymbol)}-${encodeURIComponent(buySymbol)}?${searchQuery}`
+}
+
 const TradeWidget: React.FC = () => {
   const { networkId, networkIdOrDefault, isConnected, userAddress } = useWalletConnection()
   const { connectWallet } = useConnectWallet()
@@ -469,7 +489,9 @@ const TradeWidget: React.FC = () => {
       : tokenListApi.getTokens(networkIdOrDefault)
 
   // Listen on manual changes to URL search query
-  const { sell: sellTokenSymbol, buy: receiveTokenSymbol } = useParams()
+  const { sell: dirtySellTokenSymbol, buy: dirtyReceiveTokenSymbol } = useParams()
+  const sellTokenSymbol = decodeURIComponent(dirtySellTokenSymbol || '')
+  const receiveTokenSymbol = decodeURIComponent(dirtyReceiveTokenSymbol || '')
   const {
     sellAmount: sellParam,
     price: priceParam,
@@ -621,13 +643,15 @@ const TradeWidget: React.FC = () => {
     setValue(receiveInputId, calculateReceiveAmount(priceValue, sellValue))
   }, [priceValue, priceInverseValue, setValue, receiveInputId, sellValue])
 
-  const searchQuery = buildSearchQuery({
+  const url = buildUrl({
     sell: sellValue,
     price: priceValue,
     from: validFromValue,
     expires: validUntilValue,
+    sellSymbol: sellToken.symbol as string,
+    buySymbol: receiveToken.symbol as string,
   })
-  const url = `/trade/${sellToken.symbol}-${receiveToken.symbol}?${searchQuery}`
+  // Updates page URL with parameters from context
   useURLParams(url, true)
 
   // TESTING
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -57,6 +57,7 @@ import { tokenListApi } from 'api'
 
 import validationSchema from './validationSchema'
 import { useBetterAddTokenModal } from 'hooks/useBetterAddTokenModal'
+import { encodeTokenSymbol, decodeSymbol } from '@gnosis.pm/dex-js'
 
 const WrappedWidget = styled(Widget)`
   overflow-x: visible;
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr924--dexreact.review.gnosisdev.com

Velenir:
We still have a major problem with this approach
While `USDC-FTK**+|-` works, the reverse doesn't. Try http://localhost:8080/trade/FTK%20%2F21%2F10-DAI%20%2499%20%26%20%2B%7C--USDC?sell=0&price=0.99203&from=&expires=2880 See the double-dash there, herein lies the problem

The problem is that `-`(dash) is not encoded and we use dash for url params in `<Route path="/trade/:sell-:buy"`

Now that means we either use only encodable symbols for `:sell-:buy` separation, for example `:sell|:buy`
Or we encode it ourselves, as `%2D`, that is `str.replace(/-/g, '%2D')`. decoding though works properly, no need to change it (`decodeURIComponent('%2D') === decodeURI('%2D') === '-'`)

I vote for the second approach

Velenir:
Also, will we need this logic in telegram bot for url formatting? Eventually

alfetopito:
@Velenir agree and agree.
Will move this logic up to `dex-js` with tests.

As for the approach, I would favor something that makes the URL readable.

> Now that means we either use only encodable symbols for :sell-:buy separation, for example :sell|:buy
> Or we encode it ourselves, as %2D, that is str.replace(/-/g, '%2D'). decoding though works properly, no need to change it (decodeURIComponent('%2D') === decodeURI('%2D') === '-')

Wouldn't option 2 make the URL (DAI, USDC for example) be `/DAI%2DUSDC?sell=1&...`?

Velenir:
No, no. A token symbol `T-KEN` encoded becomes `T%6DKEN`and that's it. The URL stays `T%2DKEN-DAI`. You only decide encore symbols. And replace dashes in symbols not entire url

src/components/TradeWidget/index.tsx
```diff
@@ -425,6 +425,26 @@ const preprocessTokenAddressesToAdd = (addresses: (string | undefined)[], networ
...
+  const { sell, price, from, expires, sellSymbol, buySymbol } = params
+
+  const searchQuery = buildSearchQuery({
+    sell,
+    price,
+    from,
+    expires,
+  })
+
+  return `/trade/${encodeURIComponent(sellSymbol)}-${encodeURIComponent(buySymbol)}?${searchQuery}`
```
Velenir, line 445: Here's where you'll do `${encodeURIComponent(symbol).replace(dash, ...)}`, not the whole url

alfetopito: ohhhhh got it!

yep, agree :)

alfetopito: What if the browser disagrees with encoding `-` and replaces `%2D` with a dash anyway in the URL? :/

Velenir: What do you mean disagrees? Browser doesn't do that. What you put in is what gets displayed. Browser doesn't care for anything except url specific symbols like ?#/ that delineate path/?search#hash and locale specific letters ( urls in languages other than English, which is a horrible idea IMHO)

Velenir: % in url is just a %, to decode or take at face value is up to us not browser. Even react-router doesn't automatically decode url params, though it's a frequent subject of issues in their repo

alfetopito: Well, something is changing `%2D` back to a `-`.
Try it out now on localhost.

I left a console.log to print the generated URL.

That has the escaped version, but the url still shows a `-`.
screenshot_2020-04-16_14-15-51
screenshot_2020-04-16_14-16-00

Velenir: Shit, you are right, browser auto-decodes `-~'` symbols, so we can't rely on replacing them
Then we'll have to change what we use in `path="/trade/:sell-:buy"`, can choose from `$&`:<>[]{}"+%@/;=\^|\',` symbols
```sh
> encodeURIComponent(' $&`:<>[]{}"+#%@/;=?\^|,~\'-')
"%20%24%26%60%3A%3C%3E%5B%5D%7B%7D%22%2B%23%25%40%2F%3B%3D%3F%5E%7C%2C~'-"
```
As you can see `-~'` stayed unencoded

on the bright side, won't have to do `replace(dash`

Velenir: @anxolin @alfetopito @W3stside 
need to decide if we are changing url scheme because some tokens may include dash `-` and browser auto-decodes it from under our feet

Velenir: See #936 for a working example

anxolin: I think this is an edge case, I wouldn't change our beautiful scheme just for that.
If there's not a better solution, I even prefer to do the decoding in two steps using, using `%_2D` instead of `%2D` to encode the parameters
But proably there's a more elegant solution....

alfetopito: As I mentioned on https://github.com/gnosis/dex-react/pull/936#issuecomment-615295388.
I'm in favor of using `|` instead, keeping keeping a redirect rule for `-` for backward compatibility.

Although, Anxo's proposal might work...
`replace('-', '%dashdonotreplaceyoudamnpokyingbrowser')`

alfetopito: Much better, right?

screenshot_2020-04-17_10-29-23

PR #928, opened 2020-04-16
Add sUSD to liquidity and allow to modify the list using env vars

anxolin:
* Adds sUSD
* Adds an env var to easily mantain this list

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -102,7 +102,8 @@ export const GP_ORDER_TX_HASHES = {
   4: 'GP_ORDER_TX_HASHES_4',
 }
 
-export const LIQUIDITY_TOKEN_LIST = new Set(['USDT', 'TUSD', 'USDC', 'PAX', 'GUSD', 'DAI'])
+const LIQUIDITY_TOKEN_LIST_VALUES = process.env.LIQUIDITY_TOKEN_LIST || 'USDT,TUSD,USDC,PAX,GUSD,DAI,sUSD'
+export const LIQUIDITY_TOKEN_LIST = new Set(LIQUIDITY_TOKEN_LIST_VALUES.split(',').map(symbol => symbol.trim()))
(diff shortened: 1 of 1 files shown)


src/const.ts
```diff
@@ -102,7 +102,8 @@ export const GP_ORDER_TX_HASHES = {
   4: 'GP_ORDER_TX_HASHES_4',
 }
 
-export const LIQUIDITY_TOKEN_LIST = new Set(['USDT', 'TUSD', 'USDC', 'PAX', 'GUSD', 'DAI'])
+const LIQUIDITY_TOKEN_LIST_VALUES = process.env.LIQUIDITY_TOKEN_LIST || 'USDT,TUSD,USDC,PAX,GUSD,DAI,sUSD'
+export const LIQUIDITY_TOKEN_LIST = new Set(LIQUIDITY_TOKEN_LIST_VALUES.split(',').map(symbol => symbol.trim()))
```
Velenir, line 106: `trim` just in case?

Velenir: that makes me think, can there be token symbols with a space?
yes there can
but probably not starting or ending with (trimmable) space, that would be ridiculous

anxolin: yes, because anyone can add a comma separated list of tokens. So in case there's a space somewhere, it won't break it

gnosis-info:
Travis automatic deployment:
 https://pr928--dexreact.review.gnosisdev.com

src/const.ts
Velenir: what if there's a token with comma in symbol, `T,KEN`, the horror :scream:

anxolin: What if the token is a whitespace??

alfetopito: >What if the token is a whitespace?? 
Doesn't deserve to be listed.

PR #929, opened 2020-04-16
Fix stale tokens

Velenir:
When switching from mainnet to rinkeby old tokens stay in TokenSelector even if they are not available on rinkeby
Same when starting on rinkeby with URL containing mainnet-only token
try https://dex.dev.gnosisdev.com/trade/DAI-PAN?sell=0&price=0&from=&expires=2880 on rinkeby

ezgif-3-19cb5151a038

This PR adds a check for tokens from url on network change

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -499,6 +499,41 @@ const TradeWidget: React.FC = () => {
       defaultTokenSymbol: 'USDC',
     }),
   )
+
+  useEffect(() => {
+    console.log('TRADE', trade)
+    if (trade.sellToken) {
+      const sellTokenOrFallback = chooseTokenWithFallback({
+        token: tokenListApi.hasToken({ tokenAddress: trade.sellToken.address, networkId: networkIdOrDefault })
+          ? trade.sellToken
+          : null,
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr929--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -425,6 +425,28 @@ const preprocessTokenAddressesToAdd = (addresses: (string | undefined)[], networ
...
+
+const chooseTokenWithFallback = ({
+  tokens,
+  token,
+  tokenSymbol,
+  defaultTokenSymbol,
+}: ChooseTokenInput): TokenDetails => {
+  return (
+    token ||
+    (tokenSymbol && isAddress(tokenSymbol?.toLowerCase())
```
alfetopito, line 443: ~Why check if `tokenSymbol` is an address? Typo?~

Ohhh, because that's what's coming from the URL.

Can you maybe rename this var to something less confusing? Like `tokenIdentifierFromUrl` (too long) or `identifier` or `tokenSymbolFromUrl`. Hard to find a good name...

PR #930, opened 2020-04-16
797/improve orders state

### LINKED ISSUE #797: Improve state of the orders
image

Chris wrote in slack:
----------
We just had a good feedback with Martin and there is one change that should be prioritised over the adding-addresses to be included in this last pre-launch-sprint: 
It’s about seeing the status on submitted orders:

1. Transaction submitted --> Same as now.
2. Waiting to be included in the first batch, rather than saying Active, change to: “Active starting next batch: COUNTDOWN”
3. In the very first batch the order is part of: “Waiting for settlement: COUNTDOWN” (counting down the 4 minutes, not the 5).
4. From the second batch onwards, as usual: filled, partially filled and active. 
Regarding this, the two COUNTDOWNS are optionally and it would be good to see how cluttered it would look in the tables. Any chance, @Michel Bio could mock this up?
5. ~~Additionally, if not much work, fully filled orders should move to the tab “closed”.~~ (addressed on https://github.com/gnosis/dex-react/issues/907)
Do you know what we mean? Else, let’s have a quick chat about it.

Martin wrote:
------------------

A user comes to the exchange and wants to trade token A for B. They don’t want to place a order and see if it matches - they simply really want to have token B. So the idea is that you order should be matched immediately - if it is not matched immediatly the user want to submit the order again with a different price so that it gets matched.
Currently the user has no idea how long they need to wait untill they should expect a match. The answer is: 4-9 minutes after your order was mined. However - wether the order was mined at 4:59 ( last second of the batch) or at 5:01 (in the next batch) - you don’t see that. So it is really hard to get this info how long you should wait for a match.
The infos we are suggesting to add here should make it much easier to understand what is going on and when to expect a match.

alfetopito:
Closes #797 

Step 1: pending

screenshot_2020-04-16_16-12-45

Step 2: Active next batch

screenshot_2020-04-16_16-15-00

Step 3: Looking for match

screenshot_2020-04-16_16-15-20

From then on, follow regular states

### Update 1

- [x] `Looking for match` countdown stops at 4min. `Active` from then on.
- [x] Renamed `Looking for match` to `Waiting for settlement`
- [x] `clearInterval` instead of `clearTimeout`
- [x] Small refactoring
- [x] Order row component now only updates once time is up
- [x] New Status countdown component updates every second using `useTimeRemainingInBatch` hook
- [x] Starting countdowns at half a second mark, to try and keep all in sync

### Update 2

- [x] Renamed `Waiting for settlement` to `Pending solver submission`
- [x] Besides countdown stopping at 4min on `Pending solver submission`, now actually counting for 4min instead of 5min
- [x] Addressed Dima's comments

Notice the time difference between `Active in next batch` and `Pending solver submission` countdowns:
screenshot_2020-04-20_10-15-39

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -202,12 +202,20 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
     clear()
 
     const ms = getTimeRemainingInBatch(true)
-    console.log(`timeout useEffect activated, ${ms}ms left`)
 
     if (isActiveNextBatch) {
       refreshTimeout.current = setTimeout(() => forceUpdate({}), ms)
     } else if (isFirstActiveBatch) {
-      refreshTimeout.current = setTimeout(() => forceUpdate({}), Math.max(0, ms - 60 * 1000))
+      // we want to display this for 4min (1min less than batch duration)
+      const timeout = ms - 60 * 1000
+
+      if (timeout <= 0) {
+        // already less than 1min less left, update right now
+        forceUpdate({})
+      } else {
+        // more than 1min left, update when it's over
+        refreshTimeout.current = setTimeout(() => forceUpdate({}), timeout)
+      }
     }
 
     return clear
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -225,7 +225,7 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
         </>
       ) : isFirstActiveBatch ? (
         <>
-          {`Waiting for settlement: `} <StatusCountdown />
+          {`Pending solver submission: `} <StatusCountdown />
         </>
       ) : isScheduled ? (
         <>
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -234,6 +234,8 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
         pending
       ) : isExpiredOrder ? (
         'Expired'
+      ) : isFilled ? (
+        'Filled'
       ) : isActiveNextBatch ? (
         <>
           {`Active in next batch: `} <StatusCountdown />
@@ -248,8 +250,6 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
           <br />
           {formatDateFromBatchId(order.validFrom)}
         </>
-      ) : isFilled ? (
-        'Filled'
       ) : isActive ? (
         'Active'
       ) : (
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -141,11 +141,18 @@ const Expires: React.FC<Pick<Props, 'order' | 'pending' | 'isPendingOrder'>> = (
   return <td data-label="Expires">{isNeverExpires ? <span>Never</span> : <span>{expiresOn}</span>}</td>
 }
 
-const StatusCountdown: React.FC = () => {
+const StatusCountdown: React.FC<{ timeoutDelta?: number }> = ({ timeoutDelta }) => {
   // If it's rendered, it means it should display the countdown
   const timeRemainingInBatch = useTimeRemainingInBatch()
 
-  return <>{formatSeconds(timeRemainingInBatch)}</>
+  // `timeoutDelta` use case is for a countdown that's shorter than batch duration
+  // instead of counting all the way down to (currently 5min) batch time, we count instead to
+  // batchTime - timeoutDelta.
+  // When this countdown is over but there's still time left in the batch, return 0 for safety.
+  // Up to parent component to stop rendering at that time
+  const timeRemaining = timeoutDelta ? Math.max(0, timeRemainingInBatch - timeoutDelta) : timeRemainingInBatch
+
+  return <>{formatSeconds(timeRemaining)}</>
 }
 
 const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash' | 'isPendingOrder'>> = ({
@@ -233,7 +240,7 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
         </>
       ) : isFirstActiveBatch ? (
         <>
-          {`Pending solver submission: `} <StatusCountdown />
+          {`Pending solver submission: `} <StatusCountdown timeoutDelta={60} />
         </>
       ) : isScheduled ? (
         <>
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useEffect } from 'react'
+import React, { useMemo, useEffect, useRef } from 'react'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 
 import lowBalanceIcon from 'assets/img/lowBalance.svg'
@@ -155,6 +155,7 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
   const isScheduled = batchIdToDate(order.validFrom) > now
   const isActiveNextBatch = batchId === order.validFrom
   const isFirstActiveBatch = batchId === order.validFrom + 1 && secondsRemainingInBatch > 60 // up until minute 4
+
   const isUnlimited = useMemo(() => isOrderUnlimited(order.priceNumerator, order.priceDenominator), [
     order.priceDenominator,
     order.priceNumerator,
@@ -182,18 +183,18 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
   // Dima's trick to force component update
   const [, forceUpdate] = useSafeState({})
 
+  const countDownInterval = useRef<null | NodeJS.Timeout>(null)
+
   // Updates the component every second, if it has one of the states where it should happen
   useEffect(() => {
-    let timer: null | NodeJS.Timeout = null
-
     if (isActiveNextBatch || isFirstActiveBatch) {
-      timer = setInterval(() => forceUpdate({}), 1000)
-    } else if (!!timer) {
-      clearTimeout(timer)
+      countDownInterval.current = setInterval(() => forceUpdate({}), 1000)
+    } else if (!!countDownInterval.current) {
+      clearInterval(countDownInterval.current)
     }
 
     return (): void => {
-      if (timer) clearTimeout(timer)
+      if (countDownInterval.current) clearInterval(countDownInterval.current)
     }
   }, [forceUpdate, isActiveNextBatch, isFirstActiveBatch])
 
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -163,7 +163,7 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
 }) => {
   const now = new Date()
   const batchId = dateToBatchId(now)
-  const msRemainingInBatch = getTimeRemainingInBatch(true)
+  const msRemainingInBatch = getTimeRemainingInBatch({ inMilliseconds: true })
 
   const isExpiredOrder = batchIdToDate(order.validUntil) <= now
   const isScheduled = batchIdToDate(order.validFrom) > now
@@ -208,7 +208,7 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
 
     clear()
 
-    const ms = getTimeRemainingInBatch(true)
+    const ms = getTimeRemainingInBatch({ inMilliseconds: true })
 
     if (isActiveNextBatch) {
       refreshTimeout.current = setTimeout(() => forceUpdate({}), ms)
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr930--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +162,50 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
+    let timer: null | NodeJS.Timeout = null
+
+    if (isActiveNextBatch || isFirstActiveBatch) {
+      timer = setInterval(() => forceUpdate({}), 1000)
+    } else if (!!timer) {
+      clearTimeout(timer)
+    }
+
+    return (): void => {
+      if (timer) clearTimeout(timer)
```
Velenir, line 193: don't have to care about `if (timer)`, `clearTimeout(any nonsense)` is noop if not given a timer id
Don't even have to init `timer` as `null`

Velenir: Not to mention, should be `clearInterval`, :rofl:

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +162,50 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
+  // Dima's trick to force component update
+  const [, forceUpdate] = useSafeState({})
+
+  // Updates the component every second, if it has one of the states where it should happen
+  useEffect(() => {
+    let timer: null | NodeJS.Timeout = null
+
+    if (isActiveNextBatch || isFirstActiveBatch) {
+      timer = setInterval(() => forceUpdate({}), 1000)
+    } else if (!!timer) {
```
Velenir, line 188: This is always `false`, because this `timer` is always `L184: let timer`, never `L187: timer = ...` because of scope

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +162,50 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
 
   const pending = useMemo(() => isPendingOrder && <PendingLink transactionHash={transactionHash} />, [
     isPendingOrder,
     transactionHash,
   ])
 
+  // Dima's trick to force component update
+  const [, forceUpdate] = useSafeState({})
+
+  // Updates the component every second, if it has one of the states where it should happen
```
Velenir, line 182: Wouldn't you rather update after batch ended instead of rerendering every second
```
setTimeout(..., (BATCH_TIME - secondsRemaining + 1) * 1000) // to fall on after the batch ended 
//and then again for every batch end
setInterval(..., BATCH_TIME * 1000)
```

Take [useTimeRemainingInBatch](https://github.com/gnosis/dex-react/blob/ead3ff4e2e60d1d69bc2b2425ff116e34d68e4be/src/hooks/useTimeRemainingInBatch.ts#L57) for reference

And while you are at it can you change ~window.~ `setTimeout|setInterval` in there. I had problems with TS and id: NodeJS.Timer, but here I see TS didn't complain for you. Please :pleading_face:

src/components/OrdersWidget/OrderRow.tsx
Velenir: You are doing this to `clearTimeout(lnterval` :wink: ) after next active batch and first active batch has passed (both `isActiveNextBatch, isFirstActiveBatch` are false)? Then you need `timeout` in outer scope to keep it over rerenders, put it in `useRef`

src/components/OrdersWidget/OrderRow.tsx
alfetopito: Let me tell you a secret. I first used `setTimeout` hihi

src/components/OrdersWidget/OrderRow.tsx
alfetopito: `(both isActiveNextBatch, isFirstActiveBatch are false)? ` <- yes!

`useRef`, will do!

src/components/OrdersWidget/OrderRow.tsx
alfetopito: But but, to render the countdown, I'll need to refresh every second, no?

Velenir: Oh, shit, I didn't think about render
Yep, that's true
Hate that we have so much to display

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +162,50 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
+    } else if (!!timer) {
+      clearTimeout(timer)
+    }
+
+    return (): void => {
+      if (timer) clearTimeout(timer)
+    }
+  }, [forceUpdate, isActiveNextBatch, isFirstActiveBatch])
+
+  const secondsCountdown = formatSeconds(getSecondsRemainingInBatch())
```
anxolin, line 197: nit: This is not necesary unless you `pending && (isActiveNextBatch || isFirstActiveBatch)` but is executed for every row independently of those values

Would become simpler markup if you make a static function that returns the string of the status text?

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -141,6 +150,8 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
 
   const isExpiredOrder = batchIdToDate(order.validUntil) <= now
   const isScheduled = batchIdToDate(order.validFrom) > now
+  const isActiveNextBatch = dateToBatchId(now) === order.validFrom
```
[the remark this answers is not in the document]
alfetopito: I don't get it. Both are dynamic.
`now` is a const in the component set to `new Date()` and `order.validFrom` comes from the order.

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +162,50 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
+    !isPendingOrder &&
+    !isExpiredOrder
 
   const pending = useMemo(() => isPendingOrder && <PendingLink transactionHash={transactionHash} />, [
     isPendingOrder,
     transactionHash,
   ])
 
+  // Dima's trick to force component update
+  const [, forceUpdate] = useSafeState({})
```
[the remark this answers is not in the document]
alfetopito: :smiling_imp:

src/components/OrdersWidget/OrderRow.tsx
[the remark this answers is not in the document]
anxolin: I mean, you calculate batchId 2 times, right? 
`dateToBatchId(now)` in line 153, 154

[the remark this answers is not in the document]
alfetopito: ahhhhh right :+1:

src/components/OrdersWidget/OrderRow.tsx
alfetopito: > And while you are at it can you change window. setTimeout|setInterval in there. I had problems with TS and id: NodeJS.Timer, but here I see TS didn't complain for you. Please pleading_face

It didn't complain because of `if (timer) clearTimeout(timer)`

If I apply your suggestion from https://github.com/gnosis/dex-react/pull/930#discussion_r410036030, then the typing goes to shit.

I'd rather keep the `if (timer)` and not have TS annoyances

alfetopito: Check the latest verison

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,19 +173,60 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
+
+    clear()
+
+    const ms = getTimeRemainingInBatch(true)
+    console.log(`timeout useEffect activated, ${ms}ms left`)
+
+    if (isActiveNextBatch) {
+      refreshTimeout.current = setTimeout(() => forceUpdate({}), ms)
+    } else if (isFirstActiveBatch) {
+      refreshTimeout.current = setTimeout(() => forceUpdate({}), Math.max(0, ms - 60 * 1000))
```
Velenir, line 210: Presumably you don't need to schedule an update if `Math.max(0, ms - 60 * 1000) === 0`

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -1,13 +1,19 @@
...
 
   useEffect(() => {
-    const interval = setInterval(() => setTimeRemaining(getSecondsRemainingInBatch()), 1000)
+    // timeout to start the timer exactly at half second
+    let interval = setTimeout(() => {
+      // update once
+      setTimeRemaining(getTimeRemainingInBatch())
+      // update every second from now on
+      interval = setInterval(() => setTimeRemaining(getTimeRemainingInBatch()), 1000)
+    }, Date.now() % 500)
```
Velenir, line 16: if `Date.now() % 500 === 0` better not schedule an update but update right away. Otherwise there's a delay because of task queuing. 
Check with
```js
function ttime() {
    const now = Date.now()
    setTimeout(() => console.log('delay', Date.now() - now), 0)
}
```
but know that our delay will be longer because of all sync code on stack that needs to run first, then the microtask queue that needs to empty, then any older tasks from the tasks queue

src/components/OrdersWidget/OrderRow.tsx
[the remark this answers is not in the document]
Velenir: Maybe we need a `useForceUpdate` hook

src/components/OrdersWidget/OrderRow.tsx
alfetopito: So in that case I just execute it straight away? Or skip it entirely?

src/hooks/useTimeRemainingInBatch.ts
alfetopito: > if Date.now() % 500 === 0 better not schedule an update but update right away. 

Hmm yes makes sense.

> but know that our delay will be longer because of all sync code on stack that needs to run first

About this, any better way of syncing all of these countdowns?

src/components/OrdersWidget/OrderRow.tsx
Velenir: Straight away I think, hard to decide if that extra update will ultimately be useless, better to update

src/hooks/useTimeRemainingInBatch.ts
Velenir: requestAnimationFrame would be more precise, but also more annoying
setTimeout is good enough

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -1,15 +1,35 @@
...
+
+    // timeout to start the timer exactly at half second
+    const intervalStart = Date.now() % 500
+
+    if (intervalStart === 0) {
+      // to avoid possible scheduling delays, execute right now if exactly at 500ms mark
+      updateImmediatelyAndStartInterval()
+    } else {
+      // otherwise, schedule starting
+      interval = setTimeout(updateImmediatelyAndStartInterval, intervalStart)
```
anxolin, line 27: Can you clearInterval using a `timeoutId`??

I mean, if this component is unmounted before the timeout fires, you are calling clearInterval with a timeoutId

anxolin: edit: I learned something today, JS is very weird. Maybe add a comment in line 31 to say, that clears the interval that can contain the intervalId or the timeoutId

image

Velenir: So [I can do it](https://github.com/gnosis/dex-react/blob/d18b43ac14145c1a3ff00a5a1b25f9dfd552f81c/src/hooks/useTimeRemainingInBatch.ts#L90-L92) and none bats an eye, but when @alfetopito does you notice :rofl:

alfetopito: > So [I can do it](https://github.com/gnosis/dex-react/blob/d18b43ac14145c1a3ff00a5a1b25f9dfd552f81c/src/hooks/useTimeRemainingInBatch.ts#L90-L92) and none bats an eye, but when @alfetopito does you notice rofl

No one cares about you @Velenir 

OR

No one trusts my code

:thinking:

src/hooks/useTimeRemainingInBatch.ts
```diff
@@ -1,15 +1,35 @@
...
-    return (): void => clearInterval(interval)
+    function updateImmediatelyAndStartInterval(): void {
+      // update once
+      setTimeRemaining(getTimeRemainingInBatch())
+      // update every second from now on
+      interval = setInterval(() => setTimeRemaining(getTimeRemainingInBatch()), 1000)
+    }
+
+    // timeout to start the timer exactly at half second
+    const intervalStart = Date.now() % 500
```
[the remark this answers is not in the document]
alfetopito: > Why half a second?

More likely that all timers will be in the same second when the update happens, reducing the chance of having mismatches due to how JS schedules the timers.

> Also, if we are in the millisecond 400 , then Date.now() % 500 = 400
If I wait 400ms then I'll execute updateImmediatelyAndStartInterval at the ms 800 right? 

Ah shit, yes. `500` should be the delta, not the modulus.

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -151,27 +180,76 @@ const Status: React.FC<Pick<Props, 'order' | 'isOverBalance' | 'transactionHash'
...
     <td className="status">
       {pending ? (
         pending
       ) : isExpiredOrder ? (
         'Expired'
+      ) : isFilled ? (
+        'Filled'
+      ) : isActiveNextBatch ? (
+        <>
+          {`Active in next batch: `} <StatusCountdown />
```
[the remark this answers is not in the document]
alfetopito: Only when the timer is active. Otherwise not rendered.

src/hooks/useTimeRemainingInBatch.ts
anxolin: It just depend on the number of cups of coffee I had ☕️☕️☕️☕️☕️☕️☕️☕️

PR #933, opened 2020-04-17
switch to literal regexp

Velenir:
A pet peeve of mine.
Let's use literal notation for regexp when possible

- `new RegExp(string)` is useful when string is dynamically constructed
- `new RegExp(/reg_exp/)` is copy-constructed
- `/pattern/` is genrally prefered for static patterns

[Reference 1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Description), [2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Creating_a_regular_expression)

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -58,9 +58,9 @@ export function sanitizeNegativeAndMakeMultipleOf(value?: string | null, default
 export function validatePositiveConstructor(message: string) {
   return (value: string, constraint = 0): true | string => Number(value) > constraint || message
 }
-export const validInputPattern = new RegExp(/^\d+\.?\d*$/) // allows leading and trailing zeros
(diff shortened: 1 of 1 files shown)


W3stside:
is there any difference beyond the dynamic thing between the two? is it a performance thing?

gnosis-info:
Travis automatic deployment:
 https://pr933--dexreact.review.gnosisdev.com

Velenir:
> is there any difference beyond the dynamic thing between the two? is it a performance thing?

Supposedly so
> The literal notation provides a compilation of the regular expression when the expression is evaluated. Use literal notation when the regular expression will remain constant. For example, if you use literal notation to construct a regular expression used in a loop, the regular expression won't be recompiled on each iteration.

PR #936, opened 2020-04-17
Different separator

Velenir:
Trying a different separator
`:sell\+:buy` , + must be escaped because it has special meaning in [path-to-regex](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0#one-or-more)

Safe separator can be any of `$&:<>[]{}"+%@/;=^|',` symbols

http://localhost:8080/trade/FTK%20%2F21%2F10-DAI%20%2499%20%26%20%2B%7C-+DAI?sell=0&price=0&from=&expires=2880
http://localhost:8080/trade/DAI+FTK%20%2F21%2F10-DAI%20%2499%20%26%20%2B%7C-?sell=0&price=0&from=&expires=2880
with `yarn mock`

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.12.0",
     "@fortawesome/react-fontawesome": "^0.1.8",
     "@gnosis.pm/dapp-ui": "^0.5.3",
-    "@gnosis.pm/dex-js": "0.2.2",
+    "@gnosis.pm/dex-js": "0.2.3-RC.0",
     "@hapi/joi": "^17.1.1",
     "@hot-loader/react-dom": "^16.11.0",
     "@popperjs/core": "^2.0.6",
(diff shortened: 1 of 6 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr936--dexreact.review.gnosisdev.com

anxolin:
buuuuuuuut, what if now you get a token like `Grid+`

I wouldn't change the URL for an edge case, right? Can we just escape the dash or dash and slash if it's they are the only problematic encoding params?

Velenir:
> buuuuuuuut, what if now you get a token like `Grid+`

```js
encodeURIComponent('Grid+') === "Grid%2B"
// it's Grid<minus> that is a problem
encodeURIComponent('Grid-')==="Grid-"

alfetopito:
I prefer a `|`. More neutral, doesn't need escaping.

2 more things:

1. Better have a constant for that
2. We should keep a redirect for the `-` for at least 1 version, until we deploy a change to `dex-telegram` that uses the new separator.

Velenir:
Closing because of a different approach in #924

PR #937, opened 2020-04-17
Add hops to order book

anxolin:
Just adds a input field for allowing to change the hops in the order book

image

This order book is not exposed in the app directly, but it's accessible in the hidden page `/book`

TEST:
https://pr937--dexreact.review.gnosisdev.com/book


gnosis-info:
Travis automatic deployment:
 https://pr937--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -67,9 +69,9 @@ interface ProcessedItem {
   price: number
 }
 
-const orderbookUrl = (baseToken: TokenDetails, quoteToken: TokenDetails, networkId?: number): string => {
+const orderbookUrl = (baseToken: TokenDetails, quoteToken: TokenDetails, hops: number, networkId?: number): string => {
   const network = getNetworkFromId(networkId || 1)
-  return `https://price-estimate-${network}.dev.gnosisdev.com/api/v1/markets/${baseToken.id}-${quoteToken.id}?atoms=true`
+  return `https://price-estimate-${network}.dev.gnosisdev.com/api/v1/markets/${baseToken.id}-${quoteToken.id}?atoms=true&hops=${hops}`
```
alfetopito, line 74: Should we be using the prod endpoints here?

src/const.ts
```diff
@@ -108,3 +108,5 @@ export const INPUT_PRECISION_SIZE = 6
 export const VALID_UNTIL_DEFAULT = '2880'
 export const VALID_FROM_DEFAULT = '30'
 export const WETH_ADDRESS_MAINNET = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
+export const ORDER_BOOK_HOPS_DEFAULT = 2
+export const ORDER_BOOK_HOPS_MAX = 2
```
alfetopito, line 112: oh, so only 1 or 2? what a let down...

anxolin: Ahaha, I double checked with @fleupold , if this changes he will notify us, but for now 2 hops tops. I'm not sure if you would get a great difference with more hops in the book, but sure there would be some CPU bill to pay by devops.

I think it's interesting to set it to 0, and see how it changes already with 1 hop.

This is a hidden functionality for debugging, and impress people in the bar 🍻

src/components/OrderBookWidget.tsx
anxolin: I'll do it in a different PR then

PR #949, opened 2020-04-21
948/update orders instead of replacing

### LINKED ISSUE #948: Merge/update new orders instead of reseting
Another improvement I thought about (considering my horrible first fix attempt) that might be useful to avoid flickering when we fetch orders from start is to have the option to merge the current list of orders.

I'm thinking about moving the filtering to the global state, and there (when merging) update the existing orders, removing ones that have potentially been cancelled.

This way, the list is always constant, while only modified orders are updated/removed.

What do ya think?

_Originally posted by @alfetopito in https://github.com/gnosis/dex-react/issues/909#issuecomment-614748126_

alfetopito:
Closes #948 

UX improvement. Was annoyed seeing my orders flash on every 4min mark of the batch.

This change makes so we only reset orders when changing network/address.
All other updates will do it "in place", sort of.

Moved deleted orders filtering logic from `useOrders` hook to orders reducer.

### DIFF
--- a/src/hooks/useOrders.ts
+++ b/src/hooks/useOrders.ts
@@ -3,41 +3,18 @@ import { useEffect, useCallback, useRef } from 'react'
 import { unstable_batchedUpdates } from 'react-dom'
 
 import useGlobalState from './useGlobalState'
-import { overwriteOrders, appendOrders, updateOffset } from 'reducers-actions/orders'
+import { overwriteOrders, updateOffset, updateOrders } from 'reducers-actions/orders'
 
 import { useWalletConnection } from './useWalletConnection'
 import useSafeState from './useSafeState'
 import { exchangeApi } from 'api'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
-import { ZERO } from 'const'
 import { useCheckWhenTimeRemainingInBatch } from './useTimeRemainingInBatch'
 
 interface Result {
   orders: AuctionElement[]
   forceOrdersRefresh: () => void
   isLoading: boolean
-}
-
-/**
- * Filter out deleted orders.
- *
- * When orders are `deleted` from the contract, they are still returned, but with all fields set to zero.
- * We will not display such orders.
- *
- * @param orders all orders returned by the contract
- */
-function filterDeletedOrders(orders: AuctionElement[]): AuctionElement[] {
-  return orders.filter(
-    order =>
-      !(
-        order.buyTokenId === 0 &&
-        order.sellTokenId === 0 &&
-        order.priceDenominator.eq(ZERO) &&
-        order.priceNumerator.eq(ZERO) &&
-        order.validFrom === 0 &&
-        order.validUntil === 0
-      ),
-  )
 }
 
 const REFRESH_WHEN_SECONDS_LEFT = 60 // 1min before batch done
@@ -81,19 +58,13 @@ export function useOrders(): Result {
         // check cancelled bool from parent scope
         if (cancelled) return
 
-        // Apply filters (remove deleted orders)
-        const filteredOrders = filterDeletedOrders(orders)
-
         // ensures we don't have multiple reruns for each update
         // i.e. offset change -> render
         //      isLoading change -> another render
         unstable_batchedUpdates(() => {
-          if (offset === 0) {
-            // fresh start/refresh: replace whatever is stored
-            dispatch(overwriteOrders(filteredOrders))
-          } else if (filteredOrders.length > 0) {
-            // incremental update: append
-            dispatch(appendOrders(filteredOrders))
+          if (orders.length > 0) {
+            // update
+            dispatch(updateOrders(orders))
           }
 
           if (!nextIndex) {
--- a/src/reducers-actions/orders.ts
+++ b/src/reducers-actions/orders.ts
@@ -1,7 +1,8 @@
 import { Actions } from 'reducers-actions'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
+import { ZERO } from '@gnosis.pm/dex-js'
 
-export type ActionTypes = 'OVERWRITE_ORDERS' | 'APPEND_ORDERS' | 'UPDATE_OFFSET'
+export type ActionTypes = 'OVERWRITE_ORDERS' | 'APPEND_ORDERS' | 'UPDATE_ORDERS' | 'UPDATE_OFFSET'
 
 export interface OrdersState {
   orders: AuctionElement[]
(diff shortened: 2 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr949--dexreact.review.gnosisdev.com

src/reducers-actions/orders.ts
```diff
@@ -22,13 +23,53 @@ export const appendOrders = (orders: AuctionElement[]): UpdateOrdersActionType =
...
+  allOrders: AuctionElement[],
+): { activeOrders: { [id: string]: number }; deletedOrdersIds: Set<string> } {
+  const activeOrders: { [id: string]: number } = {}
+  const deletedOrdersIds = new Set<string>()
+
+  allOrders.forEach((order, index) => {
+    if (isOrderDeleted(order)) {
+      deletedOrdersIds.add(order.id)
+    } else {
+      activeOrders[order.id] = index
```
W3stside, line 66: are indices in `Set` guaranteed to be in order?

src/reducers-actions/orders.ts
```diff
@@ -38,21 +79,67 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
...
-      // make a copy first (slice(0)) to avoid mutating current state
-      const reversedOrders = orders.slice(0).reverse()
+      const reversedOrders = orders
+        // make a copy first (slice(0)) to avoid mutating current state
+        .slice(0)
+        // default sorting order is ascending by order id
+        // since we are starting from 0, reverse to have the latest on top
+        .reverse()
+        // remove deleted orders
+        .filter(order => !isOrderDeleted(order))
```
W3stside, line 89: just filter first, that returns a new array, then reverse

src/hooks/useOrders.ts
```diff
@@ -81,19 +58,13 @@ export function useOrders(): Result {
...
-        // Apply filters (remove deleted orders)
-        const filteredOrders = filterDeletedOrders(orders)
-
         // ensures we don't have multiple reruns for each update
         // i.e. offset change -> render
         //      isLoading change -> another render
         unstable_batchedUpdates(() => {
-          if (offset === 0) {
-            // fresh start/refresh: replace whatever is stored
-            dispatch(overwriteOrders(filteredOrders))
```
Velenir, line 93: Shouldn't we still overwrite orders when `offset == 0`. I don't think `updateOrders` quite replaces that

src/reducers-actions/orders.ts
```diff
@@ -22,13 +23,53 @@ export const appendOrders = (orders: AuctionElement[]): UpdateOrdersActionType =
...
+    order.priceNumerator.eq(ZERO) &&
+    order.validFrom === 0 &&
+    order.validUntil === 0
+  )
+}
+
+function classifyOrders(
+  allOrders: AuctionElement[],
+): { activeOrders: { [id: string]: number }; deletedOrdersIds: Set<string> } {
+  const activeOrders: { [id: string]: number } = {}
```
W3stside, line 59: why not also make this is a `Set`?

src/reducers-actions/orders.ts
```diff
@@ -38,21 +79,67 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
...
+      const { activeOrders, deletedOrdersIds } = classifyOrders(newOrders)
+
+      // Filter deleted orders and update existing orders
+      const filteredAndUpdatedOrders = currentOrders.reduce((acc: AuctionElement[], order) => {
+        if (activeOrders[order.id] !== undefined) {
+          // It's on activeOrders map? it has been updated.
+          // Get the updated from newOrders list by the index in the activeOrders map
+          acc.push(newOrders[activeOrders[order.id]])
+          // Remove from activeOrders map so we know it's not a new order
+          delete activeOrders[order.id]
```
W3stside, line 110: would be preferable as a `Set` with `set.delete(order.id)`

src/reducers-actions/orders.ts
```diff
@@ -38,21 +79,67 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
...
         payload: { orders: newOrders },
       } = action
       const { orders: currentOrders } = state
 
       // reverse new orders
-      const reversedNewOrders = newOrders.slice(0).reverse()
+      const reversedNewOrders = newOrders
+        .slice(0)
+        .reverse()
+        .filter(order => !isOrderDeleted(order))
```
W3stside, line 142: same thing here, can filter then reverse

src/reducers-actions/orders.ts
```diff
@@ -38,21 +79,67 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
...
+      const reversedNewOrders = Object.values(activeOrders)
+        // Use their indexes to fetch from newOrders list
+        .map(index => newOrders[index])
+        // Reverse the result to have newest on top
+        .reverse()
+
+      // Add new orders first
+      const orders = reversedNewOrders.concat(filteredAndUpdatedOrders)
+
+      return { ...state, orders }
```
Velenir, line 130: This is kinda complicated. Wouldn't the following produce the same result?
```ts
   // put newest (by id) on top
  const reversedNewOrders = newOrders.slice(0).reverse()
  // keep track so that we don't duplicate orders
  const processedOrderIds = new Set<string>()

  // first we process reversedNewOrders
  // then currentOrders
  // but thanks to processedOrderIds newOrders override currentOrders with same id
  const updatedOrders = reversedNewOrders.concat(currentOrders).reduce<AuctionElement[]>((accum, order) => {
    // already included a potentially updated order
    // or the order was deleted
    if (processedOrderIds.has(order.id)) return accum

    // don't include deleted orders
    if (isOrderDeleted(order)) {
      // but consider them processed
      processedOrderIds.add(order.id)
      return accum
    }

    // this order is either one of the newOrders
    // or one of the currentOrders that wasn't overriden by newOrders
    // and it's not deleted
    accum.push(order)
    processedOrderIds.add(order.id)
    return accum
  }, [])

  return { ...state, orders: updatedOrders }
```

src/reducers-actions/orders.ts
```diff
@@ -1,7 +1,8 @@
 import { Actions } from 'reducers-actions'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
+import { ZERO } from '@gnosis.pm/dex-js'
 
-export type ActionTypes = 'OVERWRITE_ORDERS' | 'APPEND_ORDERS' | 'UPDATE_OFFSET'
+export type ActionTypes = 'OVERWRITE_ORDERS' | 'APPEND_ORDERS' | 'UPDATE_ORDERS' | 'UPDATE_OFFSET'
```
Velenir, line 5: Are we even using `'APPEND_ORDERS'` at this point?

src/reducers-actions/orders.ts
anxolin: I guess, you mean a Map, right?

Or ou mean that he wouldn't be storing the index. I think he did that to be more efficient updating. I also have some doubts if the optimization is worth it for the number of orders we expect. Arguably, you could search in a collection every time you want to update an order, and decide if it's addition/update. Anyways, I'm ok with the optimization too, and for sure is more efficient :)

src/reducers-actions/orders.ts
anxolin: Good point David, although I think it does. 
And that's another weird thing about JS, sets are unordered in any other language, list have order 🤯

image

alfetopito: Does not matter in this case. I just want a quick access to identify the ones that should be removed. I'm relying on the existing orders list instead.

src/reducers-actions/orders.ts
alfetopito: Ohhhhh thanks for the tip

src/reducers-actions/orders.ts
alfetopito: > why not also make this is a Set?

As Anxo noted, yeah, this is sort of a Map, but I'm using an Object instead. Guess I should go for a map?

Regardless, not a set because I need the index in the original array, and the order id is the key for quicker access.

src/reducers-actions/orders.ts
alfetopito: But in that case, I wouldn't have all the info I need. Having only the order.id in the set means I'd have to search `newOrders` array for every order found.

I can give it a try using a Map for a better interface.

src/hooks/useOrders.ts
alfetopito: Not really. The point of this change is exactly that. To update the list of orders 'in place' without flashing the UI.

`updateOrders` updates existing, remove deleted and appends new orders. That should cover all cases during an update, no?

src/reducers-actions/orders.ts
alfetopito: Nope

src/reducers-actions/orders.ts
alfetopito: Yes, I like your suggestion.
Much simpler, thanks!

src/reducers-actions/orders.ts
alfetopito: No longer in use with Dima's suggestion

src/hooks/useOrders.ts
Velenir: You are right. On second thought, shouldn't updateOrders be called unconditionally, without if orders.length
Because orders can be deleted all at once from the contact, if they were all previously cancelled

alfetopito: > Because orders can be deleted all at once from the contact, if they were all previously cancelled

If orders are deleted from the contract, even if in a batch, they will still be returned but with all fields zeroed out.

That means, if orders.length === 0, there is really nothing to update.

Velenir: Cool. I need to reread the contract

PR #951, opened 2020-04-22
More robust token id fetching

alfetopito:
Clearing this todo entry:

TODO: try again only for the ones that failed

### DIFF
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -18,7 +18,7 @@ export function getTokensFactory(factoryParams: {
     failed?: true
   }
 
-  async function updateTokenId(networkId: number, token: TokenDetails): Promise<UpdateTokenIdResult> {
+  async function updateTokenIdOrIndicateFailure(networkId: number, token: TokenDetails): Promise<UpdateTokenIdResult> {
     try {
       const hasToken = await exchangeApi.hasToken({ networkId, tokenAddress: token.address })
       if (!hasToken) {
@@ -47,7 +47,7 @@ export function getTokensFactory(factoryParams: {
       )
     }
 
-    const promises = tokens.map(token => updateTokenId(networkId, token))
+    const promises = tokens.map(token => updateTokenIdOrIndicateFailure(networkId, token))
 
     const results = await Promise.all(promises)
 
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -12,21 +12,91 @@ export function getTokensFactory(factoryParams: {
 
   const updatedIdsForNetwork = new Set<number>()
 
-  interface GetTokenIdResult {
-    value: number | null
+  interface UpdateTokenIdResult {
+    token: TokenDetails
     remove?: true
+    failed?: true
   }
 
-  async function getTokenId(networkId: number, tokenAddress: string): Promise<GetTokenIdResult> {
+  async function updateTokenId(networkId: number, token: TokenDetails): Promise<UpdateTokenIdResult> {
     try {
-      const hasToken = await exchangeApi.hasToken({ networkId, tokenAddress })
+      const hasToken = await exchangeApi.hasToken({ networkId, tokenAddress: token.address })
       if (!hasToken) {
-        return { value: null, remove: true }
+        return { token, remove: true }
       }
-      return { value: await exchangeApi.getTokenIdByAddress({ networkId, tokenAddress }) }
+      token.id = await exchangeApi.getTokenIdByAddress({ networkId, tokenAddress: token.address })
+      return { token }
     } catch (e) {
-      logDebug(`[network:${networkId}][address:${tokenAddress}] Failed to fetch id from contract`, e.message)
-      return { value: null }
+      logDebug(`[network:${networkId}][address:${token.address}] Failed to fetch id from contract`, e.message)
+      return { token, failed: true }
+    }
+  }
+
+  async function updateWithRetries(
+    networkId: number,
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr951--dexreact.review.gnosisdev.com

src/services/factories/tokenList.ts
```diff
@@ -12,58 +12,106 @@ export function getTokensFactory(factoryParams: {
...
+      const tokenList = tokenListApi.getTokens(networkId).reduce<TokenDetails[]>((acc, token) => {
+        if (toRemove.has(token.address)) {
+          // removing tokens not registered in the exchange for this network
+          return acc
+        } else if (updated.has(token.address)) {
+          // updating
+          acc.push(updated.get(token.address) as TokenDetails)
+        } else {
+          // not changed OR failed to fetch
+          acc.push(token)
```
Velenir, line 82:
```suggestion
        } else {
          // changed or old token
          acc.push(updated.get(token.address) || token)
```
not that it matters much

src/services/factories/tokenList.ts
```diff
@@ -12,58 +12,106 @@ export function getTokensFactory(factoryParams: {
...
+
+    // if something failed...
+    if (failedToUpdate.length > 0) {
+      logDebug(
+        `[network:${networkId}] Failed to fetch ids for ${failedToUpdate.length} tokens. Trying again in ${retryDelay}ms`,
+      )
+      // calculate how long to wait next time
+      const nextDelay = retryDelay * (exponentialBackOff ? 2 : 1)
+      // try again failed tokens after retryDelay
+      setTimeout(() => updateWithRetries(networkId, failedToUpdate, maxRetries - 1, nextDelay), retryDelay)
```
Velenir, line 99: really like exponential backoff :+1:

src/services/factories/tokenList.ts
```diff
@@ -12,58 +12,106 @@ export function getTokensFactory(factoryParams: {
...
+    tokens: TokenDetails[],
+    maxRetries = 3,
+    retryDelay = 1000,
+    exponentialBackOff = true,
+  ): Promise<void> {
+    if (maxRetries <= 0) {
+      throw new Error(
+        `[network:${networkId}] Max retries exceeded trying to fetch token ids for tokens ${tokens.map(
+          t => t.address,
+        )}`,
```
Velenir, line 46: do we really want to hard-fail here, as opposed to a warning?
not that uncaught promise rejection is a very hard-fail, but still
we've already removed the always-failing token at this point anyway

alfetopito: Not really hard failing. This exception is caught on https://github.com/gnosis/dex-react/pull/951/files#diff-e5e8192286e6ad1d544324c3dce8ed97R109

When it does fail, I clear the flag so next time the service is called we try to fetch again.

Velenir: Cool, good idea :+1:

src/services/factories/tokenList.ts
```diff
@@ -12,58 +12,103 @@ export function getTokensFactory(factoryParams: {
...
-        tokens.push(tokenListApi.getTokens(networkId)[index])
-        // Try again next time
-        // TODO: try again only for the ones that failed
-        failedToUpdate = true
+    const updated = new Map<string, TokenDetails>()
+    const failedToUpdate: TokenDetails[] = []
+    const toRemove = new Set<string>()
+
+    // classify results into map/list/set for easier access
+    results.forEach(({ token, remove, failed }) => {
```
anxolin, line 59: Maybe I'm mistaken, but if `result` is the result of Promise.all, then in case of an error it will throw already in line `52` and you are not catching that error. The retry logic depends on `failedToUpdate` that is initialized after

anxolin: edit: All right, I see, `updateTokenId` in his implementation make sure we never throw. so I guess is just a nitpick to make it more clear, so you don't need to check the implementation.
Either:
* making it more evident in the name `updateTokenIdOrXXXXXX` (being XXXXXX whatever you think clarifies here, too mentally tired now)
* ...or, make the catch inline `tokens.map(token => updateTokenId(networkId, token).catch(_handleFailedToken))`

src/services/factories/tokenList.ts
```diff
@@ -12,58 +12,103 @@ export function getTokensFactory(factoryParams: {
...
+        `[network:${networkId}] Max retries exceeded trying to fetch token ids for tokens ${tokens.map(
+          t => t.address,
+        )}`,
+      )
+    }
 
-    const promises = tokenListApi.getTokens(networkId).map(token => getTokenId(networkId, token.address))
+    const promises = tokens.map(token => updateTokenId(networkId, token))
 
     const results = await Promise.all(promises)
```
W3stside, line 52: just wondering - what's the point of doing this if you're resolving right after? why not just
`const results = await Promise.all(tokens.map(token => updateTokenId(networkId, token)))`?

i could see it making sense if u were resolving much lower

src/services/factories/tokenList.ts
W3stside: TIL you can multi-line comment

src/services/factories/tokenList.ts
alfetopito: In a previous PR, @Velenir suggested to do it using `Promise.all` instead of what you are suggesting.
I honestly can't say what's better.

src/services/factories/tokenList.ts
alfetopito: Updated the function name.

PR #966, opened 2020-04-24
759/save pending orders fix

### LINKED ISSUE #759: Sometimes there's a big delay showing the order in pending
Observed behaviour:

1. A user submitted an order
2. The order didn't show up as pending right away, not even after some seconds
3. User was confused and refreshed several times (until it was mined) and showed up in the orders list

W3stside:
Closes #759 

Here is the problem and why this PR was created. Imagine:
1. User heads to `Liquidity` and creates a large order
2. Signs TX and orders are, within the `onSubmit` handler, sent to global state as `pendingOrders` to be carried over and shown inside the `OrdersWidget` inside `TradeWidget`, for example.
3. After TX signature, still on LiquidityWidget, user immediately accidentally refreshes page
4. Now all those pending transactions are lost from global state
5. So, we save to `localStorage` via `onSubmit`, in this case we do so inside the `reducer` (arguably a bad choice right @Velenir?)
6. Now, if step 3 happens again, we simply pull from `localStorage` inside `reducers/pendingOrders.ts` on app load to maintain old orders
7. But how do we check for pending orders being cleared/mined? We can't rely on the closure callback from the `onSubmit` handler as was done before since that closure is lost on page refresh.
8. So we use `(await web3.eth.getBlock(blockNumber || 'latest')).transactions` to get get an array of `txHash`es that have been mined and filter that from our global state `txHash` list of pending orders which we insert in a hook that listens for changes in `blockNumber`, `networkId`, and `userAddress` to remove
9. For concise reasons, we add it to `useOrders` to keep order related logic inside one hook and not require the need for 2 hooks in places we need only 1


gnosis-info:
Travis automatic deployment:
 https://pr966--dexreact.review.gnosisdev.com

src/hooks/useOrders.ts
```diff
@@ -118,5 +146,12 @@ export function useOrders(): Result {
     dispatch(overwriteOrders([]))
   }, [userAddress, networkId, forceOrdersRefresh, dispatch])
 
-  return { orders, isLoading, forceOrdersRefresh }
+  return {
+    orders,
+    isLoading,
+    forceOrdersRefresh,
+    get pendingOrders(): AuctionElement[] {
+      return networkId && userAddress ? pendingOrders[networkId][userAddress] : emptyArray
```
Velenir, line 154: This was the problem with rerender?
Stable reference will fix this :+1:

W3stside: yeah all this has changed locally, ill update now

Velenir:
can wrap side effects over slice reducers, would be even easier
```js
pendingOrders: withSideEffect(PendingOrderReducer, saveToStorage)
```

src/hooks/useOrders.ts
```diff
@@ -1,34 +1,91 @@
...
+  // pending orders are saved in global app state as well as local storage
+  // users can refresh page after creating orders and see them after reload thanks for local storage
+  // uses this effect hook to:
+  //   1. listen to block updates and filter recently mined transactions based on their hash
+  //   2. remove any orders that may have been re-sent at a higher gas price
+  const currentPendingOrders: PendingTxObj[] =
+    (networkId && userAddress && pendingOrders[networkId] && pendingOrders[networkId][userAddress]) || emptyArray
+  useEffect(() => {
+    // Don't fire if there are no pending orders...
+    if (networkId && userAddress && currentPendingOrders.length > 0) {
```
anxolin, line 46: I'm wondering. The logic of updating the orders is in `useOrders`, so if anyone uses the global state instead of `useOrders` he won't get the updates, right?

src/hooks/useOrders.ts
```diff
@@ -1,34 +1,91 @@
...
+      const managePendingOrders = async (): Promise<void> => {
```
anxolin, line 47: Concern 2: If we implement it like this, what happens if 10 parts of the UI uses `useOrders`, does it execute this logic 10 times?
@alfetopito @W3stside @Velenir

Velenir: It does
Where would you rather put it?
Or maybe use cache?

anxolin: Cache would work, although also we have the "Concern 1" above, I wonder if this is more central piece

src/hooks/useOrders.ts
```diff
@@ -1,34 +1,91 @@
...
+        const [latestBlock, pendingOrderStatuses] = await Promise.all([
+          web3.eth.getBlock(blockNumber || 'latest'),
+          Promise.all(currentPendingOrders.map(({ txHash }) => web3.eth.getTransaction(txHash))),
+        ])
```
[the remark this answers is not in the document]
W3stside: good idea! :+1:

W3stside:
Should close in favor for #999

W3stside:
Closing in favor for #999

PR #967, opened 2020-04-24
960/dynamic token fetching

### LINKED ISSUE #960: Dynamically load tokens from contract
This is the initial idea for dynamically loading tokens from https://github.com/gnosis/dex-react/issues/467:

---

1. Ignore completely list of tokens from `dex-js`
2. Whenever not connected, default to `mainnet` list of tokens
3. Store list of tokens for each network on local storage
4. ~On `TokenList` constructor~
    1. Load list of tokens from local storage, if any
    2. Async; Fetch [number of tokens registered in each network](https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L54)
    3. If number of tokens == local list length, stop (per network)
    4. Else, fetch token details from token id = localList.length -1 until numTokensInExchange
        1. Fetch address from exchange contract
        1. Fetch symbol, name, decimals from ERC20 contract
    5. Update local storage

- Token list can start with 0 tokens and be updated async.
- ~Might need to update a bunch of stuff that has sync assumptions about the token list.~ done
- ~Once loaded, list of tokens can be updated if adding new tokens in the UI.~ done

_Originally posted by @alfetopito in https://github.com/gnosis/dex-react/issues/467#issuecomment-604591406_

On step `4`, will use the approach done on #467 and instead trigger the fetch when `getTokens` is first called.

---

Steps from this comment https://github.com/gnosis/dex-react/issues/467#issuecomment-605166348
- [x] Step 1 completed (#467)
- [ ] This is the step 2 
- [ ] Step 3 is to intersect loaded list with TCR/Kleros/whatever (https://github.com/gnosis/dex-react/issues/181 https://github.com/gnosis/dex-react/issues/180)

alfetopito:
Closes #960 

Dynamically loading all tokens from the contract.

When local copy list of tokens length match number of registered token on the contract, update only the token ids.

### Note
1. ~Some tokens do not have symbol/name due to how they are encoded in the contract. For example: https://etherscan.io/address/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359#readContract (check symbol and name)~
2. Some tokens fail to get the balance https://rinkeby.etherscan.io/address/0x2A642a7D6Fb2be3F2ADbcb3457F7953483c28213#code (don't know yet why)
3. Rinkeby contract has 3 addresses registered which are just plain addresses, not contracts:
```
[4] Address 0x4A642a7d6fb2be3f2adbcb3457F7953483c28223 is not a valid ERC20 token 
[4] Address 0xD0DC005d31C2979CC0d38718e23c82D1A50004C0 is not a valid ERC20 token 
[4] Address 0xAe38b81459d74A8C16eAa968c792207603D84480 is not a valid ERC20 token
```

### Next step: 
Cross reference tokens listed on the contract with whitelisted list of tokens.

### Edit 1

Addressing `1`. Hard coded a local list of token details to overcame issue with tokens not having proper symbol/name set.

So many tokens now:

screenshot_2020-04-27_13-45-03
screenshot_2020-04-27_13-45-16

### DIFF
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -268,19 +268,6 @@ export class Erc20ApiImpl implements Erc20Api {
 
     return erc20Details ? erc20Details[prop] : undefined
   }
-
-  private async withLocalErc20Details<P extends keyof Erc20Props>(
-    params: BaseParams,
-    fn: (params: BaseParams) => Promise<Erc20Props[P]>,
-    prop: P,
-  ): Promise<Erc20Props[P]> {
-    const erc20Prop = this._getLocalErc20Property(params.tokenAddress, prop)
-
-    if (erc20Prop) {
-      return erc20Prop
-    }
-    return fn(params)
-  }
 }
 
 export default Erc20ApiImpl
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -139,7 +139,7 @@ export class Erc20ApiImpl implements Erc20Api {
 
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
-    return await erc20.methods.symbol().call()
+    return erc20.methods.symbol().call()
   }
 
   public async decimals({ tokenAddress, networkId }: DecimalsParams): Promise<number> {
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -7,7 +7,19 @@ import { TxOptionalParams, Receipt } from 'types'
 import { ZERO } from 'const'
 import { toBN } from 'utils'
 
+import ERC20_DETAILS from './erc20Details.json'
+
 import Web3 from 'web3'
+
+export interface Erc20Props {
+  name: string
+  symbol: string
+  decimals: number
+}
+
+export interface Erc20Details {
+  [address: string]: Erc20Props
+}
 
 interface BaseParams {
   tokenAddress: string
@@ -77,6 +89,7 @@ export interface Params {
 export class Erc20ApiImpl implements Erc20Api {
   private _contractPrototype: Erc20Contract
   private web3: Web3
+  private readonly localErc20Details: Erc20Details
 
   private static _contractsCache: { [network: number]: { [address: string]: Erc20Contract } } = {}
 
@@ -84,6 +97,11 @@ export class Erc20ApiImpl implements Erc20Api {
 
   public constructor(injectedDependencies: Params) {
     Object.assign(this, injectedDependencies)
+
+    // Local overwrites for token details
+    // Usually that shouldn't be needed but some tokens do not abide by the standard
+    // and return symbol/name as bytes32 as opposed to string
+    this.localErc20Details = ERC20_DETAILS
 
     this._contractPrototype = new this.web3.eth.Contract(erc20Abi as AbiItem[]) as Erc20Contract
 
@@ -103,23 +121,38 @@ export class Erc20ApiImpl implements Erc20Api {
   }
 
   public async name({ tokenAddress, networkId }: NameParams): Promise<string> {
+    const name = this._getLocalErc20Property(tokenAddress, 'name')
+    if (name) {
+      return name
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
     return await erc20.methods.name().call()
   }
 
   public async symbol({ tokenAddress, networkId }: SymbolParams): Promise<string> {
+    const symbol = this._getLocalErc20Property(tokenAddress, 'symbol')
+    if (symbol) {
+      return symbol
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
     return await erc20.methods.symbol().call()
   }
 
   public async decimals({ tokenAddress, networkId }: DecimalsParams): Promise<number> {
+    const decimals = this._getLocalErc20Property(tokenAddress, 'decimals')
+    if (decimals) {
+      return decimals
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
-    const decimals = await erc20.methods.decimals().call()
+    const decimalsString = await erc20.methods.decimals().call()
 
-    return Number(decimals)
+    return Number(decimalsString)
   }
 
   public async totalSupply({ tokenAddress, networkId }: TotalSupplyParams): Promise<BN> {
@@ -228,6 +261,25 @@ export class Erc20ApiImpl implements Erc20Api {
     newERC20.options.address = address
 
     return (Erc20ApiImpl._contractsCache[networkId][address] = newERC20)
+  }
+
+  private _getLocalErc20Property<P extends keyof Erc20Props>(address: string, prop: P): Erc20Props[P] | undefined {
+    const erc20Details = this.localErc20Details[address]
+
+    return erc20Details ? erc20Details[prop] : undefined
+  }
+
+  private async withLocalErc20Details<P extends keyof Erc20Props>(
+    params: BaseParams,
+    fn: (params: BaseParams) => Promise<Erc20Props[P]>,
+    prop: P,
+  ): Promise<Erc20Props[P]> {
+    const erc20Prop = this._getLocalErc20Property(params.tokenAddress, prop)
+
+    if (erc20Prop) {
+      return erc20Prop
+    }
+    return fn(params)
   }
 }
 
--- a/src/api/erc20/Erc20Api.ts
+++ b/src/api/erc20/Erc20Api.ts
@@ -128,7 +128,7 @@ export class Erc20ApiImpl implements Erc20Api {
 
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
-    return await erc20.methods.name().call()
+    return erc20.methods.name().call()
   }
 
   public async symbol({ tokenAddress, networkId }: SymbolParams): Promise<string> {
--- a/src/api/erc20/erc20Details.json
+++ b/src/api/erc20/erc20Details.json
@@ -1,0 +1,4 @@
+{
+  "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359": { "name": "Sai Stablecoin v1.0", "symbol": "SAI", "decimals": 18 },
+  "0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2": { "name": "Maker", "symbol": "MKR", "decimals": 18 }
+}
--- a/src/api/tokenList/TokenListApi.ts
+++ b/src/api/tokenList/TokenListApi.ts
@@ -143,6 +143,10 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
     // update copy in local storage
     const storageKey = TokenListApiImpl.getLocalStorageKey(networkId)
     localStorage.setItem(storageKey, JSON.stringify(tokenList))
+    // update address network set
+    tokenList.forEach(({ address: tokenAddress }) =>
+      this._tokenAddressNetworkSet.add(TokenListApiImpl.constructAddressNetworkKey({ tokenAddress, networkId })),
+    )
     // notify subscribers
     this.triggerSubscriptions(tokenList)
   }
--- a/src/api/tokenList/TokenListApi.ts
+++ b/src/api/tokenList/TokenListApi.ts
@@ -55,8 +55,10 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
     networkIds.forEach(networkId => {
       // initial value
       const tokenList = TokenListApiImpl.mergeTokenLists(
-        getTokensByNetwork(networkId),
+        // load first the local list, as this might be more up to date
         this.loadUserTokenList(networkId),
+        // then default list
+        getTokensByNetwork(networkId),
       )
       this._tokensByNetwork[networkId] = tokenList
 
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -117,13 +117,15 @@ export function getTokensFactory(
   async function fetchAddressesAndIds(networkId: number, numTokens: number): Promise<Map<string, number>> {
     logDebug(`[tokenListFactory][${networkId}] Fetching addresses for ids from 0 to ${numTokens - 1}`)
 
-    const promises = Array.from({ length: numTokens }, (_, tokenId) =>
-      exchangeApi.getTokenAddressById({ networkId, tokenId }),
+    const promises = Array.from({ length: numTokens }, async (_, tokenId) => {
+        const tokenAddress = await exchangeApi.getTokenAddressById({ networkId, tokenId })
+        return [tokenAddress, tokenId]
+      }
     )
 
-    const tokenAddresses = await Promise.all(promises)
+    const tokenAddressIdPairs = await Promise.all(promises)
 
-    return new Map(tokenAddresses.map((tokenAddress, id) => [tokenAddress, id]))
+    return new Map(tokenAddressIdPairs)
   }
 
   async function updateTokenDetails(networkId: number, numTokens: number): Promise<void> {
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -4,11 +4,21 @@ import { ExchangeApi } from 'api/exchange/ExchangeApi'
 import { TokenDetails, Command } from 'types'
 import { logDebug } from 'utils'
 
-export function getTokensFactory(factoryParams: {
-  tokenListApi: TokenList
-  exchangeApi: ExchangeApi
-}): (tokenId: number) => TokenDetails[] {
+import { TokenFromErc20Params, TokenFromErc20 } from './'
+
+interface Injects {
+  getTokenFromErc20(params: TokenFromErc20Params): Promise<TokenFromErc20>
+}
+
+export function getTokensFactory(
+  factoryParams: {
+    tokenListApi: TokenList
+    exchangeApi: ExchangeApi
+  },
+  injects: Injects,
+): (tokenId: number) => TokenDetails[] {
   const { tokenListApi, exchangeApi } = factoryParams
+  const { getTokenFromErc20 } = injects
 
   const updatedIdsForNetwork = new Set<number>()
 
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -37,7 +37,7 @@ export function getTokensFactory(
       token.id = await exchangeApi.getTokenIdByAddress({ networkId, tokenAddress: token.address })
       return { token }
     } catch (e) {
-      logDebug(`[network:${networkId}][address:${token.address}] Failed to fetch id from contract`, e.message)
+      logDebug(`[tokenListFactory][${networkId}][address:${token.address}] Failed to fetch id from contract`, e.message)
       return { token, failed: true }
     }
   }
@@ -78,7 +78,7 @@ export function getTokensFactory(
 
     // only bother update the list if there was anything updated/removed
     if (toRemove.size > 0 || updated.size > 0) {
-      logDebug(`[network:${networkId}] Updated ${updated.size} ids and removed ${toRemove.size} tokens`)
+      logDebug(`[tokenListFactory][${networkId}] Updated ${updated.size} ids and removed ${toRemove.size} tokens`)
       // build a new list from current list of tokens
       const tokenList = tokenListApi.getTokens(networkId).reduce<TokenDetails[]>((acc, token) => {
         if (toRemove.has(token.address)) {
@@ -98,7 +98,7 @@ export function getTokensFactory(
     // if something failed...
     if (failedToUpdate.length > 0) {
       logDebug(
-        `[network:${networkId}] Failed to fetch ids for ${failedToUpdate.length} tokens. Trying again in ${retryDelay}ms`,
+        `[tokenListFactory][${networkId}] Failed to fetch ids for ${failedToUpdate.length} tokens. Trying again in ${retryDelay}ms`,
       )
       // calculate how long to wait next time
       const nextDelay = retryDelay * (exponentialBackOff ? 2 : 1)
@@ -114,7 +114,7 @@ export function getTokensFactory(
   }
 
   async function fetchAddressesAndIds(networkId: number, numTokens: number): Promise<Map<string, number>> {
-    logDebug(`[${networkId}] Fetching addresses for ids from 0 to ${numTokens - 1}`)
+    logDebug(`[tokenListFactory][${networkId}] Fetching addresses for ids from 0 to ${numTokens - 1}`)
 
     const promises = Array.from({ length: numTokens }, (_, tokenId) =>
       exchangeApi.getTokenAddressById({ networkId, tokenId }),
@@ -131,8 +131,8 @@ export function getTokensFactory(
       fn: fetchAddressesAndIds,
       fnParams: [networkId, numTokens],
     })
-    logDebug(`[${networkId}] Token id and address mapping:`)
-    addressesAndIds.forEach((id, address) => logDebug(`[${networkId}] ${id} : ${address}`))
+    logDebug(`[tokenListFactory][${networkId}] Token id and address mapping:`)
+    addressesAndIds.forEach((id, address) => logDebug(`[tokenListFactory][${networkId}] ${id} : ${address}`))
 
     // Create a map of current token list addresses and tokens
     const localAddressesMap = new Map<string, TokenDetails>(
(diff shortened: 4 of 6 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr967--dexreact.review.gnosisdev.com

src/services/factories/tokenList.ts
```diff
@@ -93,27 +103,119 @@ export function getTokensFactory(factoryParams: {
...
+  async function getErc20DetailsOrAddress(networkId: number, tokenAddress: string): Promise<TokenFromErc20 | string> {
+    // Simple wrapper function to return original address instead of null for make logging easier
+    const erc20Details = await getTokenFromErc20({ networkId, tokenAddress })
+    return erc20Details || tokenAddress
+  }
+
+  async function fetchAddressesAndIds(networkId: number, numTokens: number): Promise<Map<string, number>> {
+    logDebug(`[${networkId}] Fetching addresses for ids from 0 to ${numTokens - 1}`)
+
+    const promises = [...Array(numTokens).keys()].map((_, tokenId) =>
```
Velenir, line 119:
```suggestion
    const promises = Array.from({ length: numTokens }, (_, tokenId) =>
```

anxolin:
> Some tokens fail to get the balance https://rinkeby.etherscan.io/address/0x2A642a7D6Fb2be3F2ADbcb3457F7953483c28213#code (don't know yet why)

That's not an ERC20, it's a 721

anxolin:
Not sure if it will make sense to have some local info for things like, SAI name, or black lists for this non ERC20 tokens, so we don't make users to fetch those

src/services/factories/tokenList.ts
```diff
@@ -93,27 +103,119 @@ export function getTokensFactory(factoryParams: {
...
-      updatedIdsForNetwork.delete(networkId)
+      updatedTokensForNetwork.delete(networkId)
     }
   }
 
   return function(networkId: number): TokenDetails[] {
-    if (!updatedIdsForNetwork.has(networkId)) {
-      updateTokenIds(networkId)
+    if (!updatedTokensForNetwork.has(networkId)) {
+      console.log(`[${networkId}] Will update tokens for network`)
```
anxolin, line 217: use logDebug? also, add sth like `[tokenListFactory]` ?

src/services/factories/tokenList.ts
```diff
@@ -2,15 +2,25 @@ import { TokenList } from 'api/tokenList/TokenListApi'
...
+    tokenListApi: TokenList
+    exchangeApi: ExchangeApi
+  },
+  injects: Injects,
+): (tokenId: number) => TokenDetails[] {
   const { tokenListApi, exchangeApi } = factoryParams
+  const { getTokenFromErc20 } = injects
 
-  const updatedIdsForNetwork = new Set<number>()
+  const updatedTokensForNetwork = new Set<number>()
```
anxolin, line 23: The name made confused me a bit at the beginning. I was thinking it was a list of tokens. 
Although a `Set` is enough, I find sth like `areTokensUpdated = new Map<Network, boolean>` a bit easier to read

src/services/factories/tokenList.ts
```diff
@@ -93,27 +103,119 @@ export function getTokensFactory(factoryParams: {
...
+
+    // Convert map values to a list. Map keeps insertion order, so new tokens are added at the end
+    const tokenList = Array.from(localAddressesMap.values())
+
+    // Persist it \o/
+    tokenListApi.persistTokens({ networkId, tokenList })
+  }
+
+  async function updateTokens(networkId: number): Promise<void> {
+    updatedTokensForNetwork.add(networkId)
```
anxolin, line 193: should we update `updatedTokensForNetwork` first thing? or should we do it once they are updated?

How would it behave with a concurrent get, for example when it's awaiting in line 196

src/services/factories/tokenList.ts
```diff
@@ -93,27 +103,119 @@ export function getTokensFactory(factoryParams: {
...
+      if (numTokens > tokens.length) {
+        // When there are more tokens in the contract than locally, fetch the new tokens
+        await updateTokenDetails(networkId, numTokens)
+      } else {
+        // Otherwise, only update the ids
+        await updateTokenIds(networkId, tokens)
+      }
     } catch (e) {
       // Failed to update after retries.
       logDebug(e.message)
```
anxolin, line 209: Maybe a bit more of context in this message? also, should it be logDebug, or it's console.error?

src/services/factories/tokenList.ts
```diff
@@ -32,7 +42,7 @@ export function getTokensFactory(factoryParams: {
     }
   }
 
-  async function updateWithRetries(
+  async function updateTokenIds(
     networkId: number,
     tokens: TokenDetails[],
     maxRetries = 3,
```
anxolin, line 48: why it needs retry logic here? it was refactored to an util, right?

src/api/tokenList/TokenListApi.ts
```diff
@@ -141,6 +143,10 @@ export class TokenListApiImpl extends GenericSubscriptions<TokenDetails[]> imple
     // update copy in local storage
     const storageKey = TokenListApiImpl.getLocalStorageKey(networkId)
     localStorage.setItem(storageKey, JSON.stringify(tokenList))
+    // update address network set
+    tokenList.forEach(({ address: tokenAddress }) =>
+      this._tokenAddressNetworkSet.add(TokenListApiImpl.constructAddressNetworkKey({ tokenAddress, networkId })),
```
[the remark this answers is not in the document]
alfetopito: @Velenir likes static :zap:

src/services/factories/tokenList.ts
```diff
@@ -2,15 +2,25 @@ import { TokenList } from 'api/tokenList/TokenListApi'
...
-import { logDebug } from 'utils'
+import { logDebug, retry } from 'utils'
 
-export function getTokensFactory(factoryParams: {
-  tokenListApi: TokenList
-  exchangeApi: ExchangeApi
-}): (tokenId: number) => TokenDetails[] {
+import { TokenFromErc20Params, TokenFromErc20 } from './'
+
+interface Injects {
```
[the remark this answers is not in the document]
alfetopito: Following the same pattern as other services. Happy to update them all if you think that's better.

src/services/factories/tokenList.ts
alfetopito: Good catch. I added them mostly for debugging and thought about removing them later, but I got to like these logs, so I'll make them `logDebug`

src/services/factories/tokenList.ts
alfetopito: 1. Historical reasons. On the previous version, it was the first network call that could cause failures. With this change, there is at least one other call before that can fail/retried
2. It all started from a TODO to only retry fetching the ids for tokens that failed, instead of query all tokens again. So, this is a sort of a custom retry function, doesn't quite fit into the generic one.

src/utils/miscellaneous.ts
```diff
@@ -88,7 +88,48 @@ export async function silentPromise<T>(promise: Promise<T>, customMessage?: stri
...
+ *
+ * Inspired by: https://gitlab.com/snippets/1775781
+ *
+ * @param fn Function to retry
+ * @param fnParams Optional parameters list to pass to function
+ * @param retriesLeft How many retries. Defaults to 3
+ * @param delay How long to wait between retries. Defaults to 1s
+ * @param exponentialBackOff Whether to use exponential back off, doubling wait interval. Defaults to true
+ */
+export async function retry<T>(params: RetryParams<T>): Promise<T> {
```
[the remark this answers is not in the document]
alfetopito: I like the separation of the fn/fn params from retry params. But I think #retries should be kept as this should be configurable by the caller.

src/services/factories/tokenList.ts
alfetopito: Yes, this is something I wasn't happy about. Although is redundant, makes sense to have a Map for clarity.

src/services/factories/tokenList.ts
alfetopito: I'm doing as a first thing to avoid having multiple update calls in parallel.

Consider the situation:
1. getTokens is called, updateTokens is triggered async and tokens are returned to the caller.
1. getTokens is called again. updateTokens haven't yet finished, flag is false, updateTokens is executed again, tokens are returned to the caller...

And so on. Makes sense?

src/services/factories/tokenList.ts
alfetopito: Ok, will update the message.

I'm not sure about `console.error` though. This is expected under some circumstances (first load and network connection is not established yet, changing networks and query is triggered with the older network)

alfetopito:
> > Some tokens fail to get the balance https://rinkeby.etherscan.io/address/0x2A642a7D6Fb2be3F2ADbcb3457F7953483c28213#code (don't know yet why)
> 
> That's not an ERC20, it's a 721

How can we tell it's a 721 token? Check something like whether it has `ownerOf` function? http://erc721.org/

src/services/factories/tokenList.ts
alfetopito: Actually, can't have myself to use Map here, Set is enough. 
But, I did rename and added a comment to make the intent more clear.

src/api/erc20/Erc20Api.ts
```diff
@@ -103,23 +121,38 @@ export class Erc20ApiImpl implements Erc20Api {
...
 
   public async name({ tokenAddress, networkId }: NameParams): Promise<string> {
+    const name = this._getLocalErc20Property(tokenAddress, 'name')
+    if (name) {
+      return name
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
     return await erc20.methods.name().call()
```
W3stside, line 131: no need to return await, just return

W3stside, line 131:
```suggestion
    return erc20.methods.name().call()
```

src/api/erc20/Erc20Api.ts
```diff
@@ -103,23 +121,38 @@ export class Erc20ApiImpl implements Erc20Api {
...
 
   public async symbol({ tokenAddress, networkId }: SymbolParams): Promise<string> {
+    const symbol = this._getLocalErc20Property(tokenAddress, 'symbol')
+    if (symbol) {
+      return symbol
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
     return await erc20.methods.symbol().call()
```
W3stside, line 142:
```suggestion
    return erc20.methods.symbol().call()
```

src/api/erc20/Erc20Api.ts
```diff
@@ -103,23 +121,38 @@ export class Erc20ApiImpl implements Erc20Api {
...
+      return decimals
+    }
+
     const erc20 = this._getERC20AtAddress(networkId, tokenAddress)
 
-    const decimals = await erc20.methods.decimals().call()
+    const decimalsString = await erc20.methods.decimals().call()
 
-    return Number(decimals)
+    return Number(decimalsString)
```
W3stside, line 155: Not sure if `undefined` is possible as a return here but this will return `NaN` if so - maybe better with a try catch and no error handling beyond rebubbling the error

W3stside: and because JS is litty, `typeof NaN === 'number'`

src/api/tokenList/TokenListApi.ts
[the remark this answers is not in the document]
W3stside: static allows this to be used outside a class instantiation right? what benefit does a private static method have over just static? unless the calling function is also static?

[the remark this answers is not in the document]
W3stside: let me just type static one more time

src/services/factories/tokenList.ts
```diff
@@ -2,15 +2,26 @@ import { TokenList } from 'api/tokenList/TokenListApi'
...
+    exchangeApi: ExchangeApi
+  },
+  injects: Injects,
+): (tokenId: number) => TokenDetails[] {
   const { tokenListApi, exchangeApi } = factoryParams
+  const { getTokenFromErc20 } = injects
 
-  const updatedIdsForNetwork = new Set<number>()
+  // Set containing ids for networks which we successfully updated the tokens from the contract
+  const areTokensUpdated = new Set<number>()
```
W3stside, line 24: nitty: `areBlahBlah` makes me auto assume `boolean` - why not `updatedTokenSet`? or other. Anyways, doesn't really matter

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+      setTimeout(() => updateTokenIds(networkId, failedToUpdate, maxRetries - 1, nextDelay), retryDelay)
     }
   }
 
-  async function updateTokenIds(networkId: number): Promise<void> {
-    // Set as updated on start to prevent concurrent queries
-    updatedIdsForNetwork.add(networkId)
+  async function getErc20DetailsOrAddress(networkId: number, tokenAddress: string): Promise<TokenFromErc20 | string> {
+    // Simple wrapper function to return original address instead of null for make logging easier
+    const erc20Details = await getTokenFromErc20({ networkId, tokenAddress })
```
W3stside, line 113: Careful what `erc20Details` returns, is it possible that this promise always resolves to something `truthy`? then this fn will fail on `114` in the sense that `tokenAddress` will never return - but then I guess TS would catch this?

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+    })
+
+    const partialTokens = await Promise.all(promises)
+
+    // For all newly fetched tokens
+    partialTokens.forEach(partialToken => {
+      if (typeof partialToken === 'string') {
+        // We replaced potential null responses with original tokenAddress string for logging purposes
+        logDebug(`[tokenListFactory][${networkId}] Address ${partialToken} is not a valid ERC20 token`)
+      } else if (partialToken) {
```
W3stside, line 172: can just be `else` - since you're not handling any other type anyways

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+        const token = localAddressesMap.get(tokenAddress) as TokenDetails
+        token.id = id
+      }
+    })
+
+    const partialTokens = await Promise.all(promises)
+
+    // For all newly fetched tokens
+    partialTokens.forEach(partialToken => {
+      if (typeof partialToken === 'string') {
```
W3stside, line 169: Can `partialToken` ever be `null`? or anything `falsy`? If not, can just do `else` after first `if`, no need for `partialToken` check

src/utils/miscellaneous.ts
```diff
@@ -88,7 +88,50 @@ export async function silentPromise<T>(promise: Promise<T>, customMessage?: stri
   try {
     return await promise
   } catch (e) {
-    console.error(customMessage || 'Failed to fetch promise', e)
+    logDebug(customMessage || 'Failed to fetch promise', e.message)
     return
   }
 }
+
+interface RetryFnParams<T> {
```
W3stside, line 96: cooooooooooool

src/api/tokenList/TokenListApi.ts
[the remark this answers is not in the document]
Velenir: private static vs outside function -- only benefit is namespacing
it literally doesn't matter

src/api/erc20/Erc20Api.ts
Velenir: Don't think `undefined` is possible, it should just break if result isn't a string according to ABI.
But I wonder if it can return `0x`

src/api/erc20/Erc20Api.ts
```diff
@@ -229,6 +262,12 @@ export class Erc20ApiImpl implements Erc20Api {
 
     return (Erc20ApiImpl._contractsCache[networkId][address] = newERC20)
   }
+
+  private _getLocalErc20Property<P extends keyof Erc20Props>(address: string, prop: P): Erc20Props[P] | undefined {
+    const erc20Details = this.localErc20Details[address]
+
+    return erc20Details ? erc20Details[prop] : undefined
```
Velenir, line 269: Please try
```ts
return this.localErc20Details[address]?.[prop]
```
I just want to see if it works. Don't actually use it, not very readable.

src/api/erc20/erc20Details.json
```diff
@@ -0,0 +1,4 @@
+{
+  "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359": { "name": "Sai Stablecoin v1.0", "symbol": "SAI", "decimals": 18 },
+  "0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2": { "name": "Maker", "symbol": "MKR", "decimals": 18 }
```
Velenir, line 3: So only for mainnet?

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+  async function fetchAddressesAndIds(networkId: number, numTokens: number): Promise<Map<string, number>> {
+    logDebug(`[tokenListFactory][${networkId}] Fetching addresses for ids from 0 to ${numTokens - 1}`)
+
+    const promises = Array.from({ length: numTokens }, (_, tokenId) =>
+      exchangeApi.getTokenAddressById({ networkId, tokenId }),
+    )
+
+    const tokenAddresses = await Promise.all(promises)
+
+    return new Map(tokenAddresses.map((tokenAddress, id) => [tokenAddress, id]))
```
Velenir, line 126:
```suggestion
    const promises = Array.from({ length: numTokens }, async (_, tokenId) => {
        const tokenAddress = await exchangeApi.getTokenAddressById({ networkId, tokenId })
        return [tokenAddress, tokenId]
      }
    )

    const tokenAddressIdPairs = await Promise.all(promises)

    return new Map(tokenAddressIdPairs)
```
minus one loop

src/utils/miscellaneous.ts
```diff
@@ -88,7 +88,50 @@ export async function silentPromise<T>(promise: Promise<T>, customMessage?: stri
...
+ */
+export async function retry<T>(params: RetryFnParams<T>, options?: RetryOptions): Promise<T> {
+  const { fn, fnParams } = params
+  const { retriesLeft = 3, delay = 1000, exponentialBackOff = true } = options || {}
+
+  try {
+    return fn(...fnParams)
+  } catch (error) {
+    if (retriesLeft) {
+      await new Promise(r => setTimeout(r, delay))
```
Velenir, line 126: didn't we have a util for that
somewhere in `utils/miscellaneous` just a few lines above :grin:

src/utils/miscellaneous.ts
```diff
@@ -88,7 +88,50 @@ export async function silentPromise<T>(promise: Promise<T>, customMessage?: stri
...
   } catch (e) {
-    console.error(customMessage || 'Failed to fetch promise', e)
+    logDebug(customMessage || 'Failed to fetch promise', e.message)
     return
   }
 }
+
+interface RetryFnParams<T> {
+  fn: (...fnParams: unknown[]) => Promise<T>
+  fnParams?: unknown[]
```
Velenir, line 98: This kind of typing allows for incorrect inputs like
```ts
retry({
  fn: async (a: number, b: number) => a + b,
  fnParams: ['', true],
})
```
You'll have to use type inference to make it work. Kinda like [here](https://github.com/gnosis/dex-price-estimator/blob/a4c2e34b6f94378df937916cc65c96040d25a7c3/src/utilities.ts#L57L58)

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+  }
+
+  async function updateTokenDetails(networkId: number, numTokens: number): Promise<void> {
+    // Fetch addresses from contract given numTokens count
+    const addressesAndIds = await retry({
+      fn: fetchAddressesAndIds,
+      fnParams: [networkId, numTokens],
+    })
```
Velenir, line 134: I would prefer a simpler kind of signature:
```ts
const addressesAndIds = await retry(() => fetchAddressesAndIds(networkId, numTokens))
```
Easier to reason about, imho

src/services/factories/tokenList.ts
```diff
@@ -88,32 +99,124 @@ export function getTokensFactory(factoryParams: {
...
+    // Persist it \o/
+    tokenListApi.persistTokens({ networkId, tokenList })
+  }
+
+  async function updateTokens(networkId: number): Promise<void> {
+    areTokensUpdated.add(networkId)
 
     try {
-      await updateWithRetries(networkId, tokenListApi.getTokens(networkId))
+      const numTokens = await retry<number>({ fn: exchangeApi.getNumTokens.bind(exchangeApi), fnParams: [networkId] })
```
Velenir, line 197: Isn't this better?
```ts
const numTokens = await retry(() => exchangeApi.getNumTokens(networkId))
```
and no need for `.bind` that destroys function type

src/api/erc20/Erc20Api.ts
W3stside: doesn't work i was trying something similar in my code for usePending

src/services/factories/tokenList.ts
W3stside: agree

src/api/erc20/erc20Details.json
alfetopito: The likelihood of having the same address for different networks AND that address not pointing to the same token is very low in my opinion.
If it is, I'd say, yes, mainnet has precedence.

Of course, I could just split by network. But is it necessary?

Velenir: nah
:+1:

src/utils/miscellaneous.ts
alfetopito: Anxo asked the same thing: https://github.com/gnosis/dex-react/pull/967#discussion_r415998111

If you guys think this is overkill, I can just drop the 'retry only failed' to retry the whole function.

Velenir: no, sorry
I meant why not use `delay` util func from this very file

src/services/factories/tokenList.ts
alfetopito: The idea was to have the nice fn name when logging https://github.com/gnosis/dex-react/pull/967/files/9eddabf6906da351a54211bb43f8a8f498d020d1#diff-ee40819644f3b6e21adb3f10e5481ff4R134

But I see the point, will simplify.

src/api/erc20/Erc20Api.ts
alfetopito: Oh well /shrug

src/services/factories/tokenList.ts
alfetopito: It cannot be null because on the wrapper promise above I replace `null` by the `tokenAddress`.
But, the type of the response can still be `null`, that's why I have the additional check after this.

I'm happy yo update that if I manage to make TS understand `partialToken` cannot be null.

src/services/factories/tokenList.ts
alfetopito: `getTokenFromErc20` returns `null` when the token is not a valid Erc20.

This function just wraps the response to, instead of null, return the token address so we can log which token failed to load.

src/utils/miscellaneous.ts
alfetopito: Ohhhhhh right. Will do :+1:

src/api/tokenList/TokenListApi.ts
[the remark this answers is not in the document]
anxolin: S * T * A  * T * I * C

Good to know. I have slight mix feelings, in one hand if we use OO approach, it make sense to not mix a class definition with some private helper function, in the other, it seems to verbose. On first sight, It looks like it's calling other file or impl
But I'm good with the approach

S * T * A  * T * I * C
S * T * A  * T * I * C

src/services/factories/tokenList.ts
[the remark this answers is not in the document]
anxolin: If it's the same as others, leave it :)

Sounded a bit verbose. Normally in IoC you just are injected directly the dependencies, I've never seen an Injects wrapper.

src/utils/miscellaneous.ts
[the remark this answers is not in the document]
anxolin: yep, you can add `retries` into the optional `RetryParams`

```js
# Flavour 1
const amazingFunction = withRetry(flakyFunction)

# Flavour 2
const amazingFunction = withRetry(flakyFunction, { delay: 500, retries: 5 })

# Call the amazingFunction
amazingFunction(param1, param2)
```

src/services/factories/tokenList.ts
anxolin: Make sense, although second call can return the same promise if it's updating?

```js
if (updatingPromise) {
  return updatingPromise 
} else {
  updatingPromise = update()
}
```

src/services/factories/tokenList.ts
anxolin: Aligned with my proposal above too

anxolin: Wait, maybe we don't say exactly the same. I was suggesting a HoF approach, so we can wrap functions? does it make sense? something like the one you've made for the price feed

src/services/factories/tokenList.ts
Velenir: Either approach works

src/services/factories/tokenList.ts
alfetopito: So you rather have a separated state for tracking when it's being updated, besides the state where we track when it had successfully updated, or have the former instead of the later?

Both are amounting to very similar results in my mind, unless you mean something different and I'm not following?

src/utils/miscellaneous.ts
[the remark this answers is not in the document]
alfetopito: It is like that now, except that the function is not a wrapper but instead it calls the passed on function.

src/api/erc20/Erc20Api.ts
alfetopito: If it does break, or return something which isn't convertable to a number, we silently ignore it and default to 18 decimals instead.

src/services/factories/tokenList.ts
W3stside: nyeh leave it

PR #981, opened 2020-04-30
Manage tokens

### LINKED ISSUE #379: Add token in the deposit page
In the bottom, a banner. "Not in the list?". 
When clicked, it shows a plain input to add it for that user (not needed to add in the contract for this issue)
Verifies that the token is in the exchange. Shows an error otherwise.
Persist in the browser the token in a different list, so it survives a refresh.

image

Related:
* Handle better tokens not in the exchange: https://github.com/gnosis/dex-react/issues/375
* Allow to add any token in the token selector: https://github.com/gnosis/dex-react/issues/378
* Accept any address in the URL params of the trade page: https://github.com/gnosis/dex-react/issues/376

Velenir:
- [x] Can add new tokens
- [x] Can disable/enable tokens displayed in `/balances`
- [x] Disabled list of addresses persists in **localStorage**
- [x] Double **Modali** breaks style
- [x] Toggle component in place of checkbox

Try with any non-added token as usual
For example, the only Token in contract but not in TCR -- 0x7de91b204c1c737bcee6f000aaa6569cf7061cb7 on mainnet

I think it's ok to maintain one list of disabled addresses regardless of **networkId**, as disabling an address can be considered disabling either that unique token or or that token address for both networks.

Persisting to **localStorage** is done through a simplistic implementation of side effects in `reducer-actions/index.ts`. Can be later migrated to @W3stside's middlewares in #999

Anyone is welcome to style **Toggles** further

Includes terrible **Modali** hack, because when there are two **Modali** modals open, second one closing removes `.modali-open` class from `body` + double overlay dark transition is hard on the eyes

Closes #379


gnosis-info:
Travis automatic deployment:
 https://pr981--dexreact.review.gnosisdev.com

anxolin:
It would be a good idea to make it very visible a link inviting the user to list any ERC20 token:
image

src/reducers-actions/localTokens.ts
```diff
@@ -0,0 +1,93 @@
...
+export const INITIAL_LOCAL_TOKENS_STATE: LocalTokensState = {
+  disabled: initializeDisabled(),
+}
+
+export function reducer(state: LocalTokensState, action: ReducerActionType): LocalTokensState {
+  const { disabled } = state
+
+  switch (action.type) {
+    case ActionTypes.ENABLE_TOKEN: {
+      const { tokenAddress } = action.payload
```
Velenir, line 59: Thanks to `enums`, this `action` is unequivocally determined to be `EnableLocalToken` and we know its payload's shape.
Same doesn't work for type union -- `action` would be `ReducerActionType = EnableLocalToken | DisableLocalToken | UpdateAllDisabled`

src/hooks/useManageTokens.tsx
```diff
@@ -0,0 +1,242 @@
...
+        autoFocus
+        autoCorrect="off"
+        autoComplete="off"
+        spellCheck="false"
+        type="text"
+        placeholder="Search by token Name, Symbol or Address"
+        {...props}
+        // value={inputValue}
+        // onChange={onChange}
+        // onKeyPress={wrappedOnKeyDown}
```
alfetopito, line 130: Dead code?

Velenir:
> After adding the token and reloading the page, the token is nowhere to be found now

Turns out that's a separate bug that got introduced with TCR. #1004

> and don't want to be added back again

This is probably because of caching from some api, probably in `services/factories/tokenList`. Should be resolved together with ^above.

src/reducers-actions/localTokens.ts
anxolin: TS is very smart, I like it 👍 

image

Great use of enums. 
... however, I wonder what happens if I use strings instead of Enums 🤔
Just for fun, let me try... 

...It works!!! (disclaimer alert, I tried, saw it worked, and then wrote "let me try" to pretend to be interesting)
And it looks simpler, we just go straight into defining the actions
```ts
type EnableLocalToken = Actions<'ENABLE_TOKEN', TokenAddressPayload>
type DisableLocalToken = Actions<'DISABLE_TOKEN', TokenAddressPayload>
type UpdateAllDisabled = Actions<'UPDATE_ALL_DISABLED', Partial<LocalTokensState>>
```

image


Although I love enums in any programming language, I don't see the point in TS yet.

Velenir:
> Since we highlight the whole token, probably it should be all clickable, not just the checkbox.

You commented just as an update with that was building on Travis :laughing: Have another look

src/reducers-actions/localTokens.ts
Velenir: > I don't see the point in TS yet

Why? They are not much different than in any other language

alfetopito: I think what he means is that with the types, there's really no need for enums.

anxolin: Exactly, I just feel is always more verbose, with the same level of type safety, the auto-suggest is worse  --> cause for enums I need to import and then I get the autosuggest (opposed to getting it right away with string approach)

PR #987, opened 2020-05-01
978/config file

### LINKED ISSUE #978: Add config file for improve forkability
So we can...
Fork 'em

@alfetopito you seemed keen on this one?

alfetopito:
Closes #978 

Most of the work done by Anxo on #1001 

I'm just giving the final touches.

Additions:
1. Load any type of file (it was enforcing a file named `config.yaml`)
2. Disregarding custom config on tests
3. Moving env vars from Define to Environment plugin on webpack config.
4. Added `logoPath` config.

Check it out:

The custom `JSON` config:
screenshot_2020-05-07_15-14-23

The result:
screenshot_2020-05-07_15-14-05

That's all I can think of. Anything else?


gnosis-info:
Travis automatic deployment:
 https://pr987--dexreact.review.gnosisdev.com

config.json
```diff
@@ -0,0 +1,14 @@
...
+  },
+  "dexPriceEstimator": {
+    "1": "https://dex-price-estimator.gnosis.io/",
+    "4": "https://dex-price-estimator.rinkeby.gnosis.io/"
+  },
+  "theGraphApi": {
+    "1": "https://api.thegraph.com/subgraphs/name/gnosis/protocol",
+    "4": "https://api.thegraph.com/subgraphs/name/gnosis/protocol-rinkeby"
+  }
+}
```
anxolin, line 14: I'll make a proposal with an example of a config. 
I think every piece of the config should fully optional, so if you create an empty yaml, or no yaml at all, would have sensible defaults.


```yaml
# Use a supported implementation with the default config
name: My cool web
logo:
  path: custom/my-logo.png
  width: 100
  height: 100

# Use a supported implementation with the default config
tokenList:
  type: 'json-file'

# Use a supported implementation with the default config
tokenList:
  type: 'gnosis-protocol-contract'


# Use a supported implementation with any other config
tokenList:
  type: 'json-file'
  config:
    path: custom/tokenList.json


# Example for the TCR you included in this PR
tcr:
  type: 'multi-tcr' # json-file, kleros, multi-tcr, none (for full list, it's not the default)
  config:    
    - networkId: 1
      listId: 1
      contractAddress: "0x1854dae560abb0f399d8badca456663ca5c309d0"
    - networkId: 4
      contractAddress: "0xBb840456546496E7640DC09ba9fE06E67C157E1b"


# Example of gas station config, although as any other config there should be a default (actually the following would be the default)
dexPriceEstimation:
  type: 'gnosis-gas-station'  
  config:
    defaultGas: standard # standard, lowest, safeLow, fast, fastest
    endpoints:
      - networkId: 1
        url: "https://dex-price-estimator.gnosis.io"
      - networkId: 4
        url: https://dex-price-estimator.rinkeby.gnosis.io

    
# Override any API implementation
tokenList:
  factory: custom/tokenList.ts
  config:
    constructorParamA: 1
    constructorParamB: 'Hi'


# All of the above configs can be used by the service/api factories to:
#   - Create an instance merging this config with the default config
#   - Normally it will be enough to provide just one implementation, ideally with some basic config parameters
#   - Ideally, we would allow to provide a "factory", that would be any custom implementation of the interface
```

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -1,7 +1,7 @@
 import BigNumber from 'bignumber.js'
 import { assert, TEN_BIG_NUMBER, ONE_BIG_NUMBER } from '@gnosis.pm/dex-js'
-import { Network } from 'types'
 import { ORDER_BOOK_HOPS_DEFAULT, ORDER_BOOK_HOPS_MAX } from 'const'
+import { DexPriceEstimatorConfig } from 'types/config'
```
anxolin, line 4: I think the config for the API should be fully defined in this file. 
In my opinion it should be the `Params`.

Also, note that this `Params` is the config for an specific implementation, maybe is time to move it out of the interface file and into something more like `RestPriceEstimationApi` or whatever we think it's convenient.

The factory responsibility would be to see the implementation the user wants to use (`type` for know implementation, `factory` for custom implementation --> probably too much for this PR)

Depending on the implementation, it would merge the defaults with the provider config, then it would hand all those parameters to the constructor of the implementation.

config.json
```diff
@@ -0,0 +1,14 @@
+{
```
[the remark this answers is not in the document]
alfetopito: Well, not a big fan on YAML. Too much ambiguity. Right, json doesn't allow comments, but there are ways of doing so.

What I really wanted was something like Django's settings where we can define the settings as code and use Typescript for ensuring correct configs are in place.

If we switch to YAML, can we have the same type safety magic we have for JSON?

Also, if we do end up going with YAML, I'd prefer to going only with YAML. Don't want to support two different types of config files.

config.json
alfetopito: Yes, I 100% agree with the optional part, for the **custom** config.

Default app config though, I prefer to have some data there for two reasons:

1. separates code from configuration
2. helps to give concrete examples of what to change when customizing

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
alfetopito: > I think the config for the API should be fully defined in this file.
> In my opinion it should be the `Params`.

You mean, to have instead of a `config` object, to expand the config into the params themselves sort of like it was before?
 
> Also, note that this `Params` is the config for an specific implementation, maybe is time to move it out of the interface file and into something more like `RestPriceEstimationApi` or whatever we think it's convenient.
> 
> The factory responsibility would be to see the implementation the user wants to use (`type` for know implementation, `factory` for custom implementation --> probably too much for this PR)

Don't fully get this part.

> Depending on the implementation, it would merge the defaults with the provider config, then it would hand all those parameters to the constructor of the implementation.

That's one thing I'd like to avoid. In my view, the config provider should take care of using/overwriting defaults. Then we don't need any extra logic on the classes to do this merge.

config.json
alfetopito: > ```
> # Use a supported implementation with the default config
> tokenList:
>   type: 'gnosis-protocol-contract'
> ```

And I don't get the `type` part on the config. How does that work?

anxolin: > Yes, I 100% agree with the optional part, for the custom config.

Not sure I follow what you like or dislikes :) 
I understand that you prefer if this doen't work by default with no-config.

My argument is that new version might bring new config, and to be backwards compatible we should have defaults.

The defaults can be just the `config-default.yaml` that get merged with `config.yaml`.
An user woud copy the default into `config.yaml` to start a configuration.

We could use YAML comments to document possible values

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
anxolin: Mmm, nup, I think we should try to see it in a call or sth. Or I can try to do a counter PR, cause it's not easy to explain

Right now I'm super tired, I'll do a tiny attempt to do a PR, if I don't manage to do that now in literally 10min, I'll send it tomorrow

config.json
alfetopito: > The defaults can be just the config-default.yaml that get merged with config.yaml.
An user woud copy the default into config.yaml to start a configuration.

Yes, that's what I mean :)

- custom user config: optional
- default app config: always present

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
alfetopito: We better then book a call for when you are available.

config.json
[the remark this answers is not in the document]
Velenir: TS understands JSON and infers types correctly. I doubt the same is true for yaml
YAML will require an additional parser, which is meh
I vote JSON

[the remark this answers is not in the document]
alfetopito: Talking about JSON with comments, there are a couple of ways of doing that:
- https://www.npmjs.com/package/strip-json-comments
- https://github.com/microsoft/node-jsonc-parser
- https://www.npmjs.com/package/comment-json
- ...

But, don't think they will integrate as nicely with TS as vanilla JSON does.

In other words, only pure JSON does not require an additional parser.
Using JSON with comments might amount the same as using YAML :/

src/config.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Contains all possible configs for the app
+ *
+ * When forking, update the file at `/custom/config.json`.
+ */
+
+const config: Config = defaultConfig
+
+// Load optional overwrites, if any
+try {
+  const customConfig = JSON.parse(fs.readFileSync('../custom/config.json', 'utf8'))
```
Velenir, line 19: OMG why? Just `require` it

Velenir: You are also pulling in `fs` even if you mock it with `empty` in `webpack.config`
And when you are mocking, it throws, and when it throws you can't get custom config.
So in web custom config is useless.

All is solved if you use `require`

webpack.config.babel.js
```diff
@@ -185,4 +185,8 @@ module.exports = ({ stats = false } = {}) => ({
     },
     runtimeChunk: true,
   },
+  node: {
+    // to be able to use `fs` module
+    fs: 'empty',
+  },
```
Velenir, line 191: There shouldn't be a need for this, it's useless when web is target anyway

src/config.ts
alfetopito: Just because I knew you would be bothered ;)

If `require` works, then it works.

test/config.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { MultiTcrConfig, DexPriceEstimatorConfig, TheGraphApiConfig } from 'types/config'
+
+describe('Test config defaults', () => {
+  it('name', () => {
+    expect(CONFIG.name).toEqual('Mesa - Gnosis Protocol DApp')
```
alfetopito, line 5: @anxolin just realized a problem with these tests.

You are comparing default configs against the global configs, which might have settings overwritten by custom config.
This test will fail and need to be updated on downstream, causing conflicts when syncing with our base branch.

For example, replacing the name in the custom config breaks it:
screenshot_2020-05-07_14-40-18

I addressed this by conditionally loading custom config.

anxolin: Good point!

jest.config.js
```diff
@@ -0,0 +1,27 @@
+const config = require('./src/loadConfig')(true)
```
anxolin, line 1: Cool ;)

src/loadConfig.js
```diff
@@ -0,0 +1,45 @@
...
+  } else {
+    throw new Error(`Unknown file extension "${extension}". Supported JSON or YAML: ${filePath} `)
+  }
+}
+
+function getCustomConfigFilePath() {
+  const customPath = path.resolve(CUSTOM_FOLDER_PATH)
+  const regex = new RegExp(`${CONFIG_FILE_OVERRIDE_NAME}\.(${SUPPORTED_EXTENSIONS})`)
+
+  const fileName = fs.readdirSync(customPath).find(fileName => regex.test(fileName))
```
anxolin, line 24: fine for me, although it looks simpler just to test the existance of the file for any of the extenssions, instead of going over all files in the dir, and testing for a pattern

sth in this direction:
```js
const extension = SUPPORTED_EXTENSIONS.find(extension => _existFile(CONFIG_FILE_OVERRIDE_NAME +  extension))
if (extension) {
   parseJsonOrYaml(CONFIG_FILE_OVERRIDE_NAME +  extension)
} else {
...
}
```

alfetopito:
@anxolin added default provider settings https://github.com/gnosis/dex-react/pull/987/commits/3d9da1f901f7081e4bbe42d5dcbdcae67893cad4

anxolin:
> @anxolin added default provider settings 3d9da1f

That's great!

One candidate for moving parts of the readme out of there if we see it big at some point is using the WIKI: https://github.com/gnosis/dex-react/wiki

PR #999, opened 2020-05-05
759.1/add middleware / fix pending orders stuff

W3stside:
Fixes saving `pendingOrders` to `localStorage` on action submission - `delete`s orders as well

Adds middleware support to global state reducer setup

1. adds middleware `pendingOrdersAndLocalStorage` which tracks action types and side effects into localstorage, if necessary - removes hook logic that previously handled this
2. fixes some inconsistency between reducer types and makes it easier to use types as action types (think constant action types also as types)
3. dont loop send pooling orders to redux, use batched array push
4. optimised `useOrders` as per @Velenir 's suggestion, should be less hefty now!
5. moved files out to separate concerns


gnosis-info:
Travis automatic deployment:
 https://pr999--dexreact.review.gnosisdev.com

src/middlewares/types.ts
```diff
@@ -0,0 +1,11 @@
...
+import { GlobalState } from 'reducers-actions'
+
+/**
+ * Middleware
+ * next: React.Dispatch<AnyAction> === the NEXT dispatch in the middleware thru-put
+ */
+export type Middleware = (
+  getState: () => GlobalState,
+  dispatch: React.Dispatch<AnyAction>,
+) => (next: React.Dispatch<AnyAction>) => (action: AnyAction) => void
```
alfetopito, line 11: This sentence is causing me brainfreeze

src/reducers-actions/orders.ts
```diff
@@ -2,7 +2,12 @@ import { Actions } from 'reducers-actions'
 import { AuctionElement } from 'api/exchange/ExchangeApi'
 import { ZERO } from '@gnosis.pm/dex-js'
 
-export type ActionTypes = 'OVERWRITE_ORDERS' | 'APPEND_ORDERS' | 'UPDATE_ORDERS' | 'UPDATE_OFFSET'
```
alfetopito, line 5: Why you no like types?
Can we keep 'em types?

W3stside: I prefer being able to use the types directly as shown inside one of the other commits. But I guess with TS it's the same? Makes it easier anyways inside Middleware for example

src/middlewares/types.ts
W3stside: Yeah it's the curried AF redux way, basically setup inside useGlobalState uses the first params getState and dispatch to pass state and dispatch, then next is the "next" Middleware in the pipeline, with action being obv

src/hooks/useGlobalState.tsx
```diff
@@ -1,25 +1,82 @@
...
+    // returns ======> e.g loggerDispatch(next)(action)
+    .map(middleware => middleware(getState, dispatch))
+
+  const enhancedDispatch = compose(...middlewareChain)(dispatchThatReturnsAction(dispatch))
+
+  return enhancedDispatch
+}
+
+function useDeepCompareRef<R>(value?: R): R | undefined {
+  const ref = useRef<R | undefined>()
```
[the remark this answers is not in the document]
W3stside: The closure is lost on rerender, ref allows persistence thru lifecycle

src/reducers-actions/orders.ts
alfetopito: I do get auto-completes for types as much as everything else, as long as the editor knows what I'm talking about. Or is this a different context where the type is not known?

W3stside:
> Don't follow what's happening here. I skipped the advanced middleware class I guess.
Essentially we needed separation of concerns specifically concerning the adding and removing of pending orders into local storage and the. Dima and I realised a Middleware approach akin to redux days would be best considering our current useReducer flow. So that is what Middleware enables us to do. In this case the pendingOrdersToLocalStorage Middleware handles adding/removing orders side effect to local storage and moves it out of hooks or inside reducers. It intercepts our action(s) concerning pending orders and adds the data to local storage before finishing it's job inside our reducers

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
alfetopito: But, why do you need to keep the reference inside the function is what I don't get.

Don't you just call the function for a given state and are done with it?

alfetopito:
> > Don't follow what's happening here. I skipped the advanced middleware class I guess.
> 
> Essentially we needed separation of concerns specifically concerning the adding and removing of pending orders into local storage and the. Dima and I realised a Middleware approach akin to redux days would be best considering our current useReducer flow. So that is what Middleware enables us to do. In this case the pendingOrdersToLocalStorage Middleware handles adding/removing orders side effect to local storage and moves it out of hooks or inside reducers. It intercepts our action(s) concerning pending orders and adds the data to local storage before finishing it's job inside our reducers

I get the middleware concept, I just don't understand the implementation :sweat_smile:

W3stside:
> > > Don't follow what's happening here. I skipped the advanced middleware class I guess.
> > 
> > 
> > Essentially we needed separation of concerns specifically concerning the adding and removing of pending orders into local storage and the. Dima and I realised a Middleware approach akin to redux days would be best considering our current useReducer flow. So that is what Middleware enables us to do. In this case the pendingOrdersToLocalStorage Middleware handles adding/removing orders side effect to local storage and moves it out of hooks or inside reducers. It intercepts our action(s) concerning pending orders and adds the data to local storage before finishing it's job inside our reducers
> 
> I get the middleware concept, I just don't understand the implementation

🥺🥺🥺🥺🥺🥺🥺🥺🥺🥺🥺🥺🥺🥺

src/reducers-actions/orders.ts
W3stside: How do you grab a ActionTypes when it's a conditional type?  Was my problem. What is your preference over the previous implementation?

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
W3stside: Haha yeah good point but the answer is so dumb. There's an infinite loop happening from the state dep being passed inside `enhancedReducer` callback. So this method with useRef allows me to pass state updates WITHOUT passing state from `useReducer` into the callback dep list. Even with deep state check it infinite loops. I'd love help with that

src/reducers-actions/orders.ts
alfetopito: Oh boy. So much I don't understand going on here :sweat_smile: 

I'd prefer to keep them as types because it's less verbose, but I won't block you from changing to Enum if that's necessary.

src/hooks/useOrders.ts
```diff
@@ -39,39 +35,47 @@ export function useOrders(): Result {
...
+          )
+
+          // Remove orders that are NULL and/or orders that have a blockNumber (indicating mined status)
+          // e.g [null, { blockNumber: 19082137, ... }, ...]
+          blockTransactionsFilteredPendingOrders = blockTransactionsFilteredPendingOrders.filter(
+            (_, index) => !!transactionStatusArray[index] && !transactionStatusArray[index].blockNumber,
+          )
+        }
+
         // don't dispatch unneccessarily
```
anxolin, line 71: Maybe say: Dispatch if a pending transaction was mined instead of "don't dispatch unneccessarily"

src/middlewares/index.ts
```diff
@@ -0,0 +1,60 @@
+import { Middleware } from './types'
+import { ActionTypes as PendingOrdersActionTypes } from 'reducers-actions/pendingOrders'
+
+import { setStorageItem } from 'utils'
+import { GP_PENDING_ORDER_KEY } from 'const'
+import { GlobalState } from 'reducers-actions'
+
+const logger: Middleware = getState => next => (action): void => {
+  process.env.NODE_ENV === 'development' &&
+    console.debug(`
```
anxolin, line 10: use logDebug

Also, there's already a logging middleware, right? why do we want to do our own?

src/middlewares/index.ts
```diff
@@ -0,0 +1,60 @@
...
+        PENDING ORDERS UPDATED
+    
+        P-ORDERS:
+        ${JSON.stringify(action.payload.orders, undefined, 2)}
+        =====>
+    `)
+
+    const { networkId, orders, userAddress } = action.payload
+
+    let newState: GlobalState['pendingOrders']
```
anxolin, line 43: Wouldn't be simpler to just add here `PendingOrdersState` as the type?

src/middlewares/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    const userPendingOrdersArr = state[networkId][userAddress] ? state[networkId][userAddress] : []
+    // add/remove pending orders to localStorage
+    if (action.type === PendingOrdersActionTypes.SAVE_PENDING_ORDERS) {
+      // save 'em
+      const newPendingTxArray = userPendingOrdersArr.concat(orders)
```
anxolin, line 48: Let's not shorten `Arr`. If it's an array, normally using the plural is enough. Otherwise, adding `list` suffix, or even if you want to add the `Array` sufix. But not `Arr`

`userPendingOrders` is not good?

src/utils/miscellaneous.ts
```diff
@@ -99,3 +99,18 @@ export function setStorageItem(key: string, data: unknown): void {
   const formattedData = JSON.stringify(data)
   return localStorage.setItem(key, formattedData)
 }
+
+// Used e.g inside middleware composition in useGlobalState
+export function compose(...funcs: Function[]): Function {
```
anxolin, line 104: redux has already a way of composing, right?
Both for middleware and for reducers

src/reducers-actions/orders.ts
W3stside: Can use `ActionTypes.APPEND_ORDERS` directly instead of `string`

src/const.ts
```diff
@@ -113,3 +113,8 @@ export const VALID_FROM_DEFAULT = '30'
 export const WETH_ADDRESS_MAINNET = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
 export const ORDER_BOOK_HOPS_DEFAULT = 2
 export const ORDER_BOOK_HOPS_MAX = 2
+export const GP_PENDING_ORDER_KEY = 'GP_ORDER_TX_HASHES'
```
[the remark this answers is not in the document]
W3stside: That's what GP is, a prefix denoting Gnosis Protocol

src/const.ts
```diff
@@ -113,3 +113,8 @@ export const VALID_FROM_DEFAULT = '30'
...
+export const REFRESH_WHEN_SECONDS_LEFT = 60 // 1min before batch done
+// for stable reference
+// to avoid updates on setState([])
+export const EMPTY_ARRAY = []
```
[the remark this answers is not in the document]
W3stside: hah yeah idk it doesn't work when empty array is created within the hook. breaks in more than this place as well, happy for someone else to look at it

src/hooks/useGlobalState.tsx
```diff
@@ -1,25 +1,82 @@
...
+import { compose } from 'utils'
 
 const GlobalStateContext = React.createContext({})
 
+const dispatchThatReturnsAction = (dispatch: React.Dispatch<AnyAction>) => (action: AnyAction): AnyAction => {
+  dispatch(action)
+  return action
+}
+
+const applyMiddlewares = (...middlewares: Middleware[]) => (
```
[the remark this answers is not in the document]
W3stside: just pipes middleware: `middleware1 > middleware2 > ... > dispatch(action)`

is there already an original `applyMiddleware`? we don't use `redux`

src/middlewares/index.ts
W3stside: true!

W3stside: artifact from testing anways, ill remove

src/middlewares/index.ts
W3stside: yep

src/middlewares/index.ts
W3stside: that works, i always use `arr` as it's clear to me but I'll change it

src/utils/miscellaneous.ts
W3stside: we don't use `redux` but i'll check if it's shipped with `react`, very likely actually

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
W3stside: > I feel like this kind of middleware implementation complicates `dispatch` handling
> 
> Wouldn't a post-effect like middleware for reducer result be a simpler approach
> [Like this](https://www.typescriptlang.org/play/?target=99#code/PTAEEsFsAcHsCcAuoDGtICNwDsCmAlXAEwFcVd4BnAGlAG9RDTz5aBBbATzZUXFmzte-bKACGlUAGEefAaAC+oAGbx0oAORpMOXAFp4xMhUoaAUGdwAPOEgjZEFZWPLTZIgDwAVUAF5xXAB89GagYaCInNC4AFygXgDcZgoW1rbIOI7wzq4c3MLy1o7YRJIyBaJ0oeEgoGwANvWwAO6SYlygRfBioNBq0UjguJKIsKAYuKBEuMq6RPYB4hUAdNVhANpdYgAK-ZRxlIjwOADmALpx7ZxJKZY2CMiR0YxGLB4Ayn4BnOydVsWlNwVL55dwCYL+AAUhzEjji71oLjk2DibAAlH5gu8kiBcXj8QTCal7nZMk4XJMAOJNDBiervRCwyZVMKjADWuGw+3isA52AAMrAUHSGUzqtESqcAPLwaZUOLbTlEaWykyixzVBBy7ky7Xq3DVI5iaZxLzdab65IWWq4cCIAAWFAi9vAkmUCFAkAQk200Hq1lQEmGZjQXOQvtglAIrxMXw+vyKSrKYNE-lBFUCkOW2cMzBMcSYxng8bqgXWZzRBZjxYRpcxITC4GUoEhuaLlGW-uwJwdfl8-gADBjDIgSPBRNDGXDQO9KzP6zCNdUmy22ywO12e-a+-4AIzD3Cj8egNcmda7s4WMIjseiU9UZanyGQvq4AButBQY-gGN8wRZ4Q3sek5MvCiIVKic6fH+DbhGEoaHKAeDNPqXyvm+IGOOByJomsgGHreqDfpCyH6thIi4XBKRhAouG3AhyCUGIb64Iqkrdl8mGxKA1KwLS9JTrgc5vrA4DzDBAEIbA-qdrAJyQho7zMacGi0IuuDLBKyrdrqJiUaAtRMSxERjIcCBiCcBr0QIiFwIchYsJx6lxLx-FkUsyKolwKZzq5IqCfWAGUM0dooNukJIiIyxPEJsHwUGoBsdpJy6VQKZeFEwzLO8bAAGoAKIAPrbPlAByAAiACSpWUoVUr4OV+X4O8MR4WEhnMZMoygGZ3SWW1PWdUlpxcfpcFAaI6lXuE0zOCQ9SIK1cHXgRx5TTRVohjZ4boHZxBfBGUYOSYrawLAiDHawvSRhd1a4WYtQeg6roqB6lBQH6kzvdMnTKMouC8JQW1huIRBEO8Ym4Plf0A8g-glmwfwAsmwJpt5GaQtU95VnmNbsIE1DVN9UMw7wcQkbgKGCWBHkiJB9YiWJhNzpdCMQv+WOrROzkzuRAiQfCgV4QxSGU6h-j3lxfPYPdcHE9D-28BTVNMtLsv4UeoikYJ1QpNZIN2YgCuw051M8TS-mq7T-N1OjOFxIz4kc2EwWheFkUCNFmUYgB8VRolSoqtq6WZR2OUFcVZVVTVdUNU1LUDR1xndb1FkGstg0scN3ajQNGCGGIbK65tIshQ6gonJZ8BfMa4OQ8bStqOdl20IbDeIPdtTCqI76cqAzTdNA9jKm+YkkHSJ7VkDwO2TdqWUO3pugaA8-uR7KK2-k9ugI7Qsu2XYUtuvXvRD7wsJdnKWqmlFQZdEYd5UVJUVdVtX1Y1zVLctSddaZox9enZaRlWKBxzupMa4R864ELsXW4aQHioG2ieM6t1cYHXQFgPAl1KCQgAuyTk3IvC8k5PgFoLdxSgKvtqBUlDUrkLCFqfMoM67THbpCeeLdrqHHnu3fSRoTTxHNNGXGhNaJAA)

imo this is easier to keep in one place succinctly

src/middlewares/index.ts
Velenir: `Yarr` :pirate_flag:

src/hooks/useGlobalState.tsx
```diff
@@ -1,25 +1,82 @@
...
+    // ==============> e.g logger(getState, dispatch)
+    // returns ======> e.g loggerDispatch(next)(action)
+    .map(middleware => middleware(getState, dispatch))
+
+  const enhancedDispatch = compose(...middlewareChain)(dispatchThatReturnsAction(dispatch))
+
+  return enhancedDispatch
+}
+
+function useDeepCompareRef<R>(value?: R): R | undefined {
```
Velenir, line 31: To remove `R | undefined` ambiguity would be nice to use overloads:
```suggestion
function useDeepCompareRef<R>(value: R): R
function useDeepCompareRef(value: undefined): undefined
function useDeepCompareRef<R>(value?: R): R | undefined {
```
or something

src/hooks/useGlobalState.tsx
```diff
@@ -1,25 +1,82 @@
...
+    // hold a reference to state here
+    const stateRef = useRef(state)
+    // deep check there are no majour changes
+    // between updates
+    const deepState = useDeepCompareRef(state)
+    // set new ref current value if no deep changes
+    useMemo(() => {
+      stateRef.current = deepState as GlobalState
+    }, [deepState])
+    const enhanceDispatchWithMiddleware = applyMiddlewares(pendingOrdersAndLocalStorage)
```
Velenir, line 62: This line can be outside of the hook?

src/const.ts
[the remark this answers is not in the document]
anxolin: but everything is GP, what's the point of this prefix? I would think  prefix that could make sense is `STORAGE_`, like in `STORAGE_PENDING_ORDER_KEY` or something that would that it's a key used as to persist it in the browser

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
anxolin: I mean, this is not what u want? https://redux.js.org/api/applymiddleware

src/utils/miscellaneous.ts
anxolin: ok, I mean this two guys:

https://redux.js.org/api/applymiddleware

https://redux.js.org/api/combinereducers

anxolin:
What are we doing with this PR?

W3stside:
> What are we doing with this PR?

I would love to add it once i address comments - it really facilitates actions handling if we need some external sideeffect and it's easy as hell to add new functionality to it

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
W3stside: only exported from redux which we dont have in this project - as it's just one or 2 functions i don't feel the need to install the lib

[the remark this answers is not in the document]
anxolin: okay, if it's not in React no issue. Yes, no need to use that library just for this, sure!

src/hooks/useDeepCompareRef.ts
```diff
@@ -0,0 +1,16 @@
+import { useRef } from 'react'
+import { isDeepEqual } from 'utils'
+
+function useDeepCompareRef(value: undefined): undefined
+function useDeepCompareRef<R>(value: R): R
+function useDeepCompareRef<R>(value?: R): R | undefined {
+  const ref = useRef<R | undefined>()
+
+  if (!isDeepEqual(value, ref.current)) {
```
anxolin, line 9: shouldn't this code be wrapped in a useEffect? or since there's only on e parameter doesn't matter?

src/hooks/useGlobalState.tsx
```diff
@@ -1,25 +1,82 @@
...
+    // apply middlewares and pass a state getting fn via ref and NO DEPS
+    const enhancedDispatch = useCallback(
+      enhanceDispatchWithMiddleware(() => stateRef.current, dispatch),
+      [],
+    )
+
+    process.env.NODE_ENV === 'development' && (window['eDispatch'] = enhancedDispatch)
```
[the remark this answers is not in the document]
anxolin: 🤷

src/middlewares/pendingOrdersToLocalStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+    let newState: PendingOrdersState
+    const userPendingOrders = state[networkId][userAddress] ? state[networkId][userAddress] : []
+    // add/remove pending orders to localStorage
+    if (action.type === 'SAVE_PENDING_ORDERS') {
+      // save 'em
+      const newPendingTxArray = userPendingOrders.concat(orders)
+      newState = { ...state, [networkId]: { ...state[networkId], [userAddress]: newPendingTxArray } }
+    } else {
+      // delete 'em
+      newState = { ...state, [networkId]: { ...state[networkId], [userAddress]: orders } }
```
anxolin, line 30: This line and 27 can be refactorized. Only the orders are different

package.json
```diff
@@ -71,6 +71,7 @@
     "@types/enzyme-adapter-react-16": "^1.0.5",
     "@types/hapi__joi": "^16.0.12",
     "@types/jest": "^25.1.1",
+    "@types/lodash.isequal": "^4.5.5",
```
anxolin, line 74: Did @Velenir approve this ;)

W3stside: nope, but it just imports `isEqual` types for the package we already have so I can just import that `fn` and not the entire lib. I think it's up to par with what @Velenir likes to see ;)

Velenir: No, and I'd rather you
```js
import isEqual from 'lodash/isEqual'
```
`lodash.isequal` as well as all other function-specific packages are old

Not to mention I'm not completely sold on using it for deep equality of redux-like state, which should be immutable in the first place

src/hooks/useGlobalState.tsx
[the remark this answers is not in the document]
alfetopito: enhanced dispatch?

package.json
W3stside: fixed

src/hooks/useDeepCompareRef.ts
W3stside: no its not really a hook, it's using `useRef` and the `use` prepend just tells react that it should follow `hook` rules e.g not being callled outside the function or being called conditionally, etc

https://reactjs.org/docs/hooks-custom.html#using-a-custom-hook

src/hooks/useGlobalState.tsx
```diff
@@ -11,15 +15,34 @@ export function withGlobalContext<P>(
...
   return function WrappedComponentWithGlobalState(props: P): JSX.Element {
     const [state, dispatch] = useReducer(reducer, initialStateFunc())
+    // hold a reference to state here
+    const stateRef = useRef(state)
+    // deep check there are no majour changes
+    // between updates
+    const deepState = useDeepCompareRef(state)
+    // set new ref current value if no deep changes
+    useMemo(() => {
+      stateRef.current = deepState as GlobalState
```
Velenir, line 25: Why even do it? `deepCompare` can get expensive depending on depth
If it's only to ensure `getState` inside middlewares is current, just do
```js
const stateRef = useRef(state)
stateRef.current = deepState
```

without extra hooks and checking

src/hooks/useDeepCompareRef.ts
anxolin: what I mean is `isDeepEqual` would be executed on every re-render of `useDeepCompareRef` that if `value` and `ref.current` don't change (the reference), you know you don't need to update the `ref.current = value` therefore you also save some calls to `isDeepEqual`
--> probably is not too important if the object has not too much deph

W3stside:
i will close this as it doesn't seem to quite make the cut - can re-open later potentially

Velenir:
But at least `pending orders stuff` could stay :cry:

PR #1001, opened 2020-05-05
978/config file proposal

anxolin:
Proposal for configuration.

It uses the original approach of defining a type with all the config, however, 

Changes a bit the approach of how we define the parameter for the APIs. One difference, is that now all the dependencies of the API is defined in the API itself, instead of importing it from `types/config`. This is related to this comment: https://github.com/gnosis/dex-react/pull/987/files#r419364366

This allows:
* That the API definition is self contain
* The API can have multiple implementation. In reality, the params can be different depending on the implementation, so it make sense to define them where they are needed

This PR moves the TCR implementation out of the definition file. This was not strictly needed, but if we decide to implement a Json based TCR or another smart-contract-based TCR, it's nice to have it in separate files. For this implementation, I gave the name of `MultiTcrApi` since it's a contract with multiple TCRs, although probably there's better names for it

For each relevant API where we allow to define config or implementation, I define a config with `type` and `config` parameter. Each type will have it's own specific config.

So, the config definitions looks like this:
```ts
// src/types/config.d.ts
export interface Config {
  tcr: MultiTcrConfig | NoTcrConfig
  dexPriceEstimator: DexPriceEstimatimatorConfig
  theGraphApi: TheGraphApiConfig
}

export type MultiTcrConfig = {
  type: 'multi-tcr'
  config: MultiTcrApiParams
}

export type NoTcrConfig = {
  type: 'none'
}

export type MultiTcrConfig = {
  type: 'multi-tcr'
  config: MultiTcrApiParams
}
...
```

And the TCR config would be:
```ts
// src/api/tcr/MultiTcrApi.ts
interface ListConfig {
  networkId: number
  contractAddress: string
  listId?: number
}

export interface MultiTcrApiParams {
  web3: Web3
  lists: ListConfig[]
}

export class MultiTcrApi implements TcrApi {
  public constructor(params: MultiTcrApiParams) {
      ...
  }
}
```

* This is what I was trying to express here: https://github.com/gnosis/dex-react/pull/987/files#r419361189

I had to fight a lot to make it work with typescrypt, specially because we want to support the tests and webpack at the same time, plus this has a GLOBAL nature that always makes things more complicated.

All the loading of the config is in `loadConfig` file, that it's used by:
* WEBPACK: To expose it as a global
* JEST: Same as ☝️

Note, that I had to move JEST config to a file so I can reuse the config loading.
I used YAML, but I'm ok changing that if it makes people sad. Also if there's a good reason like you make it more convenient cause you make the type checks work.

I've made a TEST to reduce a bit the issues with the non-type safe config, it just checks the defaults. Although, I'm sure that all this can be greatly improved. 
Trust me, it's been a struggle so far already (as always with typescript for setting things up the first time)


gnosis-info:
Travis automatic deployment:
 https://pr1001--dexreact.review.gnosisdev.com

src/types/global.ts
```diff
@@ -0,0 +1,6 @@
+// import { Config } from 'types/config'
```
Velenir, line 1: Why not `global.d.ts`?

src/types/config.d.ts
```diff
@@ -1,18 +1,2 @@
...
-
-export type TheGraphApiConfig = NetworkIdBasedConfig<string>
-
-export interface Config {
-  tcr?: TcrConfig
-  dexPriceEstimator: DexPriceEstimatorConfig
-  theGraphApi: TheGraphApiConfig
-}
+import { Config } from 'types/config'
+declare let CONFIG: Config
```
Velenir, line 2: Because you are doing `import` it changes scope of every `declare`, do
```suggestion
import { Config } from 'types/config'

declare global {
  const CONFIG: Config
}

```

src/types/config.ts
```diff
@@ -0,0 +1,29 @@
+import { TheGraphApiImplParams } from 'api/thegraph/TheGraphApi'
+import { MultiTcrApiParams } from 'api/tcr/MultiTcrApi'
+import { DexPriceEstimatorParams } from 'api/dexPriceEstimator/DexPriceEstimatorApi'
+
+export type MultiTcrConfig = {
```
Velenir, line 5: Why `type` when we usually do `interface` for such objects?

src/loadConfig.js
```diff
@@ -0,0 +1,33 @@
...
+const fs = require('fs')
+const YAML = require('yaml')
+
+const CONFIG_FILE_OVERRIDE = 'custom/config.yaml'
+const CONFIG_FILE = 'config.yaml'
+
+function parseJsonOrYaml(filePath) {
+  const content = fs.readFileSync(filePath, 'utf-8')
+  if (filePath.endsWith('.json')) {
+    return JSON.parse(content)
```
Velenir, line 11: No need for `JSON`, yaml is actually considered a superset of json, can parse it

src/loadConfig.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path')
+const fs = require('fs')
+const YAML = require('yaml')
+
+const CONFIG_FILE_OVERRIDE = 'custom/config.yaml'
+const CONFIG_FILE = 'config.yaml'
```
Velenir, line 6: We already have `.travis.yml`. Let's stick to one extension

src/api/tcr/MultiTcrApi.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    lists.forEach(list => {
+      const contract = contractPrototype.clone()
+      contract.options.address = list.contractAddress
+
+      this.contractCache[list.networkId] = {
+        // listId is optional, defaults to 0
+        listId: list.listId ?? 0,
+        contract,
+      }
```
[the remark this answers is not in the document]
anxolin: Not this PR, I just renamed the implementation and moved it to it's own file

src/types/config.d.ts
anxolin: Let me try that 😍

anxolin: It works! However, for some reason, the Global config and the other globals don't get along together, I fixed it using your tips, this is why I keep it in separate files 🤷‍♂️

src/types/config.ts
anxolin: no idea, changing!

src/loadConfig.js
anxolin: Mm, but I kind of like more yaml, it's more official. I don't see an issue
https://stackoverflow.com/questions/22268952/what-is-the-difference-between-yaml-and-yml-extension

If ir bothers u , I'll change it
Not sure if travis works with `yaml`

src/api/tcr/MultiTcrApi.ts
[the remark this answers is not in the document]
alfetopito: @Velenir no particular reason. Is this bad?

src/services/factories/tokenList.ts
```diff
@@ -133,7 +133,7 @@ export function getTokensFactory(
   }
 
   async function fetchTcrAddresses(networkId: number): Promise<Set<string>> {
-    return tcrApi ? new Set(await tcrApi.getTokens(networkId)) : new Set()
+    return new Set(tcrApi && (await tcrApi.getTokens(networkId)))
```
alfetopito, line 136: What does `new Set<string>(false)` return?

src/api/index.ts
```diff
@@ -116,26 +116,53 @@ function createTokenListApi(): TokenList {
 }
 
 function createTheGraphApi(): TheGraphApi {
-  const theGraphApi = new TheGraphApiProxy({ config: config.theGraphApi })
+  const { type, config: theGraphApiConfig } = CONFIG.theGraphApi
```
alfetopito, line 119: nit: It's scoped inside the function, why not just use `config`?

src/loadConfig.js
```diff
@@ -0,0 +1,31 @@
...
```
alfetopito, line 6: I was thinking it's better to give it a name more clear that it shouldn't be edited by hand.
What about `defaultConfig.yaml`?

Also, we should have a warning at the start of that file stating so.

src/services/factories/tokenList.ts
```diff
@@ -15,7 +15,7 @@ export function getTokensFactory(
   factoryParams: {
     tokenListApi: TokenList
     exchangeApi: ExchangeApi
-    tcrApi: TcrApi
+    tcrApi: TcrApi | undefined
```
alfetopito, line 18: Same as :point_down: ?

```suggestion
    tcrApi?: TcrApi
```

src/types/config.ts
```diff
@@ -0,0 +1,29 @@
...
+  config: DexPriceEstimatorParams
+}
+
+export interface TheGraphApiConfig {
+  type: 'the-graph'
+  config: TheGraphApiImplParams
+}
+
+export interface Config {
+  // name: string
```
alfetopito, line 25: No needed in the loaded app I guess. Remove?

webpack.config.babel.js
```diff
@@ -134,9 +137,9 @@ module.exports = ({ stats = false } = {}) => ({
       mode: 'webapp', // optional can be 'webapp' or 'light' - 'webapp' by default
       devMode: 'webapp', // optional can be 'webapp' or 'light' - 'light' by default
       favicons: {
-        appName: 'Mesa',
-        appDescription: 'Mesa',
-        developerName: 'Mesa',
+        appName: appName,
+        appDescription: appName,
+        developerName: appName,
```
alfetopito, line 142: I was going for a higher granularity, but this is fine because this is all just for the favicon I guess?

config.yaml
```diff
@@ -0,0 +1,34 @@
+# TODO: Rename Dapp into Gnosis Protocol DApp
```
alfetopito, line 1: Add warning against editing this file

src/api/tcr/MultiTcrApi.ts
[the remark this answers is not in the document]
Velenir: Not bad ,it's just lazy creation is usually better, but overhead is too small to really bother

src/types/config.d.ts
Velenir: So the genral approachis this
```ts
//something.d.ts
declare let GLOBAL_1: type
declare const GLOBAL_2 = 123456
```
variables get declared in global namespace
BUT
```ts
import something something
// or
export something
// make all declarations module scoped
// so must do
declare global {
  declare let GLOBAL_1: type
  declare const GLOBAL_2 = 123456
}
```
So in a sentence: if need `import/export` then `declare global`

src/loadConfig.js
Velenir: Ok, stick to `yaml` if you wish

src/services/factories/tokenList.ts
Velenir: `new Set(false)` breaks
but here `tcrApi` is `TcrApi | undefined`
`new Set(undefined)` is same as `new Set()`

webpack.config.babel.js
Velenir: Speaking of favicon, a more meaningful config var would be `favicon.logo` for custom icon path

src/loadConfig.js
```diff
@@ -7,9 +7,7 @@ const CONFIG_FILE = 'config.yaml'
 
 function parseJsonOrYaml(filePath) {
   const content = fs.readFileSync(filePath, 'utf-8')
-  if (filePath.endsWith('.json')) {
-    return JSON.parse(content)
-  } else if (filePath.endsWith('.yaml')) {
+  if (filePath.endsWith('.yaml') || filePath.endsWith('.json')) {
```
Velenir, line 10:
```suggestion
  if (filePath.endsWith('.yaml') || filePath.endsWith('.yml') || filePath.endsWith('.json')) {
```

Velenir: Actually a more standard way would be to use `path.extname(filePath)`

src/loadConfig.js
```diff
@@ -7,9 +7,7 @@ const CONFIG_FILE = 'config.yaml'
 
 function parseJsonOrYaml(filePath) {
   const content = fs.readFileSync(filePath, 'utf-8')
```
Velenir, line 9: Put this inside `if`, no need to even read the file if unsupported extension.

src/services/factories/tokenList.ts
alfetopito: ohhhh

src/loadConfig.js
anxolin: `.ini` or `.ion` is out of the question, right? or using `.php` but writing a yaml

src/api/index.ts
anxolin: good point, I think I had a global config before, and just wanted to go specific here. Having `config` and `CONFIG` might be slightly confusing, but I'm with you that it pays off by being simpler

src/loadConfig.js
anxolin: My thinking was, that is nice to keep everything related to the custom fork inside `custom/`
The fork steps would guide you how to create the config

> What about defaultConfig.yaml?

Yep, good idea. What about `config-default.yaml`? or `default-config.yaml`

src/types/config.ts
anxolin: It is needed, I would make it mandatory. Not sure how it works without this

anxolin: I've just made it mandatory and added a test for it

webpack.config.babel.js
anxolin: Yep, I wouldn't overcomplicate this allowsing too much granularity

For the logo, I agree, I think indeed we should allow to set/override it, OR use a naming convention, or both `custom/favicon.png`, not sure if it's good or bad :)

src/loadConfig.js
alfetopito: I'm fine with either. Just trying to drive the point that this shouldn't be edited if people want to keep the fork compatible.

config.yaml
anxolin: Good idea, done

anxolin: feel free to update the texts to your liking when you take over 👌

src/services/factories/tokenList.ts
anxolin: I had the same first impression as Leandro, then I realised
I have to admit I'm not super fan of relying on this logics, I prefer to be explicit, even on the cost of being slightly more ineficient. But in this case, the solution was good, was not that complicated, and makes dima happier

Velenir: It's ok to push harder for your way. I don't much care as long as it works
Ternary is ternary :+1:

PR #1004, opened 2020-05-08
Fix tcr override

Velenir:
On load `USER_TOKEN_LIST` was overridden with tokens from **TCR**.
As a result if user previously added an extra token not in **TCR**, it would disappear on page refresh

[Related](https://github.com/gnosis/dex-react/pull/981#issuecomment-625725927)

### DIFF
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -95,7 +95,7 @@ export function getTokensFactory(
       }, [])
 
       // persist updated list
-      tokenListApi.persistTokens({ networkId, tokenList })
+      tokenListApi.addTokens({ networkId, tokens: tokenList })
     }
 
     // if something failed...
@@ -209,7 +209,7 @@ export function getTokensFactory(
     const tokenList = Array.from(localAddressesMap.values())
 
     // Persist it \o/
-    tokenListApi.persistTokens({ networkId, tokenList })
+    tokenListApi.addTokens({ networkId, tokens: tokenList })
   }
 
   async function updateTokens(networkId: number): Promise<void> {
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -95,7 +95,7 @@ export function getTokensFactory(
       }, [])
 
       // persist updated list
-      tokenListApi.addTokens({ networkId, tokens: tokenList })
+      tokenListApi.persistTokens({ networkId, tokenList })
     }
 
     // if something failed...
@@ -209,7 +209,7 @@ export function getTokensFactory(
     const tokenList = Array.from(localAddressesMap.values())
 
     // Persist it \o/
-    tokenListApi.addTokens({ networkId, tokens: tokenList })
+    tokenListApi.persistTokens({ networkId, tokenList })
   }
 
   async function updateTokens(networkId: number): Promise<void> {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1004--dexreact.review.gnosisdev.com

src/services/factories/tokenList.ts
```diff
@@ -95,7 +95,7 @@ export function getTokensFactory(
       }, [])
 
       // persist updated list
-      tokenListApi.persistTokens({ networkId, tokenList })
+      tokenListApi.addTokens({ networkId, tokens: tokenList })
```
alfetopito, line 98: One concern though: what about tokens that should be removed? (are not registered on the contract or are not valid erc20 tokens) 
- https://github.com/gnosis/dex-react/pull/1004/files#diff-e5e8192286e6ad1d544324c3dce8ed97R87
- https://github.com/gnosis/dex-react/pull/1004/files#diff-e5e8192286e6ad1d544324c3dce8ed97R193

Velenir: I wonder. Theoretically the user won't add them as we have a check in adding logic

Velenir: Now if a token was first on the contract and then removed, hm...

alfetopito: Well, with TCR this is less likely, so might be negligible.

> I wonder. Theoretically the user won't add them as we have a check in adding logic

This is more for the (possibly faulty) lists that we provide than user input.

Let's consider when/how this might happen:

1. Token on the default list and not on the contract.
  Reason: We re-used the same list for 2 networks. In that case the token might not exist on a network at all.
  Fix: make sure the tokens in the list exist on all networks we want to use it, or include only on the ones that the token is deployed.
2. Token on TCR but not on the contract
  Reason: Self-explanatory
  Fix: either remove token from TCR or add it to the contract
3. Token on TCR but not deployed to current network OR not ERC20
  Reason: Self-explanatory
  Fix: either remove token from TCR or deploy to this network add it to the contract

I'm not worried about users adding these tokens.
It's all about we (who control the default token list) and the TCR owners keeping the house clean and not allowing this to happen.

> Now if a token was first on the contract and then removed, hm...

This can't happen. The exchange does not allow to remove tokens.

----

What if instead of appending the new tokens (this change) we add the local list of user added tokens after filtering the TCR list?

Velenir: > This can't happen. The exchange does not allow to remove tokens.

I was thinking more of a new contract address, like did happen previously on rinkeby. Shouldn't be a problem on mainnet though.

alfetopito: > What if instead of appending the new tokens (this change) we add the local list of user added tokens after filtering the TCR list?

I thought the user added tokens were separated from provided token list.
This makes things a bit messier.

Velenir: User List was separate originally and then got merged at some point when persisting to storage. When we took out tokenList.json, probably
Might be a good idea to separate it
Have at it :sunglasses:

alfetopito: Ok, I'll give it a try :)

alfetopito: https://github.com/gnosis/dex-react/pull/1006

alfetopito: Merge in my changes

PR #997, opened 2020-05-04
180/update readme with tcr info

alfetopito:
Part of #180 

Breaking off README from https://github.com/gnosis/dex-react/pull/983#discussion_r419324865

PR #1005, opened 2020-05-08
Updated README with details for Config options

alfetopito:
Depends on #987 

~Pending legal review regarding mention to DxDao's TCR and wording.~ Done

Superseding https://github.com/gnosis/dex-react/pull/997

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,8 @@ Path to the template html file.
 
 Tokens are dynamically loaded from the contract, but it might not be desirable to display everything in the interface.
 
+Gnosis Protocol is a fully permissionless trading protocol for ERC-20, as such, anyone can [enable a token for trading](https://docs.gnosis.io/protocol/docs/addtoken1/). Tokens of dubious value or nature or those not compatible with the ERC-20 standard may also be added. Accordingly, it is the responsibility of the Site operator to determine which tokens listed on the permissionless Gnosis Protocol are displayed on their Site.
+
 To dynamically control which tokens are displayed without the need of a redeployment, it's possible to use a Token Curated Registry (TCR) contract per network.
 
 The only requirement is that this contract implements the following method:
--- a/README.md
+++ b/README.md
@@ -111,7 +111,31 @@ import About from 'pages/About'
 
 ## Config
 
-### TCR (Optional)
+Default app configs can be found on [the default config file](./config-default.yaml)
+
+We recommend against editing this file directly, though.
+
+If you wish to replace any default config, create a file named `config` inside the [`custom` folder](./custom).
+
+Both JSON and YAML formats are supported.
+
+Simply replace any config found on [config-default](./config-default.yaml).
+
+Below we provide details for each config.
+
+### `name`
+
+A single string that controls the page title and favicon metadata.
+
+### `logoPath`
+
+Path to favicon logo.
+
+### `templatePath`
+
+Path to the template html file.
+
+### `tcr`
 
 Tokens are dynamically loaded from the contract, but it might not be desirable to display everything in the interface.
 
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1005--dexreact.review.gnosisdev.com

README.md
```diff
@@ -111,7 +111,31 @@ import About from 'pages/About'
 
 ## Config
 
-### TCR (Optional)
+Default app configs can be found on [the default config file](./config-default.yaml)
+
+We recommend against editing this file directly, though.
+
+If you wish to replace any default config, create a file named `config` inside the [`custom` folder](./custom).
```
anxolin, line 118: Probably we can create already the custom folder and leave it empty, just leave an empty `.gitkeep` file

README.md
```diff
@@ -111,7 +111,31 @@ import About from 'pages/About'
...
+
+Both JSON and YAML formats are supported.
+
+Simply replace any config found on [config-default](./config-default.yaml).
+
+Below we provide details for each config.
+
+### `name`
```
anxolin, line 126: Maybe we can use a table like markdown? showing the name, the description, the default value

anxolin: I continued reading, and a table maybe makes it more complicated for the TCR part, cause it's a bit complex. So maybe just ignore my comment if there's no easy way to make a compact description of the parameters

Ultimately, I think we might find ourselves replicating between the README.md and the `config-default.yaml` so it's not bad to consider being brief in the readme, and use comments in the default file to document things if we come to that point.

README.md
alfetopito: I did that at some point but I guess it was lost in a merge/rebase. Will add it.

README.md
alfetopito: We can split this into a separated readme/wiki/etc for config if this becomes to big.
For me this is still manageable. I've seem much larger readmes out there :)

anxolin: Cool!

webpack.config.babel.js
```diff
@@ -155,25 +158,24 @@ module.exports = ({ stats = false } = {}) => ({
...
-      'process.env.MOCK_WALLET': JSON.stringify(process.env.MOCK_WALLET || process.env.MOCK || 'false'),
-      'process.env.MOCK_TOKEN_LIST': JSON.stringify(process.env.MOCK_TOKEN_LIST || process.env.MOCK || 'false'),
-      'process.env.MOCK_ERC20': JSON.stringify(process.env.MOCK_ERC20 || process.env.MOCK || 'false'),
-      'process.env.MOCK_DEPOSIT': JSON.stringify(process.env.MOCK_MOCK_DEPOSIT || process.env.MOCK || 'false'),
-      'process.env.MOCK_EXCHANGE': JSON.stringify(process.env.MOCK_EXCHANGE || process.env.MOCK || 'false'),
-      'process.env.MOCK_WEB3': JSON.stringify(process.env.MOCK_WEB3 || process.env.MOCK || 'false'),
-
-      // AUTOCONNECT: only applies for mock implementation
-      'process.env.AUTOCONNECT': JSON.stringify(process.env.AUTOCONNECT || 'true'),
+      CONFIG: JSON.stringify(config),
```
Velenir, line 178: I still think this is a bad idea on account of no treeshaking
I still think a good loader would be better
I'll need experiment with that

alfetopito: You mean putting all the config into a global var?
In terms of the effects on treeshaking, how big does the config have to be in order to have an actual impact? I have no idea.

Velenir: I'll test stuff tomorrow. Will see

alfetopito: :crossed_fingers:

PR #1007, opened 2020-05-13
Inline config test

Velenir:
So inlining objects is done rather smartly
In case of 
```js
console.log('CONFIG.tcr', CONFIG.tcr)
console.log('CONFIG.tcr', CONFIG.tcr)
```
`CONFIG.tcr` will be inlined twice in whole

So it's better to use
```js
const { tcr } = CONFIG
console.log('tcr', tcr)
console.log('tcr', tcr)
```

But when used in conditions (including `switch-case`), need to use full path for better treeshaking
```js
if (CONFIG.defaultProviderConfig.type === 'infura')
// or
switch (CONFIG.tcr.type)
```
When doing
```js
const {tcr}= CONFIG
switch (tcr.type)
```
doesn't treeshake.

All in all inlining `CONFIG` works all right, unless we start spamming `CONFIG.tcr` in many files, no need for a loader.

This PR isn't for merging, of course, but to illustrate inlining.
Still, @alfetopito, cherry-pick 46ef5f9 into your PR for treeshaking and better type inference.

### DIFF
--- a/src/actualUseCase.ts
+++ b/src/actualUseCase.ts
@@ -1,0 +1,37 @@
+import { TcrApi } from 'api/tcr/TcrApi'
+
+// src/const.ts
+let ethNodeUrl
+if (CONFIG.defaultProviderConfig.type === 'infura') {
+  const { config } = CONFIG.defaultProviderConfig
+  ethNodeUrl = config.infuraEndpoint + config.infuraId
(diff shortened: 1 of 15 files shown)


anxolin:
Very informative. Thanks for checking this

alfetopito:
Dima saves the day again \o/

Velenir:
Served its purpose

PR #1009, opened 2020-05-13
943/number formatting

### LINKED ISSUE #943: Smarter decimal formatting
image

When we display some numbers, we show only a few decimals while the precisions are bigger. We should have smarter ways of representing decimals.

Also applies to the balances, where for exmple, if you have just a few dollars in WBTC you'll still see `0.0000`

W3stside:
Closes #943 

Keeping it as a draft as i dont like it at all

PLease tear it apart i spent way too much time on it juggling stupid types and edge cases left and right
@anxolin @Velenir @alfetopito


gnosis-info:
Travis automatic deployment:
 https://pr1009--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -81,12 +83,30 @@ const processData = (
...
   return (
     list
       // Account fo decimals
       .map(element => {
+        const decimals = quoteToken.decimals - baseToken.decimals
+        // get the full native string amount to pass into parsedPrice
+        // as price comes in from API as number type
+        const price = (element.price / 10 ** decimals).toLocaleString([], { maximumSignificantDigits: 18 })
+        // return it as a BN via our parseAmount
+        const parsedPrice = parseAmount(price, 18)
```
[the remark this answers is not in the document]
W3stside: the precision is handled in the line above, that should suffice, no? the `parsedPrice` is only to get it into `BN` form before passing it into `smartFormat` - the precision there needs to be the same. Probably best to just wrap `parseAmount` into `smartFormat` to avoid needing to place precision twice

src/utils/format.ts
```diff
@@ -98,3 +101,113 @@ export const formatTimeToFromBatch = (
...
+const lt1kFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 5 })
+const lt10kFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 4 })
+const lt100kFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 3 })
+const lt1mFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 2 })
+const lt10mFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 1 })
+const lt100mFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 0 })
+// same format for billions and trillions
+const lt1000tFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 3, notation: 'compact' })
+
+interface FormatAmountParams<T> {
```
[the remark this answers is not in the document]
W3stside: will do! didnt want to create a new version exporting things if it wasnt first approved

src/utils/format.ts
```diff
@@ -98,3 +101,113 @@ export const formatTimeToFromBatch = (
...
+
+  const stringAmount = formatAmount({
+    amount,
+    precision,
+    decimals,
+    isLocaleAware: false,
+    thousandSeparator: false,
+  })
+
+  return formatAmountForDisplay(stringAmount, { smallLimit })
```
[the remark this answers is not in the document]
W3stside: yeah i tried that, was a disaster. then tried to just infer types from `formatAmount` etc but either TS can't do it or my TS-fu is bad

src/utils/format.ts
```diff
@@ -98,3 +101,113 @@ export const formatTimeToFromBatch = (
   returnType: 'TIME' | 'BATCH' = 'TIME',
   batchLength = 5,
 ): number => (returnType === 'TIME' ? Number(value) * batchLength : Number(value) / batchLength)
+
+// TODO: move
+// Locale is an empty array because we want it to use user's locale
+const ltFractionFormatter = new Intl.NumberFormat([], { maximumFractionDigits: 18 })
```
anxolin, line 107: Are we sure we want to go this road? multiple formatters?

The formatters we have didn't use Intl probably for the issue with big numbers. right?

In my mind it was reusing our other formatter, so we don't have to solve the issues again:
```js
formatSmart (...same arguments as formatAmount) {
   const decimals = getSmartDecimals(parameters)
   return formatAmount({...parameters, decimals })
}
```

Maybe I'm oversimplifying and `getSmartDecimals` should return an optional sufix like `M` or `K`, but I find it simpler

src/utils/format.ts
```diff
@@ -98,3 +101,113 @@ export const formatTimeToFromBatch = (
...
+  let decimals = 4
+  let smallLimit = DEFAULT_SMALL_LIMIT
+
+  if (!params || ('amount' in params && !params.amount)) {
+    return null
+  } else if (BN.isBN(params)) {
+    amount = params
+    precision = _amountPrecision as number
+  } else {
+    amount = params.amount as BN
```
anxolin, line 189: why as BN?

src/utils/format.ts
anxolin: EDIT: What, I'm confused. Now I see there's 2 methods: `formatSmart` and `formatAmountForDisplay`. I thought `formatSmart` would be exactly what we use for displaying the amount

src/utils/format.ts
alfetopito: This is the same as `formatAmount` on `dex-js`. The function has 2 signatures:
1. `formatAmount(amount, precision)`
2. `formatAmount(params)` where params is an object

This part of the code is decomposing the `params`

W3stside:
closing in favour of a PR in dex-js - also reworks lots of stuff

PR #1010, opened 2020-05-14
Fix TradePage warning label size

W3stside:
1. warning could potentially grow too large in size
2. would push elements everywhere
3. this fixes the dynamically sized text to ellipsis

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -23,6 +23,7 @@ const Wrapper = styled.div`
   justify-content: space-between;
   align-items: flex-start;
   flex-flow: column wrap;
+  width: 100%;
 
   > div:first-of-type {
     width: 100%;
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1010--dexreact.review.gnosisdev.com

W3stside:
to test - just try inputting a massive amount into sell and/or changing the token symbol to some ridiculous length

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -176,17 +177,20 @@ const TokenRow: React.FC<Props> = ({
   const balanceClassName = !error && sellAmountOverMax ? 'warning' : 'success'
   const inputClassName = error ? 'error' : sellAmountOverMax ? 'warning' : ''
 
+  const fullOverMaxAmount = overMax.gt(ZERO) && formatAmountFull({ amount: overMax, precision: selectedToken.decimals })
```
Velenir, line 180: Consider
```suggestion
  const fullOverMaxAmount = overMax.gt(ZERO) ? formatAmountFull({ amount: overMax, precision: selectedToken.decimals }) : ''
```
so that you don't have to `fullOverMaxAmount as string`

anxolin:
What are we doing with this one?

PR #1011, opened 2020-05-14
952/set slippage

### LINKED ISSUE #952: Allow to select the maximum slippage
Consider making the price estimation more tight (don't subtract any value on top of the fees), and then offer these options:

image

`0.5%` would be the default option and pre-selected, so the price inputs it's already auto-filled with this value.

@koeppelmann, this percentages goes on top of the fees, right? because the `actual price - 0.2%` is the price that would make a solution feasible, so that would be the price at `0%` (which we don't offer)

Let's see the proposal with an example:
* An user wants to change `100 DAI for USDC`
* There's an order of `100 USDC for 100 DAI` in the order book
* The price estimation will return `99.8`

The UI will offer:
* 0.1% = 99.7002
* 0.5% (default) = 99.301
* 1% = 98.802

For the custom field, 0% would be a valid value, but also negative amounts, that would give you a better price, but the order won't probably be fulfilled soon.

cc/ @fleupold @alfetopito @c3rnst

W3stside:
Closes #952 

Set Slippage

- [x] make component
- [x] style the thing
- [x] make smart

### DIFF
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -4,7 +4,7 @@ import { usePopperOnClick, usePopperDefault, TOOLTIP_OFFSET } from 'hooks/usePop
 import { State, Placement } from '@popperjs/core'
 import styled from 'styled-components'
 import { isElement, isFragment } from 'react-is'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { FontAwesomeIcon, FontAwesomeIconProps } from '@fortawesome/react-fontawesome'
 import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons'
 
 // visibility necessary for correct boundingRect calculation by popper
@@ -188,6 +188,7 @@ interface HelpTooltipProps {
   tooltip: ReactNode
   placement?: Placement
   offset?: number
+  iconSize?: FontAwesomeIconProps['size']
 }
 
 const HelperSpan = styled.span`
@@ -206,7 +207,7 @@ export const HelpTooltipContainer = styled(LongTooltipContainer)`
   color: black;
 `
 
-export const HelpTooltip: React.FC<HelpTooltipProps> = ({ tooltip, placement = 'top', offset }) => {
+export const HelpTooltip: React.FC<HelpTooltipProps> = ({ tooltip, placement = 'top', offset, iconSize }) => {
   const {
     targetProps: { ref, onClick },
     tooltipProps,
@@ -223,7 +224,7 @@ export const HelpTooltip: React.FC<HelpTooltipProps> = ({ tooltip, placement = '
   return (
     <>
       <HelperSpan ref={ref} onClick={handleClick}>
-        <FontAwesomeIcon icon={faQuestionCircle} />
+        <FontAwesomeIcon icon={faQuestionCircle} size={iconSize} />
       </HelperSpan>
       <Tooltip {...tooltipProps} bgColor="#bfd6ef">
         {tooltip}
--- a/src/components/TradeWidget/MaximumSlippage.tsx
+++ b/src/components/TradeWidget/MaximumSlippage.tsx
@@ -167,4 +167,13 @@ const MaximumSlippage: React.FC<MaximumSlippageProps> = ({ setNewSlippage, price
   )
 }
 
+/*
+  // to add in Price component to show current selected slippage
+  {priceSlippage && (
+    <FormMessage className="warning">
+      <small>{priceSlippage}% slippage</small>
+    </FormMessage>
+  )}
+*/
+
 export default MaximumSlippage
--- a/src/components/TradeWidget/MaximumSlippage.tsx
+++ b/src/components/TradeWidget/MaximumSlippage.tsx
@@ -2,9 +2,10 @@ import React from 'react'
 import styled from 'styled-components'
 
 import { HelpTooltip } from 'components/Tooltip'
-import { MEDIA } from 'const'
+import { MEDIA, SLIPPAGE_MAP } from 'const'
+import { PriceSlippageState } from 'reducers-actions/priceSlippage'
 
-const MaximumSlippageWrapper = styled.div`
+const Wrapper = styled.div`
   background: var(--color-background-input);
   border-radius: var(--border-radius-top);
   padding: 1rem;
@@ -21,8 +22,12 @@ const MaximumSlippageWrapper = styled.div`
     margin: 1rem auto 0.2rem;
     width: 100%;
 
+    > button {
+      border-radius: 3rem;
+    }
+
     > button,
-    > label > input[type='text'] {
+    > label > input {
       flex: 1;
 
       background: var(--color-background-pageWrapper);
@@ -30,7 +35,6 @@ const MaximumSlippageWrapper = styled.div`
       font-size: inherit;
       font-weight: normal;
 
-      border-radius: 3rem;
       height: 3rem;
 
       padding: 0.65rem 1.5rem;
@@ -41,6 +45,7 @@ const MaximumSlippageWrapper = styled.div`
       white-space: nowrap;
 
       &.selected,
+      &.selected ~ small,
       &:hover:not(input),
       &:focus,
       &:focus ~ small {
@@ -48,8 +53,12 @@ const MaximumSlippageWrapper = styled.div`
       }
 
       &.selected,
-      &:hover:not(input),
+      &:hover:not(input):not(.selected),
       &:focus {
+        background: var(--color-background-button-hover);
+      }
+
+      &.selected {
         background: var(--color-background-CTA);
       }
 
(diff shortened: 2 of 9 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1011--dexreact.review.gnosisdev.com

src/components/TradeWidget/MaximumSlippage.tsx
```diff
@@ -0,0 +1,132 @@
...
+  suggested?: boolean
+}
+
+interface MaximumSlippageProps {
+  availableSlippage?: SlippageParam[]
+}
+
+const DEFAULT_AVAILABLE_SLIPPAGE = [{ value: 0.1 }, { value: 0.5, suggested: true }, { value: 1 }]
+
+const MaximumSlippage: React.FC<MaximumSlippageProps> = ({ availableSlippage = DEFAULT_AVAILABLE_SLIPPAGE }) => {
```
anxolin, line 110: I guess you'll pass the `value` and `onChangeSlippage` as parameters,

Rafanator:
Limit additional slippage is clear enough. I don't think it needs a tooltip.

src/components/TradeWidget/MaximumSlippage.tsx
```diff
@@ -0,0 +1,132 @@
...
+      &:not(:last-child) {
+        margin-right: 1rem;
+      }
+
+      white-space: nowrap;
+
+      &.selected,
+      &:hover:not(input),
+      &:focus,
+      &:focus ~ small {
```
Velenir, line 46: Love these CSS selectors :+1:

src/components/TradeWidget/MaximumSlippage.tsx
```diff
@@ -0,0 +1,132 @@
...
+const DEFAULT_AVAILABLE_SLIPPAGE = [{ value: 0.1 }, { value: 0.5, suggested: true }, { value: 1 }]
+
+const MaximumSlippage: React.FC<MaximumSlippageProps> = ({ availableSlippage = DEFAULT_AVAILABLE_SLIPPAGE }) => {
+  return (
+    <MaximumSlippageWrapper>
+      <small>Limit additional price slippage</small>{' '}
+      <HelpTooltip iconSize="xs" tooltip={'Set additional slippage parameters'} />
+      <div>
+        {availableSlippage.map(({ value, suggested }, index) => (
+          <button key={value + index} type="button" onClick={(): void => alert(`Slippage Value:: ${value}`)}>
```
Velenir, line 117: Why `key={value + index}`? Not that it matters but this is no more unique than just `key={index}`

src/components/TradeWidget/MaximumSlippage.tsx
W3stside: :grimacing:

src/components/TradeWidget/MaximumSlippage.tsx
W3stside: :thinking:

W3stside:
> The styles can be improved a bit still
> 
> * tooltip in dark mode is bad, in light mode it's very different than the others??
> * tooltip icon has different size
> * The title, should be like the other labels, right?
> * Also the font of the buttons, feels a bit different
> * The square has the color we use for INPUTS, and the button has the color that we should use for the background
> 
> @Rafanator, maybe can help with the tooltip there.
> image

Just styled it as you did:
image

will re-style as suggested

src/reducers-actions/index.ts
```diff
@@ -15,9 +20,14 @@ export interface Actions<T, P> {
   payload: P
 }
 
+export type ActionCreator<T, P> = (payload: P) => Actions<T, P>
+
+export type ReducerCreator<S, A> = (state: S, action: A) => S
```
Velenir, line 25:
```suggestion
export type ReducerCreator<S, A extends Actions> = (state: S, action: A) => S
```
maybe

src/components/TradeWidget/Price.tsx
```diff
@@ -188,10 +202,19 @@ const Price: React.FC<Props> = ({ sellToken, receiveToken, priceInputId, priceIn
     precision: DEFAULT_PRECISION,
   })
 
+  // PRICE SLIPPAGE
+  const dispatchNewSlippage = (payload: string): void => dispatch(setPriceSlippage(payload))
```
Velenir, line 206: Memoize?

src/components/TradeWidget/Price.tsx
```diff
@@ -21,17 +25,26 @@ const Wrapper = styled.div`
...
     font-size: 1.5rem;
+
     @media ${MEDIA.mobile} {
       font-size: 1.3rem;
     }
 
+    > ${FormMessage} {
+      width: min-content;
+      white-space: nowrap;
+      font-size: x-small;
```
Velenir, line 44: Do we ever use font-size keywords? We usually do `*(r)em`, no? Not that it matters

src/components/TradeWidget/MaximumSlippage.tsx
```diff
@@ -0,0 +1,170 @@
...
+      <strong>Limit additional price slippage</strong>
+      <div>
+        {slippagePercentages.map((slippage, index) => (
+          <button
+            key={index}
+            type="button"
+            onClick={(): void => setNewSlippage(slippage)}
+            className={slippage === priceSlippage ? 'selected' : ''}
+          >
+            {slippage}%{SLIPPAGE_MAP.get(slippage) && <small>(suggested)</small>}
```
Velenir, line 151:
```js
SLIPPAGE_MAP.get(slippage)
```
is always true as you construct `slippagePercentages` array from that same map

Rafanator:
Hi, just wanted to say that in general I would oppose having a slippage section on Mesa. 

Reasons:
- GP has no slippage 
- Adding this will cramp up an already full interface 
- I believe we could tweak the price estimator to suggest worst prices to achieve what we are doing with the slippage calculator (Getting users to accept worst order fills) 

 cc/ @c3rnst @anxolin

anxolin:
This was suggested by @koeppelmann, happy to re-review:

* GP has no slippage --> well, you create limit prices, there's not "traditional slippage", but what this PR does is to create an order **below the current Gnosis Protocol price** --> The order will likely execute at a better price, so you are basically limiting the maximum slippage
* Adding this will cramp up an already full interface --> 100% agree. Make it an advanced option makes it any better?
* I believe we could tweak the price estimator to suggest worst prices to achieve what we are doing with the slippage calculator (Getting users to accept worst order fills) --> Yes, but that would offer always a worse price than available. I think the reasoning is to allow the user decide his slippage.

W3stside:
> This was suggested by @koeppelmann, happy to re-review:
> 
> > GP has no slippage
> 
> Well, you create limit prices, there's not "traditional slippage", but what this PR does is to create an order **below the current Gnosis Protocol price** --> The order will likely execute at a better price, so you are basically limiting the maximum slippage
> 
> > Adding this will cramp up an already full interface
> 
> 100% agree. Make it an advanced option makes it any better?
> 
> > I believe we could tweak the price estimator to suggest worst prices to achieve what we are doing with the slippage calculator (Getting users to accept worst order fills)
> 
> Yes, but that would offer always a worse price than available. I think the reasoning is to allow the user decide his slippage.

Okay now I need to ask, did we not clear it with anyone beforehand?

src/reducers-actions/priceSlippage.ts
```diff
@@ -0,0 +1,30 @@
+import { Actions, ActionCreator, ReducerCreator } from 'reducers-actions'
+import { DEFAULT_SUGGESTED_SLIPPAGE } from 'const'
+
+const ActionsList = ['SET_PRICE_SLIPPAGE'] as const
```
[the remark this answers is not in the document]
W3stside: i know right!? found it while thinking why can't i infer types from an array? then i can use directly AND have them inferred/autocompleted. looks a bit weird tho

src/components/Tooltip.tsx
```diff
@@ -188,6 +188,7 @@ interface HelpTooltipProps {
   tooltip: ReactNode
   placement?: Placement
   offset?: number
+  iconSize?: FontAwesomeIconProps['size']
```
anxolin, line 191: I had to do something similar. Probably better to use the type:
here https://github.com/gnosis/dex-react/pull/1025/files#diff-493194965e4afe11bf4448365069f0aeR9

src/hooks/usePriceEstimation.ts
```diff
@@ -73,7 +104,9 @@ export function usePriceEstimationWithSlippage(params: SlippageParams): Result {
           getPriceParams['amountInUnits'] = amount
         }
 
-        const price = await dexPriceEstimatorApi.getPrice(getPriceParams)
+        const preSlippagePrice = await dexPriceEstimatorApi.getPrice(getPriceParams)
+
+        const price = checkSlippageAgainstPrice(priceSlippage, preSlippagePrice)
```
anxolin, line 109: I think we should undo the changes in this file, right?

If any piece needs to apply a slippage, he would get the estimation, and then apply the slippage.

W3stside: done

PR #1012, opened 2020-05-14
Trades view proposal

alfetopito:
## Update (Jun 1):

screenshot_2020-06-01_15-07-37

Couple improvements:

- Shuffled columns around
- Display data for deleted orders (NaN limit price on some trades)
- CSV download!
- (Ugly) order type pills
- Showing full date, prices and amounts on hover
- Relative date
- Updating trades only once per batch, after it is finalized
- Many minor refactors/improvements

## Update (May 27): 

screenshot_2020-05-27_12-34-58

Page is working, loading trade events with the minimal functionality needed.

Ready for testing with your own trades on https://pr1012--dexreact.review.gnosisdev.com/trades/
(remember to connect your wallet)

### To be decided based on feedback on this PR:

- [ ] Where to put this page?
- [ ] Are the current columns ok?
- [ ] Styling
- [ ] Call it Fills instead of Trades

### Follow ups to this PR

- [ ] Trade reverts:
  - Show reverted trades for recent batches
  - Do not show reverted trades for past batches (>5 batches ago?)
- [ ] [Good to have] Order detail page
- [ ] [Good to have] Link trade to order detail page
- [x] [Good to have] Download as csv

---

### :warning: Styling is non-existent. Just a very crude (and ugly) mock :warning: 

# Proposal for a Trades view

Added a very basic table with ideas for a trades view (page or tab, to be decided).

For now listing a few fields I think might be useful for users.

One idea as to where to put this is as an additional tab on the Trade page, such as:

screenshot_2020-05-14_14-49-25

(Keep in mind the image above is not in the PR)

The current mock is under `/trades` path.

screenshot_2020-05-14_14-59-25

## Notes
1. `Settled` column indicates whether the trade is final or can still be reverted (batch still open for submissions). Possibly display the countdown for how long until settlement.
2. I considered showing the batchId but this info is not displayed anywhere else in the UI, so doesn't make sense to have it here.
3. We could have for each trade a link to the transaction where it originated (although I don't see much value as a user other than transparency)
4. Still considering what to display - if should display anything at all - when a trade is reverted
5. In a regular exchange scenario, I would expect a link from the trade to an order, but here we don't have that. We could show the `orderId`, but like `batchId`, this is not displayed anywhere else. Might not be worth starting it now.
6. Alternatively, we could link each trade (and the orders as well) to an *Order detail page* where one could see all trades/reverts for a single order, and be able to cancel it too.

---

While we discuss this proposal, I'll fiddle around with blockchain events to populate this table.


gnosis-info:
Travis automatic deployment:
 https://pr1012--dexreact.review.gnosisdev.com

anxolin:
Amazing proposal.

> 1. Settled column indicates whether the trade is final or can still be reverted (batch still open for submissions). Possibly display the countdown for how long until settlement.

Very good suggestion!


> 2. I considered showing the batchId but this info is not displayed anywhere else in the UI, so doesn't make sense to have it here.
👌

> 3. We could have for each trade a link to the transaction where it originated (although I don't see much value as a user other than transparency)

I think we should do it. It offers transparency, and a user can start to understand how his trade "magically" happens because someone submits a solution

> 4. Still considering what to display - if should display anything at all - when a trade is reverted
Depending on the effort, but ideally, yes for a recent batch, because otherwise, the user would just see the trade disapear or change slightly his values.
However, I don't see the point for reverted trades of some batches ago.
(I would say, it's a good to have)

> 5. In a regular exchange scenario, I would expect a link from the trade to an order, but here we don't have that. We could show the orderId, but like batchId, this is not displayed anywhere else. Might not be worth starting it now.

Probably for now we can keep it simple and not show it.
A good to have is a link to the order --> it would take us to the orders tab and highlight the row. But I would say that is not that important for a first version if it makes things complicated.


6. Alternatively, we could link each trade (and the orders as well) to an Order detail page where one could see all trades/reverts for a single order, and be able to cancel it too.
This would be amazing. 

------- 


Not from this proposal, but it looks that filters by tokens, dates, filled/unfilled are really good aditions, both for ORDERS and for TRADES.

By the way, @Rafanator likes to call TRADES "Filled", probably is better.

Also, instead of Bough, I would use "Received" to be coherent with the trading widget.

Regarding the dates, I see UTC, I guess we are showing the local time of the user, right?

The layout, I would ask @biocom, but I would vote for get rid of the Liquidity orders here, and move it to the Liquidity section (there's a proposal already) and use the space for the "Filled" 

Lastly, although we don't show the order id and the batch id, but I think it's a good idea to at least add it hidden in the HTML5 data in the row (to easily check this thigns): `data-batch-id`,  `data-order-id`.

c3rnst:
Hi Leandro - looks smooth and clearer, thanks!!
Just one thing that came up a couple of times over the past week: In the your trades tab, could it also show the price the user wanted? ("Limit price") (as one of the value proposition is, that the final price might indeed be better than the limit price? :) 
Just another thought, which has NO priority at all but a though occurred to me: as this is the trade history for the user, possibly for accounting purposes, it might be useful to download it. Is there an easy way to make this a downloadable csv file?

Rafanator:
Very nice Leandro! 

Regarding what Anxo said, I think a more proper term to use here is "Fills" instead of "Trades". 

`Fills` is often used in most trading platforms, alternatively, you could use `Trade History` but it is too long, I'd go for `Fills`. `Trades` by itself is a more generic term that can lead to confusion. 

Using fills enables you to have a relevant `Type` column, where you can indicate `Partial` and `Filled/Full`.. 

Here is DY/DX as an example of the columns in their `Fills` tab:
Screenshot 2020-05-15 at 10 57 10

In the context of GP, I would suggest that our `Trades` (hopefully soon `Fills`) tab has the following columns and order:

Limit Price | Fill Price | Amount | Type | Time | Transaction 

Limit Price -> Limit Price of the original order (i.e. 150 WETH/DAI) 
Fill Price -> Price at which the order got filled (even if it didn't get filled in its entirety)
Amount -> Amount of the order that got filled, expressed in the sell token (If user wants to buy 1000 DAI worth of WETH at 150 and only 500 Dai of the order get filled, you would display Amount -> 500 Dai) 
Type -> `Partial` or `Full` 
Time -> Date and time of fill 
Transaction -> Transaction ID/url to etherscan

Observations: 

- I would discard the 'Bought' amount included by @alfetopito , as a user can check it in its balance and it is not common to show bought amount in most trading platforms. 

- I add 6 columns instead of 5 originally suggested by Leandro. If we are to disregard one because of space constraints, I would disregard `Time`. `Transaction` probably looks nicer if we just add a link to etherscan. Users would also find the time of their fill by checking that info. 

I hope this makes sense :)

alfetopito:
Lot's of great feedback, thank you all!

> Regarding the dates, I see UTC, I guess we are showing the local time of the user, right?

@anxolin sure, was just an example of showing the full date time as opposed to what we have else where which is "x minutes ago" etc.

> as this is the trade history for the user, possibly for accounting purposes, it might be useful to download it. Is there an easy way to make this a downloadable csv file?

@c3rnst we sure can do that. I'll split this into a separated task though, to not over extend the scope of this one

> Type -> Partial or Full

@Rafanator that information is not part of the event. We can of course get the data from the original order. May or may not have a performance impact. I'll try it out.

Also, in your suggested columns Rafa, you don't mention *Settled*. Would you rather not have it or are you indifferent?

## Summary:

- [ ] Show countdown for Settled while batch is still open for solutions
- [ ] Add link to transaction on etherscan
- [ ] [Good to have] Show reverted trades for recent batches
- [ ] [Good to have] Do not show reverted trades for past batches (>5 batches ago?)
- [ ] [Good to have] Order detail page
- [ ] [Good to have] Link trade to order detail page
- [ ] Show date on local time -- possibly hide date if low on space
- [ ] Add batchId, orderId etc as data props
- [ ] Show limit price
- [ ] [Good to have] Download as csv
- [ ] Call it *Fills* instead of *Trades*
- [ ] Columns now (not clear on *Settled* yet):

  Limit Price | Fill Price | Amount | Type | Time | Tx | Settled

anxolin:
Rafa knows more about trading than me, I just want to point out, that regarding this:

> I would discard the 'Bought' amount included by @alfetopito , as a user can check it in its balance and it is not common to show bought amount in most trading platforms.

This is what they show in Changly:
Screenshot at May 13 11-35-14


> , I would disregard Time. Transaction probably looks nicer if we just add a link to etherscan. Users would also find the time of their fill by checking that info

For me time is important, I don't understand why would we remove it. You shouldn't be having to click on all the fills and check the date in Etherescan for see if the last fills are from today or not.

This is for example Kraken, they show fills too. By the way, they call them trades 🤷‍♂️
image

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +114,156 @@ export interface Order {
...
+      `[ExchangeApiImpl] Fetched ${tradeEvents.length} trades for address ${userAddress} on network ${networkId}`,
+    )
+
+    return tradeEvents.filter(event => !event['removed']).map(this.parseTradeEvent)
+  }
+
+  public async subscribeToTradeEvent(params: SubscriptionParams): Promise<() => void> {
+    const { userAddress, networkId, callback } = params
+
+    const subscription = await this.getTradeSubscription(params)
```
Velenir, line 202: This one is not fully typed on the **typechain** side
It's actually an [extended EventEmitter](https://web3js.readthedocs.io/en/v1.2.7/web3-eth-subscribe.html#returns)
So you would unsubscribe with `subscription.unsubscribe()`

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +114,156 @@ export interface Order {
...
     super(injectedDependencies)
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ;(window as any).exchange = this._contractPrototype
   }
 
+  public async getPastTrades({ userAddress, networkId }: PastEventsParams): Promise<BaseTradeEvent[]> {
+    const contract = await this._getContract(networkId)
+
+    // to get all past events
+    const tradeEvents = await contract.getPastEvents('Trade', {
```
Velenir, line 187: Not typed by **typechain**, but we can and probably should type it ourselves in dex-js
I'll look into it

src/api/exchange/ExchangeApi.ts
alfetopito: Done!

src/api/deposit/DepositApi.ts
```diff
@@ -67,7 +67,7 @@ export class DepositApiImpl implements DepositApi {
   public constructor(injectedDependencies: Params) {
     Object.assign(this, injectedDependencies)
 
-    this._contractPrototype = new this.web3.eth.Contract(batchExchangeAbi) as BatchExchangeContract
+    this._contractPrototype = (new this.web3.eth.Contract(batchExchangeAbi) as unknown) as BatchExchangeContract
```
anxolin, line 70: Do you know why this worked and now we require this?

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +115,175 @@ export interface Order {
   remainingAmount: BN
 }
 
+/**
+ * BaseTradeEvent uses only info available on the emitted Trade event
+ */
+export interface BaseTradeEvent {
```
anxolin, line 121: Why do we need to redeclare the type?

In dex-telegram we have the typed event
https://github.com/gnosis/dex-telegram/blob/develop/src/services/DfusionService.ts#L166

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +115,175 @@ export interface Order {
...
 }
 
+// TODO: move to const/config
+const CONTRACT_DEPLOYMENT_BLOCK = {
+  1: 9340147,
+  4: 5844678,
+}
+
+// Syntactic sugar. Aliasing event types to make code cleaner
+type TradeSubscription = ReturnType<BatchExchangeContract['events']['Trade']>
```
anxolin, line 162: I like it 👌

Although I think in one PR I suggested something like this, and @Velenir commented that this could bring problems. I think it was with the flattening of the Promises, so it might be different case  🤷‍♂️

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +115,175 @@ export interface Order {
...
+    // Remove any active subscription
+    this.unsubscribeToTradeEvent()
+
+    logDebug(`[ExchangeApiImpl] subscribing to trade events for address ${userAddress} and networkId ${networkId}`)
+
+    const subscription = await this.getTradeSubscription(params)
+    // Two ways of subscribing
+
+    // 1: with error handling
+    subscription.subscribe((error, event) => {
```
anxolin, line 208: Are we using subscription? 
I thought one issue with that according to @Velenir  is that some wallets don't allow to subscribe.
If I recall from the sprint planning, The alternative was using a default provider, but that was a bit less scalable.

This is to keep up with the new trade events?
The idea we said it might be worth trying was, to use getPastLogs with a range of logs, and only do that if the bloom filter says so. 

For some reason, we trust more in `getPastLogs` than subscribe. There's testimonies of missed events?? Also, are we sure our strategy on reorgs?

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +115,175 @@ export interface Order {
...
+    //   Type '{ owner: string; orderId: string; sellToken: string; buyToken: string; executedSellAmount: string; executedBuyAmount: string; 0: string; 1: string; 2: string; 3: string; 4: string; 5: string; }' is missing the following properties from type 'ContractEventLog<{ owner: string; orderId: string; sellToken: string; buyToken: string; executedSellAmount: string; executedBuyAmount: string; 0: string; 1: string; 2: string; 3: string; 4: string; 5: string; }>': returnValues, event, address, logIndex, and 4 more.ts(2345)
+
+    return (): void => this.unsubscribeToTradeEvent()
+  }
+
+  public unsubscribeToTradeEvent(): void {
+    Object.keys(this.subscriptions.trade).forEach(networkId => {
+      logDebug(`[ExchangeApiImpl] Unsubscribing trade events for network ${networkId}`)
+      this.subscriptions.trade[networkId].unsubscribe()
```
anxolin, line 246: Is this fn idempotent?

src/hooks/useTrades.ts
```diff
@@ -0,0 +1,41 @@
+import { useState, useEffect, useCallback } from 'react'
+import { BaseTradeEvent } from 'api/exchange/ExchangeApi'
+import { useWalletConnection } from './useWalletConnection'
+import { exchangeApi } from 'api'
+
+// TODO: move trades to global state
+// TODO: (on global state) store trades to local storage
+// TODO: account for reverts
+// TODO: move trades/revert logic to a service
+// TODO: enrich trade event data with data from block and order
```
anxolin, line 10: ❤️ these todos, very clear implementation steps

src/api/exchange/ExchangeApi.ts
```diff
@@ -101,21 +115,175 @@ export interface Order {
...
 export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
+  private subscriptions: Subscriptions = { trade: {} }
+
   public constructor(injectedDependencies: Params) {
     super(injectedDependencies)
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ;(window as any).exchange = this._contractPrototype
   }
 
+  public async getPastTrades({ userAddress, networkId }: PastEventsParams): Promise<BaseTradeEvent[]> {
```
anxolin, line 180: Maybe this question could be asked in other APIs, but what happens if we pass a networkId that is not the one the web3 provider is connected to?

src/api/deposit/DepositApi.ts
Velenir: Previously our `ExtendedContract` could still be considered `web3.Contract` and could be cast to it
Because `web3.Contract.events is any`
But now typechain overrites not only something that was `any` but also `once` function + we override `getPastEvents`. So types become incompatible

Velenir:
@alfetopito take a look at #1038 with type fixes

src/api/exchange/ExchangeApi.ts
alfetopito: 2 main reasons:
1. I want to decouple the smart contract event type from the API return type
2. Initially I thought I could add more info like batchId, time, and so on, but that is not available here. In the end, the only custom filed I added is `id`, which is simply `txHash | txIndex`

Wdyt, should it stay or should it go?

src/api/exchange/ExchangeApi.ts
alfetopito: > Are we using subscription?
I thought one issue with that according to @Velenir is that some wallets don't allow to subscribe.
If I recall from the sprint planning, The alternative was using a default provider, but that was a bit less scalable.

Yes, we talked about that and to be honest I did not remember about it until now :)
Did not get to test it with any mobile wallet though.

> This is to keep up with the new trade events?

Yep.

> The idea we said it might be worth trying was, to use getPastLogs with a range of logs, and only do that if the bloom filter says so.

I thought the bloom filter was for getting balances.

> For some reason, we trust more in getPastLogs than subscribe. There's testimonies of missed events?? Also, are we sure our strategy on reorgs?

I can't testify in favor nor against any approach.

What I can tell is that, while testing, I'm somehow subscribing more than once to the events, or at least getting more than one notification per event. Tried to always unsubscribe before new subscriptions but that still did not help.

---

In summary, you are saying that we might just not have this method, and instead rely on `getPastTrades` with the addition of an optional `fromBlock` param. To know when to call though, we'd subscribe to bloom filters.
Is that it?

src/hooks/useTrades.ts
alfetopito: Will get there, eventually...

src/api/exchange/ExchangeApi.ts
alfetopito: Don't really know what happens if you call unsubscribe on something that is already unsubscribed... I guess nothing?

src/api/exchange/ExchangeApi.ts
alfetopito: Contract will not be found on the network and the call will fail.
Same for all APIs I believe.

src/api/exchange/ExchangeApi.ts
alfetopito: Oh, I'm also doing the amount conversions from string to BN.

anxolin:
leandro, I was taking a early look, and I would like to make a small suggestion. Actually this suggestion could apply to orders too. It's part of received feedback.

image

The red square above ☝️is the market. I think we should have a column with it, and then the prices can either completely omit it, or if it's not clear enough add the quote token (OWL) in this case. I'd probably try first just omitting to simplify, and get some feedback on that.

@Rafanator @c3rnst  what do you think?

Rafanator:
@anxolin I completely agree. It will clear a lot of repeated text and look better. 

Most platforms have a column for `Market`. Here are two examples:

- dy/dx (of course hehe):
 
Screenshot 2020-05-25 at 12 34 33

- Oanda (traditional forex/commodity/index broker):

Screenshot 2020-05-25 at 12 35 13

src/api/exchange/ExchangeApi.ts
anxolin: > What I can tell is that, while testing, I'm somehow subscribing more than once to the events, or at least getting more than one notification per event. Tried to always unsubscribe before new subscriptions but that still did not help.

that's strange, probably we do more than one subscription? maybe is the hook, that allows to reuse the logic, but if you for example do a subscription in a hook, and this hook is being used in 2 components, you end up with 2 subscriptions (of course, you can handle it with some logic)

> In summary, you are saying that we might just not have this method, and instead rely on getPastTrades with the addition of an optional fromBlock param. To know when to call though, we'd subscribe to bloom filters.
Is that it?

As a wise man once said, "I can't testify in favor nor against any approach.". But yes, I was suggesting that my gut was telling me less problems and more control if we use `getPastEvents` 😗

alfetopito: The last version is using only `getPastEvents`, still have to add the polling logic though.
It think for now (while I write the other parts) I'll do a simple interval.
Later we can improve with the bloom filter approach.

src/api/exchange/ExchangeApi.ts
anxolin: I like how we are declarative with the parameters. But still there's sth odd in that we have the parameters, but the "internal state" also plays a role.

If we pass `networkId = 1` and web3 points to rinkeby, it'll have a contract, just the `userAddress`, so we will just get other events, or no events if the user didn't trade.

Again. I think this is weird, but is not from this PR, and maybe is not something we want to worry too much right now. Just flagging it

alfetopito: Yeah, this is most likely the source for all of our "did you run out of gas?" exceptions we get when switching networks or loading the page.

alfetopito:
> @anxolin I completely agree. It will clear a lot of repeated text and look better.
> 
> Most platforms have a column for `Market`. Here are two examples:
> 
>     * dy/dx (of course hehe):
> 
> 
> Screenshot 2020-05-25 at 12 34 33
> 
>     * Oanda (traditional forex/commodity/index broker):
> 
> 
> Screenshot 2020-05-25 at 12 35 13

@anxolin @Rafanator Added a market column and removed reference to pair on both prices.
Kept the token for amount though.

Also, now the type column is populated.

The `batch Id | trade Id` column is there now just for help with my testing, don't intend to keep it there.
screenshot_2020-05-25_10-40-39

anxolin:
@alfetopito I'm curious about the `330` order in your screenshot. Why does it say filled and then you get a parcial? what's the logic in the type?

image

anxolin:
Also, do we have room somewhere for the receive amount.
dydx doesn't show it, I think.

Changelly, presents this info very clear: 
image

@Rafanator the screenshot you pasted from dydx I think it's from an open sell orders, that is not what Leandro is doing right now. I see the Fills are slightly different:
image

by the way! they can save space because they have a market selected!
That makes the UI simpler too --> less things
Although I guess it makes it more complicated if u trade in many many markets

alfetopito:
> @alfetopito I'm curious about the `330` order in your screenshot. Why does it say filled and then you get a parcial? what's the logic in the type?
> 
> image

I was using the wrong calculation for considering a trade filled.
Was smply using the `remainingAmount` from the order. Which is wrong, because each trade should calculate how much from the order it has filled.

Fixed it a few minutes ago.

Now I only have 1 order that was fully matched at once:
screenshot_2020-05-25_11-06-57

anxolin:
☝️this dydx screenshot, gives also an idea of an alternative way to show things.

For example if I sell `0.5 WETH` in `WETH-DAI` at `190`

```
 type       amount           Fill Price     Limit Price    received     Time           Tx
------   ----------      ------------   ----------    ---------   ------     ---------
SELL       0.5 WETH      190 DAI       180 DAI           95 DAI     2h ago     0x12..33
```

alfetopito:
> point_upthis dydx screenshot, gives also an idea of an alternative way to show things.
> 
> For example if I sell `0.5 WETH` in `WETH-DAI` at `190`
> 
> ```
>  type     amount      Fill Price     Limit Price     received     Time       Tx
> ------   ----------    ------------   ----------    ---------   ------     ---------
> SELL       0.5 WETH    190 DAI       180 DAI           95 DAI     2h ago     0x12..33
> ```

Experimenting with this idea (pushed on 1cdaf28f5d771f73637d51f7dce4496a7e4ed398):
screenshot_2020-05-25_14-27-33

bh2smith:
- I really like the `Your Orders | Your Trades` sub-tabs! I was thinking to suggest "Trade History" until I saw what you proposed

> Are the current columns ok? 
They are alright except when I see `Nan`on the limit price.  

Screenshot 2020-05-28 at 9 02 01 AM


> Call it Fills instead of Trades 
 Meh, I'm on the fence about this one.


> Show reverted trades for recent batches
I personally do not care about this at all. In fact, this is irrelevant information its like showing someone what ["All that could have been"](https://www.youtube.com/watch?v=mMtYyLJ1dCw) A great track by NIN btw. No sense in dwelling on what isn't.

> [Good to have] Download as CSV

This is a **must have** from my perspective.

anxolin:
That's a greate progress!

Some minor feedback:

1. It looks like there's some rounding issues
image

@W3stside, for the smart format, I thought we were truncating (floor) the number, right? It looks like we are rounding. I would think is better to be in the safe side and truncate but not sure if there's an "official standard"

2. The titles are missing
It would be nice if I hover to see the full amount, without tuncating any info

3. The market is the other way around
You show above USDC/OWL, but is actually OWL/USDC

4. I'm having second thought about the `settled` column
It's only relevant while the trade can still be reverted but has too much protagonism in the table.
Maybe we can say it in the Type by adding an icon there "Partial ⏱" and maybe highlight the row with different color showing that is not settled... We can think about this

5. The price you are showing is the inverse price. In this example the price was `0.502136008 USDC`

6. Should Date be the first column? Also I would use "relative dates with tooltip" like:
image

7. ubernit: It would be nice to style the type column values differently. Something similar to how Etherscan shows "In" "Out" transactions --> like a tag/pill

8. At some point I think we need to deal with pagination. So it don't become sluggish on some devices.
This can be handled on it's own issue, cause it's not so trivial and the patter here can be also used for the orders. 

One simple way is to use a "infinity scroll" to load pages.
This way we would show only the last page to the user AKA "most recent trades" and only if he scrolls we show more.

One possible issue with the above is that if there's too many elements it can make the browser to have too many markup, although we could try it out. And there's ways around it.
Alternatively, if we have issues with the above or we don't like the idea, a normal pagination as in Etherescan would suffice. 

----

I ❤️ you did the html5 data fields here:
image

Velenir:
> One simple way is to use a "infinity scroll" to load pages.

I'd like to avoid this. _infinity scroll_ is a variant of scroll hijacking and is arguably bad UX

> One possible issue with the above is that if there's too many elements it can make the browser to have too many markup

This can get unwieldy really fast with rerenders of long list. We'd have to implement virtual lists

anxolin:
> I'd like to avoid this. infinity scroll is a variant of scroll hijacking and is arguably bad UX

I don't have a strong opinion, and I think normal pagination would suffice, but "infinity scroll" is used in apps like instagram or facebook with lot's of happy users behind. Why you say "bad UX"?

> This can get unwieldy really fast with rerenders of long list. We'd have to implement virtual lists

I think there's ways around it, but if it's a concern, I'm happy to do page-by-page pagination if it simplifies things.

Velenir:
> Why you say "bad UX"?

It lies about expectation of what is and will be displayed, implicitly loads more. An alternative could be a `Load More` button, very explicit

src/services/factories/getTrades.ts
```diff
@@ -0,0 +1,109 @@
...
+    })
+
+    // TODO: maybe combine both promise lists into 1?
+    const blockTimes = new Map<number, number>(await Promise.all(Array.from(blocksSet).map(getBlockTimePair)))
+    // TODO: we probably already have this info in global state and don't need to fetch from the contract again?
+    const orders = new Map<string, Order>(
+      await Promise.all(Array.from(orderIdsSet).map(orderId => getOrderPair(userAddress, networkId, orderId))),
+    )
+    // TODO: list might not be up to date. Handle case where possibly tokens are not found
+    // In that case, what to do? retry? get token from contract directly? use the `add to local list` functionality?
```
Velenir, line 66: Yes, good idea. Add to local USER_LIST what is missing

src/services/index.ts
```diff
@@ -39,6 +40,8 @@ export const subscribeToTokenList = subscribeToTokenListFactory(apis)
 
 export const getTokens = getTokensFactory(apis, { getTokenFromErc20 })
 
+export const getTrades = getTradesFactory({ ...apis, getTokens })
```
Velenir, line 43: Why not use signature like in other services `(apis, { getTokens })`

src/hooks/useTrades.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  useEffect(() => {
+    // Check only up to `latest - CONFIRMATION_BLOCKS` to reduce chance of re-orgs
+    const toBlock = blockNumber && blockNumber - CONFIRMATION_BLOCKS
+
+    if (userAddress && networkId && toBlock && (!lastCheckedBlock || lastCheckedBlock < toBlock)) {
+      getTrades({
+        userAddress,
+        networkId,
+        fromBlock: lastCheckedBlock,
```
Velenir, line 32: `fromBlock and toBlock` are inclusive boundaries
Try
```js
await exchangeApi.getPastTrades({networkId:1, fromBlock: 10154000, toBlock: 10154154})
await exchangeApi.getPastTrades({networkId:1, fromBlock: 10154154})
```
you'll get dulicate events in the second result

Have to do `fromBlock: lastCheckedBlock+1`

src/api/exchange/ExchangeApi.ts
```diff
@@ -97,21 +118,183 @@ export interface Order {
...
     ;(window as any).exchange = this._contractPrototype
   }
 
+  // TODO: Not very happy with this method. Can't be used inside the class because batchId is only known with block data
+  // TODO: Don't really know where to put it
+  public static buildTradeRevertKey(batchId: number, orderId: string): string {
+    return batchId + '|' + orderId
+  }
+
+  public async getPastTrades({
```
Velenir, line 189: You will eventually run into 
```
inpage.js:1 MetaMask - RPC Error: query returned more than 10000 results {code: -32005, message: "query returned more than 10000 results"}
```

Can get that even now with

```js
await exchangeApi.getPastTrades({networkId:1})
```

And as soon as any `userAddress` accumulates more than 10000 events, it'll break for that address

src/hooks/useTrades.ts
Velenir: This is the only comment that really needs to be addressed before merge

src/hooks/useTrades.ts
```diff
@@ -0,0 +1,40 @@
...
+        toBlock,
+      }).then(newTrades => dispatch(appendTrades(newTrades, toBlock)))
```
Velenir, line 34: This would dispatch even if `newTrades` is empty, causing reducer run and rerender. Better to opt out if no new trades

Velenir: Or don't change state inside reducer when incoming trades is empty

src/api/exchange/ExchangeApi.ts
```diff
@@ -97,21 +118,183 @@ export interface Order {
...
+        )
+      })
+
+    return (): void => this.unsubscribeToTradeEvent()
+  }
+
+  public unsubscribeToTradeEvent(): void {
+    Object.keys(this.subscriptions.trade).forEach(networkId => {
+      logDebug(`[ExchangeApiImpl] Unsubscribing trade events for network ${networkId}`)
+      this.subscriptions.trade[networkId].unsubscribe()
```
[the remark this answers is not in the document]
alfetopito: Not actually used. In anyway, did not get the joke. Can you draw?

src/hooks/useTrades.ts
Velenir: Since we still need to update `toBlock` anyway. So either a new action for `lastCheckedBlock` update only. Or check ifEmpty in reducer.
I prefer new action, like we already do for 'UPDATE_OFFSET` in orders

src/components/TradesWidget/TradeRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (trade.settlingDate > now) {
+      const timeout = trade.settlingDate.getTime() - now.getTime()
+      refreshTimeout.current = setTimeout(() => forceUpdate({}), timeout)
+    }
+
+    return clear
+  }, [forceUpdate, now, trade.settlingDate])
+
+  return (
+    <tr key={trade.id} data-order-id={trade.orderId} data-batch-id={trade.batchId}>
```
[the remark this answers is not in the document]
alfetopito: Left over from when this was only one component in the trades widget. Will remove.

src/services/factories/tokenList.ts
```diff
@@ -18,7 +18,7 @@ export function getTokensFactory(
     tcrApi?: TcrApi
   },
   injects: Injects,
-): (tokenId: number) => TokenDetails[] {
+): (networkId: number) => TokenDetails[] {
```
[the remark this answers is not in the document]
alfetopito: As long as it was there I guess. No harm though :)

src/services/factories/getTrades.ts
```diff
@@ -0,0 +1,109 @@
...
+    tradeEvents.forEach(event => {
+      blocksSet.add(event.blockNumber)
+      orderIdsSet.add(event.orderId)
+      tokenIdsSet.add(event.buyTokenId)
+      tokenIdsSet.add(event.sellTokenId)
+    })
+
+    // TODO: maybe combine both promise lists into 1?
+    const blockTimes = new Map<number, number>(await Promise.all(Array.from(blocksSet).map(getBlockTimePair)))
+    // TODO: we probably already have this info in global state and don't need to fetch from the contract again?
```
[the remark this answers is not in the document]
alfetopito: But, how messy can it get if I consume the global state inside the service?

src/services/index.ts
alfetopito: @anxolin commented in other PR, why use a different signature for this injection?

So I thought about adding everything into the same param. Refactor the others next.

Arguments in favor or against?

src/hooks/useTrades.ts
alfetopito: I hoped it would be smart enough.
Alright then, I'll update only the `lastCheckedBlock` when no new trades.

src/api/exchange/ExchangeApi.ts
alfetopito: Scheisse. 

I don't see a pagination API for this on web3. 

The best approach then is to fetch events in batches of arbitrary size and hope there are no more than 1k events per batch?

src/services/factories/getTrades.ts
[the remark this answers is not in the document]
Velenir: Very messy. Rather than consume can accept as input param, defaulting to empty array

src/services/index.ts
Velenir: Only argument -- consistency

src/api/exchange/ExchangeApi.ts
Velenir: you can do what we did somewhere before, @anxolin remembers
```js
const fetch = (...) => try {
 // fetch everything
} catch(error is our error) {
 fetch(from block to middle of prev range).concat(
   fetch(from middle of prev range)
  )
}
```
Binary fetch, like binary search

src/api/exchange/ExchangeApi.ts
Velenir: Universe entropy increases

This is where it was supposed to go. Commented through VSCode, didn't work out :disappointed:

src/services/factories/getTrades.ts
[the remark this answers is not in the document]
alfetopito: Great idea, take it as input param!!!

alfetopito:
> > Are the current columns ok?
> 
> They are alright except when I see `Nan`on the limit price. Could it be that these correspond to orders I "deleted" after they had been filled.
> 
> Screenshot 2020-05-28 at 9 02 01 AM

Oh boy, if that's case, we won't be able to tell what was the limit price.
Could you share the tx and your address for me to take a look?

> > [Good to have] Download as CSV
> 
> This is a **must have** from my perspective.

[CSV download is coming sooner than I expected](https://github.com/gnosis/dex-react/pull/1071).
You can thank @Velenir.

alfetopito:
@anxolin 
>     2. The market is the other way around
>        You show above USDC/OWL, but is actually OWL/USDC

Fun with markets... :scream: 

---

>     3. I'm having second thought about the `settled` column
>        It's only relevant while the trade can still be reverted but has too much protagonism in the table.
>        Maybe we can say it in the Type by adding an icon there "Partial stopwatch" and maybe highlight the row with different color showing that is not settled... We can think about this

Agree, will get rid of the settled column. Should we still have the countdown?

On the other hand, I'm thinking of a slightly different approach to avoid having `unsettled` trades altogether.

What about not displaying trades that are not yet settled?

*Pros*:
1. No need for extra UI to indicate trade is "pending".
2. If's displayed it's final. 
3. No chance of a trade "disappear" after being displayed. Think about when we add trade notifications: "You got a new trade! (later) Nah, just kidding, no trades for you"

*Cons*:
1. Trades not notified in real time.

---

>     4. The price you are showing is the inverse price. In this example the price was `0.502136008 USDC`

The more we talk about the prices/markets/etc the more confused I get.

---

>     6. ubernit: It would be nice to style the type column values differently. Something similar to how Etherscan shows "In" "Out" transactions --> like a tag/pill

I'll give it a try. Hopefully Michel will come to the rescue when everyone looks away in terror from the monstrosity that will result.

---

>     7. At some point I think we need to deal with pagination. So it don't become sluggish on some devices.
>        This can be handled on it's own issue, cause it's not so trivial and the patter here can be also used for the orders.

Agree and agree. To be dealt with on https://github.com/gnosis/dex-react/issues/1072

---

> I heart you did the html5 data fields here:
> image

Just need to find a use for it...

src/components/TradesWidget/index.tsx
```diff
@@ -1,20 +1,119 @@
...
-  const { networkId } = useWalletConnection()
+  const { networkId, userAddress } = useWalletConnection()
   const trades = useTrades()
 
+  const generateCsv = useCallback(
+    () =>
+      toCsv({
+        headers: csvHeaders,
+        data: trades.filter(isTradeSettled),
+        transformer: csvTransformer,
```
Velenir, line 99: Really like how generic this func is :+1:

src/components/TradesWidget/index.tsx
```diff
@@ -1,20 +1,119 @@
...
+      }),
+    [trades],
+  )
+
+  const filename = useMemo(
+    () => `trades_${getNetworkFromId(networkId as number).toLowerCase()}_${userAddress}_${new Date().getTime()}.csv`,
```
Velenir, line 105: Does

```suggestion
    () => `trades_${getNetworkFromId(networkId!).toLowerCase()}_${userAddress}_${new Date().getTime()}.csv`,
```

work? It's less of a hack than `as` type cast

src/hooks/useTrades.ts
```diff
@@ -19,21 +25,69 @@ export function useTrades(): Trade[] {
...
+      }
+    }
+
+    function clear(): void {
+      if (intervalId.current) clearInterval(intervalId.current)
+      intervalId.current = null
     }
-  }, [userAddress, networkId, blockNumber, lastCheckedBlock, dispatch])
+
+    clear()
```
Velenir, line 59: There's no need to call `clear` here, since interval is always cleared in clean-up phase of `useEffect()`
At this point there's no valid intervalId to clear

src/hooks/useTrades.ts
```diff
@@ -19,21 +25,69 @@ export function useTrades(): Trade[] {
...
+
+      // Update now to not leave the interface empty
+      if (trades.length === 0 || (delay >= 0 && delay < 10)) updateTrades()
+
+      if (delay >= 0 && delay < 10) {
+        // If time left in batch within this window, start the interval
+        intervalId.current = setInterval(updateTrades, BATCH_TIME_IN_MS)
+      } else {
+        // Otherwise, set a timeout to, update, then set interval
+        intervalId.current = setTimeout(() => {
```
Velenir, line 82: `intervalId` can live inside `useEffect`, since it's only used inside and cleared between calls

src/reducers-actions/trades.ts
```diff
@@ -1,20 +1,26 @@
...
-export type ActionTypes = 'APPEND_TRADES'
+export type ActionTypes = 'APPEND_TRADES' | 'UPDATE_BLOCK'
 
 export interface TradesState {
   trades: Trade[]
   lastCheckedBlock?: number
 }
 
+type UpdateBlockActionType = Actions<ActionTypes, Required<Pick<TradesState, 'lastCheckedBlock'>>>
 type ReducerActionType = Actions<ActionTypes, TradesState>
```
Velenir: Need more specificity
```suggestion
type UpdateBlockActionType = Actions<'UPDATE_BLOCK', Required<Pick<TradesState, 'lastCheckedBlock'>>>
type ReducerActionType = Actions<'APPEND_TRADES', TradesState>
```

like in` src/reducers-actions/localTokens.ts`
Otherwise in `switch..case` in reducer TS won't be able to differentiate between actions, and see action as `UpdateBlockActionType | ReducerActionType`

Also rename `ReducerActionType` to `UpdateTradeStateActionType` or something

src/utils/csv.ts
```diff
@@ -0,0 +1,57 @@
+import { logDebug } from 'utils/miscellaneous'
+
+// TODO: can I make sure - using types - that headers length matches transformer return type length?
```
Velenir, line 3: Can't typecheck length unless we do tuples, and we can't do tuples
A compromise could be adding `assert` for a runtime check inside `toCsv`

src/utils/miscellaneous.ts
```diff
@@ -88,6 +88,10 @@ export function isTradeFilled(trade: Trade): boolean {
   return isAmountDifferenceGreaterThanNegligibleAmount(trade.sellAmount, trade.remainingAmount)
 }
 
+export function isTradeSettled(trade: Trade): boolean {
+  return trade.settlingDate <= new Date()
```
Velenir, line 92: not that it matters, but why did you choose to deal with Date as object rather than timestamp as number.
Like here `trade.settlingDate <= Date.now()` and in Trade type
It ultimately doesn't matter, but ? remains

src/utils/csv.ts
```diff
@@ -0,0 +1,57 @@
...
+        try {
+          values = transformer(item)
+
+          // Make sure the returned values match.
+          if (headers && values.length !== headers.length) {
+            throw new Error(`Values length (${values.length}) doesn't match headers length ${headers.length}`)
+          }
+        } catch (e) {
+          logDebug(`[utils:toCsv] Not able to transform into csv: ${item}`, e)
+          return acc
```
Velenir, line 34: Since you are throwing and catching immediately + fallback, no need to even throw
Do `console.error` or `console.warn`
Or even better `console.assert`, we never get to use `console.assert`

src/utils/csv.ts
```diff
@@ -0,0 +1,57 @@
+import { logDebug } from 'utils/miscellaneous'
+
+// TODO: can I make sure - using types - that headers length matches transformer return type length?
+// Alternatively, I could make headers be the keys of a second type parameter
+// which is the result of transformer, such as transformer<T, R>(item: T) => R[]
```
Velenir, line 5: Don't see the use for it. Headers can be completely arbitrary

Elaborate how you would do it

bh2smith:
@alfetopito: Based on my `NAN`comment from above. I have confirmed that this does indeed correspond to a deleted order:

My Address: `0x7f01d9b227593e033bf8d6fc86e634d27aa85568`
Order Id: `114`

Solution TX: https://etherscan.io/tx/0xaa9ec59055e53b137930e75812046dd16dc03bcd03dfafe1abb53fc8fa9ca707


Screenshot:

Screenshot 2020-05-29 at 11 18 36 AM

NAN is probably ok in this scenario, but we could probably uniquely identify those orders which have been deleted.

anxolin:
@alfetopito 

> What about not displaying trades that are not yet settled?

It looks you have more PROs and CONS in your list. Fine with me to go with settled if it make things easier and less flickery. 

> The more we talk about the prices/markets/etc the more confused I get.

It's an ups and downs way, you think you understand, you are up, then you go down again

> I'll give it a try. Hopefully Michel will come to the rescue when everyone looks away in terror from the monstrosity that will result.

No shame in "getting inspiration in etherscan" ;) 
image

src/api/exchange/ExchangeApi.ts
alfetopito: 70e7cdb82b8c342b369d818c35ab848d96c7d495

src/components/TradesWidget/index.tsx
alfetopito: screenshot_2020-05-29_14-17-49

Either a cast or TS eslint warning

src/utils/csv.ts
alfetopito: The idea is that the transformer returns an object instead of a list.
Better for readability and now that I think of it, we might as well get rid of the headers column.

I just don't know about the order in the object.

Let's say I want to convert to csv the following dumb headers, data and transformer:
- headers: ['times1', 'times2', 'times3']
- data: [1, 2, 3]

```ts
transformer<T>(item:T):string[] {
  return [
    item * 1,
    item * 2,
    item * 3
  ]
}
```

Resulting csv should be:
```
times1,times2,times3
1,2,3
2,4,6
3,6,9
```

That's how it's right now. With the amount of columns we have, it's easy to get confused and pass the wrong order/incorrect #of values.

What I'm thinking about is like this, for same data, but without headers and with transformer:

```ts
transformer<T, R>(item:T):R {
  return {
    times1: item*1,
    times2: item*2,
    times3: item*3
}
```

We can extract the headers from the transformer return type `R`.

Now, is the key order guaranteed?

I'm out of time today, might try it out on Monday.

src/utils/miscellaneous.ts
alfetopito: Didn't really think about it. Will revise for consistency.

src/utils/csv.ts
alfetopito: Wrote down my idea to address that here https://github.com/gnosis/dex-react/pull/1012#discussion_r432756355

src/utils/csv.ts
alfetopito: Didn't know about `console.assert`. Cool!

Changed (although it might not be required if I get this working https://github.com/gnosis/dex-react/pull/1012#discussion_r432756355)

src/utils/csv.ts
alfetopito: The requirement for `R` is that keys and values are strings. There's a way to express that in TS for sure but not right now.

alfetopito:
> @alfetopito: Based on my `NAN`comment from above. I have confirmed that this does indeed correspond to a deleted order:
> 
> My Address: `0x7f01d9b227593e033bf8d6fc86e634d27aa85568`
> Order Id: `114`
> 
> Solution TX: https://etherscan.io/tx/0xaa9ec59055e53b137930e75812046dd16dc03bcd03dfafe1abb53fc8fa9ca707
> ### Screenshots:
> 
> My cancelled order:
> Screenshot 2020-05-29 at 11 18 36 AM
> 
> How it appears in trade page:
> 
> Screenshot 2020-05-29 at 11 22 47 AM
> 
> NAN is probably ok in this scenario, but we could probably uniquely identify those orders which have been deleted.

Thanks for checking it out, Ben.
I did not get time to fix this today, but I'm thinking of how to address it.

NaN is definitely not good. I could for now just say N/A, but that is not optimal.
Rather than leave it empty, I think we can fetch this data, from the Order placement event.
Need to investigate further how to query that, shouldn't be too hard.

alfetopito:
@anxolin 

> @alfetopito
> 
> > What about not displaying trades that are not yet settled?
> 
> It looks you have more PROs and CONS in your list. Fine with me to go with settled if it make things easier and less flickery.

It's beautiful! After all, this is a BATCH exchange, where trades happen in BATCHES! hehe
Checking once is enough.

> > The more we talk about the prices/markets/etc the more confused I get.
> 
> It's an ups and downs walk, you think you understand, you are up, then you go down again

But really, the prices look fine to me. Same as what I see when I place an order.
Why should they be inverted here?

> No shame in "getting inspiration in etherscan" ;)
> image

[Can't say that I haven't triiiiiiiiiiiieeed](https://youtu.be/YQHsXMglC9A?t=173)

screenshot_2020-05-29_15-21-20

src/services/factories/getTrades.ts
```diff
@@ -56,14 +62,28 @@ export function getTradesFactory(factoryParams: {
...
+    // Fetch block info
+    const blockTimes = new Map<number, number>(await Promise.all(Array.from(blocksSet).map(getBlockTimePair)))
+
+    // Add whatever token might be missing. Will not try to add tokens already in the list.
+    await addUnlistedTokensToUserTokenListById(Array.from(tokenIdsSet))
+
+    // Get the tokens we need
     const tokens = new Map(
       getTokens(networkId)
         .filter(token => tokenIdsSet.has(token.id))
```
Velenir, line 89: Why even filter if you just cherry-pick from map later (`tokens.get(event.buyTokenId)`).
Just feed everything into map. Can even do `.forEach` if you want for <sub><sup>micro</sup></sub>optimization

alfetopito: Good point :+1:

src/components/FileDownloaderLink.tsx
```diff
@@ -26,22 +26,22 @@ export const FileDownloaderLink: React.FC<FileSaverLinkProps & AnchorHTMLAttribu
   // only create download onClick because
   // creating and retaining BLob and ObjectURL can be expensive on each render
   const handleClick = (e: React.MouseEvent<HTMLAnchorElement>): void => {
+    const data = _data instanceof Function ? _data() : _data
```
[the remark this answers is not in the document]
alfetopito: Hmmm, didn't know `typeof` worked for functions. Sure, I can change that.

src/services/factories/getTrades.ts
```diff
@@ -56,14 +62,28 @@ export function getTradesFactory(factoryParams: {
...
+    // Filter orders that we might already have
     const orders = new Map<string, Order>(
-      await Promise.all(Array.from(orderIdsSet).map(orderId => getOrderPair(userAddress, networkId, orderId))),
+      existingOrders
+        .filter(order => orderIdsSet.has(order.id))
+        .map(order => {
+          orderIdsSet.delete(order.id)
+          return [order.id, order]
+        }),
     )
```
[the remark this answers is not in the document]
alfetopito: I love optimizations! Yes, do like it, thanks!

src/utils/csv.ts
alfetopito: :feelsgood: c9b2e38e3ce69343c43a31cba8ac4f5091cec592

alfetopito:
> Thanks for checking it out, Ben.
> I did not get time to fix this today, but I'm thinking of how to address it.
> 
> NaN is definitely not good. I could for now just say N/A, but that is not optimal.
> Rather than leave it empty, I think we can fetch this data, from the Order placement event.
> Need to investigate further how to query that, shouldn't be too hard.

@bh2smith the latest version should not show you any more NaNs.
Fetching event data (for deleted orders) from [OrderPlacement events](https://github.com/gnosis/dex-contracts/blob/master/contracts/BatchExchange.sol#L97).

src/services/factories/getTrades.ts
```diff
@@ -52,44 +49,73 @@ export function getTradesFactory(factoryParams: {
...
+          try {
+            order = await exchangeApi.getOrderFromOrderPlacementEvent({
+              userAddress,
+              networkId,
+              orderId,
+              // Parameters bellow not required, but help narrow down the search
+              buyTokenId,
+              sellTokenId,
+              toBlock: blockNumber,
+            })
```
Velenir, line 97: There may be a better way
we can make calls to check contract state at a specific block
You have `blockNumber` from `TradeEvent` and `Order` corresponding to that event must be active at that point, of course. So you can do
```ts
contract.methods.getEncodedUserOrders(userAddress).call({}, blockNumber)
// then you'll have to decode orders as usual and find the one you need
```
But maybe even better since we have paginated orders and know `orderId`
```ts
exchangeApi.getOrdersPaginated({
 userAddress.
 networkId,
 offset: orderId,
 pageSize:1,
}, blockNumber) // fetches that exact order as it was at that block
```

I wonder if to have fewer fetches we could, if we have multiple deleted orders close by in id, fetch them all together with appropriate `pageSize`, but probably not, cause will be hard to guess at what `blockNumber` they all would be alive.

This is much cheaper (probably) than getting a bunch of unindexed events.

`blockNumber` isn't typed though, [it's not even in the docs](https://github.com/ethereum/web3.js/pull/3558)
So try with `//@ts-ignore` first and if it works for you, once again I'll need to hack around types in **dex-js**, ugh...

src/services/factories/getTrades.ts
```diff
@@ -52,44 +49,73 @@ export function getTradesFactory(factoryParams: {
...
+    // Add all existing orders to map. No problem to have more than what we need since it'll be picked by id later.
+    existingOrders.forEach(order => orders.set(order.id, order))
+
+    // Fetch from contract the ones we don't have locally, and add it to the map
+    await Promise.all(
+      Array.from(orderIdsMap.keys()).map(async orderId => {
+        // We already have this order, ignore
+        if (orders.has(orderId)) {
+          return
+        }
```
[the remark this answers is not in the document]
alfetopito: We don't:
- https://github.com/gnosis/dex-react/blob/trades-view/src/reducers-actions/orders.ts#L50
- https://github.com/gnosis/dex-react/blob/trades-view/src/reducers-actions/orders.ts#L78
- https://github.com/gnosis/dex-react/blob/trades-view/src/reducers-actions/orders.ts#L101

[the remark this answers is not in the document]
Velenir: Great! I thought maybe we do.

src/utils/csv.ts
```diff
@@ -1,58 +1,72 @@
 import { logDebug } from 'utils/miscellaneous'
 
-// TODO: can I make sure - using types - that headers length matches transformer return type length?
-// Alternatively, I could make headers be the keys of a second type parameter
-// which is the result of transformer, such as transformer<T, R>(item: T) => R[]
-interface ToCsvParams<T> {
-  headers?: string[]
+export interface CsvColumns {
+  [key: string]: string
+}
```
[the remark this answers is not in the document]
alfetopito: Fancy!

src/components/TradesWidget/index.tsx
```diff
@@ -111,22 +97,37 @@ const Trades: React.FC = () => {
...
-            <th>Time</th>
-            <th>Tx</th>
+            <th>
+              <div
+                style={{
+                  display: 'flex',
+                  justifyContent: 'space-around',
+                  alignItems: 'center',
+                }}
+              >
```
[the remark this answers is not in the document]
alfetopito: No, this is bad and shouldn't be here. Styling is a big TODO that needs fixing :D

(but ok, I might add that just for the sake of inline style being bad)

src/services/factories/getTrades.ts
alfetopito: Well, if that's the case, I could use the `getOrder` directly with the `blockNumber`.
Then orders will never be empty and I can skip this additional step entirely.

I'll give this a try :+1:

alfetopito: Yes, that does work!

No need to check for this event then, I'll remove the method `getOrderFromOrderPlacementEvent` and this extra hop.

And, as expected, the build does fail:

```
ERROR in /home/alfeto/work/dfusion/dex-react/src/api/exchange/ExchangeApi.ts
ERROR in /home/alfeto/work/dfusion/dex-react/src/api/exchange/ExchangeApi.ts(347,83):
TS2554: Expected 0-1 arguments, but got 2.
```
screenshot_2020-06-02_09-25-08

alfetopito: Several layers of ignore later :D
1cb96712bb95ae8063aeaeff7876978418f579f1

I'm wondering if I should still leave the check for empty orders there as an additional check.

But maybe not. There shouldn't be a case that a Trade event is emitted while an order does not exist on the contract.

src/services/index.ts
anxolin: yes, for me it's a strange to add the dependencies like `getTradesFactory({ ...apis, getTokens })` instead of `getTradesFactory({ ..anyIocDepencency })`

But if it's how we did things, I'm ok in pro of consistency

src/api/exchange/ExchangeApi.ts
anxolin: > You will eventually run into...

Hehe, @Velenir you have good old memories from dxDAO times :)

src/api/exchange/ExchangeApi.ts
```diff
@@ -296,6 +426,76 @@ export class ExchangeApiImpl extends DepositApiImpl implements ExchangeApi {
...
+  private async safeGetEvents<T>(
+    fn: (options: PastEventOptions) => Promise<T[]>,
+    options: PastEventOptions,
+  ): Promise<T[]> {
+    try {
+      return await fn(options)
+    } catch (e) {
+      // Error `-32005` means too many results in range.
+      // Let's split it up into 2 smaller requests
+      if (e.code === -32005) {
```
anxolin, line 441: Cool, should we move it to a const and maybe add a link to this https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md

anxolin:
Not for this PR, but once it's merge we can style it further.

I did some clumsy test, and I think it looks better with a bit more of air/padding for the rows:
image


Also I tried to play a bit with the pills. 
In my test, I added: 
```css
    display: inline-block;
    font-size: 0.9rem;
    padding: 0.2rem 0.8rem;
    min-width: 5em;
    font-weight: 800;
}
```

alfetopito:
As Anxo suggested, merged branch as it currently is. Too big to review, will address comments in new PRs.

PR #1013, opened 2020-05-15
gasPrice and tx.data middlewares

Velenir:
Moves gasPrice and earmarking logic into provider middlewares
For earmarking uses existing encoder, but now we can use not only numbers but hex values. Which presents a good opportunity to change `SENTINEL` to something interesting
For example, `feed` or `dad` or even `feeddad`

Works for me, bu I would appreciate some extra testing for all transactions we have in UI :pleading_face:


gnosis-info:
Travis automatic deployment:
 https://pr1013--dexreact.review.gnosisdev.com

Velenir:
@alfetopito does it freeze on develop too?

Velenir:
> * Trade
> * Liquidity
> * Enable
> * Deposit
> * Withdraw
> * Claim
> * Cancel
> * Delete

Tested with Gnosis Safe. Works!

Velenir:
> on MM the page freezes

Couldn't reproduce no matter what I tried
@anxolin , @W3stside can you try?

alfetopito:
> @alfetopito does it freeze on develop too?

Yes it does...

At least on FF 
screenshot_2020-05-19_10-04-10

But only when switching from Rinkeby to Mainnet, not the other way around ... :/

Had no issues whatsoever on Chrome... :(

PR #1025, opened 2020-05-19
Refactor spinners into a new component

### LINKED ISSUE #844: Refactor spinners
A component spinner sounds like a good idea

Out of Dimas comment here: https://github.com/gnosis/dex-react/pull/840#discussion_r405359525

anxolin:
Refactor spinners into it's own component

Closes #844

### DIFF
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -1,0 +1,16 @@
+import React from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faSpinner } from '@fortawesome/free-solid-svg-icons'
+import { SizeProp } from '@fortawesome/fontawesome-svg-core'
+
+interface SpinnerProps {
+  spin?: boolean
+  style?: React.CSSProperties
+  size?: SizeProp
+}
+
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1025--dexreact.review.gnosisdev.com

src/components/Spinner.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { SizeProp } from '@fortawesome/fontawesome-svg-core'
+
+interface SpinnerProps {
+  spin?: boolean
+  style?: React.CSSProperties
+  size?: SizeProp
+}
+
+export const Spinner: React.FC<SpinnerProps> = ({ style, spin = true, size }) => (
+  <FontAwesomeIcon icon={faSpinner} style={style} spin={spin} size={size} />
```
W3stside, line 13: i would argue to make this more flexible is to allow icon to be a prop and set a default

anxolin: I just wanted to do the refactor. I think is better not to get ahead of ourselves. If you have an use case in mind, like have a  "🤘" spinner good, otherwise, I would think is better to add it when we need it

PR #1026, opened 2020-05-19
Proxy composed provider

Velenir:
I couldn't connect some mobile providers(including TrustWallet)

My guess is that the minimal Provider interface that `eth-json-rpc-middleware/providerFromEngine` implements, which is only `send` and `sendAsync` functions, is not enough. Even if we reemit events

Maybe adding `enable` function would suffice. Maybe some more. I just wrapped `composedProvider` in a `Proxy`


gnosis-info:
Travis automatic deployment:
 https://pr1026--dexreact.review.gnosisdev.com

W3stside:
i have no idea what's going in this PR but there are some console logs and comments that should be removed

W3stside:
middleware seems to be the devil

PR #1027, opened 2020-05-20
Autoadd tokens from Orders

Velenir:
When an unlisted token appears in orders it's now added to USER_TOKEN_LIST automatically


gnosis-info:
Travis automatic deployment:
 https://pr1027--dexreact.review.gnosisdev.com

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -339,6 +350,17 @@ const OrderRow: React.FC<Props> = props => {
...
+  useEffect(() => {
+    // if token somehow not in list
+    // as it was traded already we assume it to be a valid token
+    // and know to user
+    // and add it to USER_TOKEN_LIST automatically
+    addTokenToListIfNotAlready({ token: sellToken, networkId })
+  }, [networkId, sellToken])
+  useEffect(() => {
+    addTokenToListIfNotAlready({ token: buyToken, networkId })
+  }, [networkId, buyToken])
```
anxolin, line 362: Although the logic is simple, I think this is not where we should add it

Imo, adding a token would be more a "service/api" job, more than a "view/component" job

Probably we should move it to a more central piece. When we load the orders. Even the hook would seem a better option.

If I make a new page, that list orders in a different way, and then it doesn't use `OrderRow` we would have this there.

Velenir: Where would you ove it? `useOrders` hook? Orders reducer?

anxolin: Maybe the reducer seems more general? It reacts to a state change
what do you think?

W3stside: i would move it to a middleware, should that be added

alfetopito: Probably a service might be best.
While building the `trades` service I found myself in a situation this can be useful.

Suppose the page is loaded directly into a `trades` view where the orders/tokenlist have not been loaded yet. There might be trades for tokens not present in the local list yet.

anxolin: I'm good, with the service too. Where wever is more central/natural

Velenir: > Suppose the page is loaded directly into a `trades` view where the orders/tokenlist have not been loaded yet

If trades include unlisted tokenId, then a similar or better yet compound side-effect(around global reducer)should be added

If we go the service way then it would either mean adding a sideeffect to orders fetching, and we currently just call `exchangeApi.getOrdersPaginated` without going through a service
And then another side-effect for trades fetching.
Or some bigger abstraction around any fetching that returns token ids.

I prefer to keep logic closer to where new tokenIds can appear -- in reducers, also because it only becomes relevant when we both display orders/trade and later user wants to see tokens. No reason to check for new unlisted tokens on every call to service.
Logic will become:
```js
case 'OVERWRITE_ORDERS':
case 'UPDATE_ORDERS':
case 'APPEND_ORDERS':
+ case 'APPEND_TRADES':
 // add new tokens
```

src/reducers-actions/orders.ts
```diff
@@ -126,3 +129,29 @@ export const reducer = (state: OrdersState, action: ReducerActionType): OrdersSt
...
+        const tokensFromExchange = await Promise.all(
+          // calls are already cached, so we don't get duplicate calls later in components
+          Array.from(newTokenIdsFromOrders).map(tokenId => getTokenFromExchangeById({ tokenId, networkId })),
+        )
+
+        // only unlisted tokens get added inside
+        tokenListApi.addTokens({
+          tokens: tokensFromExchange.filter<TokenDetails>((token): token is TokenDetails => !!token),
+          networkId,
+        })
```
Velenir, line 154: This can be a service, if you want
@alfetopito @anxolin

src/components/OrdersWidget/OrderRow.tsx
alfetopito: I'm currently using the token ids before getting into the global state: https://github.com/gnosis/dex-react/pull/1012/files#diff-1027fd89b6327be60ccab6646b57d502R65

The hook is calling the service and updating the global state. 
I didn't stop yet to think this through, but one option might be to re-use part of your logic here.

src/reducers-actions/orders.ts
alfetopito: Whatever works to make this reusable, I'll be happy :)

src/components/OrdersWidget/OrderRow.tsx
Velenir: You will touch reducer afterwards, right? To persisting trades to global state, to not refresh every time

alfetopito: Yes, but at that point, I should have the tokens populated already.
And the reducer is already implemented: https://github.com/gnosis/dex-react/pull/1012/files#diff-7ef769cc38af3906b033b24919b284b2R21

PR #1037, opened 2020-05-20
Handle dimas edge case

anxolin:
Okay, this is a little experiment to tackle @Velenir  edge case.
Try to follow his instructions and see the console:
1. Wrap 0.1 using metamask, but don't send the tx yet
2. Close the modal
3. Now send the transaction! --> See the console 😅

image

* According to the modal, it's open (that's a lie)
* According to the hook I created out of angryness (that's a lie, also concerning, cause in React devtool I see it as `false` --> I must be overlooking something)


gnosis-info:
Travis automatic deployment:
 https://pr1037--dexreact.review.gnosisdev.com

src/components/WrapEtherBtn.tsx
```diff
@@ -270,6 +270,17 @@ const WrapUnwrapEtherBtn: React.FC<WrapUnwrapEtherBtnProps> = (props: WrapUnwrap
...
                   })}
+                  onBlur={(e): void => {
+                    // react-hook-form does something onBlur that interferes with button clicks
+                    // at the same time modali buttons rerender at every opportunity
+                    // so click never arrives where we expect it to
+                    // at least that is my guess
+                    const { relatedTarget } = e
+                    // here be hacks
+                    if (relatedTarget instanceof HTMLButtonElement) {
+                      relatedTarget.click()
```
anxolin, line 281: oh boy! 🙈

it must be the validation that is done onBlur, but if the validation is successful, it shouldn't intercept the click.
Probably it'll be nice to do a codepen for them to show this weird thing, they must have an official explanation.

Thanks for debuging this one. It was not an easy one!

src/components/WrapEtherBtn.tsx
```diff
@@ -349,6 +329,11 @@ const WrapUnwrapEtherBtn: React.FC<WrapUnwrapEtherBtnProps> = (props: WrapUnwrap
     ],
   })
 
+  // toggleModal recreated every time, keep ref to always use current in async code
```
anxolin, line 343: So it was that?? 
🤔but with the useState should have worked then too, right? 
event if the toggle is reconstructed, the hook `isModalVisible` should have the same value. Obviously my reasoning is failing somewhere

src/components/WrapEtherBtn.tsx
```diff
@@ -67,12 +68,6 @@ const ModalWrapper = styled(ModalBodyWrapper)`
       padding-left: -0.5;
       font-size: 1.3rem;
       color: : var(--color-background-modali);
```
Velenir, line 70:
```suggestion
      color: var(--color-background-modali);
```

src/components/WrapEtherBtn.tsx
Velenir: `isModalVisible` is used inside `txOptionalParams` which is called asyncronously
So the `isModalVisible` inside `txOptionalParams` is actually  the one that is created at the same time as `txOptionalParams` function is created. `txOptionalParams` acts as closure around `isModalVisible` from that particular render.
And when `txOptionalParams` is finally called the real ``isModalVisible`` has been changed in a different render

1. Render 1:
1.1. `isModalVisible = A` created 
1.2. `txOptionalParams` and encloses over `isModalVisible = A` 
2. Render 2,3,4:
2.1.  `isModalVisible = B,C,D` created 
3. Render X:
3.1.   `txOptionalParams` called, still has ref to `isModalVisible = A`

anxolin: I see, thanks!

PR #1046, opened 2020-05-22
Mono

anxolin:
image

🚨**Highly experimental PR**🚨
* Defines a monorepo structure
* The web project now is moved to a `web` directory
* Additional packages are added as `packages/<module>` For now 2 modules:
     * `core`: An example of a core module that can be imported in all web apps.
     * `dev`: Used to simplify the monorepo package.json. See below.
* Uses yarn workspaces to wire everything up

image

**core module**
It's just an example on how `yarn` autowire the packages and we can create easily reusable packages

It's just a TS project, based in `dex-js`, that exports an OWL in ascii 🦉

image

**dev module**
So, probably most packages will use some shared  `devDependencies`.

One could think, it's a good idea to move the dev dependencies to the monorepo main package.json, but this it's a bad practice, since the repositories should be self-contained. 

Declaring the same versions of typescrypt, eslint, etc over and over is a bit boring and error prone.

One solution is to create an artificial package `dev` with the `dependencies` that you would put in `devDependencies` and the depend on this module as `devDependencies`

** Known Issues**
Yarn hoist the packages to the root `node_modules`, trying to be as efficient as possible.
I had some issues with this, because some libraries are not prepared for this.

yarn account for that by allowing you to set a `nohoist` that allows you to not hoist specific libraries. So you can specify a pattern there.

Right now, is not hoising all, and it works.
```js
"nohoist": [
  "**"
]
```

I've made some test and I've made it work with this config that target the problematic libraries. But I still run into issues with `dex-js`. It's a bit of a headache. I was hoping to use @Velenir  input on this 😀
```js
"nohoist": [
   "**/bn.js",
   "**/bn.js/**"
]
```

## Test
How to run it?? EASY! 😎
```bash
yarn
cd web
yarn start
```

then open the console and see the OWL:
image


anxolin:
Closing the PR, no-one reviewed anyway. Leaving the branch though! this MONO 🐒 might come back!

PR #1051, opened 2020-05-25
Update/dex js 0.4.0

W3stside:
Had to update types on contracts (as Dima warned)

I will update `smartFormat` into this PR so don't merge plz

PR #1055, opened 2020-05-26
Add format smart

W3stside:
Closes #1009 

Merges into #1051 

Adds format smart to dex-react

OrderBookWidget still not perfect but will work well enough for this iteration

Will likely require changes to dex-js in future to allow for more custom decimal assignment for different number thresholds

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -1,11 +1,39 @@
 import React, { useEffect, useRef } from 'react'
+import BigNumber from 'bignumber.js'
 import styled from 'styled-components'
-import { TokenDetails, Network } from 'types'
+
 import * as am4core from '@amcharts/amcharts4/core'
 import * as am4charts from '@amcharts/amcharts4/charts'
 import am4themesSpiritedaway from '@amcharts/amcharts4/themes/spiritedaway'
-import { getNetworkFromId, safeTokenName } from '@gnosis.pm/dex-js'
+
 import { dexPriceEstimatorApi } from 'api'
+
+import { getNetworkFromId, safeTokenName, formatSmart, toBN } from 'utils'
+
+import { TEN_BIG_NUMBER } from 'const'
+
+import { TokenDetails, Network } from 'types'
+
+function checkPriceAndConvert(price: number): string {
+  // prices come in as type number so need to be converted
+  // some of the WEI values are so small we need to cast to Wei twice
+  const priceAsBigNumber = new BigNumber(price)
+  const calculatedPriceAsBnWei = toBN(
+    priceAsBigNumber
+      .times(TEN_BIG_NUMBER.pow(new BigNumber('18')))
+      // we don't want any decimals
+      // before passing into BN conversion
+      .decimalPlaces(0)
+      .toString(10),
+  )
+
+  return formatSmart({
+    amount: calculatedPriceAsBnWei,
+    precision: 18,
+    decimals: 6,
+    smallLimit: '0',
+  })
+}
 
 interface OrderBookProps {
   baseToken: TokenDetails
@@ -66,7 +94,7 @@ interface ProcessedItem {
   totalVolume: number
   askValueY: number | null
   bidValueY: number | null
-  price: number
+  price: string | number
 }
 
 /**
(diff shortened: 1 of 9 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1055--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -143,13 +171,14 @@ const draw = (
     const processed = processData(data.bids, baseToken, quoteToken, Offer.Bid).concat(
       processData(data.asks, baseToken, quoteToken, Offer.Ask),
     )
-    processed.sort((lhs, rhs) => lhs.price - rhs.price)
+    processed
+      // cast as number to sort...
+      .sort((lhs, rhs) => +lhs.price - +rhs.price)
+      // show as string
+      .map(processedItem => ({ ...processedItem, price: checkPriceAndConvert(processedItem.price as number) }))
```
Velenir, line 178: can `price` even be string? I looked at `processData` and it seems price is always number. Or is there a reason?

W3stside: I'm afraid of casting to number in line 178

src/components/OrderBookWidget.tsx
```diff
@@ -1,12 +1,40 @@
...
 
+import { getNetworkFromId, safeTokenName, formatSmart, toBN } from 'utils'
+
+import { TEN_BIG_NUMBER } from 'const'
+
+import { TokenDetails, Network } from 'types'
+
+function checkPriceAndConvert(price: number): string {
+  // prices come in as type number so need to be converted
+  // some of the WEI values are so small we need to cast to Wei twice
```
alfetopito, line 19: What do you mean cast to Wei twice?

src/components/OrderBookWidget.tsx
```diff
@@ -1,12 +1,40 @@
...
+      // before passing into BN conversion
+      .decimalPlaces(0)
+      .toString(10),
+  )
+
+  return formatSmart({
+    amount: calculatedPriceAsBnWei,
+    precision: 18,
+    decimals: 6,
+    smallLimit: '0',
```
anxolin, line 34: But this will make that smaller numbers. of 0.00001 would be shown as `0.000000` right?

I'd be ok for this version. Altough there should be an issue with the improvement

src/components/OrderBookWidget.tsx
```diff
@@ -1,12 +1,40 @@
...
+
+import { TokenDetails, Network } from 'types'
+
+function checkPriceAndConvert(price: number): string {
+  // prices come in as type number so need to be converted
+  // some of the WEI values are so small we need to cast to Wei twice
+  const priceAsBigNumber = new BigNumber(price)
+  const calculatedPriceAsBnWei = toBN(
+    priceAsBigNumber
+      .times(TEN_BIG_NUMBER.pow(new BigNumber('18')))
```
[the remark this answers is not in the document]
W3stside: > nice.
> 
> Also confused by the comment Leandro pointed out.

mm because some values are so small as Wei: `0.003123897234` for example that can't be converted to `BN` and `formatSmart` requires a `BN` amount, not `BIgNumber` open to better options

W3stside:
> nice.
> 
> Also confused by the comment Leandro pointed out.

mm because some values are so small as Wei: 0.003123897234 for example that can't be converted to BN and formatSmart requires a BN amount, not BIgNumber open to better options

src/components/OrderBookWidget.tsx
W3stside: see my quote reply to Anxo - essentially some WEI values come in super small as well and can't be cast to BN and formatSmart doesn't accept `BigNumber`

src/components/OrderBookWidget.tsx
W3stside: agreed

anxolin:
> mm because some values are so small as Wei: 0.003123897234 for example that can't be converted to BN and formatSmart requires a BN amount, not BIgNumber open to better options

Ahh, yes. I think is right that formatSmart uses BN, I just didn't understand this way to express it `some of the WEI values are so small we need to cast to Wei twice`

We can review it in another moment, there's sth strange about it. anyways the displayed values are correct, even when decimals are different :)

PR #1061, opened 2020-05-27
Fixing token images on dark mode

alfetopito:
Before:

screenshot_2020-05-27_08-04-01
screenshot_2020-05-27_08-03-52

Now:

screenshot_2020-05-27_07-52-04
screenshot_2020-05-27_07-51-55


Velenir:
I only don't like how it feels in the hole in PAX. Can you try adding `background-clip: padding` and see

gnosis-info:
Travis automatic deployment:
 https://pr1061--dexreact.review.gnosisdev.com

alfetopito:
> I only don't like how it fills in the hole in PAX. Can you try adding `background-clip: padding` and see
> Nah, it'll look worse

Yeah, not all of them look better. PAX symbols are better with the default transparent background.

As for `background-clip: padding`, I see no difference.

Issue #1049, opened 2020-05-25
Order Book: Linear price for the X-axis


anxolin:
![image](https://user-images.githubusercontent.com/2352112/82791977-f1e30680-9e6e-11ea-9667-10bc1f8ff14e.png)

https://gnosisinc.slack.com/archives/CSGLUC1BN/p1590145766002600

> This view suggest that there is a big spread between buy and sell. In reality it is not, it is many very small orders (on the red side) with almost the same price. However - the scale of the x axis is not linear

Velenir:
_Estimation: 3_

alfetopito:
Moved to `gp-ui`. See https://github.com/gnosis/gp-ui/pull/6

PR #1062, opened 2020-05-27
Improve order book representation

### LINKED ISSUE #1056: `View Order Book` displays the Order Book of the opposite pair
After selecting a trading pair (A/B => Where A = buyToken, B = sellToken) on the `trade` tab, and clicking the `View Order Book` button, the orderbook of the opposite pair is displayed. 

The correct pair to express what is pictured below is DAI/WETH:
Screenshot 2020-05-26 at 16 26 46

When you click on `View order book`, you see the following image, which does not express the pair of the image above (DAI/WETH), but rather WETH/DAI:

Screenshot 2020-05-26 at 16 28 33

The correct way to express DAI/WETH in an orderbook would be the following image, which is the one I get by selecting WETH/DAI on the `Trade` page: 
Screenshot 2020-05-26 at 16 30 58

Conclusion: 

- `View Order Book` displays the Order Book of the opposite pair

Example of an orderbook for BTC/USD, from Bitmex:

Screenshot 2020-05-26 at 16 24 12

cc/ @alfetopito @anxolin

anxolin:
Fixes #1056

This PR does a refactor in the Order Book to tackle with the following issues:
- All operations are done using floats, so there's some precision errors and possible overflows
- The order book continues to be a source of confusion
- A refactor will come handy to prepare for https://github.com/gnosis/dex-react/issues/1049
- It prepares for receiving `string` from the API instead of numbers --> I guess that's coming, since Nick is working in the event-base order book. Now if they send strings instead of numbers, it should work too

## Inverts the order book
Apparently we were showing the order book the other way around

If we sell `USDC` for buying `WETH`, then we are in  `WETH-USDC` market
* `WETH`: Base token
* `USDC`: Quote token

This PR corrects this issue in the trading page

## Change the header
image

It was source of confusion. Sho now, it shows the Market (`BUY-SELL`) instead of `SELL --> BUY`

## Some changes regarding the representation.

In the previous version, we show the order book like this:

image

The proposal is to show it like this:
image

Note that:
* Y-axis shows is named after his units --> `WETH` in this case
* X-axis shows the units instead of the market --> `USDC` in this case
* The tooltip shows the Market where you are selling
* The tooltip price has been renamed and shows the units  --> `USDC`
* The tooltip volume shows the units --> `WETH` in this case
* Now volumes and prices are formatted using the `formatSmart` function, so they are more readable

## Other changes
**New types**
* `RawPricePoint`: Models the data returned by the endpoint
* `PricePoint` is a price point normalized (units instead of atoms, and uses BigNumber)
* `PricePointDetails` Models the details of the price point used for representation. It has some data replication needed to represent and format the price point (i.e. the formatted version of the volume and price)

**Debug**
Now the console shows the order book in the console. It's handy for debugging. Although in production is hidden

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -8,32 +8,15 @@ import am4themesSpiritedaway from '@amcharts/amcharts4/themes/spiritedaway'
 
 import { dexPriceEstimatorApi } from 'api'
 
-import { getNetworkFromId, safeTokenName, formatSmart, toBN } from 'utils'
+import { getNetworkFromId, safeTokenName, formatSmart, logDebug } from 'utils'
 
-import { TEN_BIG_NUMBER } from 'const'
+import { TEN_BIG_NUMBER, ZERO_BIG_NUMBER } from 'const'
 
 import { TokenDetails, Network } from 'types'
+import BN from 'bn.js'
+import { DEFAULT_PRECISION } from '@gnosis.pm/dex-js'
 
-function checkPriceAndConvert(price: number): string {
-  // prices come in as type number so need to be converted
-  // some of the WEI values are so small we need to cast to Wei twice
-  const priceAsBigNumber = new BigNumber(price)
-  const calculatedPriceAsBnWei = toBN(
-    priceAsBigNumber
-      .times(TEN_BIG_NUMBER.pow(new BigNumber('18')))
-      // we don't want any decimals
-      // before passing into BN conversion
-      .decimalPlaces(0)
-      .toString(10),
-  )
-
-  return formatSmart({
-    amount: calculatedPriceAsBnWei,
-    precision: 18,
-    decimals: 6,
-    smallLimit: '0',
-  })
-}
+const SMALL_VOLUME_THRESHOLD = 0.01
 
 interface OrderBookProps {
   baseToken: TokenDetails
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1062--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -84,67 +67,161 @@ enum Offer {
...
+  totalVolumeFormatted: string
   askValueY: number | null
   bidValueY: number | null
-  price: string | number
+}
+
+function _toPricePoint(pricePoint: RawPricePoint, quoteTokenDecimals: number, baseTokenDecimals: number): PricePoint {
+  return {
+    price: new BigNumber(pricePoint.price).dividedBy(TEN_BIG_NUMBER.pow(quoteTokenDecimals - baseTokenDecimals)),
+    volume: new BigNumber(pricePoint.volume).div(TEN_BIG_NUMBER.pow(baseTokenDecimals)),
```
Velenir, line 115: Why `L114:.dividedBy` and `L115:.div`, they are the same, choose one

src/components/OrderBookWidget.tsx
```diff
@@ -84,67 +67,161 @@ enum Offer {
...
+  //  Normalized items use decimals (BigNumber) and are given in natural units
+  let pricePoints: PricePoint[] = rawPricePoints.map(pricePoint =>
+    _toPricePoint(pricePoint, quoteTokenDecimals, baseTokenDecimals),
+  )
+
+  // Filter tiny orders
+  pricePoints = pricePoints.filter(pricePoint => pricePoint.volume.gt(SMALL_VOLUME_THRESHOLD))
+
+  // Convert the price points that can be represented in the graph (PricePointDetails)
+  const { points } = pricePoints.reduce(
```
Velenir, line 160: While this is all nicely declarative, I feel like orderbook is sometimes slow and these 3 loops can be done in one inside reduce.
Just a thought. Feel free to ignore

anxolin: https://github.com/gnosis/dex-react/pull/1069

PR #1066, opened 2020-05-28
1032/sort orders by expiration time

### LINKED ISSUE #822: Make the expires clickable: to change the order
image

I think we are not sorting by expire date, we are just getting the orders "in creation order" 

* Should we order by expire date instead? I would say so 
* If yes, add an icon showing the order. I understand it should be decendant, so newest trade
* Make it clickable so we can invert the order

W3stside:
Closes #822 

Sort `Expire` column by ascending/descending


W3stside:
closing in favour for new PR branched off `release`

gnosis-info:
Travis automatic deployment:
 https://pr1066--dexreact.review.gnosisdev.com

PR #1067, opened 2020-05-28
822/sort orders by expiration time

### LINKED ISSUE #822: Make the expires clickable: to change the order

W3stside:
Closes #822

Sort Expire column by ascending/descending


gnosis-info:
Travis automatic deployment:
 https://pr1067--dexreact.review.gnosisdev.com

anxolin:
Mm, but why this one should be merged to release/1.0?

right now we are only addressing bugs there, not adding functionality

src/hooks/useSortByTopic.ts
```diff
@@ -0,0 +1,28 @@
+import { useMemo } from 'react'
+import useSafeState from './useSafeState'
+
+interface SortTopic {
+  topic: string
+  order: boolean
```
anxolin, line 6: better say  `ascendent`?

src/hooks/useSortByTopic.ts
```diff
@@ -0,0 +1,28 @@
...
+  data: T[],
+  defaultTopic: string,
+): { sortTopic: SortTopic; setSortTopic: React.Dispatch<React.SetStateAction<SortTopic>> } {
+  // true == 'asc' && false == 'dsc'
+  const [sortTopic, setSortTopic] = useSafeState<SortTopic>({ topic: defaultTopic, order: true })
+
+  useMemo(() => {
+    if (sortTopic) {
+      data.sort((lhs, rhs) =>
+        sortTopic.order ? lhs[sortTopic.topic] - rhs[sortTopic.topic] : rhs[sortTopic.topic] - lhs[sortTopic.topic],
```
anxolin, line 20: ubernit: if extract sortTopic.topic to a const, it gets simpler

anxolin, line 20: It won't work for BigNumbers, string with mixed cases, or data where the sorting field is nested (i.e. token symbol), or would require artificial derived fields for composed data (i.e market)

I suggest adding as a parameter an optional Map that would give us a comparer for every topic

src/components/OrdersWidget/index.tsx
```diff
@@ -140,6 +141,9 @@ const OrdersWidget: React.FC = () => {
     [displayedOrders],
   )
 
+  // Sort validUntil
+  const { sortTopic, setSortTopic } = useSortByTopic<AuctionElement>(displayedOrders, 'validUntil')
```
anxolin, line 145: nit: move `validUntil` to a constant i.e. `DEFAULT_SORT_BY`

Also, what about creating a type `type OrderSortByField = 'validUntil' | '...'`
Also, I would add a 2nd generic param for the sort by topic. Instead of string, is nicer to use `OrderSortByField`

src/hooks/useSortByTopic.ts
alfetopito: `asc` for short. We are talking about sort, so it's pretty common and expected

src/components/Layout/Card/Card.tsx
```diff
@@ -149,8 +149,6 @@ export const CardTable = styled.table<{
     > tr:not(${CardRowDrawer}) {
       position: relative;
       display: grid;
-      // grid-template-columns: ${({ $columns }): string => $columns || `repeat(auto-fit, minmax(3rem, 1fr))`};
-      // grid-template-columns: minmax(2rem,.4fr) minmax(7rem,16rem) minmax(4rem,11rem) minmax(5rem, 7.5rem) minmax(3rem,9rem);
       grid-template-columns: 3.2rem 1fr 1fr minmax(3rem,8rem) minmax(5rem,9rem);
```
alfetopito, line 152: I'm touching this in my PR so we'll have conflicts (ok ok, I will have conflicts).

Anyway, I need custom column sizes, so I'd ask to keep the parameterized version.

src/components/OrdersWidget/index.tsx
```diff
@@ -258,9 +262,7 @@ const OrdersWidget: React.FC = () => {
             {ordersCount > 0 ? (
               <div className="ordersContainer">
                 <CardTable
-                  // $columns="minmax(2rem, min-content) minmax(13.625rem, 1fr) repeat(2, minmax(6.2rem, 0.6fr)) minmax(5.5rem, 1fr)"
                   $columns="minmax(2rem,.4fr)  minmax(11rem,1fr)  minmax(11rem,1.3fr)  minmax(5rem,.9fr)  minmax(auto,1.4fr)"
```
alfetopito, line 265: This for example is useless at the moment since `$columns` is not being used.

src/hooks/useSortByTopic.ts
```diff
@@ -0,0 +1,28 @@
+import { useMemo } from 'react'
+import useSafeState from './useSafeState'
+
+interface SortTopic {
+  topic: string
+  order: boolean
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
alfetopito, line 9: There's no `any`

src/hooks/useSortByTopic.ts
```diff
@@ -0,0 +1,28 @@
...
+function useSortByTopic<T>(
+  data: T[],
+  defaultTopic: string,
+): { sortTopic: SortTopic; setSortTopic: React.Dispatch<React.SetStateAction<SortTopic>> } {
+  // true == 'asc' && false == 'dsc'
```
alfetopito, line 14: Adding to Anxo's comment above, renaming `order` to `asc` makes it clearer. No need for the comment.

src/hooks/useSortByTopic.ts
```diff
@@ -0,0 +1,28 @@
...
+
+interface SortTopic {
+  topic: string
+  order: boolean
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function useSortByTopic<T>(
+  data: T[],
+  defaultTopic: string,
```
alfetopito, line 12: Could use something like this to avoid passing in a topic that doesn't exist:

```suggestion
function useSortByTopic<T, K extends keyof T>(
  data: T[],
  defaultTopic: K,
```

:warning: Code above was not tested, might require changes.

anxolin: I was suggesting a similar thing in this other comment (https://github.com/gnosis/dex-react/pull/1067#discussion_r431851117), but:

instead of `K extends keyof T` 
I would say `K extends string` 

so `type OrderSortByField = 'validUntil' | 'otherSorting' | ...` would work with it too

W3stside:
> Mm, but why this one should be merged to release/1.0?
> 
> right now we are only addressing bugs there, not adding functionality

changed

src/hooks/useSortByTopic.ts
W3stside: good idea, can you give me an example of the comparer parameter?

W3stside: i think it get it, a `cb` parameter to pass into sort, yeah?

Velenir: Yes, accepting predicate/comparator/callback is the more generic solution
Would also be easier on the types with `K extends keyof T`, as TS would refuse to compare apples with oranges (BN with string)

src/hooks/useSortByTopic.ts
```diff
@@ -1,28 +1,41 @@
...
+  const { asc, topic } = sortTopic
 
   useMemo(() => {
-    if (sortTopic) {
-      data.sort((lhs, rhs) =>
-        sortTopic.order ? lhs[sortTopic.topic] - rhs[sortTopic.topic] : rhs[sortTopic.topic] - lhs[sortTopic.topic],
-      )
+    if (topic) {
+      // make a copy as not to mutate
+      const dataCopy = Object.assign([], data)
```
Velenir, line 23: Why?
```suggestion
      const dataCopy = data.slice(0)
```

W3stside: is one faster than the other?

W3stside: also i don't need the 0 there i think

Velenir: `0` is there because once upon a time @alfetopito said it's faster. He even cited something.

It is faster though insignificantly, but more importantly it's a standard way to clone an array

Velenir: Look what I found
https://jsben.ch/Wncwa

W3stside: lol, that's convincing enough

alfetopito: Found it! https://stackoverflow.com/a/21514254/1272513
Doesn't seem to make that big of a difference though 

screenshot_2020-05-29_08-01-11

alfetopito:
Build failure... TS as always

src/hooks/useSortByTopic.ts
Velenir: @alfetopito 
in your image `.slice()` is faster than `.slice(0)` :laughing:

W3stside:
> way better!! 
> 
> You recreated this PR, right? maybe is good in this cases you link the old one to keep track of the comments

mm no i didn't

PR #1068, opened 2020-05-28
dont target radio container div

### LINKED ISSUE #1065: CSS issue in validity inputs
The long token symbol pr which added ellipses into token name at max length caused an issue in another input as seen below:

Screenshot from 2020-05-28 11-22-38

Very simple fix:
```css
label > div:not(.radio-container) {
    position: absolute;
    width: 7.7rem;
    ...
```

does this need a hotfix? @anxolin

W3stside:
Closes #1065 

Fixes inputs css issues in PriceValidity

### DIFF
--- a/src/components/TradeWidget/Price.tsx
+++ b/src/components/TradeWidget/Price.tsx
@@ -84,7 +84,7 @@ export const PriceInputBox = styled.div`
     }
   }
 
-  label > div {
+  label > div:not(.radio-container) {
     position: absolute;
     width: 7.7rem;
     right: 1rem;


gnosis-info:
Travis automatic deployment:
 https://pr1068--dexreact.review.gnosisdev.com

src/components/TradeWidget/Price.tsx
```diff
@@ -84,7 +84,7 @@ export const PriceInputBox = styled.div`
     }
   }
 
-  label > div {
+  label > div:not(.radio-container) {
```
Velenir, line 87: Maybe rather than target everything but (`:not(*)`), better target a specific class.
Otherwise this might bite us back at another time

Or not, doesn't matter much

W3stside: yeah i also considered this but this the only use place where it differs, i can change it if others think it important

PR #1069, opened 2020-05-28
Move filter/map to reduce operation

anxolin:
@Velenir this PR is related to this comment

I don't think this is an issue, or the bottleneck, but I agree is slightly more performant, and is not making it to much complex the reduce.

However, for some reason, it breaks the Order Book!
I print the Order Book in the console and compare it with the 2 versions, and the points are the same! 🤷‍♂️. It looks like equivalent information! 
however, the Y-axis only show the asks and not the bids

For example, WETH-USDC with 1 hops.

This PR is not very important, it's just something that made me curious, why equivalent information would be represented differently. ❓

## before
image

## this PR
image

## If you are curious about this madness
madness --> it never is, usually there's an explanation 

Just check http://localhost:8080/book with WETH-USDC, hops 1
```bash
# change between branches, it should autorefresh
git checkout improve-order-book
git checkout move-filtering-and-map-to-reduce
```

### DIFF
--- a/src/components/OrderBookWidget.tsx
+++ b/src/components/OrderBookWidget.tsx
@@ -146,19 +146,20 @@ const processData = (
   const quoteTokenDecimals = quoteToken.decimals
   const baseTokenDecimals = baseToken.decimals
 
-  // Convert RawPricePoint into PricePoint:
-  //  Raw items use number (floats) and are given in "atoms"
-  //  Normalized items use decimals (BigNumber) and are given in natural units
-  let pricePoints: PricePoint[] = rawPricePoints.map(pricePoint =>
-    _toPricePoint(pricePoint, quoteTokenDecimals, baseTokenDecimals),
-  )
-
-  // Filter tiny orders
-  pricePoints = pricePoints.filter(pricePoint => pricePoint.volume.gt(SMALL_VOLUME_THRESHOLD))
-
   // Convert the price points that can be represented in the graph (PricePointDetails)
-  const { points } = pricePoints.reduce(
-    (acc, pricePoint, index) => {
+  const { points } = rawPricePoints.reduce(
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1069--dexreact.review.gnosisdev.com

src/components/OrderBookWidget.tsx
```diff
@@ -146,19 +146,20 @@ const processData = (
...
+      // Convert RawPricePoint into PricePoint:
+      //  Raw items use number (floats) and are given in "atoms"
+      //  Normalized items use decimals (BigNumber) and are given in natural units
+      const pricePoint = _toPricePoint(rawPricePoint, quoteTokenDecimals, baseTokenDecimals)
+
+      // Filter tiny orders
+      if (pricePoint.volume.lte(SMALL_VOLUME_THRESHOLD)) {
+        logDebug('[OrderBook] pricePoint discarded: ', pricePoint.price.toNumber(), pricePoint.volume.toNumber())
+        return acc
+      }
```
Velenir, line 161: Somehow this is the problem. If you just remove this condition everything works
I don't understand why :confounded: 

If you prefiler like on L157 -- works, if you postfilter `points` before returning from function -- works.
But inside `reduce` -- doesn't..

anxolin: Okay, thanks Dima. I think I'll close this PR.

I don't get it, I see the signature of:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

The callback returns the reduced object. If I ignore the point, I just return `acc` then.
Also, I printed the results, and it's equivalent in the console (I text compared the result) 🤷‍♂️

I'll close

PR #1071, opened 2020-05-28
Save trades as csv

Velenir:
Adds CSV saving functionality

Feel free to style/position as you see fit

Also I'm not a fan of how `formatTradeAsCSV` has to duplicate a lot of `TradeRow` functionality and make an extra loop over `trades`.
Maybe perform formatting only once and make `TradeRow` accept
```js
[
    displayTokenSymbolOrLink(trade.sellToken) + '/' + displayTokenSymbolOrLink(trade.buyToken),
    formatSmart({ amount: trade.sellAmount, precision: trade.sellToken.decimals as number }) +
      ' ' +
      displayTokenSymbolOrLink(trade.sellToken),
    formatPrice(trade.limitPrice),
    formatPrice(trade.fillPrice),
    formatSmart({ amount: trade.buyAmount, precision: trade.buyToken.decimals as number }) +
      ' ' +
      displayTokenSymbolOrLink(trade.buyToken),
    classifyTrade(trade),
    new Date(trade.timestamp).toLocaleString(),
    trade.txHash,
    trade.settlingDate > now ? 'NOT SETTLED' : 'SETTLED',
  ]
```
Up to you, @alfetopito

### DIFF
--- a/src/components/FileDownloaderLink.tsx
+++ b/src/components/FileDownloaderLink.tsx
@@ -1,14 +1,17 @@
-import React, { useRef, useEffect } from 'react'
-
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faFileCsv } from '@fortawesome/free-solid-svg-icons'
+import React, { useRef, useEffect, AnchorHTMLAttributes } from 'react'
 
 interface FileSaverLinkProps {
   data?: string
   options?: BlobPropertyBag
+  filename: string
 }
 
-export const FileDownloaderLink: React.FC<FileSaverLinkProps> = ({ data, options }) => {
+export const FileDownloaderLink: React.FC<FileSaverLinkProps & AnchorHTMLAttributes<HTMLAnchorElement>> = ({
+  data,
+  options,
+  filename,
+  ...anchorProps
+}) => {
   const lastObjectURL = useRef('')
   const lastDataUsed = useRef('')
 
@@ -41,9 +44,5 @@ export const FileDownloaderLink: React.FC<FileSaverLinkProps> = ({ data, options
     e.currentTarget.href = lastObjectURL.current = URL.createObjectURL(blob)
   }
 
-  return (
-    <a href={lastObjectURL.current} download="trades.csv" onClick={handleClick}>
-      <FontAwesomeIcon icon={faFileCsv} size="2x" />
-    </a>
-  )
+  return <a href={lastObjectURL.current} download={filename} onClick={handleClick} {...anchorProps} />
 }
--- a/src/components/FileDownloaderLink.tsx
+++ b/src/components/FileDownloaderLink.tsx
@@ -1,7 +1,7 @@
 import React, { useRef, useEffect, AnchorHTMLAttributes } from 'react'
 
 interface FileSaverLinkProps {
-  data?: string
+  data?: string | (() => string)
   options?: BlobPropertyBag
   filename: string
 }
@@ -32,15 +32,16 @@ export const FileDownloaderLink: React.FC<FileSaverLinkProps & AnchorHTMLAttribu
       e.preventDefault()
       return
     }
+    const csvData = typeof data === 'string' ? data : data()
     // data didn't change, reusing already set ObjectURL
-    if (lastDataUsed.current === data) return
-    lastDataUsed.current = data
+    if (lastDataUsed.current === csvData) return
+    lastDataUsed.current = csvData
 
     // new data, need new ObjectURL
     // release old if was set previously
     if (lastObjectURL.current) URL.revokeObjectURL(lastObjectURL.current)
 
-    const blob = new Blob([data], options)
+    const blob = new Blob([csvData], options)
     e.currentTarget.href = lastObjectURL.current = URL.createObjectURL(blob)
   }
 
--- a/src/components/FileDownloaderLink.tsx
+++ b/src/components/FileDownloaderLink.tsx
@@ -1,0 +1,49 @@
+import React, { useRef, useEffect } from 'react'
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faFileCsv } from '@fortawesome/free-solid-svg-icons'
+
+interface FileSaverLinkProps {
+  data?: string
+  options?: BlobPropertyBag
+}
+
+export const FileDownloaderLink: React.FC<FileSaverLinkProps> = ({ data, options }) => {
+  const lastObjectURL = useRef('')
+  const lastDataUsed = useRef('')
+
+  useEffect(
+    () => (): void => {
+      // on unmount release ObjectURL if any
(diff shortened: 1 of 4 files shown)


src/components/FileDownloaderLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+    // new data, need new ObjectURL
+    // release old if was set previously
+    if (lastObjectURL.current) URL.revokeObjectURL(lastObjectURL.current)
+
+    const blob = new Blob([data], options)
+    e.currentTarget.href = lastObjectURL.current = URL.createObjectURL(blob)
+  }
+
+  return (
+    <a href={lastObjectURL.current} download="trades.csv" onClick={handleClick}>
```
alfetopito, line 45: Idea, not for now necessarily.

Take the file name as an input parameter. 
We might want to use this for downloading orders as well.

Also, thinking about adding address and date to file name...

gnosis-info:
Travis automatic deployment:
 https://pr1071--dexreact.review.gnosisdev.com

src/components/TradesWidget/index.tsx
```diff
@@ -1,31 +1,71 @@
...
+import { formatPrice, formatSmart } from '@gnosis.pm/dex-js'
+
+import { Trade } from 'api/exchange/ExchangeApi'
+
+import { displayTokenSymbolOrLink } from 'utils/display'
+import { FileDownloaderLink } from 'components/FileDownloaderLink'
+
+const formatTradeAsCSV = (trade: Trade, now: Date): string => {
+  return [
+    displayTokenSymbolOrLink(trade.sellToken) + '/' + displayTokenSymbolOrLink(trade.buyToken),
```
alfetopito, line 20: I think in the CSV we should display both columns. Symbol (possibly empty) and address. More data wouldn't hurt.

src/components/TradesWidget/index.tsx
```diff
@@ -1,31 +1,71 @@
...
+    formatSmart({ amount: trade.sellAmount, precision: trade.sellToken.decimals as number }) +
```
alfetopito, line 21: Should use the full amount instead of `formatSmart`.

src/components/TradesWidget/index.tsx
```diff
@@ -1,31 +1,71 @@
...
+      ' ' +
+      displayTokenSymbolOrLink(trade.sellToken),
+    formatPrice(trade.limitPrice),
+    formatPrice(trade.fillPrice),
+    formatSmart({ amount: trade.buyAmount, precision: trade.buyToken.decimals as number }) +
+      ' ' +
+      displayTokenSymbolOrLink(trade.buyToken),
+    classifyTrade(trade),
+    new Date(trade.timestamp).toLocaleString(),
```
alfetopito, line 30: ISO date

src/components/TradesWidget/index.tsx
```diff
@@ -1,31 +1,71 @@
...
+}
+
+const tableHeads = ['Market', 'Amount', 'Limit Price', 'Fill Price', 'Received', 'Type', 'Time', 'Tx', 'Settled']
+const tableHeader = tableHeads.join(',') + '\n'
 
 const Trades: React.FC = () => {
   const { networkId } = useWalletConnection()
   const trades = useTrades()
 
+  const csvString = useMemo(() => {
```
alfetopito, line 55: Would be better to have this as a callback?

Otherwise on every change to `trades` we'll be re-creating the csv data, right?

Velenir: `data: string | (() => string)` ?

src/components/FileDownloaderLink.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useRef, useEffect, AnchorHTMLAttributes } from 'react'
+
+interface FileSaverLinkProps {
+  data?: string | (() => string)
+  options?: BlobPropertyBag
+  filename: string
+}
+
+export const FileDownloaderLink: React.FC<FileSaverLinkProps & AnchorHTMLAttributes<HTMLAnchorElement>> = ({
```
anxolin, line 9: wouldn't be a bit more readable like this?

```ts
interface FileSaverLinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
   ... 
}

export const FileDownloaderLink: React.FC<FileSaverLinkProps> = ({ ... })
```

src/components/FileDownloaderLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const csvData = typeof data === 'string' ? data : data()
+    // data didn't change, reusing already set ObjectURL
+    if (lastDataUsed.current === csvData) return
+    lastDataUsed.current = csvData
+
+    // new data, need new ObjectURL
+    // release old if was set previously
+    if (lastObjectURL.current) URL.revokeObjectURL(lastObjectURL.current)
+
+    const blob = new Blob([csvData], options)
```
anxolin, line 44: Wouldn't make sense to use a more stream approach?

We usually represent lists of something, so probably make sense, so we don't run into issues in bigger files?

src/utils/csv.ts
```diff
@@ -0,0 +1,57 @@
...
+        try {
+          values = transformer(item)
+
+          // Make sure the returned values match.
+          if (headers && values.length !== headers.length) {
+            throw new Error(`Values length (${values.length}) doesn't match headers length ${headers.length}`)
+          }
+        } catch (e) {
+          logDebug(`[utils:toCsv] Not able to transform into csv: ${item}`, e)
+          return acc
```
anxolin, line 34: not sure if this is the best error handling.

I think if we find an error, we should raise it so the user gets some input back
Relying on the log is not good, cause we will not realise if 1 registry breaks the transformer

src/utils/csv.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Util function that transforms given data and transformer function into a csv string,
+ * taking care of escaping and error handling
+ *
+ * `headers` (optional). If present, length must match transformer return type length
+ * `data` is a list of items to be converted to CSV by `transformer`
+ * `transformer` is a function that receives the item and returns a list of strings
+ */
+export function toCsv<T>({ headers = [], data, transformer }: ToCsvParams<T>): string {
```
anxolin, line 20: I know you love to use fancy stuff, is this a job for generators? 

You don't need to change this, only if it really makes it more robust  (if it makes a difference), i.e. for big files, instead of creating all the array of string and joining you process the elements and create the blob from the steam.

Velenir: I don't understand what you are suggesting.
How would you stream to a blob, I don't think you can
Neither can you stream from a blob.
And `URL.createObjectURL` is sync

src/components/FileDownloaderLink.tsx
Velenir: How would you use a stream here?

PR #1087, opened 2020-06-04
1030/reverted trades

### LINKED ISSUE #1030: Show filled orders (trades) - Handle reverts
* Depends on #1029
* Fetches revert events
* Merges the revert events with the trade events, and show it beautifully in the new Filled tab

**Implementation note:** For matching trades and reverts, we assign a cardinal to trades and reverts depending on the order according to (order id, block number, event index)

alfetopito:
Closes #1030

## Update Jun 08

- No longer removing reverted trades
- No longer storing reverts on global state
- Only for testing/debugging, added an icon next to reverted trades. Will filter reverted trades out before merging and remove this entirely.

screenshot_2020-06-08_14-09-35

- Addressed minor PR comments

screenshot_2020-06-08_15-03-58

---

Removing reverted trades.

screenshot_2020-06-05_13-56-29

Todo:
- [x] Use trade reverts on global state to filter out reverted trades
- [x] Fix type issue on `getTrades` service return
screenshot_2020-06-04_15-12-31

### DIFF
--- a/src/reducers-actions/trades.ts
+++ b/src/reducers-actions/trades.ts
@@ -94,24 +94,38 @@ function applyRevertsToTrades(trades: Trade[], reverts: TradeReversion[]): [Trad
     const reverts = (revertsByRevertKey.get(revertKey) as TradeReversion[]).sort(sortByTimeAndPosition)
     const trades = tradesByRevertKey.get(revertKey)?.sort(sortByTimeAndPosition)
 
-    // Given the assumptions above, the possibilities are:
-    // 1. # trades > # reverts
-    // 2. # trades == # reverts
-    // Not possible:
-    // 3. # trades < # reverts
-    // 4. There are reverts but there are no trades for given revertKey <- should NOT ever happen
-    // 5. There are trades but no reverts <- wouldn't be in this loop in that case
     if (trades) {
-      const remainingTrades = trades.length - reverts.length
-      if (remainingTrades < 1) {
-        // Case 2 and 3, all trades reverted
-        tradesByRevertKey.delete(revertKey)
-        logDebug(`[reducers:trade][${revertKey}] All ${trades.length} trade(s) reverted by ${reverts.length} revert(s)`)
-      } else {
-        // Case 1. One or more trades left, pick from the end
-        const filteredTrades = trades.slice(trades.length - remainingTrades)
-        tradesByRevertKey.set(revertKey, filteredTrades)
-        logDebug(`[reducers:trade][${revertKey}] Reverted ${reverts.length} trade(s), ${filteredTrades.length} left`)
+      let tradesIndex = 0
+      let revertsIndex = 0
+      // Iterate over both trades and reverts while there are any
+      while (tradesIndex < trades.length && revertsIndex < reverts.length) {
+        if (!trades[tradesIndex].revertId) {
+          // Trade was not reverted. Now it is.
+          // Update trade, move both indices forward
+          trades[tradesIndex].revertId = reverts[revertsIndex].id
+          trades[tradesIndex].revertTimestamp = reverts[revertsIndex].timestamp
+          logDebug(
+            `[reducers:trade][${revertKey}] Trade ${trades[tradesIndex].id} matched to Revert ${reverts[revertsIndex].id}`,
+          )
+          tradesIndex++
+          revertsIndex++
+        } else if (trades[tradesIndex].revertId === reverts[revertsIndex].id) {
+          // Trade was already reverted by this Revert
+          // Move both indices forward
+          logDebug(
+            `[reducers:trade][${revertKey}] Trade ${trades[tradesIndex].id} was already matched to Revert ${reverts[revertsIndex].id}`,
+          )
+          tradesIndex++
+          revertsIndex++
+        } else {
+          // Trade was already reverted, not by this Revert
+          // Move on to next Trade, keep same Revert
+          tradesIndex++
+        }
+      }
+      // Shouldn't happen, here for debugging
+      if (tradesIndex === trades.length && revertsIndex < reverts.length) {
+        console.error(`There are reverts not matched to trades`)
       }
     }
     // TODO: this effectively makes the reverts global state to always be empty.
--- a/src/reducers-actions/trades.ts
+++ b/src/reducers-actions/trades.ts
@@ -46,14 +46,11 @@ function groupByRevertKey<T extends EventWithBlockInfo>(list: T[]): Map<string, 
 
   list.forEach(item => {
     if (map.has(item.revertKey)) {
-      // Do I have to pop it out?
       const subList = map.get(item.revertKey) as T[]
 
       // Do not insert duplicates
       if (!subList.find(({ id }) => item.id === id)) {
         subList.push(item)
-        // Do I have to put it back in?
-        map.set(item.revertKey, subList)
       }
     } else {
       map.set(item.revertKey, [item])
--- a/src/reducers-actions/trades.ts
+++ b/src/reducers-actions/trades.ts
@@ -41,7 +41,9 @@ export const updateLastCheckedBlock = (lastCheckedBlock: number): UpdateBlockAct
 // TODO: store to/load from localStorage
 export const INITIAL_TRADES_STATE = { trades: [], reverts: [] }
 
-function groupByRevertKey<T extends EventWithBlockInfo>(list: T[], map: Map<string, T[]>): void {
+function groupByRevertKey<T extends EventWithBlockInfo>(list: T[]): Map<string, T[]> {
+  const map = new Map<string, T[]>()
+
   list.forEach(item => {
     if (map.has(item.revertKey)) {
       // Do I have to pop it out?
@@ -57,6 +59,8 @@ function groupByRevertKey<T extends EventWithBlockInfo>(list: T[], map: Map<stri
       map.set(item.revertKey, [item])
     }
   })
+
+  return map
 }
 
 function flattenGroup<T extends EventWithBlockInfo>(map: Map<string, T[]>): T[] {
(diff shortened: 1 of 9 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1087--dexreact.review.gnosisdev.com

src/reducers-actions/trades.ts
```diff
@@ -29,30 +39,99 @@ export const updateLastCheckedBlock = (lastCheckedBlock: number): UpdateBlockAct
...
+    return acc.concat(list)
+  }, [])
+}
+
+function applyRevertsToTrades(trades: Trade[], reverts: TradeReversion[]): [Trade[], TradeReversion[]] {
+  const tradesByRevertKey = new Map<string, Trade[]>()
+  const revertsByRevertKey = new Map<string, TradeReversion[]>()
+
+  // Group trades by revertKey
+  groupByRevertKey(trades, tradesByRevertKey)
```
anxolin, line 76: it's not super good practice to mutate things inside a method. Imo, groupByRevertKey would need to receive only the trades, and do a "reduce" operation starting the reduce with `new Map<string, Trade[]>()`

src/reducers-actions/trades.ts
```diff
@@ -29,30 +39,99 @@ export const updateLastCheckedBlock = (lastCheckedBlock: number): UpdateBlockAct
...
-          seenTradeIdsSet.add(trade.id)
-        }
-        return acc
-      }, [])
-
-      return { trades, lastCheckedBlock }
+      const { trades: currTrades, reverts: currReverts } = state
+      const { trades: newTrades, reverts: newReverts, lastCheckedBlock } = action.payload
+
+      const [trades, reverts] = applyRevertsToTrades(currTrades.concat(newTrades), currReverts.concat(newReverts))
```
anxolin, line 125: This is not incremental, right?
I mean, you apply, and repply for past batches every time you have new trades

wouldn't it make sense to just have a global state with trades, this trades have a revertDate (in case they were reverted), and then you only need to use the reducer to appply the new information:
- new trades: You add them to the global state
- new reverts: You match them in order to the trades of that user/batch/orderId using blockNumber and the index of the event to sort them and decide which one is next

This way:
* You keep all the information as a single entity TRADES, that you know they can be reverted or not (checking the revertDate field)
* You process things incrementally, no need to reprocess past batches
* The hook can return this trades, so the view can decide to filter out and remove the reverted

Alternativelly, the hook could return the not reverted ones only, if we think we won't ever need the other ones. Maybe is good

src/services/factories/getTrades.ts
```diff
@@ -2,42 +2,114 @@ import Web3 from 'web3'
...
+      const sellToken = tokens.get(event.sellTokenId) as TokenDetails
+
+      // Maybe we couldn't find the order /shrug
+      const order = orders.get(event.orderId)
+
+      return {
+        ...event,
+        batchId,
+        timestamp,
+        revertKey,
```
anxolin, line 79: I don't think this should be part of the data object, right?
this seems more like an internal thing we use for calculating things, so we don't want to expose it.

src/reducers-actions/trades.ts
alfetopito: I see. Will give the idea of not deleting and instead setting a `revertDate` a try.

alfetopito: Done: d3c154ab0e31df91097ae1f084bf0ba557438ff8

To illustrate, I added an icon for reverted trades (will remove this before merging)
screenshot_2020-06-08_14-09-35

src/services/factories/getTrades.ts
alfetopito: Why not? And which one specifically are you talking about?

src/reducers-actions/trades.ts
```diff
@@ -29,30 +39,99 @@ export const updateLastCheckedBlock = (lastCheckedBlock: number): UpdateBlockAct
...
+        // Do I have to put it back in?
+        map.set(item.revertKey, subList)
+      }
+    } else {
+      map.set(item.revertKey, [item])
+    }
+  })
+}
+
+function flattenGroup<T extends EventWithBlockInfo>(map: Map<string, T[]>): T[] {
```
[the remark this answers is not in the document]
alfetopito: Done 2920f28afd94242a1012d3fbf4058d4ea8bf3695

src/reducers-actions/trades.ts
anxolin: Well, part of this comment was also originated by the line:

```js
const [trades, reverts] = applyRevertsToTrades(currTrades.concat(newTrades), currReverts.concat(newReverts))
```

It seams that on every `APPEND_TRADES`, so on every new trade, it looks like we reprocess all trades.

Issues with that:

* It's imposible to revert from the last batch. So when we change batch trades are final. An user with trades expanding months could reprocess only the last batch
* Maybe is just keeping track of the last batch you know, and the list of unreverted trades on it. Then you map 1<>1 in order for the reverts and remove them from the list (they never need to be processed again either)
* The double loop is a bit complex, iterating with 2 indexes. Maybe if you just keep the `unreverted` you can use a reduce like this:

```js
// Recalculate unreverted trades for the current batch
unreverted = reverts.reduce(youReduceLogic, new Set(unreverted))
```

anxolin: Also, for first processing, I understand we need to process old batches, so the logic is a bit more complex.

I tried to write this code in case it gives you any ideas, not for using it right away, cause I know it won't work (I think I ignored for example that the matching should consider the orderIds), so take it just as an idea on incremental updating.

```js
// Group new trades and reverts by batch
// Note how I make sure I also consider my previous unrevertedTrades  from my comment above  ☝️
const tradesByBatch: BatchInfo<Trade> = trades.concat(unrevertedTrades).reduce(groupByBatchId, [])
const revertsByBatch: BatchInfo< EventWithBlockInfo> = reverts.reduce(groupByBatchId, [])

// not sure if we can assume order, otherwise
tradesByBatch = tradesByBatch.sort(sortByBatchId)


// Mark some trades as reverted, and save the unreverted of the batch
tradesByBatch.tradesByBatch(batch => {
   let reverts = revertsByBatch.find(batchRevert => batchRevert.batchId === batch.batchId) // you can use a map, if u find u want to be more performant
  
     // Make sure revers and trades are in order
    const trades = batch.items.sort(sortByBlockNumberAndEventIndex)
    reverts = reverts ? reverts.sort(sortByBlockNumberAndEventIndex) : []

    // Get the unreverted trades
    let unreverted = trades.filter(notReverted)

    // Recalculate unreverted trades for the current batch
    unreverted = reverts.reduce(yourReduceLogic, new Set(unreverted))
    batch.unreverted = unreverted
})

// Keep track always of the latests unreverted trades (only for the current batch), so when we get new trades and reverts we can consider them for the incremental update
const latestBatch = tradesByBatch[tradesByBatch.length -1]
unrevertedTrades = latestBatch >= currentBatch ? latestBatch.unreverted : []
```

alfetopito: I get your concern, and I can address parts, but not all of it.

Even if I chose to (re)process only the last N batches (N = 1, 2, etc) on insertion, I'll have to go over the current list and find trades of given batches.
That's the same cost already "paid" for grouping trades (by batch|order).
I could go back only X trades looking for N batches to shorten the search range, to improve this a bit.

One thing that I don't think I can do is to ignore already reverted trades.
That's because for a given Revert, I can only tell it has been consumed when I find a trade with a matching `revertId`. Otherwise I risk (in the worst case) reverting all trades (of a batch|order) with a single Revert event.

One solution (to not processing all trades on every updates) would be to store in the global state the map with groups of trades for faster insertion, at the cost of flattening the map on read.

---

Alternatively, (re-read multiple times your points) trying to interpret your suggestion again.

1. On first load, match reverts to trades. 
1. Store on global state all trades from current + last batch as `pendingTrades`.
1. On every new group of trades/reverts:
    1. Add `pendingTrades` to new `trades`
    1. Apply `reverts` to them.
    1. Update global trades list with the result
    1. Repeat step `2`

Is this more inline with what you are thinking?
I'll give this a try.

---

Last point: double index loop is necessary! 
The alternative would be a less efficient nested loop, where for each trade I'll loop through (potentially all) reverts. Even if I remove consumed reverts from this list, it's not as good.

anxolin: I think we discuss it in the chat. let me know if you want to discuss anything further

> Last point: double index loop is necessary!
😂

src/services/factories/getTrades.ts
anxolin: I mean `revertKey` looks like a derived field, and implementation detail not part of the trade data.

Imo, It doesn't belong in any component, service, or hook. Maybe is just part of the reducer, and maybe not even that is required if we consider other ways to update the trades. 

Also, not a big deal, and nitpick so if it makes it too complicated, ignore it

alfetopito: Ah yes, this makes sense. Removed.

src/reducers-actions/trades.ts
alfetopito: Hopefully now I got what you meant. I addressed somewhat as I described on **Part 2** above.

And **Part 3** remains!

src/services/factories/getTrades.ts
```diff
@@ -122,47 +193,16 @@ export function getTradesFactory(factoryParams: {
...
-    }, [])
+    // ### TRADES and TRADE REVERSIONS ###
+    // Final step, put all together
+    const [trades, reverts] = await Promise.all([
+      assembleTrades(tradeEvents, blockTimes, orders, tokens),
+      assembleTradeReversions(tradeReversionEvents, blockTimes),
+    ])
 
-    return trades
+    //TODO: Can't figure out why, but TS doesn't accept `trades` is of type `Trade[]`
```
Velenir, line 203: Try `Promise.all<Trade[], TradeReversion[]>()`

PR #1089, opened 2020-06-08
Drop unused amchart deps

Velenir:
removes some deps from bundle


gnosis-info:
Travis automatic deployment:
 https://pr1089--dexreact.review.gnosisdev.com

PR #1091, opened 2020-06-09
759/side effect pending orders

### LINKED ISSUE #759: Sometimes there's a big delay showing the order in pending

W3stside:
Closes #759 

Removes Middleware code in favour for sideEffect Reducer logic

1. pendingOrders can be stuck, so logic inside `useOrders` checks for tx hash in block tx list to remove if necessary
2. logic happens after reducer for reduced logic implementation as was seen in middleware
3. pending saved to local storage
4. changes key names as previously mentioned in older PR (all comments were addressed thus not showing here)


W3stside:
A lot of commits here from older branch, the last 3 commits are the only relevant ones

gnosis-info:
Travis automatic deployment:
 https://pr1091--dexreact.review.gnosisdev.com

src/reducers-actions/pendingOrders.ts
```diff
@@ -89,18 +83,60 @@ export const reducer = (state: PendingOrdersState, action: ReducerType): Pending
 
       return newState
     }
-    case ActionTypes.REMOVE_PENDING_ORDERS: {
-      const { networkId, pendingTxHash, userAddress } = action.payload
+    case 'REMOVE_PENDING_ORDERS': {
```
Velenir, line 86: This is more `'REPLACE_PENDING_ORDERS'` than `REMOVE`

src/hooks/useOrders.ts
```diff
@@ -1,34 +1,90 @@
...
+  //   2. remove any orders that may have been re-sent at a higher gas price
+  const currentPendingOrders = (userAddress && networkId && pendingOrders[networkId]?.[userAddress]) || EMPTY_ARRAY
+  useEffect(() => {
+    // Don't fire if there are no pending orders...
+    if (userAddress && networkId && currentPendingOrders.length > 0) {
+      const managePendingOrders = async (): Promise<void> => {
+        const latestBlock = await web3.eth.getBlock(blockNumber || 'latest')
+
+        // Set from block transaction array for easy lookup
+        const transactionsSet = new Set(latestBlock.transactions)
```
anxolin, line 46: Is it really necesary to check on every block over the whole list of transactions?
It looks a bit of overkilling

src/hooks/useOrders.ts
```diff
@@ -1,34 +1,90 @@
...
     dispatch,
   ] = useGlobalState()
 
+  // Handle Pending Orders
+  // pending orders are saved in global app state as well as local storage
+  // users can refresh page after creating orders and see them after reload thanks for local storage
+  // uses this effect hook to:
+  //   1. listen to block updates and filter recently mined transactions based on their hash
+  //   2. remove any orders that may have been re-sent at a higher gas price
+  const currentPendingOrders = (userAddress && networkId && pendingOrders[networkId]?.[userAddress]) || EMPTY_ARRAY
```
anxolin, line 38: why is the main reason to delete te usePendingOrders hook and merge it with the useOrders?
I mean, the complexity of the hook grew and it look like isolated code

src/reducers-actions/pendingOrders.ts
```diff
@@ -1,52 +1,50 @@
 import { PendingTxObj } from 'api/exchange/ExchangeApi'
 import { Actions } from 'reducers-actions'
+import { toBN, setStorageItem } from 'utils'
+import { STORAGE_PENDING_ORDER_KEY } from 'const'
 
-export const enum ActionTypes {
-  SAVE_PENDING_ORDERS = 'SAVE_PENDING_ORDERS',
-  REMOVE_PENDING_ORDERS = 'REMOVE_PENDING_ORDERS',
-}
+export type ActionTypes = 'SAVE_PENDING_ORDERS' | 'REMOVE_PENDING_ORDERS'
```
anxolin, line 6: ❤️ me like!

src/utils/miscellaneous.ts
```diff
@@ -128,6 +128,13 @@ export async function silentPromise<T>(promise: Promise<T>, customMessage?: stri
   }
 }
 
+export function setStorageItem(key: string, data: unknown): void {
+  // localStorage API accepts only strings
+  // TODO: consider switching to localForage API (accepts all types)
```
anxolin, line 133: Maybe remove the TODO and make an issue if u want.
I don't think is a big issue though

src/const.ts
```diff
@@ -135,13 +135,13 @@ export const ETH_NODE_URL = ethNodeUrl
 
 export const STORAGE_KEY_LAST_PROVIDER = 'lastProvider'
 
-export const STORAGE_KEY_DISABLED_TOKENS_ADDRESSES = 'disabledTokens'
-
-export const GP_ORDER_TX_HASHES = {
-  1: 'GP_ORDER_TX_HASHES_1',
-  4: 'GP_ORDER_TX_HASHES_4',
+export const STORAGE_PENDING_ORDER_TX_HASHES = {
```
anxolin, line 138: not sure if this belong here, cause doesn't look like a general const that different components may use. It looks just like part of a reducers implementation

src/const.ts
```diff
@@ -135,13 +135,13 @@ export const ETH_NODE_URL = ethNodeUrl
...
+  1: 'STORAGE_PENDING_ORDER_TX_HASHES_1',
+  4: 'STORAGE_PENDING_ORDER_TX_HASHES_4',
 }
```
anxolin, line 142: Nice job, I just have concerns on how we check if a pending transaction was mined

src/hooks/useOrders.ts
W3stside: well, only if there are pending orders i guess :shrug:

src/hooks/useOrders.ts
W3stside: i agree it's bigger, maybe can extrapolate into own hook then implement hook in here, but there wasn't any scenario in my head where we'd only use `usePendingOrders` ina  component. Guess the former is a solution tho

W3stside: moved into own file and using in useOrders

PR #1094, opened 2020-06-10
useDataFilter hook

W3stside:
Changes what used to be this code inside `DepositWidget/index` for filtering into:
```ts
// ... DepositWidget Code ...
const [search, setSearch] = useState('')
  const [hideZeroBalances, setHideZeroBalances] = useState(false)

  const handleSearch = (e: React.ChangeEvent<HTMLInputElement>): void => setSearch(e.target.value)

  const handleHideZeroBalances = (e: React.ChangeEvent<HTMLInputElement>): void => setHideZeroBalances(e.target.checked)

  const { value: debouncedSearch, setImmediate: setDebouncedSearch } = useDebounce(search, 500)

  const clearFilters = (): void => {
    setSearch('')
    setDebouncedSearch('')
    setHideZeroBalances(false)
  }

  const [{ localTokens }] = useGlobalState()

  const filteredBalances = useMemo(() => {
    if ((!debouncedSearch && localTokens.disabled.size === 0) || !balances || balances.length === 0) return balances

    const searchTxt = debouncedSearch.trim().toLowerCase()

    return balances.filter(({ symbol, name, address }) => {
      if (localTokens.disabled.has(address)) return false

      if (searchTxt === '') return true

      return (
        symbol?.toLowerCase().includes(searchTxt) ||
        name?.toLowerCase().includes(searchTxt) ||
        address.toLowerCase().includes(searchTxt)
      )
    })
  }, [debouncedSearch, balances, localTokens.disabled])

  const displayedBalances = useMemo(() => {
    if (!hideZeroBalances || !filteredBalances || filteredBalances.length === 0) return filteredBalances
    return filteredBalances.filter(({ totalExchangeBalance, pendingWithdraw, walletBalance }) => {
      return !totalExchangeBalance.isZero() || !pendingWithdraw.amount.isZero() || !walletBalance.isZero()
    })
  }, [hideZeroBalances, filteredBalances])
```
into
```ts
// ... DepositWidget Code ...
const [{ localTokens }] = useGlobalState()

  const { filteredData: filteredBalances, search, handleSearch } = useDataFilter({
    data: balances,
    filterFnFactory: customFilterFnFactory((params: TokenDetails) => localTokens.disabled.has(params.address)),
    customStopCheck: () => localTokens.disabled.size === 0,
  })

  const {
    filteredData: displayedBalances,
    showFilter: hideZeroBalances,
    handleToggleFilter: handleHideZeroBalances,
    clearFilters,
  } = useDataFilter({
    data: filteredBalances,
    filterFnFactory: () => customHideZeroFilterFn,
    isSearchFilter: false,
  })
// ... rest of DepositWidget Code ...
```

From the `useDataFilter` hook

Will reuuse this logic inside `OrdersWidget` so made sense to move into own file

Though I am not sure if this is the cleanest of even a necessary implementation so feedback obv welcome


src/components/DepositWidget/index.tsx
```diff
@@ -352,48 +385,24 @@ const BalancesDisplay: React.FC<BalanceDisplayProps> = ({
...
-      )
-    })
-  }, [debouncedSearch, balances, localTokens.disabled])
-
-  const displayedBalances = useMemo(() => {
-    if (!hideZeroBalances || !filteredBalances || filteredBalances.length === 0) return filteredBalances
+  const { filteredData: filteredBalances, search, handleSearch } = useDataFilter({
+    data: balances,
+    filterFnFactory: customFilterFnFactory((params: TokenDetails) => localTokens.disabled.has(params.address)),
+    customStopCheck: () => localTokens.disabled.size === 0,
```
Velenir, line 393: Can be memoized, so that they don't cause rerenders in [useDataFilter::filteredData](https://github.com/gnosis/dex-react/blob/8ad03b8173febe013d7eda0934fee473c608cd48/src/hooks/useDataFilter.tsx#L53)
```ts
const dataFilterFuncs = useMemo(() => ({
  filterFnFactory: customFilterFnFactory((params: TokenDetails) => localTokens.disabled.has(params.address)),
  customStopCheck: () => localTokens.disabled.size === 0,
}), [localTokens.disabled])
```
or something

src/components/DepositWidget/index.tsx
```diff
@@ -352,48 +385,24 @@ const BalancesDisplay: React.FC<BalanceDisplayProps> = ({
...
-    })
-  }, [hideZeroBalances, filteredBalances])
+  const {
+    filteredData: displayedBalances,
+    showFilter: hideZeroBalances,
+    handleToggleFilter: handleHideZeroBalances,
+    clearFilters,
+  } = useDataFilter({
+    data: filteredBalances,
+    filterFnFactory: () => customHideZeroFilterFn,
```
Velenir, line 403: memoize please
Can even just define outside the component

src/hooks/useDataFilter.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [showFilter, setShowFilter] = useState(isSearchFilter)
+  const { value: debouncedSearch, setImmediate: setDebouncedSearch } = useDebounce(search, searchDebounceTime)
+
+  const handleSearch = (e: React.ChangeEvent<HTMLInputElement>): void => setSearch(e.target.value)
+  const handleToggleFilter = (): void => setShowFilter(state => !state)
+  const clearFilters = (): void => {
+    setSearch('')
+    setDebouncedSearch('')
+    setShowFilter(false)
+  }
```
Velenir, line 39: Memoize together, as setState are stable refs.
It's nicer to have these functions as const when passing to components later

src/components/DepositWidget/index.tsx
```diff
@@ -335,6 +335,39 @@ interface BalanceDisplayProps extends TokenLocalState {
   ): Promise<void>
 }
 
+const customFilterFnFactory = (customStopCheck: (...any: unknown[]) => boolean) => (searchTxt: string) => ({
+  symbol,
+  name,
+  address,
+}: TokenBalanceDetails): boolean => {
+  if (
+    customStopCheck &&
```
alfetopito, line 344: Why do you need this check? From what I understand this param is not optional, so it should always be true, no?

src/components/DepositWidget/index.tsx
```diff
@@ -335,6 +335,39 @@ interface BalanceDisplayProps extends TokenLocalState {
   ): Promise<void>
 }
 
+const customFilterFnFactory = (customStopCheck: (...any: unknown[]) => boolean) => (searchTxt: string) => ({
+  symbol,
+  name,
+  address,
+}: TokenBalanceDetails): boolean => {
```
alfetopito, line 342: Shit, another brain freeze. WTH is going on here?

1. function `customFilterFnFactory` takes as argument a function that returns a boolean.
2. then `customFilterFnFactory` returns a function that:
    1. takes a string as param
    1. then returns a function which:
        1. takes `TokensBalanceDetails` as params
        1. then returns a boolean

Did I get this right?

alfetopito: I don't think I got this right. The function body only returns `boolean`s.  :thinking: :question: :confused:

src/hooks/useDataFilter.tsx
W3stside: you mean like:
```ts
const handlers = useMemo(() => {
    const handleSearch = (e: React.ChangeEvent<HTMLInputElement>): void => setSearch(e.target.value)
    const handleToggleFilter = (): void => setShowFilter(state => !state)
    const clearFilters = (): void => {
      setSearch('')
      setDebouncedSearch('')
      setShowFilter(false)
    }

    return {
      handleSearch,
      handleToggleFilter,
      clearFilters,
    }
  }, [setDebouncedSearch])
```
? or  just the clear filters

W3stside: what do you mean by easier to pass as `const`s?

src/components/DepositWidget/index.tsx
W3stside: you caught sth here actually, it should just take the `localTokens` as the first curried param as it's defined outside the scope. So basically you're right, but it's `localTokens` => `searchText` => `filterFn`

src/components/DepositWidget/index.tsx
W3stside: good catch, this is pointless what im doing, fixed it

src/hooks/useDataFilter.tsx
Velenir: exactly like you described, `handlers` 👍 
`const` worded awkwardly, I meant stable references from `handlers`

W3stside: thanks man

Issue #1029, opened 2020-05-20
Show filled orders (trades) - Basic implementation trades


anxolin:
* Implement proposal: #1028
* HTML/CSS Markup 
* Get trade information (not reverts)
* Implement hooks/reducers/persistance logics

PR #1096, opened 2020-06-10
1029/styling take 2

alfetopito:
Part of #1029 

## Update Jun 16

- Added padding to left of table
- Merged price columns and Amount/Received columns
- Renamed `Amount` and `Received` columns respectively to `Sold` and `Bought`. -- Anyone like this?
- Fixed bug reported by Federico where on account change trades were not updated.

screenshot_2020-06-16_10-12-50

## Update Jun 12

- Applied awesome mobile friendly update by David
- Added it to `/trade` page under a new tab: `Fills`
- Still available under `/trades` with a slightly different look, although there's no direct link to it.

screenshot_2020-06-12_11-05-27

### Notes:
- :warning: location not yet final :warning: 
- Ui touches still needed.
- Feedbacks are ~welcome~ expected ;)
- Moving `Liquidity` tab to Liquidity page out of the scope of this PR.

---

## Original

Working to make it look better.

Throw suggestions at me. What could be better?

So far:
- Added column gap
- Added row gap

### Before:

screenshot_2020-06-10_15-13-44

### After:

screenshot_2020-06-10_15-12-32


anxolin:
Looks better! Would make the table headers coherent with the balance table.

Maybe we could start working on seeing where to show this. What are the options?
* Standalone trades page with his own menu?
* New tab: ACTIVE, FILLS, LIQUIDITY, CLOSED?
* New Tab and move liquidity to Liquidity part? ACTIVE, FILLS, CLOSED (only closed normal orders, and fill normal orders)
* Or maybe sth like this is a simple one:
image

@W3stside @Velenir @alfetopito what do you think.

David, can you add the filters here too?

Velenir:
> * Standalone trades page with his own menu?
> * New tab: ACTIVE, FILLS, LIQUIDITY, CLOSED?
> * New Tab and move liquidity to Liquidity part? ACTIVE, FILLS, CLOSED (only closed normal orders, and fill normal orders)
> * Or maybe sth like this is a simple one:

Interface is already busy as it is, so something without adding a new tab.
@anxolin's suggestion is good
or something like a new tab in an accordion wouldn't take much extra space

<sub><sub>I'm not saying let's use jQuery</sub></sub>

alfetopito:
>     * New Tab and move liquidity to Liquidity part? ACTIVE, FILLS, CLOSED (only closed normal orders, and fill normal orders)

I vote for :point_up: 

Although it would make sense to also not show liquidity trades here and move them to liquidity tab as well.

Velenir:
ezgif-5-d1c9d59ebb53
Not a fan of how Fills tab makes the whole widget change width, but not much you can do about it. Save for adding a horizontal scroll, which is ugly.

src/components/OrdersWidget/index.tsx
```diff
@@ -101,12 +103,15 @@ const OrdersWidget: React.FC = () => {
...
+
   // syntactic sugar
   const { displayedOrders, displayedPendingOrders, markedForDeletion } = useMemo(
     () => ({
-      displayedOrders: filteredOrders[selectedTab].orders,
-      displayedPendingOrders: filteredOrders[selectedTab].pendingOrders,
-      markedForDeletion: filteredOrders[selectedTab].markedForDeletion,
+      displayedOrders: selectedTab === 'fills' ? [] : filteredOrders[selectedTab].orders,
+      displayedPendingOrders: selectedTab === 'fills' ? [] : filteredOrders[selectedTab].pendingOrders,
+      markedForDeletion: selectedTab === 'fills' ? new Set<string>() : filteredOrders[selectedTab].markedForDeletion,
```
Velenir, line 114: I wonder if it could be a good idea to have `emptyArray, emptySet` here fr stable references, like @W3stside did somewhere else.
Save a couple useMemo calls
But ugly

W3stside:
> ezgif-5-d1c9d59ebb53
> Not a fan of how Fills tab makes the whole widget change width, but not much you can do about it. Save for adding a horizontal scroll, which is ugly.

I would just make the default width the max width of the `Filled` tab.

Maybe we can combine `Fill Price/Limit Price` like we do for filled/total? meh

alfetopito:
> > ezgif-5-d1c9d59ebb53
> > Not a fan of how Fills tab makes the whole widget change width, but not much you can do about it. Save for adding a horizontal scroll, which is ugly.
> 
> I would just make the default width the max width of the `Filled` tab.
> 
> Maybe we can combine `Fill Price/Limit Price` like we do for filled/total? meh

I like it!

So much that I went all in and combined **Amount** and **Received** columns as well.

Too far?
screenshot_2020-06-15_15-28-47

(Just need to make the headers look better. Too close together IMO)

W3stside:
> > > ezgif-5-d1c9d59ebb53
> > > Not a fan of how Fills tab makes the whole widget change width, but not much you can do about it. Save for adding a horizontal scroll, which is ugly.
> > 
> > 
> > I would just make the default width the max width of the `Filled` tab.
> > Maybe we can combine `Fill Price/Limit Price` like we do for filled/total? meh
> 
> I like it!
> 
> So much that I went all in and combined **Amount** and **Received** columns as well.
> 
> Too far?
> screenshot_2020-06-15_15-28-47
> 
> (Just need to make the headers look better. Too close together IMO)

Looks nice! I would not stack `Amount Received` tho - looks like it's saying the `Amount received` haha. Maybe just `Amount / Received`

alfetopito:
> At some point it might make sense to move out the liquidity tab. Feels like a lot of tabs and we have a page just for that.
>
> Although maybe if it's not trivial, looks like we won't have capacity.

Indeed, I do mention that somewhere in the PR description.
I'm working my way towards that with https://github.com/gnosis/dex-react/pull/1105 and #1103
Filtering liquidity orders and trades out is easy. The "hard" part is the UI part for adding them to the Liquidity page.
 
> Should the FILLS be the second tab? In order of importance ACTIVE, FILLS, LIQUIDITY, CLOSED

You got it.

W3stside:
Some notes:

1. the `thead` is too "tall" because of the `SVG` pushing the outer container. (Compare it to the other tabs)
  a. An easy solution is to set the `min-height` of the `thead > tr` to `min-height: 6rem` or sth. 
  b. Other solution is to use absolute position of the svg like:
    ```ts
  const CsvButtonContainer = styled.div`
  display: flex;
  justify-content: space-around;
  align-items: center;

  a {
    > svg {
      position: absolute;
      top: 23%;
      right: 4%;
    }
  }
`
    ```
3. `thead` headers need some breathing room on left, notable the `date` column:
  a. ```ts
const TradesCardTable = styled(CardTable)`
  tr {
    min-height: 4rem;
    > th:first-child,
    > td:first-child {
      padding-left: 2rem;
    }
  }
`
```

PR #1097, opened 2020-06-11
1032/filter tokens orders widget

### LINKED ISSUE #1032: Filter orders/trades
Filter orders / trades.

At least by status, token

W3stside:
Closes #1032 

Dependent on #1094 

To test: head to `/orders` or `/trade` page and filter tokens by address/symbol/name

`useOrders`:

- Changes hook to call for token details inside `useEffect` and replace `OrderRow`'s previous responsibility on this
- orders returned now adhere to `DetailedAuctionElement` interface which is `AuctionElement` + `TokenDetails` for both `buyToken` and `sellToken`

Filter at work - `/orders`:
Screenshot from 2020-06-11 11-54-44

`/trade`:
Screenshot from 2020-06-11 11-55-49

### DIFF
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -1,10 +1,10 @@
 import BN from 'bn.js'
 import { PastEventOptions } from 'web3-eth-contract'
 
-import { assert, BatchExchangeEvents, TokenDetails } from '@gnosis.pm/dex-js'
+import { assert, BatchExchangeEvents } from '@gnosis.pm/dex-js'
 
 import { DepositApiImpl, DepositApi, DepositApiDependencies } from 'api/deposit/DepositApi'
-import { Receipt, WithTxOptionalParams } from 'types'
+import { Receipt, TokenDetails, WithTxOptionalParams } from 'types'
 import { logDebug } from 'utils'
 import { decodeAuctionElements, decodeOrder } from './utils/decodeAuctionElements'
 import { DEFAULT_ORDERS_PAGE_SIZE, LIMIT_EXCEEDED_ERROR_CODE } from 'const'
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -104,6 +104,11 @@ export interface ExchangeApi extends DepositApi {
   cancelOrders(params: CancelOrdersParams): Promise<Receipt>
 }
 
+export interface DetailedAuctionElement extends AuctionElement {
+  buyToken: TokenDetails | null
+  sellToken: TokenDetails | null
+}
+
 export interface AuctionElement extends Order {
   user: string
   sellTokenBalance: BN
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -257,6 +257,10 @@ const OrdersWidget: React.FC = () => {
                   // onChange={handleSearch}
                 />
               </label>
+              <label className="balances-hideZero">
+                <input type="checkbox" /* checked={hideZeroBalances} onChange={handleHideZeroBalances} */ />
+                <b>Hide untouched orders</b>
+              </label>
             </BalanceTools>
             <div className="infoContainer">
               <div className="countContainer">
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -97,8 +97,8 @@ function checkTokenAgainstSearch(token: TokenDetails | null, searchText: string)
   )
 }
 
-const filterOrdersFn = (searchTxt: string) => ({ id, buyToken, sellToken }: DetailedAuctionElement): boolean => {
-  if (searchTxt === '') return true
+const filterOrdersFn = (searchTxt: string) => ({ id, buyToken, sellToken }: DetailedAuctionElement): boolean | null => {
+  if (searchTxt === '') return null
 
   return (
     !!id.includes(searchTxt) ||
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -14,12 +14,12 @@ import { DEFAULT_ORDERS_SORTABLE_TOPIC } from 'const'
 // Hooks
 import { useOrders } from 'hooks/useOrders'
 import useSafeState from 'hooks/useSafeState'
-import usePendingOrders from 'hooks/usePendingOrders'
+import usePendingOrders, { DetailedPendingOrder } from 'hooks/usePendingOrders'
 import { useWalletConnection } from 'hooks/useWalletConnection'
 import useSortByTopic from 'hooks/useSortByTopic'
 
 // Api
-import { AuctionElement, PendingTxObj } from 'api/exchange/ExchangeApi'
+import { DetailedAuctionElement } from 'api/exchange/ExchangeApi'
 
 // Components
 import { ConnectWalletBanner } from 'components/ConnectWalletBanner'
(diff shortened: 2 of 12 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1097--dexreact.review.gnosisdev.com

W3stside:
> Nice, although the tooltip should show when you filter also by text, right?
> 
> See how it's 1 result in both screenshots, but only one shows the tooltip
> 
> Screenshot at Jun 11 13-15-57
> 
> Screenshot at Jun 11 13-16-12

I didn't add it here since users will look up 1 token at a time ussually? or no? i can re-add it

src/hooks/useOrders.ts
```diff
@@ -53,11 +54,21 @@ export function useOrders(): Result {
...
+        const { orders: ordersPreTokenDetails, nextIndex } = await exchangeApi.getOrdersPaginated({
+          userAddress,
+          networkId,
+          offset,
+        })
 
+        const ordersPromises = ordersPreTokenDetails.map(async order => ({
+          ...order,
+          sellToken: await getTokenFromExchangeById({ tokenId: order.sellTokenId, networkId }),
+          buyToken: await getTokenFromExchangeById({ tokenId: order.buyTokenId, networkId }),
```
Velenir, line 66: Promise.all?

src/hooks/usePendingOrders.ts
```diff
@@ -3,18 +3,34 @@ import { useEffect } from 'react'
...
   useEffect(() => {
-    if (userAddress && networkId) {
-      setPendingOrders(pendingOrdersGlobal[networkId][userAddress] || [])
+    async function getDetailedPendingOrders(): Promise<void> {
+      if (userAddress && networkId) {
+        const pendingOrdersByNetwork = pendingOrdersGlobal[networkId][userAddress] || []
+        const ordersPromises = pendingOrdersByNetwork.map(async order => ({
+          ...order,
+          sellToken: await getTokenFromExchangeById({ tokenId: order.sellTokenId, networkId }),
+          buyToken: await getTokenFromExchangeById({ tokenId: order.buyTokenId, networkId }),
```
Velenir, line 26: Promise.all?

anxolin:
Additional comment (discussed in the chat):

Can we also include `orderId` as part of the filter. It's just super convenient. 
Also what leandro did in the trades with the HTML5 data to see the order id would be good

src/components/OrdersWidget/index.tsx
```diff
@@ -245,6 +274,36 @@ const OrdersWidget: React.FC = () => {
...
+                .balances-searchTokens { 
+                    height: 3.6rem; 
+                    margin: 0.8rem; 
+                    width: 100%; 
+                    
+                    > input { 
+                      width: 100%; 
+                    } 
+                }`
+              }
```
alfetopito, line 294: Wouldn't be better to have a different widget instead of inlining the style, even if as a prop?

src/api/exchange/ExchangeApi.ts
```diff
@@ -104,6 +104,11 @@ export interface ExchangeApi extends DepositApi {
   cancelOrders(params: CancelOrdersParams): Promise<Receipt>
 }
 
+export interface DetailedAuctionElement extends AuctionElement {
+  buyToken: TokenDetails | null
+  sellToken: TokenDetails | null
```
alfetopito, line 109: What about:

```suggestion
  buyToken?: TokenDetails
  sellToken?: TokenDetails
```

Unless you effectively need `null` for some reason.

src/hooks/useOrders.ts
W3stside: Like this? or

```ts
        const ordersPromises = ordersPreTokenDetails.map(async order => {
          const [sellToken, buyToken] = await Promise.all([
            getTokenFromExchangeById({ tokenId: order.sellTokenId, networkId }),
            getTokenFromExchangeById({ tokenId: order.buyTokenId, networkId }),
          ])
          return {
            ...order,
            sellToken,
            buyToken,
          }
        })
```

src/api/exchange/ExchangeApi.ts
W3stside: mm the `getTokenFromExchangeById` fn can return `null`, any other way?

src/components/OrdersWidget/index.tsx
W3stside: oki

src/api/exchange/ExchangeApi.ts
alfetopito: Then keep it as it is. `null` is fine.

alfetopito:
I still see it upside down :/

screenshot_2020-06-12_08-24-49

W3stside:
@alfetopito fixed it, wasn't moz related but dumb (me) related, good catch!

src/components/OrdersWidget/index.tsx
```diff
@@ -83,15 +88,41 @@ function classifyOrders(
...
+  if (!token) return false
+  return (
+    token?.symbol?.toLowerCase().includes(searchText) ||
+    token?.name?.toLowerCase().includes(searchText) ||
+    token?.address.toLowerCase().includes(searchText)
+  )
+}
+
+const filterOrdersFn = (searchTxt: string) => ({ id, buyToken, sellToken }: DetailedAuctionElement): boolean => {
+  if (searchTxt === '') return true
```
Velenir, line 101: Should this check be here as opposed to [useDataFilter](https://github.com/gnosis/dex-react/blob/ae8ab0c521/src/hooks/useDataFilter.tsx#L64-L65)?
By doing `data.filter(() => true)` we essentially clone array and mutate refrence.

src/components/OrdersWidget/index.tsx
```diff
@@ -83,15 +88,41 @@ function classifyOrders(
...
+
+  return (
+    !!id.includes(searchTxt) ||
+    checkTokenAgainstSearch(buyToken as TokenDetails, searchTxt) ||
+    checkTokenAgainstSearch(sellToken as TokenDetails, searchTxt)
```
Velenir, line 106: Would
```suggestion
    (buyToken && checkTokenAgainstSearch(buyToken, searchTxt)) ||
    (sellToken && checkTokenAgainstSearch(sellToken, searchTxt))
```
work? Eliminating type casting

src/components/OrdersWidget/index.tsx
W3stside: what if we need a custom `searchTxt` condition? can I use the built in 3rd `array` param of the filter function to use the reference somewhere?

W3stside: Also i don't really understand how that reference mutation happens, could you expand?

Velenir: > can I use the built in 3rd array
I don't see how

Velenir: if you need a `if(searchTxt.includes(*)` condition, then by all means include in `filterOrdersFn`
But as empty `searchTxt` is essentially `no filtering`, you can opt out early
```ts
const filterOrdersFn = (searchTxt: string) => {
  if (searchTxt === '') return null
  return ({ id, buyToken, sellToken }: DetailedAuctionElement): boolean => {
```
and then
```js
const customFilterFn = filterFnFactory(searchTxt)
if (!customFilterFn) return unfiltered_data
```

Velenir: > Also i don't really understand how that reference mutation happens, could you expand?
```js
const filteredData = useMemo(() => {
  return data.filter(() => true)
})
/// filteredData same as data, but
filteredData !== data


// optOut = true
const filteredData = useMemo(() => {
  return optOut ? data : data.filter(fn)
})
/// filteredData same as data and
filteredData === data
// reference stays the same
if we have useMemo(*, [filteredDate]) it won't rerurn
```

alfetopito:
> @alfetopito fixed it, wasn't moz related but dumb (me) related, good catch!

Still there. Do you have anything local pending `git push`?

W3stside:
> > @alfetopito fixed it, wasn't moz related but dumb (me) related, good catch!
> 
> Still there. Do you have anything local pending `git push`?

indeed ;)

src/components/OrdersWidget/index.tsx
W3stside: thanks man

PR #1099, opened 2020-06-11
1029/local storage

alfetopito:
Part of #1029 

Storing trades to / loading from localStorage.
Storing both networks on globalStorage at the same time.

### DIFF
--- a/src/hooks/useTrades.ts
+++ b/src/hooks/useTrades.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useMemo } from 'react'
 
 import { BATCH_TIME } from '@gnosis.pm/dex-js'
 
@@ -7,7 +7,7 @@ import { web3 } from 'api'
 
 import { getTradesAndTradeReversions } from 'services'
 
-import { appendTrades, updateLastCheckedBlock, overwriteTrades } from 'reducers-actions/trades'
+import { appendTrades, updateLastCheckedBlock } from 'reducers-actions/trades'
 
 import { useWalletConnection } from 'hooks/useWalletConnection'
 import useGlobalState from 'hooks/useGlobalState'
@@ -16,21 +16,20 @@ import { BATCH_TIME_IN_MS } from 'const'
 
 import { getTimeRemainingInBatch } from 'utils'
 
-// TODO: (on global state) store trades to local storage
 export function useTrades(): Trade[] {
   const [
     {
-      trades: { trades, lastCheckedBlock },
+      trades: globalStateTrades,
       orders: { orders },
     },
     dispatch,
   ] = useGlobalState()
   const { userAddress, networkId } = useWalletConnection()
 
-  useEffect(() => {
-    // Resetting trades on network change
-    dispatch(overwriteTrades({ trades: [], reverts: [], lastCheckedBlock: undefined }))
-  }, [dispatch, networkId, userAddress])
+  const { lastCheckedBlock, trades } = useMemo(
+    () => (networkId ? globalStateTrades[networkId] : { lastCheckedBlock: undefined, trades: [] }),
+    [globalStateTrades, networkId],
+  )
 
   useEffect(() => {
     // Flow control. Cancel query/state update on unmount
@@ -64,8 +63,8 @@ export function useTrades(): Trade[] {
 
         dispatch(
           newTrades.length > 0 || reverts.length > 0
-            ? appendTrades({ trades: newTrades, reverts, lastCheckedBlock: toBlock })
-            : updateLastCheckedBlock(toBlock),
+            ? appendTrades({ trades: newTrades, reverts, lastCheckedBlock: toBlock, networkId })
+            : updateLastCheckedBlock(toBlock, networkId),
         )
       }
     }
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1099--dexreact.review.gnosisdev.com

src/reducers-actions/trades.ts
```diff
@@ -161,28 +186,97 @@ function applyRevertsToTrades(
...
+
+// ******** SIDE EFFECT
+
+export async function sideEffect(state: TradesState, action: ReducerActionType): Promise<void> {
+  switch (action.type) {
+    case 'APPEND_TRADES':
+    case 'OVERWRITE_TRADES':
+    case 'UPDATE_BLOCK':
+      setStorageItem(TRADES_LOCAL_STORAGE_KEY, state)
+  }
```
W3stside, line 235: consider adding a default here in case

src/reducers-actions/trades.ts
```diff
@@ -14,39 +23,55 @@ interface WithReverts {
...
-// TODO: store to/load from localStorage
-export const INITIAL_TRADES_STATE = { trades: [], pendingTrades: new Map<string, Trade[]>() }
-
 function buildTradeRevertKey(batchId: number, orderId: string): string {
   return batchId + '|' + orderId
 }
 
+// ******** HELPERS
+
 function groupByRevertKey<T extends EventWithBlockInfo>(list: T[], initial?: Map<string, T[]>): Map<string, T[]> {
```
W3stside, line 75: idk why buy this function just screams Leandro, and not in any bad way, just is maybe sth I've seen you handling/man-handling/coding out for this app a alot in other PRs. I have these kinds for Anxo and Dima, now I found your fav type! ;)

src/reducers-actions/trades.ts
```diff
@@ -161,28 +186,97 @@ function applyRevertsToTrades(
...
+
+      return { ...state, [networkId]: { ...state[networkId], lastCheckedBlock } }
     }
     default: {
       return state
     }
   }
 }
+
+// TODO: use the one from David once his changes are merged https://github.com/gnosis/dex-react/pull/1091
```
anxolin, line 219: +1 👌

src/reducers-actions/trades.ts
```diff
@@ -161,28 +186,97 @@ function applyRevertsToTrades(
...
+// ******** INITIAL STATE / LOCAL STORAGE
+
+const INITIAL_TRADES_STATE_SINGLE_NETWORK = { trades: [], pendingTrades: new Map<string, Trade[]>() }
+
+/**
+ * Custom json parser for BN and BigNumber values.
+ * Since there's no context on what is BN/BigNumber,
+ * we have to keep track of keys and parse accordingly
+ */
+function reviver(key: string, value: unknown): unknown {
```
anxolin, line 247: there's no way of typing this more?

I'm fine with this solution if there's not. 
Other alternative would be to define a Stringify frieendly type version of the trade, where we replace the BN and BigNumbers by strings, so instead of the reviver we would have a a couple of convertertes from "Trade <--> TradeStrigifyable"

src/reducers-actions/trades.ts
alfetopito: Wait, what type? Now I'm curious. Do I do that a lot? :thinking:

src/reducers-actions/trades.ts
alfetopito: > there's no way of typing this more?

There might be, but then I'll have to map all possible JSON serialized types to all JS deserialized types. Which feels like unnecessary to me.
Unless @Velenir has a better idea? (Of course he has)

> Other alternative would be to define a Stringify frieendly type version of the trade, where we replace the BN and BigNumbers by strings

I considered a slightly different approach: 
1. When serializing JSON, add a "watermark" to identify BN/BigNumber, such as:
    ```
    if (value is BN) return `BN|${value.toString()}`
    ```
2. On deserialization, we can have a more generic `reviver`, such as:
    ```
    if (typeof value === 'string' && /^BN|/.test(value)) return toBn(value)
    ```

I decided not to because @W3stside had a TODO for using a different type of storage:
`// TODO: consider switching to localForage API (accepts all types)`

So that might address the issue in another way.

src/reducers-actions/trades.ts
alfetopito: Actually, the function returns `void`, so having a default makes no difference.

src/reducers-actions/trades.ts
Velenir: > there's no way of typing this more?

Don't think there is. JSON.parse, not a generic, `string` in `any` out, can only cast or assign to known type, as `TradesState` here. Same for `reviver`

Can't do a string subtype `"BN:*"`

src/hooks/useTrades.ts
```diff
@@ -16,21 +16,20 @@ import { BATCH_TIME_IN_MS } from 'const'
...
   const { userAddress, networkId } = useWalletConnection()
 
-  useEffect(() => {
-    // Resetting trades on network change
-    dispatch(overwriteTrades({ trades: [], reverts: [], lastCheckedBlock: undefined }))
-  }, [dispatch, networkId, userAddress])
+  const { lastCheckedBlock, trades } = useMemo(
+    () => (networkId ? globalStateTrades[networkId] : { lastCheckedBlock: undefined, trades: [] }),
+    [globalStateTrades, networkId],
+  )
```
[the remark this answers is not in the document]
alfetopito: I accept your nitpick

PR #1104, opened 2020-06-12
`Filled` status takes precedence over `Expired`

### LINKED ISSUE #1093: Label filled orders as filled (even after expiry)
It appears, that expired orders that were “filled” appear as “expired” in the "Closed" tab after their expiry date. I would think that it is much more intuitive to override the “Expired” label in favour of “Filled” when this happens.

Screenshot 2020-06-10 at 11 19 07 AM

alfetopito:
Closes #1093

screenshot_2020-06-12_15-37-05


gnosis-info:
Travis automatic deployment:
 https://pr1104--dexreact.review.gnosisdev.com

PR #1109, opened 2020-06-16
Filter pending orders

W3stside:
Filtering was ignoring pending orders - now considers them albeit they may disappear quickly - could be useful if there is a long tx time for pending and user wants to search

Can be tested by submitting an order and setting a very low gas


gnosis-info:
Travis automatic deployment:
 https://pr1109--dexreact.review.gnosisdev.com

Issue #932, opened 2020-04-17
Proposal: Increase awareness of mechanics


anxolin:
The proposal is to show the current batch number, and the state:
* min 0-4: Open for solutions
* min 4-5: Closed

This is a small mock up (a proper design would make it look better 👨‍🎨)

![image](https://user-images.githubusercontent.com/2352112/79547180-d6c4f000-8093-11ea-8d1b-660531df491d.png)

![image](https://user-images.githubusercontent.com/2352112/79547541-5bb00980-8094-11ea-845e-c44e07ba0816.png)

**Not part of this issue, but worth mentioning**

In the future, in this part, we could link to the block explorer to see info about this batch or past batches.

Once a batch is closed, if we have the batch explorer, it's nice to be able to see the:
* Trades
* Closing prices
* Graph representing all the trades

PR #1110, opened 2020-06-16
Added batch id to header

alfetopito:
What do you think?

screenshot_2020-06-16_08-58-31

### DIFF
--- a/src/components/Layout/Header/index.tsx
+++ b/src/components/Layout/Header/index.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 
 // utils, const
-import { formatSeconds } from 'utils'
+import { formatSeconds, dateToBatchId } from 'utils'
 import { MEDIA } from 'const'
 
 // Components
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1110--dexreact.review.gnosisdev.com

anxolin:
This reminds me of something https://github.com/gnosis/dex-react/issues/932

alfetopito:
> Would it make sense to add a title and show as a title/tooltip the time `From 12:30 to 12:35`?

Seems redundant with the countdown.

> What about showing when it's closed/open for solutions?

Also redundant. The orders have a countdown too.

What I see as good to have is a FAQ entry? Or a tooltip.

josojo:
This is cool. Thanks for doing it so proactively!

I would prefer it if we would see this information of the batch number even when users have opened the orderbook modal. But probably, this would imply bigger design changes, right? It's not critical, just another nice to have.

alfetopito:
> This is cool. Thanks for doing it so proactively!
> 
> I would prefer it if we would see this information of the batch number even when users have opened the orderbook modal. But probably, this would imply bigger design changes, right? It's not critical, just another nice to have.

That would require to either change the modal size so the countdown is always visible or insert the batch time into the orderbook modal, and refresh it at every orderbook update.

Does require a bit more work and thinking besides this 2min fix I did here, hehe. Not THAT much, just a bit more I'm not sure we want to invest right now.

PR #1112, opened 2020-06-17
1029.2/david style take

W3stside:
Merges into #1096 

Optimises CSS

`/trades` formatting was a bit strange due to `CardWidget`s default CSS which was optimised for `Balances` page
  a. edited to better fit 

Added `isTab` prop to `TradesWidget` to optimise column sizes for both formats

Standardised `thead` size across all tables

# Update 19.06
**Waterfall PRs**

1. adds filtering for `trades` aka Fills
2. uses inverse price for fills widget/tab
3. adds market filtering e.g `WETH/DAI` or `WETH-DAI` or `WETHDAI` and inverse (inverse search only for `Active/Liquidity/Closed` orders, not for fills
4. css fixes on table styling etc

### DIFF
--- a/src/components/DepositWidget/index.tsx
+++ b/src/components/DepositWidget/index.tsx
@@ -57,9 +57,13 @@ const BalancesWidget = styled(CardWidgetWrapper)`
         flex-flow: row nowrap;
       }
 
-      &[data-label='Token'] > div > b {
-        display: block;
-        color: var(--color-text-primary);
+      &[data-label='Token'] > div {
+        word-break: break-word;
+
+        > b {
+          display: block;
+          color: var(--color-text-primary);
+        }
       }
     }
   }
@@ -224,7 +228,7 @@ const BalancesDisplay: React.FC<BalanceDisplayProps> = ({
   const { modalProps, toggleModal } = useManageTokens()
 
   return (
-    <BalancesWidget $columns="repeat(auto-fit, minmax(5rem, 1fr));">
+    <BalancesWidget $columns="minmax(13.2rem,0.8fr) repeat(2,minmax(10rem,1fr)) minmax(14.5rem, 1fr) minmax(13.8rem, 0.8fr)">
       <FilterTools
         resultName="tokens"
         searchValue={search}
@@ -243,7 +247,7 @@ const BalancesDisplay: React.FC<BalanceDisplayProps> = ({
       {error ? (
         <ErrorMsg title="oops..." message="Something happened while loading the balances" />
       ) : (
-        <CardTable className="balancesOverview">
+        <CardTable className="balancesOverview" $gap="0 1rem">
           <thead>
             <tr>
               <th>Token</th>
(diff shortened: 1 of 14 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1112--dexreact.review.gnosisdev.com

alfetopito:
Nice, thanks for the fixes!

One thing though, the low balance indicator is fucked up:
screenshot_2020-06-17_10-07-49

Should we do `flex-direction:column;` for all `td`s?

W3stside:
> Nice, thanks for the fixes!
> 
> One thing though, the low balance indicator is fucked up:
> screenshot_2020-06-17_10-07-49
> 
> Should we do `flex-direction:column;` for all `td`s?

@alfetopito just for this col is good, done in latest commit

src/components/DepositWidget/index.tsx
```diff
@@ -37,17 +37,30 @@ interface WithdrawState {
...
-  ${CardTable} > tbody > tr:not([class^="Card__CardRowDrawer"]) > td {
-    &[data-label='Token'] {
-      font-family: var(--font-default);
-      letter-spacing: 0;
-      line-height: 1.2;
-      flex-flow: row nowrap;
+  ${CardTable} {
+    > thead,
+    > tbody {
+      > tr:not([class^='Card__CardRowDrawer']) {
```
Velenir, line 43: Is `Card__CardRowDrawer` generated by styled-components? I'm not sure it won't change at some point without us noticing.
Would be better to use `> tr:not([class^='${CardRowDrawer or whatever}'])` if possible

W3stside: i think this was maybe when Michel was changing CSS, you're right. good catch

src/components/Layout/Card/Card.tsx
```diff
@@ -102,7 +102,7 @@ export const CardDrawer = React.forwardRef<HTMLTableRowElement, CardDrawerProps>
   ref,
 ) {
   return (
-    <CardRowDrawer ref={ref}>
+    <CardRowDrawer className=".cardRowDrawer" ref={ref}>
```
Velenir, line 105: Typo
```suggestion
    <CardRowDrawer className="cardRowDrawer" ref={ref}>
```

W3stside:
> Looking great, everything working nice 
> 
> Only one minor thing that I noticed.
> 
> When I filter on active tab, it says: `Filter: Showing N result`
> On fills tab, it says: `Filter: Showing N trades`
> On liquidity/closed tab: it says: `Filter: Showing N orders`
> 
> It's odd that active text is different from liquidity and closed, since both are essentially the same type of objects.

Is it because you are only getting 1 result? I intentionally made it show `result` when `filter array length === 1` otherwise it uses the passed data title prop

anxolin:
Also getting this error:

image

specifically in this line:
image

alfetopito:
> Also getting this error:
> 
> image
> 
> specifically in this line:
> image

That's the same issue you reported last week. It's not merged, and I haven't yet had the time to work on it again.
I hope I can do so today :)

alfetopito:
It seems this is the only pending issue https://github.com/gnosis/dex-react/pull/1112#issuecomment-648251632

Once that's addressed we can merge the waterfalls :droplet: :fallen_leaf:

anxolin:
Approved since @alfetopito fixed the issue in another PR

PR #1119, opened 2020-06-18
Filter /Orders by market

### LINKED ISSUE #1118: Filter /Orders by a market prop
Filter by WETH-USD for example

W3stside:
Closes #1118

Test:
1. head to /orders or on /trade
2. search via:
  a. WETH-TUSD
  b. WETH/TUSD
  c. WETH TUSD
  d. (`/orders` and/or `Active` | `Liquidity` | `Closed` tabs only): 
      i. WETH-TUSD (or variations with "/" or " ") and the inverse market too: TUSD-WETH

Query: <sellToken>[SEPARATOR]<buyToken>

### DIFF
--- a/src/utils/display.tsx
+++ b/src/utils/display.tsx
@@ -10,12 +10,31 @@ export function displayTokenSymbolOrLink(token: TokenDetails): React.ReactNode |
   return displayName
 }
 
+/**
+ * computeMarketProp
+ * @description returns array of potentially accepted market names by:: SELLTOKEN <SEPARATOR> BUYTOKEN
+ * @param { sellToken, buyToken, acceptedSeparators: string[] }
+ */
 export function computeMarketProp({
   sellToken,
   buyToken,
+  acceptedSeparators = ['-', '/', ' '],
+  inverseMarket = false,
 }: {
   sellToken: TokenDetails
   buyToken: TokenDetails
-}): string {
-  return `${safeTokenName(sellToken).toLowerCase()}-${safeTokenName(buyToken).toLowerCase()}`
+  acceptedSeparators?: string[]
+  inverseMarket?: boolean
+}): string[] {
+  const sellTokenFormatted = safeTokenName(sellToken).toLowerCase()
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1119--dexreact.review.gnosisdev.com

src/utils/display.tsx
```diff
@@ -10,14 +10,31 @@ export function displayTokenSymbolOrLink(token: TokenDetails): React.ReactNode |
...
 
+/**
+ * computeMarketProp
+ * @description returns array of potentially accepted market names by:: SELLTOKEN <SEPARATOR> BUYTOKEN
+ * @param { sellToken, buyToken, acceptedSeparators: string[] }
+ */
 export function computeMarketProp({
   sellToken,
   buyToken,
+  acceptedSeparators = ['-', '/', ' '],
```
anxolin, line 21: I think, they usually don't use space as a separator. - and / are good, but the third I would say is empty string if that's easy, otherwise this is great already

image

W3stside: ok :+1:

src/utils/display.tsx
```diff
@@ -10,14 +10,31 @@ export function displayTokenSymbolOrLink(token: TokenDetails): React.ReactNode |
...
 
+/**
+ * computeMarketProp
+ * @description returns array of potentially accepted market names by:: SELLTOKEN <SEPARATOR> BUYTOKEN
+ * @param { sellToken, buyToken, acceptedSeparators: string[] }
+ */
 export function computeMarketProp({
   sellToken,
   buyToken,
+  acceptedSeparators = ['-', '/', ''],
```
Velenir, line 21: This would trigger false positives for tokens symbols with  `'-/ '` in the symbol, won't it?

W3stside: well it spits out e.g `WETH-DAI` so if the user is searching for market `WETH-D` - `DAI` it wouldn't find it, which is intended, no?

PR #1127, opened 2020-06-19
reverse market computation

W3stside:
Reverses the market computation for filtering: WETH/DAI is now filtered by DAI/WETH etc


gnosis-info:
Travis automatic deployment:
 https://pr1127--dexreact.review.gnosisdev.com

PR #1135, opened 2020-06-21
Fix error removing trailing zeros for price

### LINKED ISSUE #1134: Error in the price calculation when trailing zeros
The left price in the trade page is wrong after inputting a value that ends in zero.

Reproduce by inputting `0.40` into the right price. You should be getting 2.5 in the left price, but it's setting 0.4 instead.

anxolin:
Fixes #1134


gnosis-info:
Travis automatic deployment:
 https://pr1135--dexreact.review.gnosisdev.com

PR #1137, opened 2020-06-22
add enableToken to TokenRow (TradeWidget)

### LINKED ISSUE #611: Implement enable token in Trade widget
Implement enable token in Trade widget

image

W3stside:
Closes #611 

To test: pick a non enabled token (check Balances page) and enable

Not enabled:
Screenshot from 2020-06-22 11-42-02

Enabling:
Screenshot from 2020-06-22 11-42-29

### DIFF
--- a/src/components/DepositWidget/Form.styled.ts
+++ b/src/components/DepositWidget/Form.styled.ts
@@ -19,6 +19,7 @@ export const WalletDrawerInnerWrapper = styled.div`
     font-family: var(--font-default);
     letter-spacing: 0;
     line-height: 1.4;
+    text-align: left;
   }
 
   p.error {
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1137--dexreact.review.gnosisdev.com

alfetopito:
Oh, something else.

In my view, the `enable` button should not be displayed unless a wallet is connected.
OR
It should prompt the wallet connect, then enable the token.

Currently it just fails.

W3stside:
Issues fixed in #1139

anxolin:
> I just wonder whether this will confuse more our users since they might think a deposit is needed in order to place an order.
Even though it's required to trade...

I think it's good they think that in general :) 
Probably it would be a nice default event to make the user to create orders backed by funding. But as a "general" app, this allows what the protocol allows, at least for now

Issue #1132, opened 2020-06-21
Limit Price is Confusing to Users


jpkcambridge:
Some users are placing the inverse of the limit price they actually intend due to confusion about which box has the quote currency they intend to use.  Could be better to have one box with the price in the quote currency.  For example instead of what is shown in the screenshot below, just have one box with the price "0.36 USDT"

<img width="533" alt="Screen Shot 2020-06-20 at 9 35 18 PM" src="https://user-images.githubusercontent.com/874683/85214802-00252500-b33e-11ea-80de-5c4fd732363d.png">

koeppelmann:
@jpkcambridge the problem is that it is very hard to make a general rule for this.
Agree, in this case, the 0.36 representation is more intuitive but e.g. what in a USDT/DAI use case

What does a price 1.02 express here? 
As soon as you fill in the "sell" amount it will fill in the "receive" amount and that should make it in principal clear.

jpkcambridge:
The price would be in the quote currency, so 1.02 DAI would be the amount of DAI you are paying for 1 USDT.  Examples of this are https://loopring.io/trade/USDT-DAI and https://pro.coinbase.com/trade/DAI-USDC (Couldn't find USDT on Coinbase)

Of course it would make sense for a user to verify the sell and receive amounts make sense, but I think the goal should be to stick to popular conventions and make it as clear as possible.

besoeasy:
no limit price makes perfect sense, if someone can install metamask and enable coins he can understand limit order.

limit order is general term used widely while trading

anxolin:
We brought that up several times. I think the underlying issue is that normally exchanges decide for each token pair, which one is base and which one is quote.
In Mesa, there’s no such a thing, anyone can list any token pair. Actually, this would make a nice addition dxDAO, defining the markets in a decentralized way.

If you set the base/quote in a canonical way, then you could sell/buy for that market. The price would refer always to the quote, and people would get less confusion.

If the UI cannot decide which is the quote, you could see "SELL A for B" order as:
- Sell in A-B market (so price is given in B => you are interested in the UI's left price)
- Buy in B-A market (so price is given in A => you are interested in the UI's right price)

anxolin:
About the solution, it might be good to show just one price, but I think probably wouldn't solve the issue for everyone, imagine for example a user is selling in DAI-WETH market.

[...]

PR #1138, opened 2020-06-22
One price field

Velenir:
Displays only reverse price filed, that is BUY/SELL by default
Also changes `BUY/SELL`  to `SELL per BUY`
Allows to swap between `SELL per BUY <-> BUY per SELL`

ezgif-6-b4d2ad150188

Related to #1132

### DIFF
--- a/src/components/TradeWidget/Price.tsx
+++ b/src/components/TradeWidget/Price.tsx
@@ -194,12 +194,12 @@ const Price: React.FC<Props> = ({ sellToken, receiveToken, priceInputId, priceIn
     [setValue],
   )
 
-  const onChangePrice = useCallback(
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1138--dexreact.review.gnosisdev.com

koeppelmann:
to me, it is not clear at all that this will help with the issue.

It is simply the case that for a token price e.g. GNO/DAI the price people think of is e.g. 28$ and for sure not 0.03571428571 (this is how much GNO you need to pay for 1 DAI)
So, unless we have some logic that figures out which perspective people expect (which is not always obvious in e.g. the stable coin/stable coin example) showing one 1 price will make it WORSE (if it is the price people do not expect)

PR #1146, opened 2020-06-22
Trade Page: Do not expire by default

anxolin:
The default params fo the trade sets "Never" as the default expiration date


gnosis-info:
Travis automatic deployment:
 https://pr1146--dexreact.review.gnosisdev.com

PR #1147, opened 2020-06-22
Updating how we calculate whether a token is enabled.

alfetopito:
Instead of a hard value of max uint128, we now check
whether allowance is >= 10**(token.decimals + 12)

The assumption we had before did not hold for DMG token.
We set the allowance to maxint256, but on their contract the amount is downcast to utin128 (https://etherscan.io/address/0xed91879919b71bb6905f23af0a68d231ecf87b14#code)

So doesn't matter how many times the allowance is called, the UI still considers it as not enabled.

## Testing:

1. Enable DMG in mainnet https://dex.dev.gnosisdev.com
    Refresh the page.
    Should still show the `Enable` button.

2. Check DMG balance on https://pr1147--dexreact.review.gnosisdev.com/wallet
    Should show the `+` sign

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -31,6 +31,8 @@ export const ONE_HUNDRED_BIG_NUMBER = new BigNumber(100)
 // ∴ when the amount is < 5 the order will be considered filled.
 export const ORDER_FILLED_FACTOR = new BN(10000) // 0.01%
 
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1147--dexreact.review.gnosisdev.com

PR #1145, opened 2020-06-22
Wallet connect bridge

anxolin:
Adds a new config for wallet connect bridge

Changes also the default to use Gnosis one, since it's more reliable

@nlordell , let me know if this solves your issue

PR #1149, opened 2020-06-23
Hotfix/1.0.3

anxolin:
* Trade Page: Do not expire by default (#1146)
* Wallet connect bridge (#1145)


Velenir:
Why target this at develop.
Should just merge master into develop, without squashing.

anxolin:
Better from master

PR #1154, opened 2020-06-24
Substitute acc

Velenir:
Hack for easier debugging from an arbitrary account's perspective

`window.loginAs('0x1234ef...')` to enable for account
call `window.loginAs('')` with anything falsy to disable

Read-only access obviously, no sending txs.
Development only.

### DIFF
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -106,6 +106,28 @@ export const composeProvider = <T extends Provider>(
     ),
   )
 
+  if (process.env.NODE_ENV === 'development') {
+    // hack to be able to use interface as if from a different account
+    // read-only of course
+    // account will update on the next eth_accounts call
+    // normally on new block in a few seconds
+    engine.push(
+      createConditionalMiddleware(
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1154--dexreact.review.gnosisdev.com

src/api/wallet/composeProvider.ts
```diff
@@ -106,6 +106,28 @@ export const composeProvider = <T extends Provider>(
...
+    // hack to be able to use interface as if from a different account
+    // read-only of course
+    // account will update on the next eth_accounts call
+    // normally on new block in a few seconds
+    engine.push(
+      createConditionalMiddleware(
+        req => req.method === 'eth_accounts',
+        (_req, res) => {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          if ((window as any).__acc) {
```
anxolin, line 119: Can we change it so what we expose is a function `window.logicAs(address: string)` I think it's easier to remember.

Then the function of course would just set the var

Velenir: Done

PR #1156, opened 2020-06-24
Using transaction time instead of batch time for trades date

alfetopito:
### The problem

Date displayed for a trade was based on batch time, which is the start of the 5min interval.
It's misleading though, since in fact the trade happens up to minute 4 of the batch.
So when the trade is actually detected, it' shows as having happened 4min+ ago.

With this change, the time displayed is the transaction timestamp of when the solver executed the trade.


gnosis-info:
Travis automatic deployment:
 https://pr1156--dexreact.review.gnosisdev.com

PR #1162, opened 2020-06-29
Fix filter orders delete

### LINKED ISSUE #1107: Filtering Orders + delete orders bad UX
When filtering orders and choosing to delete orders, selecting "ALL" should use filtered data set, not the entire classified orders set

to reproduce: on PR <insert PR # here> filter a set of tokens and then select all to delete. Will highlight all orders, not filtered ones

Velenir:
Fix #1107 

Now checking `delete/cancel all` only checks currently displayed orders
And clears previous selection if any.


gnosis-info:
Travis automatic deployment:
 https://pr1162--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -235,19 +233,22 @@ const OrdersWidget: React.FC<Props> = ({ isWidget = false }) => {
       if (selectedTab === 'fills') return
 
       setClassifiedOrders(curr => {
-        const state = emptyState()
-
         // copy full state
-        Object.keys(curr).forEach(tab => (state[tab] = curr[tab]))
```
W3stside, line 241: You know, I was always wondering at the complexity of this code Vs doing what you did. Personally was too scared/lazy to actually do it. Absolute mad man

PR #1164, opened 2020-06-29
1163/non intrusive price suggestion

### LINKED ISSUE #1163: Make price estimation non-intrusive
## The problem
There has been negative feedback lately regarding the price estimation feature.

Currently, it replaces the price input with the estimation when it's received from `dex-price-estimator` service.

The summary of raised points are:
1. User input is overwritten without warning
1. Some times the estimation arrives too late, and users end up placing orders with bad prices (might have been solved already)
1. Estimation for larger amounts might return in worse prices, making the `receive amount` smaller than when selling a smaller amount.
1. Users do not know how far off their prices are from the estimation when placing the orders

More details and suggestions can be found on this spreadsheet containing [Mesa feedback and improvement suggestions](https://docs.google.com/spreadsheets/d/1Z8D9Bka4ewojkstrK7bZKu7yB1iof1rnUDxBSQC6cKU/edit#gid=0)

alfetopito:
## Update July 2

- Added swap icon
- No longer showing `infinity` prices
- Added tooltip. Please validate text contents:
screenshot_2020-07-02_13-53-28

## Update Jun 30

- Did a minimal styling, properly aligning things and matching other parts of the app
- Adjusted widget height to fit the new entry
- No longer displaying `infinity` for bad prices. Using `N/A` instead
- Disabling input when `N/A`
- Added base/quote token suffix to be explicit on what this price refers too (although being a bit redundant)
- Added `selling` word to sentence to make it clear what the amount refers to: `Onchain orderbook price for selling <amount> <symbol|name|address>`
- Fixed the displayed symbol in the sentence. Previously it was changing every time the price was inverted
- Displaying second price estimation only when amount is different from default of 1 unit. 

Worst case scenario (long token symbols):
screenshot_2020-06-30_14-49-37

Regular use case (short token symbols):
screenshot_2020-06-30_15-05-39

### TODO:
- [ ] Show % difference between prices (slippage?)
- [x] Add tooltips
- [x] Add swap price buttons

## Proposal

No longer automatically update price inputs with price suggestion.
Instead, show a button and allow the user to make the choice.

Closes #1163

This change adds mostly the logic of fetching the prices and displaying it in a button.

I added 1 price for a single unit and one price for the selected amount.
This way the user has a reference of the `slippage` given his amount.

When switching to the inverse price, the suggestion is also inverted.
When clicking in a button, both prices are filled.

screenshot_2020-06-29_15-09-52

It's hideous, I know. Will work on that.

As always, open for suggestions/feedback.

## TODO:
- [x] Style
- [x] Do not show `infinity` when one of the prices is too big/too small
- [x] More testing, look for edge cases, etc

### DIFF
--- a/src/components/TradeWidget/PriceEstimations.tsx
+++ b/src/components/TradeWidget/PriceEstimations.tsx
@@ -14,6 +14,7 @@ import Spinner from 'components/Spinner'
 
 import { TradeFormData } from 'components/TradeWidget'
 import { SwapIcon } from 'components/TradeWidget/SwapIcon'
+import { HelpTooltip, HelpTooltipContainer } from 'components/Tooltip'
 
 const Wrapper = styled.div`
   > strong {
@@ -87,6 +88,14 @@ function formatPriceToPrecision(price: BigNumber): string {
   return price.toFixed(PRICE_ESTIMATION_PRECISION)
 }
 
+const OnchainOrderbookTooltip = (
+  <HelpTooltipContainer>
+    Based on existing Onchain orders, taking into account possible ring trades. <br />
+    Price is affected by sell amount. <br />
+    Higher amounts might yield worse prices, or no price at all, considering what&apos;s available in the current batch.
+  </HelpTooltipContainer>
+)
+
 const OnchainOrderbookPriceEstimation: React.FC<OnchainOrderbookPriceEstimationProps> = props => {
   const { networkId, amount, baseToken, quoteToken, isPriceInverted, updatePrices, swapPrices } = props
   const { id: baseTokenId, decimals: baseTokenDecimals } = baseToken
@@ -116,7 +125,7 @@ const OnchainOrderbookPriceEstimation: React.FC<OnchainOrderbookPriceEstimationP
   return (
     <>
       <span>
-        Onchain orderbook price for selling{' '}
+        Onchain orderbook price <HelpTooltip tooltip={OnchainOrderbookTooltip} /> for selling{' '}
         <strong>
           {+amount || '1'} {displayTokenSymbolOrLink(quoteToken)}
         </strong>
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1164--dexreact.review.gnosisdev.com

W3stside:
@alfetopito careful on the height of the div considering the added content aka price suggestion. If it gets too tall (e.g when the user puts in too much sell and gets the over balance warning) the submit button will wrap to the right and look shitty af

src/components/TradeWidget/PriceEstimations.tsx
```diff
@@ -0,0 +1,89 @@
...
+    quoteTokenDecimals,
+  })
+
+  const displayPrice = priceEstimation
+    ? (isPriceInverted ? invertPrice(priceEstimation) : priceEstimation).toFixed(PRICE_ESTIMATION_PRECISION)
+    : '0'
+
+  return (
+    <div
+      style={{ display: 'flex', justifyContent: 'space-between', padding: '0.1em 0 0.1em 0.5em', fontSize: '1.4rem' }}
```
W3stside, line 77: You know what to do here

src/components/TradeWidget/index.tsx
```diff
@@ -506,6 +499,12 @@ const TradeWidget: React.FC = () => {
   const defaultValidFrom = trade.validFrom || validFromParam
   const defaultValidUntil = trade.validUntil || validUntilParam
 
+  const [priceShown, setPriceShown] = useState<'INVERSE' | 'DIRECT'>('INVERSE')
+
+  const swapPrices = (): void => {
```
W3stside, line 504: Can be one liner lambda

src/components/TradeWidget/PriceEstimations.tsx
alfetopito: Yup :)

Silver-tape-quick-hacky-fix-first-revision

alfetopito:
> @alfetopito careful on the height of the div considering the added content aka price suggestion. If it gets too tall (e.g when the user puts in too much sell and gets the over balance warning) the submit button will wrap to the right and look shitty af

Indeed. In that case, what's the best way to handle it? Just increase the default max height?

alfetopito:
> Other than a minor things, like the INFINITE, and probably the tooltip, I think it's good to go.
> 
> The others we can add it to our backlog. Feel free to continue working on them since you are on fire
> But I think we should freeze one version with what we have and deliver

@anxolin addressed your comments. I'll wait for your feedback on the tooltip text before merging, though.

PR #1165, opened 2020-06-30
added long press capabilities

W3stside:
Waterfalls into #1158 

Allows more mobile friendly longpress to select capabilities, not sure how we feel about it

To test (it also works in desktop but can be disabled if u guys want, i saw no harm in keeping it right now)

1. head into mobile view
2. click on card and hold click down until it says delete sth
3. had a banner to show as tip


gnosis-info:
Travis automatic deployment:
 https://pr1165--dexreact.review.gnosisdev.com

PR #1168, opened 2020-06-30
Disable tokens list

Velenir:
localhost_8080_trade_DAI-USDC_sell=0 price=0 99883 from= expires=
shown in /wallet

hidden in /trade and /liquidity

Everything is subject to change. Make your suggestions


gnosis-info:
Travis automatic deployment:
 https://pr1168--dexreact.review.gnosisdev.com

Velenir:
localhost_8080_wallet (4)
localhost_8080_wallet (5)

src/components/TokenImg.tsx
```diff
@@ -7,14 +7,19 @@ function _loadFallbackTokenImage(event: React.SyntheticEvent<HTMLImageElement>):
...
+
+const TokenImg = styled.img.attrs(() => ({ onError: _loadFallbackTokenImage }))<TokenImgProps>`
   width: 2.8rem;
   height: 2.8rem;
   border-radius: 3.6rem;
   object-fit: contain;
   margin: 0 1rem 0 0;
   background-color: white;
   padding: 2px;
+  opacity: ${(props): number => (props.opaque ? 0.4 : 1)};
```
anxolin, line 22: I think the logic of the flaf is inverred:

 `opaque definition: not able to be seen through; not transparent.`

I would call it `faded` or `deprecated`

src/hooks/useManageTokens.tsx
```diff
@@ -76,7 +76,13 @@ const TokenList: React.FC<TokenListProps> = ({ tokens, onToggleToken, disabledTo
               onToggleToken(address, !checked)
             }}
           >
-            <OptionItem name={name} symbol={symbol} image={image}>
+            <OptionItem
+              name={name}
+              symbol={symbol}
+              image={image}
+              opaque={overrideReason === 'DEPRECATED'}
```
anxolin, line 83: Would it be good to move the deprecated flag to the token details so you don't need to repeat this logic?

Velenir: My thinking was, there could be overrides for reasons other than deprecation

anxolin:
Please just don't forget to add SXN, aTOKENS, Paxos Golds, and the tokens we discussed in the other thread that might be incompatible with the protocol

PR #1182, opened 2020-07-01
More disabled tokens

Velenir:
localhost_8080_wallet (6)
localhost_8080_wallet (7)


config-default.yaml
```diff
@@ -76,4 +76,12 @@ disabledTokens:
       symbol: SNX-old
       description: This token is disabled for trading and deposit. SNX will be deprecated and replaced by another token at the end of July.  Go to https://www.synthetix.io for more information
       reason: DEPRECATED
+      
+    - address: '0x45804880De22913dAFE09f4980848ECE6EcbAf78'
+      name: Paxos Gold
+      symbol: PAXG
+      description: This token is disabled for trading and deposit.
+      reason: DISABLED
```
W3stside, line 84: should we not append DISABLED to paxg name?

gnosis-info:
Travis automatic deployment:
 https://pr1182--dexreact.review.gnosisdev.com

config-default.yaml
```diff
@@ -70,4 +70,86 @@ disabledTokens:
...
+    - address: '0x3a3A65aAb0dd2A17E3F1947bA16138cd37d08c04'
+      description: This token is disabled for trading and deposit.
+      reason: DISABLED
+
+    - address: '0xfC1E690f61EFd961294b3e1Ce3313fBD8aa4f85d'
+      description: This token is disabled for trading and deposit.
+      reason: DISABLED
+
+    - address: '0x9D91BE44C06d373a8a226E1f3b146956083803eB'
+      description: This token is disabled for trading and deposit.
```
anxolin, line 112: It feels super repetitive. I would think that adding a flag `disabled` adds already that message `This token is disabled for trading and deposit`. Then, deprecated doesn't need to be modeled. We can just add optionally the `description` for some tokens. These tokens with description would show in the title. 

`"This token is disabled for trading and deposit. " + description`

config-default.yaml
Velenir: Should we? @anxolin

Velenir:
Fixed :arrow_up:

config-default.yaml
anxolin: I would say is better like it is now imo

image


you can see the warning and the faded icon. I think is good!

PR #1187, opened 2020-07-02
Fix pending orders

Velenir:
When an order is submitted with `validUntil == NEVER` due to default values on the form it becomes 0
and `validUntilWithBatchID == validFromWithBatchID`, and `isPendingOrderActive` in the part of
```js
batchIdToDate(order.validUntil) >= now
```
marks order as inactive
So that pending order is put into [closed](https://github.com/gnosis/dex-react/blob/1fb0a3110e9ccc85bc013951e5669fa625fcec95/src/components/OrdersWidget/index.tsx#L72) category

This here be fix for that

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -719,6 +719,7 @@ const TradeWidget: React.FC = () => {
         userAddress,
         validFromWithBatchID,
         validUntilWithBatchID,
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1187--dexreact.review.gnosisdev.com

PR #1191, opened 2020-07-03
Fix token name

### LINKED ISSUE #1190: Spelling error in token name
I think we overlooked this:
image

I think it's not reason enough to do another hotfix. But we should address it in develop.

Also, the synthetic USD is call Synth sUSD:
image

SNX seems alright:

image

Velenir:
Shame! :bell:  Shame! :bell: 

Closes #1190


gnosis-info:
Travis automatic deployment:
 https://pr1191--dexreact.review.gnosisdev.com

Issue #1040, opened 2020-05-22
Update orderbook automatically


anxolin:
It would be nice to be able to have a live version of the order book.

Probably once the event base order book @fleupold  and @nlordell is live, there's possible more options to implement a live order book. An ideal outcome would be to have a websocket notifying either the deltas or the full-updates. How feasable would this be?

For now, the front-end could take a simple approach of re-fetching the data automatically every 30s or so.

@alfetopito @Velenir what do you think?

Velenir:
Add `blockNumber` dep to hook?
Add `blockNumber` dep to hook

anxolin:
The orderbook changes not on every batch, but on every order/deposit/withdrawal/solution 
Right now is a bit costly for the service, this is why I didn't want a big frequency

2 users querying the same order book (base/quote/hops) could return the same result. So it's highly cacheable at both HTTP and SERVICE level. I mean, both an internal cache and using `cache-control` in the backend headers could instruct the browser and any intermediate server about the caching policy.

With that we can do queries more often without punishing the server.

The next step could be evaluate if we can use Web Sockets, so the server pushes every time he recalculates the order book, or if at some point we want to implement one super efficient, we try to push just the deltas (price points that change)

alfetopito:
> For now, the front-end could take a simple approach of re-fetching the data automatically every 30s or so.

Dumb and quick idea: cache the call response on the frontend for ~30s. Update the orderbook component every ~5 seconds.

nlordell:
> Probably once the event base order book is live

It should be currently live.

Is issue specifically for the transient orderbook (showing the fancy graph for overlapping orders between two tokens)?

anxolin:
> Dumb and quick idea: cache the call response on the frontend for ~30s. Update the orderbook component every ~5 seconds.

👌something like that is already a great start. It was kind of my suggestion. 
I was also making reference on where we could add the caches. If they for example update every 5s they can return the cache control for 1s (or even for the time to the next update). Then any request done to their service, wouldn't arrive even to the node code, because nginx can eat that up.

> Is issue specifically for the transient orderbook (showing the fancy graph for overlapping orders between two tokens)?

Yes. Someone asked us if we need to REFRESH to see the changes in the order book. Normally people is use to getting automatic refresh in the data for that.

![image](https://user-images.githubusercontent.com/2352112/82694915-65fd8e80-9c64-11ea-8146-ff81fb919a1e.png)

> It should be currently live.

[...]

PR #1196, opened 2020-07-06
1049/linear orderbook

### LINKED ISSUE #1049: Order Book: Linear price for the X-axis

alfetopito:
# Linear orderbook

Closes #1049

## Update July 31

- Adjusting Y axis for one sided markets
- Refactoring code
- Filtering out volume smaller than 10 OWL
- Fixed OWL volume calculation (I was querying the inverted value :facepalm: )

WETH - sETH market looks sooo pretty
screenshot_2020-07-31_14-48-31

## Update July 30

- Fixed another Y zoom bug
- Moved zoom controls to inside the chart
- Replaced default `zoom out` button
- Removed right side scrollbar
- Removed controls on top side scrollbar
- Added `mini map` on top side scrollbar

screenshot_2020-07-30_15-58-04

## Update July 29

- Fixed Y zoom when ask volume > bid volume
- Filtering out order with volume smaller than 1 OWL (fetching price from estimator service for quote token)
- Added zoom controls
screenshot_2020-07-29_15-01-23

### Notes :warning: 
- There are still edge cases when zooming in/out
- Haven't yet removed native zoom controls (scrollbars, zoom out button)
- Zoom does not play well when there is only one side of the market
- Of course styling is non existent 

## Update July 24

- Tooltips are back!
- Also, auto zooming to a "sensible" area on load, both and `X` and `Y` axis.
- Using as default width the equivalent to 5 times the spread. Works even if spread overlaps.
- Re-using the same chart instance, which means faster loading and no flickering
- Added scrollbars on top and right side, to indicate current zoom level (if it's crammed on a tiny corner it's because the price range is huuuuge!)

**Top**: Before; **Bottom**: After
screenshot_2020-07-24_15-48-05

### Notes:
- Further styling would be good on the scrollbars, but it's out of the scope of this PR
- Can easily turn on auto updates, just need to define how often. Potentially even allow users to turn this on and off. Also out of the scope https://github.com/gnosis/dex-react/issues/1040
- Couple of TODOs, to be addressed. Mostly regarding constants. Will do when addressing PR comments.

---

## Original

In order to adjust the graph scale to linear without having to process the data and let AMCharts lib do the job, I'm changing the X axis type from `Category` to `Value`.

`Category` axis type displays every entry with equal weigh, causing the X axis to have differing values.
`Value` axis type on the other hand distributes entries according to the value.

The downside of `Value` axis is that we can't have the nice tooltip like this anymore:
screenshot_2020-07-06_13-49-21
<sub><sup>Reference: https://www.amcharts.com/docs/v4/concepts/axes/axis-tooltips/:
> For a Category axis, it will always be the category that Cursor is currently within. For Date and Value axis, we may get some intermediate number, as the Cursor will get more granular.
</sup></sub>

The upside is this might fix the issue with sorting where depending what the backend sends the graph is screwed up.

It looks a bit weird though. I'm wondering if we should limit the range so huge prices/orders down the end of the orderbook don't make it look unintelligible.

Thoughts?

### Changing from:
screenshot_2020-07-06_13-46-19

### To:
screenshot_2020-07-06_13-46-33


Velenir:
Why aren't tooltips supported on Value axis? Sounds like a very arbitrary limitation
To the point I almost want to say let's use a different chart lib, almost

gnosis-info:
Travis automatic deployment:
 https://pr1196--dexreact.review.gnosisdev.com

alfetopito:
> Why aren't tooltips supported on Value axis? Sounds like a very arbitrary limitation
> To the point I almost want to say let's use a different chart lib, almost

LoL, this one should be fine.

Maybe I should rephrase it: In the short term I played around with the lib, **I did not find a way** to display the same tooltips Category axis provides on Value axis.

Other approach could be to do the grouping on our side and stick to Category type for the easy nice tooltip.

src/components/OrderBookWidget.tsx
```diff
@@ -271,32 +271,36 @@ const draw = (
...
   const bidSeries = chart.series.push(new am4charts.StepLineSeries())
-  bidSeries.dataFields.categoryX = 'priceNumber'
+  bidSeries.dataFields.valueX = 'priceNumber'
   bidSeries.dataFields.valueY = 'bidValueY'
   bidSeries.strokeWidth = 1
   bidSeries.stroke = am4core.color(colors.green)
   bidSeries.fill = bidSeries.stroke
-  bidSeries.startLocation = 0.5
   bidSeries.fillOpacity = 0.1
   bidSeries.tooltipText = `[bold]${market}[/]\nBid Price: [bold]{priceFormatted}[/] ${quoteTokenLabel}\nVolume: [bold]{totalVolumeFormatted}[/] ${baseTokenLabel}`
```
Velenir, line 295: this becomes essentially noop, right?

anxolin:
Yeap, that's an issue if we don't discard some values. Maybe it's not so trivial. I understand we shouñd have a price reference and place it in the middle (this price reference can be the new endpoint exposed in the price API)

Then we need a strategy to determine the max/min price to represent. Maybe is a matter of capping it to 10S (it could be any other number instead of 10), and S being the spread.

So for a Price Reference (PR) and Spread (S)
`-|--------------|--------------|-`
`PR-10S              PR                    PR + 10S

anxolin:
Super relevant conversation. It looks like the best way forward is to allow user to zoom in-out and select a good default.
https://gnosisinc.slack.com/archives/CC3C5SMB5/p1594122747186100

Different exchanges offer different ways to zoom in and out.
Also, take into account the back end is applying also some filter already. We might want to tell the backend we want a wide range, but then we apply a more restrictive filter in the FE.

One suggestion for the default:
```
spread = BestAsk (AB) - BestAsk (BA)
Price Reference = Abs(BestAsk (AB) - BestAsk (BA)) / 2
Min = Price Reference - 10 * Spread 
Max = Price Reference - 10 * Spread 
```

@fleupold shared some examples from Kraken:
> Kraken gives you a slider to chose your spread yourself. Default is 1%
image

>Binance has a top down in the top right with order of magnitudes (reasonable one is preselected)

image

> Deversify has zoom in/out buttons to change the range.

image

src/components/OrderBookWidget.tsx
alfetopito: Yup

anxolin:
Great job!
I've checked USDC-DAI in production and looks much better.
I have some concerns about how we calculate where to center the graph, I'll try to do some more testing in Rinkeby. But so far, I've seen a couple of details.

This is Rinkeby WETH-TUSD, not sure if it's related to the fact we also adjust Y-axis, but the tooltip for the ASKS is not showing up unless I zoom out:
image

If I zoom out this is how it looks:
image

Another strange thing, is that if I invert the market, it doesn't look symmetrical. It looks much better TUSD-WETH than the inverse:
image

If I press zoom out in the previous graph, I stop seeing anything because the X-axis goes all the way from range 0.0010-0.0095 which is a nice start, to 0-900. With one click I just did a 100X.
* Probably we don't need to account for prices that are way off market
* Maybe we can control the max zoom out, so we don't take this big jumps. Something like +-20%

It took me a bit to understand the scroll. I like the functionality, but I think it should be more like Tradingview, where you just scroll (https://www.tradingview.com/chart/). So I wouldn't show the component.

Last part, is it possible to add also a Zoom in button? With it, we can go +-20% with the zoom. That plus the horizontal scroll seems to me like a good option.

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+
+      // Reload data from data source re-using same chart
+      chart.dataSource.load()
+
+      // Setting up tooltips based on currently loaded tokens
+      const [bidSeries, askSeries] = chart.series.values
+      bidSeries.tooltipText = `[bold]${market}[/]\nBid Price: [bold]{priceFormatted}[/] ${quoteTokenLabel}\nVolume: [bold]{totalVolumeFormatted}[/] ${baseTokenLabel}`
+      askSeries.tooltipText = `[bold]${market}[/]\nAsk Price: [bold]{priceFormatted}[/] ${quoteTokenLabel}\nVolume: [bold]{totalVolumeFormatted}[/] ${baseTokenLabel}`
+    }
+  }, [baseToken, chart, hops, networkId, quoteToken])
```
Velenir, line 476: Would also be great to reset zoom settings when `[baseToken, hops, networkId, quoteToken]` change

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+      break
+    }
+  }
+  let asksVolume = asks.length > 0 ? asks[0].totalVolumeNumber : 0
+  for (let i = 0; i < asks.length; i++) {
+    if (upperZoomX < asks[i].priceNumber) {
+      break
+    }
+    asksVolume = asks[i].totalVolumeNumber
   }
```
Velenir, line 267:
```suggestion
  const bidsVolume = bids.find(bid => lowerZoomX < bid.priceNumber)?.totalVolumeNumber || 0
  const asksVolume = asks.find(ask => upperZoomX < ask.priceNumber)?.totalVolumeNumber || 0
```

src/components/OrderBookWidget.tsx
alfetopito: But it does. Every time there's a change, it re-subscribes to the data event which calculates the zoom.
Not working like that for you?

alfetopito:
> Great job!
> I've checked USDC-DAI in production and looks much better.
> I have some concerns about how we calculate where to center the graph, I'll try to do some more testing in Rinkeby. But so far, I've seen a couple of details.

@anxolin 

The center is based on the modulus of spread. Currently it's equal to 5x the spread, bounded on the left by min(minbid, minask) and on the right max(maxbid, maxask).
So in normal circumstances there's 2x spread on each side, divided by the spread in the middle.

In a regular exchange, there would never be an overlap of bids and asks, but we are not that fortunate.
Also can happen that - depending on the volume we consider - the spread is quite large.

I think that by improving the symmetry (filtering out orders by relative OWL price) we can get better spreads.

Other than that I don't see how to improve the display at this point.
 
> This is Rinkeby WETH-TUSD, not sure if it's related to the fact we also adjust Y-axis, but the tooltip for the ASKS is not showing up unless I zoom out:
> image
> 
> If I zoom out this is how it looks:
> image

1. For the Y axis zoom, I calculate - based on what the X axis zoom will be - what are the bid and ask volumes and pick the max (the part where Dima has a code suggestion). There might be some issues with the index, will check that.

> Another strange thing, is that if I invert the market, it doesn't look symmetrical. It looks much better TUSD-WETH than the inverse:
> image

2. For the lack of symmetry, the reason is that we filter orders by volume. Anything smaller than `0.01` is removed. That is a problem when dealing with tokens with a large price difference, such as TUSD-WETH. 
Based on Nick's comment on slack, I'll probably filter out based on token prices in relation to OWL. 
For example, remove any volume that's smaller than 1 OWL in given token units.

> If I press zoom out in the previous graph, I stop seeing anything because the X-axis goes all the way from range 0.0010-0.0095 which is a nice start, to 0-900. With one click I just did a 100X.
> 
>     * Probably we don't need to account for prices that are way off market

3. How would that work? We just drop prices that deviate x% from the suggested price?

>     * Maybe we can control the max zoom out, so we don't take this big jumps. Something like +-20%

4. Then in your opinion we should never zoom out to the full orderbook?
 
> It took me a bit to understand the scroll. I like the functionality, but I think it should be more like Tradingview, where you just scroll (https://www.tradingview.com/chart/). So I wouldn't show the component.

5. I can take a look on how to implement that using amCharts, although this might be extending the scope of this task even further.
In other words, will do now if quick, will postpone otherwise :)
 
> Last part, is it possible to add also a Zoom in button? With it, we can go +-20% with the zoom. That plus the horizontal scroll seems to me like a good option.

6. Yes! I was tinkering with that before I settled on the current approach.
I can definitely add buttons to control the zoom.

The question would be then, what would the +/- button do when the initial zoom is veeeeeery small, such as 0.01% to 0.03%?
- Zoom a percentage of initial zoom? x10% => 0.009% to 0.033%
- Add/remove a fixed amount? +10% -> 0% (can't go bellow 0) to 10.03%

src/components/OrderBookWidget.tsx
Velenir: ezgif-4-47dc9dbc405f
I it zoom setting not being reset or min/max recalculated each time to fit better?

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+        baseTokenId: baseToken.id,
+        quoteTokenId: quoteToken.id,
+        hops,
+        networkId,
+      })
+
+      // Removing any previous event handler
+      chart.dataSource.adapter.remove('parsedData')
+      // Adding new event handler
+      chart.dataSource.adapter.add('parsedData', data => {
```
anxolin, line 391: nit, would it be worth it to move `parsedData` to another function. This `iseEffect` is too big. It would be good to break it down into smaller parts for readability

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
-export default OrderBookWidget
+    setChart(_chart)
+
+    return (): void => _chart.dispose()
+    // We'll create only one instance as long as the component is not unmounted
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  useEffect(() => {
+    if (chart) {
```
anxolin, line 366: nit: if we change this into 
```
if (!chart) {
   return
}
```
I understand it would be equivalent, and you reduce the cyclomatic complexity of this useEffect. Specially, if you also take my other nitpick (below)

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+    if (upperZoomX < asks[i].priceNumber) {
+      break
+    }
+    asksVolume = asks[i].totalVolumeNumber
   }
+  console.log(`bidsVolume: ${bidsVolume}; asksVolume: ${asksVolume}`)
+  return Math.max(bidsVolume, asksVolume)
+}
+
+function calcMinX(minBid: number, minAsk: number): number {
```
anxolin, line 272: it seems a bit strange to create a wrapper of MAX and MIN functions

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+      chart.dataSource.adapter.add('parsedData', data => {
+        try {
+          const bids = processData(data.bids, baseToken, quoteToken, Offer.Bid)
+          const asks = processData(data.asks, baseToken, quoteToken, Offer.Ask)
+          const pricePoints = bids.concat(asks)
+
+          let startX = 0
+          let endX = 1
+
+          if (bids.length > 0 && asks.length > 0) {
```
anxolin, line 400: nit: if all this method is for calculating the start/end, I would suggest to move it to its own function that get bids, asks, and return these two (start/end)

src/components/OrderBookWidget.tsx
```diff
@@ -224,103 +223,257 @@ function _printOrderBook(pricePoints: PricePointDetails[], baseToken: TokenDetai
...
+            const minBid = bids[0].priceNumber
```
anxolin, line 401: Probably you don't want to deal with this in the PR, but we can ad a TODO here with an issue to select a maxBid/minAsk taking the quote price into account.

If we have the OWL or USD price for the quote, we can:
* Filter tiny orders: i.e < 5-10 OWL (it would be great to align this with the warning message for the minimal trade, or at least make this tiny orders filter be a subset of the trades that would warn if you try to send them)
* Select as the minAsk/maxBid the price where we accumulate at least 30 or 50 OWL

alfetopito: > Filter tiny orders

Yes, this is what I intend to do in this PR to resolve the issue with a fixed filter amount of 0.01, which doesn't work when dealing with WETH, creating asymmetrical graphs when inverting the token order.

> Select as the minAsk/maxBid the price where we accumulate at least 30 or 50 OWL
In that case we might have no orderbook under some circumstances.

alfetopito:
> One thing is that sometimes the reset value breaks the view but meh
@W3stside what reset value? If you could help me reproduce it would be great :+1:

src/components/OrderBookWidget.tsx
alfetopito: Nice gif, need to learn how to do that :)

Anyway, not sure yet what's wrong, it looks alright. When you change tokens the zoom is adjusted for the new data.

> ...  min/max recalculated each time to fit better?
Yes. Every time there's new data (token pair changes) the zoom is recalculated.

anxolin:
I tried again. Definitely there's something strange in the zoom selection.
I'm checking USDC-DAI right now. I'll explain with some screenshots I'm taking.

This is DEV. It looks pretty decent, although it has the problem we try to solve here. The x-axis is not linear, so in this case the ASKS (green) show liquidity that it's too off price. Other than that, we can understand there's liquidity in both sides.
image

However, we are showing this case in the current PR it looks like there's no ASKS:
image

My guess is that this is because we still have small orders that are shown. I would think that we need to discard more orders based on their USD (or OWL) price, so the threshold probably should be bigger. Knowing that our minimal trade is in the thousands of USD, I wouldn't include this orders that break the graph:
image

Maybe the another thing to take into account is that we want to show if available, an order book that shows the market for a volume that makes sense in GP. For example 20K. This could help with the strategy, maybe is a bit more complex than we thought, but including several criterias to show meaningful selection makes sense to me. This is what I mean:
image

One thing I would do for sure, is what I commented in my last review (although I made a mistake pasting the image). I meant to add a ZOOM PREVIEW. This would help to see if there's liquidity out of the current price range:
https://codepen.io/team/amcharts/pen/mjdmJX

@Rafanator, please keep an eye in this conversation for validation.

src/components/OrderBookWidget.tsx
alfetopito: To keep everything tidy, one function to calculate every parameter.

src/components/OrderBookWidget.tsx
alfetopito: Did some refactoring, still more to do.

alfetopito:
Totaly missed this comment when started this morning.

> I tried again. Definitely there's something strange in the zoom selection.
> I'm checking USDC-DAI right now. I'll explain with some screenshots I'm taking.
> 
> This is DEV. It looks pretty decent, although it has the problem we try to solve here. The x-axis is not linear, so in this case the ASKS (green) show liquidity that it's too off price. Other than that, we can understand there's liquidity in both sides.
> image
> 
> However, we are showing this case in the current PR it looks like there's no ASKS:
> image

Hopefully, the zoom controls I just pushed can help with that:

screenshot_2020-07-29_15-01-23

> My guess is that this is because we still have small orders that are shown. I would think that we need to discard more orders based on their USD (or OWL) price, so the threshold probably should be bigger. Knowing that our minimal trade is in the thousands of USD, I wouldn't include this orders that break the graph:
> image

Well, I'm filtering any order with volume less than 1 OWL. Should we go for 10? 100? I don't think thousands would be a good idea though. We'd be loosing too much information.

> Maybe the another thing to take into account is that we want to show if available, an order book that shows the market for a volume that makes sense in GP. For example 20K. This could help with the strategy, maybe is a bit more complex than we thought, but including several criterias to show meaningful selection makes sense to me. This is what I mean:
> image
> 
> One thing I would do for sure, is what I commented in my last review (although I made a mistake pasting the image). I meant to add a ZOOM PREVIEW. This would help to see if there's liquidity out of the current price range:
> https://codepen.io/team/amcharts/pen/mjdmJX

Ah, that makes more sense :)
Yes, we can add that, although... might not be very usable. Our data is not so nice like the example, so I guess most of the time it'll be unreadable. Also, the zoomed area might be so small that, like pointed out, makes it unusable.

> 
> Also. Currently the `-` button it jumps from showing prices near 1, to 24K. Obviously we should have a limit way before.
> We can see how this 24K order is infact a tiny order that doesn't improve significantly our volume.
> image
> 
> @Rafanator, can you please keep an eye in this PR for validation.

anxolin:
It certainly a more complicated "fix" than anticipated. Great progress!

i was confused by this icon, I was not understanding what it was doing for some time.
nit: I think it's good to style different the swap icon, and place it in a different location than the zoom controls. But we can review this with Michel.

The functionality works great. Zoom in and out now works as I expect!!

> Fixed Y zoom when ask volume > bid volume

I'm not sure I understood. But there's something odd with the Y-axis. You can see it in Rinkeby DAI-USDC if you swap the tokens:
image
image


I saw you added zoom for the y-axis. 
I have some mixed feelings about that. It allows more control, but it makes it also more complex. What makes more sense to me is to not have the control, and show the maximum volume within the represented values. For example here, would 330:
image


Also, what do you think of using their preview functionality. This way, we can see all the book shape (we can tell if there's a big order at a greater/smaller price)
image

anxolin:
Now there's two `-` icons. Do they allow to add the zoom controls in the graph and unify? 
image

alfetopito:
> Now there's two `-` icons. Do they allow to add the zoom controls in the graph and unify?
> image

Don't know yet. I intend to work on the native zoom controls today.

alfetopito:
> It certainly a more complicated "fix" than anticipated. Great progress!

Tell me about... =P

> i was confused by the "sync" button, I was not understanding what it was doing for some time.
> nit: I think it's good to style different the swap icon, and place it in a different location than the zoom controls. But we can review this with Michel.

Yep. Not touching that, will it up to you guys ;)
 
> The functionality works great. Zoom in and out now works as I expect!!
> 
> > Fixed Y zoom when ask volume > bid volume
> 
> I'm not sure I understood. But there's something odd with the Y-axis. You can see it in Rinkeby DAI-USDC if you swap the tokens:

There was another bug on that. Fixed.

> I saw you added zoom for the y-axis.
> I have some mixed feelings about that. It allows more control, but it makes it also more complex. What makes more sense to me is to not have the control, and show the maximum volume within the represented values. For example here, would 330:
> image

I don't understand what you mean. In order to find what is the volume for that zoom, I need to do the same calculation regardless.

> Also, what do you think of using their preview functionality. This way, we can see all the book shape (we can tell if there's a big order at a greater/smaller price)

Done. Let me know what you think.

> Now there's two - icons. Do they allow to add the zoom controls in the graph and unify?

Also, now using zoom controls inside the chart itself. Much cleaner :)

Example:
screenshot_2020-07-30_15-58-04

W3stside:
stonks

anxolin:
We need to sort out some things. Discussions here https://docs.google.com/document/d/1wKyTGIvale06cPkSHlb3p2G4poiBgZoiCjIQrdPq5Wk/edit

@Rafanator is owning this one, he will make sure we take the right decision :) 
everyone is welcome to oppinate in the doc

PR #1208, opened 2020-07-08
disable wrapEther on disconnect

### LINKED ISSUE #1193: Wrap ether with disconnected wallet
image

Wrap ether button in trade page, should follow the same workflow as "deposit" or "enable". Only visible if we have a connected wallet:

image

W3stside:
Closes #1193 

Disconnect wallet and check button isnt there

### DIFF
--- a/src/components/TradeWidget/TokenRow.tsx
+++ b/src/components/TradeWidget/TokenRow.tsx
@@ -263,9 +263,14 @@ const TokenRow: React.FC<Props> = ({
         <div>
           {editableAndConnected && !tokenDisabled && (
             <>
-              <div className="btn" onClick={(): void => showForm('deposit')}>
+              <TooltipWrapper
+                as="button"
+                type="button"
+                tooltip="Deposit tokens into the Exchange Balance so it can be used for trading"
+                onClick={(): void => showForm('deposit')}
+              >
                 + Deposit
-              </div>
+              </TooltipWrapper>
               {isDepositFormVisible && (
                 <Form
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1208--dexreact.review.gnosisdev.com

anxolin:
Perfect!

Not related to this PR, but can we just add a tooltip here?
image

See how the other links have one:

image

image



One suggestion:
"Deposit tokens into the Exchange Balance so it can be used for trading"

anxolin:
Also, ubernit! 

I think this is not aligned:
image

anxolin:
Maybe the above is even solved by the tooltip? Cause otherwise I don't see the difference in the two links

src/components/TradeWidget/TokenRow.tsx
```diff
@@ -286,7 +286,7 @@ const TokenRow: React.FC<Props> = ({
               )}
             </>
           )}
-          {!readOnly && isWeth && <WrapEtherBtn label="+ Wrap Ether" />}
+          {editableAndConnected && isWeth && <WrapEtherBtn label="+ Wrap Ether" />}
```
Velenir, line 289: `Deposit` and `Wrap Ether` buttons are unaligned because one of them isn't a button
L266: `+ Deposit` is a `<div>` for some reason

W3stside: its fixed now

PR #1209, opened 2020-07-08
Remove fixed height that breaks IOS

anxolin:
Removing this fixes the issue in IOS.

Why would the form need a fixed height?

Before:
IMG_1F44F1320999-1

After:
IMG_1E9C22141257-1


gnosis-info:
Travis automatic deployment:
 https://pr1209--dexreact.review.gnosisdev.com

Issue #1199, opened 2020-07-07
Check Metamask compatibility


alfetopito:
# Check Metamask compatibility

Make sure we are compatible with the Metamaks breaking changes

_Estimation: 1_

Velenir:
Summary of **MetaMask** deprecated API:
https://docs.metamask.io/guide/ethereum-provider.html#legacy-methods

Transaction sending methods:
* We rely on `send|sendAsync` and can't replace it,even conditionally, because afaik **web3** operates with send not request
* **Web3** providers (**HttpProvider**, **WebSocket** provider) only implement send
* **json-rpc-engine** which we use for composing providers, uses `send/sendAsync`, though that part is easy to change
* all other injected providers (on mobile), who knows when they'll implement request
* last time I checked WalletConnect.request is a bit different in signature from `send|sendAsync`, so can't just fallback` (p.request || p.send || p.sendAsync)(params)
`
`Provider.enable()`:
* we also rely on enable method, directly in autoconnect, and through **Web3Connect**
* it probably won't go away anytime soon, so it's ok to keep it
* **web3connect** will likely do something to accommodate request in the future
* so we should stiop using `.enable()` directly, only through **web3connect**
* **WalletConnect** handles enable differently from `request({ method: 'eth_requestAccounts' })`, so not equivalent for now

`Provider.on()` events:
* events, we already listen for `chainChanged`, so we are good there
* subscription results **MetaMask** actually emits in data event, not in `notification|message` event, doesn't matter **web3** uses what is right internally

`ethereum._metamask` object:
* we used to get some data from `ethereum._metamask` object back in **dapp-ui** times, not relevant anymore
* warning in console **MetaMask** will soon stop reloading pages on network change doesn't matter
* we do set `ethereum.autoRefreshOnNetworkChange= false` in `WalletApi`, but that happens a bit later than the warning
* warning appears on page load (when **MetaMask** extension is injected into window)

Summary of summary:
* new **MetaMask** methods are only available with version >= 8
* so only **Firefox** for now
* legacy will continue working

All in all because of provider feature disparity I think we should do the following:
1. Keep legacy methods
2. Wait until libraries (**web3**, **web3connect**) add support for new methods
3. Watch out for bugs as they happen
4. Stop calling `provider.enable` in `autoconnect`, instead rely on **web3connect** to do that work for us (#1210)


@alfetopito @anxolin @W3stside , thoughts?

alfetopito:
Great summary. 

[...]

PR #1210, opened 2020-07-09
Don't call provider.enable() directly

Velenir:
As part of #1199, we should try not to rely on deprecated API as much as possible.
This PR changes autoconnect logic to use `web3modal.connectors` to create providers for us.
And while internally the connectors call `provider.enable()` we can expect **web3modal** to eventually switch to `provider.request` method.


gnosis-info:
Travis automatic deployment:
 https://pr1210--dexreact.review.gnosisdev.com

PR #1216, opened 2020-07-09
1206/invert prices positions on orders

alfetopito:
Part of #1206 

Before:
screenshot_2020-07-09_09-54-29

After:
screenshot_2020-07-09_09-52-10

PR #1215, opened 2020-07-09
Inverting price estimation market

alfetopito:
Part of #1206 

Before:
screenshot_2020-07-09_09-41-31

After:
screenshot_2020-07-09_09-42-13

Issue #1206, opened 2020-07-07
Price/Market display improvements


alfetopito:
## Price/Market display improvements

Small changes before we release for price/market improvements

_Estimation: 1_

alfetopito:
Based on https://docs.google.com/document/d/1du5mgEQ1z34d2uJcXTii40U2MXWDcB9kViGDqqThFeI/edit#

For this version, we'll do:
- [ ] From `Market in Trade Widget` section: `3. The market display there is ❌ (the price is right though)`
- [ ] From `Orders: Markets and Prices` section: `2+3. The order of the prices is reversed. WETH-DAI should be on top, DAI-WETH should be on bottom`
- [ ] From `Orders: Markets and Prices` section: `Follow up question: Would it be good to make bold the WETH-DAI price, so we know it's the primary one?`
- [x] From `Export Fills: Markets and Prices` section: `nitpick: Can we just make the columns be written in Plain English instead of camel case? ⚠️`
- [ ] From `Export Fills: Markets and Prices` section: `It's showing the DAI-WETH price, instead of WETH-DAI one`
- [ ] From `Export Fills: Markets and Prices` section: `I would add the units of the price for clarification. This become important for counterintuitive markets like WETH-DAI`
- [ ] From `Export Fills: Markets and Prices` section: `Ideally it's a good idea to add the Inverse price, the CSV doesn't have issues of space, so we can just add another 2 columns for this price and his units`
- [ ] From `Export Fills: Markets and Prices` section: `Fill price ❌Same issues as above`
- [ ] From `Export Fills: Markets and Prices` section: `6. Amount correct, but we should add the units ✅❌`
- [ ] From `Export Fills: Markets and Prices` section: `7. Amount correct, but we should add the units ✅❌`
- [ ] From `Export Fills: Markets and Prices` section: `8. Show the %`
- [ ] From `Export Fills: Markets and Prices` section: `9. Missing Total amount of the order. To get a context of the previous porcentage ❌`
- [ ] From `Coherence Orders vs Trades: Market` section: `It's nice how we show the market for the trades. Then the price could just omit the base token and be expressed in quote`

PR #1220, opened 2020-07-09
1206/market column on orders tab

alfetopito:
Part of #1206

Before:
screenshot_2020-07-09_14-19-56

After:
screenshot_2020-07-09_14-18-52

Where:
1. The market here is fixed (inverted) in another PR: https://github.com/gnosis/dex-react/pull/1215
2. Buy `Market` column added
3. Building on top of #1216, displaying only the unit instead of the full market
4. Default price seen here is the same as the `buy` price on `3`

### DIFF
--- a/src/components/OrdersWidget/OrderRow.tsx
+++ b/src/components/OrdersWidget/OrderRow.tsx
@@ -86,14 +86,10 @@ const OrderDetails: React.FC<OrderDetailsProps> = ({ buyToken, sellToken, order 
     <td data-label="Price" className="showResponsive">
       <div className="order-details">
         <strong>
-          {priceInverse} {displayTokenSymbolOrLink(buyToken)}
-          {'/'}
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1220--dexreact.review.gnosisdev.com

Velenir:
For point **1** to be on the same line you can try
```
.small { // the one with USDC/WETH
  display: flex;
  align-items: baseline;

or simply
    white-space: nowrap;
}

```
whatever works best

alfetopito:
> For point **1** to be on the same line you can try
> 
> ```css
> .small { // the one with USDC/WETH
>   display: flex;
>   align-items: baseline;
> 
> or simply
>     white-space: nowrap;
> }
> ```
> 
> whatever works best

I'll address it on that PR https://github.com/gnosis/dex-react/pull/1215

PR #1226, opened 2020-07-14
Preventing submission when either sell/buy amount is 0

### LINKED ISSUE #1225: Order with buyAmount = 0
It is possible to submit orders with `buyAmount = 0`, which results in `priceNumerator= 0` and `Infinity` inverse price
localhost_8080_trade_DAI-USDC_sell=0 price=0 from= expires= (1)

To reproduce:
1. Input sellAmount
2. Before receiveAmount fills in, click Submit

As `receiveAmount` is filled async, it can take some time depending on network conditions.

This essentially allows for orders like _sell tokens for free_

alfetopito:
Closes #1225

Simple fix, prevents submission when either value is 0.
We had the check in place for `falsy` values, but since amount is a BN, we also need to check whether it's zero.

Does not address the slowness to fill in buyAmount.

Printing debug log entry when a null value is detected:
screenshot_2020-07-14_13-11-24

## Steps to reproduce
1. Fill in a sell amount
2. Click on suggested price
3. Very quickly, click on submit or hit enter.

Before: the wallet is prompted to sign the tx, before the `Receive at least` (buyAmount) is filled
After: no tx is sent, debug log message is printed.

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -21,7 +21,15 @@ import { MEDIA, PRICE_ESTIMATION_DEBOUNCE_TIME } from 'const'
 import { TokenDetails, Network } from 'types'
 
 // utils
-import { getToken, parseAmount, parseBigNumber, dateToBatchId, resolverFactory, formatTimeToFromBatch } from 'utils'
+import {
+  getToken,
+  parseAmount,
+  parseBigNumber,
+  dateToBatchId,
+  resolverFactory,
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1226--dexreact.review.gnosisdev.com

alfetopito:
> You went and fixed it clap

You went and reviewed it!
screenshot_2020-07-14_13-24-36
(I'd be long at sleep by that time =P)

PR #1228, opened 2020-07-15
Use tabs

W3stside:
Simplifies our tab component - will also use in Order Validity in next PR

available in `hooks/useTabs`

### DIFF
--- a/src/components/OrdersWidget/index.tsx
+++ b/src/components/OrdersWidget/index.tsx
@@ -18,7 +18,7 @@ import useSafeState from 'hooks/useSafeState'
 import useDataFilter from 'hooks/useDataFilter'
 import useSortByTopic from 'hooks/useSortByTopic'
 import { useWalletConnection } from 'hooks/useWalletConnection'
-import useTabs from 'hooks/useTabs'
+import { useTabs, Tabs, TabData } from 'hooks/useTabs'
 
 // Api
 import { DetailedAuctionElement, DetailedPendingOrder, Trade } from 'api/exchange/ExchangeApi'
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1228--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -354,34 +330,27 @@ const OrdersWidget: React.FC = () => {
...
+                  count: classifiedOrders.active.orders.length + classifiedOrders.active.pendingOrders.length,
+                },
+                {
+                  type: 'fills',
+                  count: trades.length,
+                },
+                {
+                  type: 'liquidity',
+                  count: classifiedOrders.liquidity.orders.length + classifiedOrders.active.pendingOrders.length,
+                },
```
alfetopito, line 346: Does it have any conflicts with the separated liquidity orders PR?

src/hooks/useTabs.tsx
```diff
@@ -0,0 +1,155 @@
...
+}
+
+function useTabs<T>(defaultTab: T): UseTabsReturn<T> {
+  const [selectedTab, setSelectedTab] = useSafeState<T>(defaultTab)
+  const setSelectedTabFactory = useCallback(
+    (type: T): ((event: React.SyntheticEvent<HTMLButtonElement | HTMLFormElement>) => void) => (
+      event: React.SyntheticEvent<HTMLButtonElement | HTMLFormElement>,
+    ): void => {
+      // form is being submitted when clicking on tab buttons, thus preventing default
+      event.preventDefault()
```
Velenir, line 121: Doesn't happen to me. And it shouldn't because you have `type="button"` not `"submit"`

src/hooks/useTabs.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+      setSelectedTab(type)
+    },
+    [setSelectedTab],
+  )
+
+  const Tabs: React.FC<TabsProps<T>> = ({ children, className, tabsList = [] }) => {
```
Velenir, line 128: It is generally not a good idea to create a component inside a hook. Causes remounting
One way to mitigate would be to memoize the component.
An arguably better way is #1231

Velenir:
> Any other tab, the header collapses

Does not happen to me. Firefox only?

src/hooks/useTabs.tsx
alfetopito: Old code, it was needed before we did the `type="button"` thingy

alfetopito:
> > Any other tab, the header collapses
> 
> Does not happen to me. Firefox only?

Yep, only FF. Chrome is fine.

alfetopito:
I still see the collapsed header

screenshot_2020-07-20_09-37-39

FF on the left, Chrome on the right

W3stside:
@alfetopito this one can def wait, no rush. thanks for the data tho i need to permanently delete FF for everyone's (mainly mine) sake

anxolin:
It doesn't make sense. I reproduced this in this PR https://github.com/gnosis/dex-react/pull/1232, but not in develop 🤔

PR #1231, opened 2020-07-16
Use tabs outside

Velenir:
Creating component inside a hook is generally not a good idea. Causes remounting.

remount (1)
See how `<buttons>` disappear on each block instead of just being updated.

Can be solved with memoization.
Or by separating component from the hook.

Modali and react-hook-form use this same pattern:
```js
const [modalHook, toggleModal] = useModali()
...
<Modali.Modal {...modalHook} />
```

```js
const methods = useForm();
...
<FormProvider {...methods} > // pass all methods into the context
```

### DIFF
--- a/src/hooks/useTabs.tsx
+++ b/src/hooks/useTabs.tsx
@@ -95,61 +95,64 @@ const TabsWrapper = styled.div`
   }
 `
 
-interface TabData<T> {
+export interface TabData<T extends string> {
   type: T
   count?: string | number
 }
 
-interface TabsProps<T> {
+interface TabsProps<T extends string> {
   className?: string
   tabsList?: TabData<T>[]
-}
-
-interface UseTabsReturn<T> {
   selectedTab: T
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1231--dexreact.review.gnosisdev.com

src/hooks/useTabs.tsx
```diff
@@ -95,61 +95,64 @@ const TabsWrapper = styled.div`
...
+      setSelectedTabFactory,
+    },
   }
 }
 
-export default useTabs
+// can pass {..tabProps} from useTabs
+// or override individual props on <Tabs {...tabProps} selectedTab="activ"/>
+// not a React.FC to allow for generic Props
+export const Tabs = <T extends string>({
```
W3stside, line 134: why this? i dont understand

Velenir: To make the Tabs component accept generic props, as it was a generic component dependent on `<T>` when it was inside `useTabs`
If you want can it make non-generic 
`interface TabsProps` ~`<T>`~

Doesn't matter much

PR #1238, opened 2020-07-17
Fix liquidity orders on rinkeby

### LINKED ISSUE #1235: Can't place liquidity orders
# Can't place liquidity orders

On #1234 

https://pr1234--dexreact.review.gnosisdev.com/liquidity

## Steps to reproduce

1. Select any token combination
2. Input any spread
3. Click on `submit transaction`

## Expected

Liquidity orders are placed

## Observed

Tx is not placed. Error shown on the UI and logged to the console:
screenshot_2020-07-16_09-50-33

## Details:
- Firefox 78.0.2 (64-bit) AND Chrome(ium) Version 83.0.4103.116 (Official Build) Arch Linux (64-bit)
- Metamask 8.0.4
- ~No issues on Chrome, same MM version.~
- No issues on FF with WallectConnect.

Velenir:
The issue happens because:

1. Page load
2. No connection => default networkId = 1
3. `selectedTokensMap` is filled with tokens from network 1
4. Connection to rinkeby established
5. Tokens updated
6. But `selectedTokensMap` not refilled

This is a quick fix, but let's switch to addresses instead of symbols soonish.
Having addresses in `LIQUIDITY_TOKEN_LIST` and that list differentiating by network would have made this issue more obvious (cause there would not have been a match for wrong address on the right network)

Closes #1235

### DIFF
--- a/src/components/PoolingWidget/index.tsx
+++ b/src/components/PoolingWidget/index.tsx
@@ -156,7 +156,13 @@ const PoolingInterface: React.FC = () => {
     )
   }, [tokenList])
 
-  const [selectedTokensMap, setSelectedTokensMap] = useSafeState<Map<number, TokenDetails>>(setFullTokenMap(tokens))
+  const [selectedTokensMap, setSelectedTokensMap] = useSafeState<Map<number, TokenDetails>>(() =>
+    setFullTokenMap(tokens),
+  )
+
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1238--dexreact.review.gnosisdev.com

PR #1234, opened 2020-07-16
Release/1.1.0

alfetopito:
:warning: DON'T SQUASH :warning: 

# Release `1.1.0`

Big release, including many new features since the `1.0` release, besides hotfixes already deployed.

## Highlights

- ~`Fills`~ `Trades` tab, showing the trade history
- Consistency across market representations throughout the app
- Batch id now displayed on top
- Enable/deposit tokens from `/trade` page
- Non intrusive price suggestion
- Orders/trades filtering
- `Hidden` settings page
- Liquidity orders moved out of `/trade` onto `/liquidity` tab.

### Bug fixes
- Trust wallet specific issues
- Locale issue on unwrap WETH
- Order validity input sanitization

## Full list of changes

8b7e5936 Decreasing the amount of token required to consider token enabled (#1230)
5fa26735 Preventing submission when either sell/buy amount is 0 (#1226)
0f59ba17 Don't call provider.enable() directly (#1210)
80f7706c change GetPriceResponse to correspond to dex-price-estimator (#1224)
dd25a6b6 1206/adjusting exported csv (#1218)
62ba9eb1 ASAP > Now and related tags (#1214)
75933b9f 1206/market column on orders tab (#1220)
eab65a1c 1206/invert prices positions on orders (#1216)
0d66bdbe Inverting price estimation market (#1215)
7cfbb9e2 Batch explanation tooltip (#1212)
4c1016ef CamelCase -> human readable (#1211)
9205e39e Make OrdersWdget moar responsive (#1158)
4a6c6c0f disable wrapEther on disconnect (#1208)
af436d53 Fix trust gas issue (#1170)
9df2dc46 ignore locale for unwrap input (#1192)
b8eb3a00 1163/non intrusive price suggestion (#1164)
f7d674f1 Settings Page: css changes (#1185)
e0f1e5e0 fix token name (#1191)
04bb98cc Estimator url prod/dev (#1183)
564d8409 Bump eslint-plugin-jest from 23.13.2 to 23.17.1
9cd3c18e Bump @babel/preset-react from 7.10.1 to 7.10.4
4b933482 Bump babel-jest from 25.5.1 to 26.1.0
b022a3fa Bump @fortawesome/fontawesome-svg-core from 1.2.28 to 1.2.29
c9f1db82 Bump eslint-plugin-import from 2.21.1 to 2.22.0
c572c8d0 Bump @popperjs/core from 2.4.1 to 2.4.3
[...]

PR #1240, opened 2020-07-17
Token for price estimator is fixed, even when showing inverted price

alfetopito:
Small fix, this got changed during a refactor.

To be merged onto #1234 

Before:
screenshot_2020-07-17_13-45-30
screenshot_2020-07-17_13-45-16

After:
screenshot_2020-07-17_13-46-48
screenshot_2020-07-17_13-46-37


gnosis-info:
Travis automatic deployment:
 https://pr1240--dexreact.review.gnosisdev.com

Velenir:
I feel like this will still get people confused with how we are switching BUY/SELL pair

PR #1243, opened 2020-07-20
css fixes

### LINKED ISSUE #1236: Responsive orders fields issues
# Responsive orders fields issues

On #1234

screenshot_2020-07-16_10-34-03

- `Market` row shows up twice
- `Unfilled amount` row title does not explain the two values. I would suggest instead the same title as the column: `Filled/Total`

- When cancel orders tx is sent, the UI changes for the orders being cancelled. Orders not selected remain the same.
screenshot_2020-07-16_10-41-51

- Can't see whole balances on token selector (same for FF and Chrome)
screenshot_2020-07-16_10-48-04

## Minor

- WRAP/UNWRAP input (might affect other inputs, haven't checked) cut the amount before half the size of the component
screenshot_2020-07-16_10-46-04

W3stside:
Closes #1236 

Width issue when in mobile

### DIFF
--- a/src/components/PoolingWidget/PoolingWidget.styled.ts
+++ b/src/components/PoolingWidget/PoolingWidget.styled.ts
@@ -58,6 +58,7 @@ export const PoolingInterfaceWrapper = styled(WrappedWidget)`
 
   @media ${MEDIA.tablet} {
     > form {
+      min-width: 100%;
       max-width: 100%;
       padding: 2rem 4rem;
     }
@@ -69,6 +70,7 @@ export const PoolingInterfaceWrapper = styled(WrappedWidget)`
     font-size: 1.3rem;
 
     > form {
+      min-width: 100%;
       max-width: 100%;
       padding: 1.6rem;
     }
--- a/src/components/PoolingWidget/PoolingWidget.styled.ts
+++ b/src/components/PoolingWidget/PoolingWidget.styled.ts
@@ -70,7 +70,8 @@ export const PoolingInterfaceWrapper = styled(WrappedWidget)`
     font-size: 1.3rem;
 
     > form {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1243--dexreact.review.gnosisdev.com

src/components/PoolingWidget/PoolingWidget.styled.ts
```diff
@@ -58,6 +58,7 @@ export const PoolingInterfaceWrapper = styled(WrappedWidget)`
 
   @media ${MEDIA.tablet} {
     > form {
+      min-width: 100%;
       max-width: 100%;
```
Velenir, line 62: So `width: 100%;`?
```suggestion
      width: 100%;
```

src/components/PoolingWidget/PoolingWidget.styled.ts
```diff
@@ -69,6 +70,7 @@ export const PoolingInterfaceWrapper = styled(WrappedWidget)`
     font-size: 1.3rem;
 
     > form {
+      min-width: 100%;
       max-width: 100%;
```
Velenir, line 74:
```suggestion
      width: 100%;
```

alfetopito:
Was this supposed to fix https://github.com/gnosis/dex-react/issues/1236#issuecomment-660303091 ?

If so, sorry dude, still the same for me, both Chrome and FF

W3stside:
> Was this supposed to fix [#1236 (comment)](https://github.com/gnosis/dex-react/issues/1236#issuecomment-660303091) ?
> 
> If so, sorry dude, still the same for me, both Chrome and FF

actually it does need the `min-width` to override the other one yay

PR #1249, opened 2020-07-21
Do not show liquidity trades on orders page

alfetopito:
Mobile `Fills` tab shows all trades, regular and liquidity.
That's because on responsive mode we show `/orders` page directly.

This change enables the filter for the `/orders` page equal to the one used on `/trade` page.


gnosis-info:
Travis automatic deployment:
 https://pr1249--dexreact.review.gnosisdev.com

PR #1251, opened 2020-07-21
Sepatate the spinner and Pending label

anxolin:
Just a tiny detail that bother me while testing and just went for it

Before:
image

After:
image

Before @Velenir says anything, `{' '}` don't work for me, cause prettier plugin turn it into a space, that then is removed when rendering. Old school `&nbsp;` works fine


gnosis-info:
Travis automatic deployment:
 https://pr1251--dexreact.review.gnosisdev.com

PR #1259, opened 2020-07-24
hotfix/1.1.1

W3stside:
Filtering for `orderId` in `Trades` tab/page was wrong - fixed

1. will still occassionally pick up wrong trades when `orderId` coincidentally is also found inside a token address, for example

### DIFF
--- a/src/utils/filter.ts
+++ b/src/utils/filter.ts
@@ -21,10 +21,9 @@ const filterTradesAndOrdersFnFactory = <
 
   const market =
     sellToken && buyToken && computeMarketProp({ sellToken, buyToken, inverseMarket: includeInverseMarket })
-
+  console.debug(id, rest)
   return (
-    !!id.includes(searchTxt) ||
-    (rest.orderId && !!rest.orderId.includes(searchTxt)) ||
+    (rest.orderId ? !!rest.orderId.includes(searchTxt) : !!id.includes(searchTxt)) ||
     (market && !!market.includes(searchTxt)) ||
     checkTokenAgainstSearch(buyToken, searchTxt) ||
     checkTokenAgainstSearch(sellToken, searchTxt)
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1259--dexreact.review.gnosisdev.com

anxolin:
> will still occassionally pick up wrong trades when orderId coincidentally is also found inside a token address, for example

Then, I would say. We shouldn't be filtering by address if the string don't look like an address: https://web3js.readthedocs.io/en/v1.2.7/web3-utils.html#isaddress

src/utils/filter.ts
```diff
@@ -1,28 +1,30 @@
...
 import { computeMarketProp } from './display'
+import { web3 } from 'api'
 
 export function checkTokenAgainstSearch(token: TokenDetails | null, searchText: string): boolean {
   if (!token) return false
   return (
     token?.symbol?.toLowerCase().includes(searchText) ||
     token?.name?.toLowerCase().includes(searchText) ||
-    token?.address.toLowerCase().includes(searchText)
+    (web3.utils.isAddress(searchText) && token?.address.toLowerCase().includes(searchText))
```
anxolin, line 11: nit: `web3.utils.isAddress(searchText)` is evaluated for every row two times. Should we pre-calculate it and just pass the flag (or use a different function depending on it)?

W3stside: but im checking both tokens, `buyToken` and sell, so it needs to eval twice, no?

W3stside: see what u mean, made changes

Velenir: `isAddress` check is checksum sensitive so if you pass lowercase to it, will return `false` always

W3stside: @Velenir it won't return `false` unless the mixed cases are wrong. if all lowercase or all uppercase it skips checksum check

W3stside: https://web3js.readthedocs.io/en/v1.2.7/web3-utils.html#isaddress

anxolin:
@W3stside I'll merge this for you

PR #1261, opened 2020-07-27
preliminary orderId column

W3stside:
Adds `ID` (orderId) column to trades


gnosis-info:
Travis automatic deployment:
 https://pr1261--dexreact.review.gnosisdev.com

src/components/OrdersWidget/index.tsx
```diff
@@ -441,6 +441,7 @@ const OrdersWidget: React.FC<Props> = ({ displayOnly }) => {
                             disabled={deleting}
                           />
                         </th>
+                        <th>ID#</th>
```
anxolin, line 444: Order ID

src/components/TradesWidget/index.tsx
```diff
@@ -170,11 +170,12 @@ export const InnerTradesWidget: React.FC<InnerTradesWidgetProps> = props => {
     <CardTable
       $rowSeparation="0"
       $gap="0 0.6rem"
-      $columns={`1fr 0.8fr minmax(6.6rem, 1fr) 1.2fr 6.5rem ${isTab ? 'minmax(9.3rem, 0.6fr)' : '0.74fr'}`}
+      $columns={`1fr 3rem 0.8fr minmax(6.6rem, 1fr) 1.2fr 6.5rem ${isTab ? 'minmax(9.3rem, 0.6fr)' : '0.74fr'}`}
     >
       <thead>
         <tr>
           <th>Date</th>
+          <th>ID#</th>
```
anxolin, line 178: I would move it to the column after type, and call it "Fill for Order Id"

src/components/TradesWidget/TradeRow.tsx
```diff
@@ -143,6 +148,9 @@ export const TradeRow: React.FC<TradeRowProps> = (params) => {
       <td data-label="Type" title={typeColumnTitle}>
         <TypePill tradeType={type}>{type}</TypePill>
       </td>
+      <td data-label="Order ID" onClick={(): void => onOrderIdClick({ target: { value: orderId } })}>
```
Velenir, line 151: `onClick` is the idea only to filter trades by that order or to also go to Orders tab?

Velenir (approved):

src/components/TradesWidget/TradeRow.tsx
W3stside: the latter, or should it auto change tabs?

Velenir: hard to say. I'm leaning towards _no_ as that would be an arbitrary change veryspecific to this use case only

Rather it makes sense to allow to fill the filter with more stuff onClick

W3stside: maybe best as you said, name onclick sth general and make every `td` clickable and fill the filter bar with it's contents (if it's searchable)

anxolin (approved):
Nice!! I really like it. It's not very conventional, but at least allows you to understand better wich trades comes from the same order, and easily filter by order between all tabs.

As a ubernit, can you make the order id a link <a />, so it looks like sth you can click? Also, if you click it, we can show a toast saying: "Filtering by order XX". It all happens too fast, so it's nice to give that feedback to the user

W3stside:
> Nice!! I really like it. It's not very conventional, but at least allows you to understand better wich trades comes from the same order, and easily filter by order between all tabs.
> 
> As a ubernit, can you make the order id a link , so it looks like sth you can click? Also, if you click it, we can show a toast saying: "Filtering by order XX". It all happens too fast, so it's nice to give that feedback to the user

mm I don't think a toast is appropriate here, toast notifications form a UX standpoint are for direct blockchain actions, not filtering. What about a much loved info button on the data bar explaining what's clickable? @Velenir @alfetopito @anxolin

W3stside:
> mm I don't think a toast is appropriate here, toast notifications form a UX standpoint are for direct blockchain actions, not filtering. What about a much loved info button on the data bar explaining what's clickable? @Velenir @alfetopito @anxolin

adding to this, we should make `Market` clickable as well

Velenir:
> As a ubernit, can you make the order id a link , so it looks like sth you can click? Also, if you click it, we can show a toast saying: "Filtering by order XX". It all happens too fast, so it's nice to give that feedback to the user

> mm I don't think a toast is appropriate here, toast notifications form a UX standpoint are for direct blockchain actions, not filtering. What about a much loved info button on the data bar explaining what's clickable? @Velenir @alfetopito @anxolin

An alternative can be a subtle animation on focus-in for the filter input. Just to draw attention that it was filled

anxolin:
> An alternative can be a subtle animation on focus-in for the filter input. Just to draw attention that it was filled

I thought about that, I think it's great idea. Suggested toast, since in my mind was easier, but your approach is better

PR #1262, opened 2020-07-28
Styling tx conf modal

W3stside:
Styles TX Confirmation modal

Screenshot from 2020-07-28 11-59-37


gnosis-info:
Travis automatic deployment:
 https://pr1262--dexreact.review.gnosisdev.com

W3stside:
Some things:

1. forewent adding `formatSmart` so there might be some huge numbers displayed (screencap kind of shows what i mean) - one option could be to simply `.toFixed(8)` or sth
2. almost wanted to use grid to align values right of their labels, but it looks weird imo

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -30,25 +31,49 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
...
-      </p>
+      <div className="intro-text">
+        <div>Carefully review the information below to make sure everything looks correct.</div>
+      </div>
+      <div className="message">
+        {/* Details */}
+        <div>
+          <strong>Order Details</strong>
+        </div>
+        <div>
```
anxolin, line 42: nit: Can we add some indentation for the content?

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -30,25 +31,49 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
...
+        <p>
+          <div>
+            <strong>Prices</strong>
+          </div>
+          <div>
+            Direct: <span>{priceInverse}</span> <strong>{displaySellToken}</strong> per{' '}
+            <strong>{displayReceiveToken}</strong>
+          </div>
+          <div>
+            Inverse: <span>{price}</span> <strong>{displayReceiveToken}</strong> per <strong>{displaySellToken}</strong>
```
anxolin, line 60: I think it's strange. Let's just use the new approach we are using in other parts of the app. 
**Offer one price, and allow to invert it**. 

The format is nice (the TOKEN1 pero TOKEN2), but you are not coherent with the web, where we are currently showing BUY/SELL as default. So sell token is the quote. For the screenshot, it should say WETH per GUSD

> IMPORTANT: Please, let's keep out of the discussion the other change we have pending for analysing that uses the stable coin as the quote. I agree it should be the opposite to what I said above once we address this, but for now, it should be always BUY/SELL market

David. If you make a component for a price, that shows a price, the units an allows to invert it we can start using it everywhere and the moment we improve the quote token selection it will be easier to update

@Rafanator

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -30,25 +31,49 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
...
+        {/* Details */}
+        <div>
+          <strong>Order Details</strong>
+        </div>
+        <div>
+          Sell: <span>{sellTokenAmount}</span> <strong>{displaySellToken}</strong> for{' '}
+          <strong>{displayReceiveToken}</strong>
+        </div>
+        <div>
+          Receive: <UnderlinedText>at least</UnderlinedText> {receiveTokenAmount} <strong>{displayReceiveToken}</strong>
```
anxolin, line 47: That's not true if there's partial executions. 

How would you express better that:
```
You will receive at least XXX 567 GUSD if your order is completely executed
⚠️ Please, remember that your order can be also partially executed, or not executed at all. Read more about how Gnosis Protocol work [here].
```

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -30,25 +31,49 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
...
+      <div className="intro-text">
+        <div>Carefully review the information below to make sure everything looks correct.</div>
+      </div>
+      <div className="message">
+        {/* Details */}
+        <div>
+          <strong>Order Details</strong>
+        </div>
+        <div>
+          Sell: <span>{sellTokenAmount}</span> <strong>{displaySellToken}</strong> for{' '}
```
anxolin, line 43: I would remove the "for GUSD" you can easily see this in the next line, where we show the amount of GUSD you may get

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -30,25 +31,49 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
...
+        <p>
+          <div>
+            <strong>Order Validity Details</strong>
+          </div>
+          <div>
+            Starts: <span>{formatTimeInHours(validFrom || 0, 'Now')}</span>
+          </div>
+          <div>
+            Expires: <span>{formatTimeInHours(validUntil || 0, 'Never')}</span>
+          </div>
```
anxolin, line 74: Maybe it's a good idea to add a link here to https://docs.gnosis.io/protocol/docs/intro-batches/#orders

Proposal:
```
⚠️ Learn more about the validity of orders in Batch Exchange [here].
```

anxolin:
Is it possible to add something similar to this?

```
How is the order executed?
* After you confirm and send the ethereum transaction, your order will be active during the expecified validity time.
* During that time, it will be matched against other orders as long as there's an overlap in the limit prices.
* The limit price specified in this order will be respected, and **no fee will be applied on top of it**
* You can cancel the order at any point after it's creation
Check [here] for more information on Gnosis Protocol.
```

W3stside:
Open
Screenshot from 2020-08-04 13-43-38

Closed
Screenshot from 2020-08-04 13-44-01

PR #1263, opened 2020-07-28
Hotfix/1.1.2

Velenir:
Moves SENTINEL to config to be easily configurable


gnosis-info:
Travis automatic deployment:
 https://pr1263--dexreact.review.gnosisdev.com

README.md
```diff
@@ -289,3 +289,18 @@ walletConnect:
...
+**Config format:**
+
+```yaml
+transactions:
+  sentinel: string
+```
+
+Where:
+
+- `sentinel` is lowercase hex value appended to transaction input data and followed by on-chain analytics flags
```
anxolin, line 306: I think we should document more the analytics part. Saying how many bytes have the sentinel. 
Also, we need to document the codes for analytics that come after the sentinel.

config-default.yaml
```diff
@@ -117,3 +117,9 @@ disabledTokens:
     - address: '0x6Fb0855c404E09c47C3fBCA25f08d4E41f9F062f'
     - address: '0xFC4B8ED459e00e5400be803A9BB3954234FD50e3'
   4: [] # Rinkeby
+
+# modifiers for transactions
+transactions:
+  # lowercase hex value appended to transaction input data
+  # and followed by on-chain analytics flags
```
anxolin, line 124: Add something like:
IMPORTANT: If you fork the project, change this into another value. See Readme for details

src/utils/flagCodes.ts
```diff
@@ -8,7 +8,7 @@ export interface Flag<T extends string> {
 
 // OUR specific flags
 // lowercase hex value to be appended to tx.data
-const SENTINEL = 'dec0de'
+const SENTINEL = CONFIG.transactions.sentinel
```
anxolin, line 11: ubernit: Does it make sense to add process.env here so we can change this value for our test servers too for example if we want too?

anxolin, line 11: We should add raise an error if:
* It's not a hex
* If it's a bad hex?

We should define what is a good code, and verify it here

README.md
```diff
@@ -289,3 +289,39 @@ walletConnect:
...
+
+Sentinel can also be passed as an environment variable, e.g. `SENTINEL=4e3a yarn start`. In which case it overrides the `sentinel` from config file.
+
+Following the `sentinel` are 6 digits: 2 for provider name, 1 for mobile or desktop flag, 2 for browser name, 1 for screen size.
```
anxolin, line 310: Can you link to the line of code per list, to know the values. Or add a new README with them. If someone wants to understand a value, it would be great to be able to know how to do it. It's even good to add an example of a transaction

src/utils/flagCodes.ts
```diff
@@ -8,7 +8,12 @@ export interface Flag<T extends string> {
 
 // OUR specific flags
 // lowercase hex value to be appended to tx.data
-const SENTINEL = 'dec0de'
+const SENTINEL = process.env.SENTINEL || CONFIG.transactions.sentinel
+console.log('SENTINEL', SENTINEL)
+
+if (!/^[0-9a-f]+$/.test(SENTINEL)) {
+  throw new Error(`SENTINEL isn't valid. Expected lowercase hex value, got ${SENTINEL}`)
```
anxolin, line 15: Shouldn't we validate that SENTINEL is 6 characters?

Velenir: It doesn't have to be 6 characters. Do we mention 6 somewhere?

README.md
```diff
@@ -289,3 +289,39 @@ walletConnect:
 ```
 
 Alternatively you can set the bridge by declaring the env var `WALLET_CONNECT_BRIDGE` that would take precedence over the config.
+
+### transactions
+
+Transactions config. Used to apply some postprocessing toevery transaction submitted through the interface
```
alfetopito, line 295:
```suggestion
Transactions config. Used to apply some post processing to every transaction submitted through the interface
```

src/utils/flagCodes.ts
anxolin: One question. When we consume this information, and in case there are sentinels from multiple forks. How do we index that info too?
Since I was in the assumption that they had fixed number of chars, my thinking was take the last 12 chars, where 6 are the analytics and 6 are the sentinel.

Can the sentinel be 0, or 00000?

I understand that knowing the ABI we can interpret the hex in the data, and know if there's additional data, but I'm not so sure about how easy is that in dune, if for example the call can have a variable size input (like an array).

Velenir:
After some discussion, in the interest of more generic sentinel decoding and to be able to differentiate between separate deployments of dex-react, `CONFIG.transactions` will have `appId`, where:
 - `appId=0` development dex-react
 - `appId=1` Mesa deployment
 - `appId>=2` other apps, not necessarily of dex-react

```js
SENTINEL = 'dec0de' + String(appId).padStart(2, '0')
```
Therefore:
 - `SENTINEL=dec0de00` development dex-react
 - `SENTINEL=dec0de01` Mesa

webpack.config.babel.js
```diff
@@ -178,6 +178,7 @@ module.exports = ({ stats = false } = {}) => ({
       // AUTOCONNECT: only applies for mock implementation
       AUTOCONNECT: 'true',
       PRICE_ESTIMATOR_URL: process.env.PRICE_ESTIMATOR_URL || 'develop',
+      APP_ID: process.env.APP_ID,
```
alfetopito, line 181: I think that if you are not setting a default value for `APP_ID`, there's no need to have this line.
`process.env.INFURA_ID` for instance is used else where and it's not here.

```suggestion
```

Velenir: Not how it works. EnvironmentPlugin never iterates over `process.env` keys
https://github.com/webpack/webpack/blob/master/lib/EnvironmentPlugin.js#L16-L24
You can try console.logging it to make sure

So `INFURA_ID` was broken all this time? Oops :no_mouth: 
Maybe other envs too

webpack.config.babel.js
```diff
@@ -179,6 +179,10 @@ module.exports = ({ stats = false } = {}) => ({
       AUTOCONNECT: 'true',
       PRICE_ESTIMATOR_URL: process.env.PRICE_ESTIMATOR_URL || 'develop',
       APP_ID: process.env.APP_ID,
+      INFURA_ID: process.env.INFURA_ID,
+      WALLET_CONNECT_BRIDGEWALLET_CONNECT_BRIDGE: process.env.WALLET_CONNECT_BRIDGEWALLET_CONNECT_BRIDGE,
```
alfetopito, line 183: Twice the same sentence? is that right?

Velenir: lol, don't know how it snuck past me

src/utils/flagCodes.ts
```diff
@@ -7,8 +7,20 @@ export interface Flag<T extends string> {
 }
 
 // OUR specific flags
+const FIXED_SENTINEL = 'dec0de'
+
+const APP_ID = Number(process.env.APP_ID || CONFIG.transactions.appId)
+if (!Number.isInteger(APP_ID) || APP_ID < 1 || APP_ID > 99) {
+  throw new Error(`APP_ID isn't valid. Expected a positive integer <= 99, got ${APP_ID}`)
```
W3stside, line 14:
```suggestion
  throw new Error(`APP_ID isn't valid. Expected a positive integer >=1 and <= 99, got ${APP_ID}`)
```

Velenir: This is changing anyway with @anxolin's PR #1267

PR #1266, opened 2020-07-30
Remove fuse

anxolin:
I bumped into this while doing another PR, so aligned with not mixing unrelated things. Here is a simple one :)

Gets rid of our old name "Fuse"


gnosis-info:
Travis automatic deployment:
 https://pr1266--dexreact.review.gnosisdev.com

Issue #1227, opened 2020-07-14
Warn about min batch fee


alfetopito:
Reference https://github.com/gnosis/dex-services/issues/1099

Discussion https://gnosisinc.slack.com/archives/CSGLUC1BN/p1594629702005200

## Summary

- Add a warning message to UI indicating how much the sell amount must be in order to be more likely to get matched.
- Warning is dependent of the sell amount.
- Add FAQ entry explaining the min fee constraint

### To define
- Sell amount warning threshold
- How to calculate given amount
- FAQ message
- Warning message

### Proposed texts
**Warning msg:** `Only trades that with a sell amount >= x TOKEN are likely to be settled. See FAQ`

**FAQ:**
_Q:_ ` What is the minimum fee per order?`
_A:_ `On the solvers owned/controlled by Gnosis, there's a min fee requirement orf the whole batch to be of X OWL, to make it economically viable. Orders might be disregarded if the fee generated does not contribute to reaching the min X OWL goal. Other competing solvers might implement different strategies where this constraint may or may not apply`

josojo:
>What about orders between tokens like WETH or *BTC where it's not straight forward to infer the OWL relationship?
Do we need an additional quote query against OWL to be able to tell? (edited) 

The backend price estimation tool will give you the price of all tokens against OWL. 

> Where the x is calculated and TOKEN is replaced by with the current selltoken
How should this calculation be done?

It would be good if you guys could implement the same logic, as we will have in the backend, described here:


Here is one example:
```
subsidize_factor:=10
Gas price: 40 Gwei
OWL price per token: 240
=> min economical viable fee = 1.2 OWL
=> Handover to the solver: --min-avg-fee-per-order= 1.2/10
=> Min tradeable amount per token: 0.12 OWL *1000 * 3/2 / 240 (min_owl_fee * fee_factor * settlement_factor / price_of_token
```
The `settlement_factor` comes from the fact, that we also need to settle the fee order. Assuming that we have 2 orders in the batch being traded against each other and 1 fee order, the fee order will decrease the average_min_fee_per_order, and hence, the trading volume of the other orders need to compensate this and we scale their min. volume by the factor 3/2.

josojo:
Could we get this issue tackled next week?

alfetopito:
Sorry @josojo , was not able to get to it yet.
Will prioritize it for next release, yes.

anxolin:
Does the 2 decimals thing affect this? 
[...]

PR #1268, opened 2020-07-30
Low volume warning

Velenir:
Ignore styling, check logic only
Additionally, what would be the best way to get `MIN_ECONOMICAL_VIABLE_FEE_IN_OWL`:
1. as a fixed value

2. or calculate like in #1227
2.1. get ETH price In OWL from [market endpoint](https://github.com/gnosis/dex-services/tree/master/price-estimator#estimated-buy-amount)
2.2. => min economical viable fee = (120k gas)(40GWEI)(Etherprice in owl)=(120000 * 40 * 10**9 * 240 *1000) =1.2 OWL

Will the solver be run with `--min-avg-fee-per-order` as start-time flag (so a fixed fee), or will the fee be dynamic? @josojo 

localhost_8080_trade_WETH-USDC_sell=0 5 price=1 01361 from= expires=

### DIFF
--- a/src/api/wallet/WalletApi.ts
+++ b/src/api/wallet/WalletApi.ts
@@ -53,6 +53,7 @@ export interface WalletApi {
   getProviderInfo(): ProviderInfo
   blockchainState: BlockchainUpdatePrompt
   userPrintAsync: Promise<string>
+  getGasPrice(): Promise<number | null>
 }
 
 export interface WalletInfo {
@@ -249,6 +250,7 @@ export class WalletApiImpl implements WalletApi {
   public blockchainState: BlockchainUpdatePrompt
 
   private _unsubscribe: Command
+  private _fetchGasPrice: ReturnType<typeof fetchGasPriceFactory> = async () => undefined
 
   public constructor(web3: Web3) {
     this._listeners = []
@@ -349,6 +351,9 @@ export class WalletApiImpl implements WalletApi {
     closeOpenWebSocketConnection(this._web3)
 
     const fetchGasPrice = fetchGasPriceFactory(this)
+
+    this._fetchGasPrice = fetchGasPrice
+
     const earmarkingFunction = async (data?: string): Promise<string> => earmarkTxData(data, await this.userPrintAsync)
 
     const composedProvider = composeProvider(provider, { fetchGasPrice, earmarkTxData: earmarkingFunction })
@@ -455,6 +460,26 @@ export class WalletApiImpl implements WalletApi {
 
     logDebug('[WalletApiImpl] Disconnected')
     await this._notifyListeners()
+  }
+
+  public async getGasPrice(): Promise<number | null> {
+    // this never errors
+    // returns undefined if unable to fetch
+    let gasPrice = await this._fetchGasPrice()
+
+    if (gasPrice) return +gasPrice
+    try {
+      // fallback to gasPrice from provider
+      // {"jsonrpc":"2.0","method":"eth_gasPrice"} call
+      gasPrice = await this._web3.eth.getGasPrice()
+
+      if (gasPrice) return +gasPrice
+    } catch (error) {
+      console.error('Error fetching gas price', error)
+    }
+
+    // unable to fetch
+    return null
   }
 
   public async getAddress(): Promise<string> {
--- a/src/api/wallet/WalletApi.ts
+++ b/src/api/wallet/WalletApi.ts
@@ -13,7 +13,7 @@ import { logDebug, toBN, txDataEncoder, generateWCOptions } from 'utils'
 import { subscribeToWeb3Event } from './subscriptionHelpers'
 import { getMatchingScreenSize, subscribeToScreenSizeChange } from 'utils/mediaQueries'
 import { composeProvider } from './composeProvider'
-import fetchGasPriceFactory from 'api/gasStation'
+import fetchGasPriceFactory, { GasPriceLevel } from 'api/gasStation'
 import { earmarkTxData } from 'api/earmark'
 import { Provider, isMetamaskProvider, isWalletConnectProvider, ProviderRpcError } from './providerUtils'
 
@@ -53,7 +53,7 @@ export interface WalletApi {
   getProviderInfo(): ProviderInfo
   blockchainState: BlockchainUpdatePrompt
   userPrintAsync: Promise<string>
-  getGasPrice(): Promise<number | null>
+  getGasPrice(gasPriceLevel?: GasPriceLevel): Promise<number | null>
 }
 
 export interface WalletInfo {
@@ -462,10 +462,10 @@ export class WalletApiImpl implements WalletApi {
     await this._notifyListeners()
   }
 
-  public async getGasPrice(): Promise<number | null> {
+  public async getGasPrice(gasPriceLevel?: GasPriceLevel): Promise<number | null> {
     // this never errors
     // returns undefined if unable to fetch
-    let gasPrice = await this._fetchGasPrice()
+    let gasPrice = await this._fetchGasPrice(gasPriceLevel)
 
     if (gasPrice) return +gasPrice
     try {
--- a/src/components/TradeWidget/TxMessage.tsx
+++ b/src/components/TradeWidget/TxMessage.tsx
@@ -10,6 +10,7 @@ import BigNumber from 'bignumber.js'
 import { ZERO_BIG_NUMBER } from 'const'
 
 import alertIcon from 'assets/img/alert.svg'
+import { useGasPrice } from 'hooks/useGasPrice'
 
 interface TxMessageProps {
   sellToken: TokenDetails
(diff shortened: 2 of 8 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1268--dexreact.review.gnosisdev.com

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,84 @@
...
+  box-shadow: inset 0 0 7px 3px #cc262647;
+  border-radius: 0.3em;
+  display: flex;
+
+  .alert {
+    min-width: 2.5em;
+  }
+`
+
+const MIN_ECONOMICAL_VIABLE_FEE_IN_OWL = 1.2
```
josojo, line 36: If you wanna make it more "sophisticated", then you are querying the current gas price for a fast transaction and then calculate this number `MIN_ECONOMICAL_VIABLE_FEE_IN_OWL`

While this would still not be perfect, as the gas prices can change quickly, it might be a little bit better .
If you think that it is over-engineered, I am also fine with the current version :)

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,98 @@
...
+  box-shadow: inset 0 0 7px 3px #cc262647;
+  border-radius: 0.3em;
+  display: flex;
+
+  .alert {
+    min-width: 2.5em;
+  }
+`
+
+const DEFAULT_GASP_RICE = 40e9 // 40 Gwei
```
anxolin, line 37: should we move it to the constants?
Why do we need a default?

src/hooks/useGasPrice.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect } from 'react'
+import useSafeState from './useSafeState'
+import { walletApi } from 'api'
+
+export const useGasPrice = (defaultGasPrice: number | null = null): number | null => {
+  const [gasPrice, setGasPrice] = useSafeState<number | null>(null)
+  useEffect(() => {
+    async function getGasPrice(): Promise<void> {
+      const gasPrice = (await walletApi.getGasPrice()) || defaultGasPrice
```
anxolin, line 9: use API better

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,98 @@
...
+const ETH_PRICE_IN_OWL = 240 * 1000
+const SUBSIDIZE_FACTOR = 10
+
+const SETTLEMENT_FACTOR = 1.5
+const FEE_FACTOR = 1000
+
+const calcMinTradableAmountInOwl = (gasPrice: number): BigNumber => {
+  //                            trade tx gasLimit
```
anxolin, line 45: not enough left spaces :)

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,98 @@
...
+  const MIN_ECONOMICAL_VIABLE_FEE_IN_OWL = 120000 * gasPrice * ETH_PRICE_IN_OWL
```
anxolin, line 46: Calculation should be moved to an util.

Unit testing plss

anxolin: Magic number

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,98 @@
...
+
+  const MIN_FEE = MIN_ECONOMICAL_VIABLE_FEE_IN_OWL / SUBSIDIZE_FACTOR
```
anxolin, line 48: I would also add a buffer. Like multiply 

`MIN_FEE = BUFFER_TO_ACCOUNT_FOR_TOKEN_AND_GAS_PRICE_CHANGES  * MIN_ECONOMICAL_VIABLE_FEE_IN_OWL / SUBSIDIZE_FACTOR`

i.e.
`BUFFER_TO_ACCOUNT_FOR_TOKEN_AND_GAS_PRICE_CHANGES  = 1.20`   ---> account for a 20% change in the time it takes to mine the tx, and start the batch

Of course with a shorter name, I just made it loooong for illustration purposes

src/components/TradeWidget/TxMessage.tsx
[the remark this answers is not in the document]
Velenir: But of course, for :curry:

Velenir: In case both gasStation and `eth_gasPrice` fail
I want to have some kind of lowVolume estimate even then

Also don't think this default should be app wide, unless we really need an app wide gasPrice
This one is good for estimates, but app-wide we better leave it to provider on each tx.
So as not to encourage app-wide use, let it stay local.

src/api/wallet/WalletApi.ts
```diff
@@ -457,6 +462,26 @@ export class WalletApiImpl implements WalletApi {
...
+  public async getGasPrice(): Promise<number | null> {
+    // this never errors
+    // returns undefined if unable to fetch
+    let gasPrice = await this._fetchGasPrice()
+
+    if (gasPrice) return +gasPrice
+    try {
+      // fallback to gasPrice from provider
+      // {"jsonrpc":"2.0","method":"eth_gasPrice"} call
+      gasPrice = await this._web3.eth.getGasPrice()
```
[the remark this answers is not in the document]
Velenir: We have `api/gasStation`, and it depends on WalletApi
1. I don't want to init it separately and pass walletApi to it. Since result of `fetchGasPriceFactory(walletApi)` needs to go inside `walletApi.connect` to be used in composed provider
2. Previously that kind of interdependence caused cyclic deps.
3. [L353](https://github.com/gnosis/dex-react/blob/313518ef08fb26eb809195fa07010a22b4e6d49b/src/api/wallet/WalletApi.ts#L353) keeps dependencies flatter

src/hooks/useGasPrice.ts
Velenir: I would prefer `walletApi.getGasPrice()` to be able to return `null` in case of failure, to better inform us whether it failed or not
And then we can decide to use default or not, in place where it really matters, like here

Velenir:
Allgood comments, but I'm more interested in opinions on
>what would be the best way to get MIN_ECONOMICAL_VIABLE_FEE_IN_OWL:
>
>as a fixed value
>
>or calculate like in #1227
>2.1. get ETH price In OWL from market endpoint
>2.2. => min economical viable fee = (120k gas)(40GWEI)(Etherprice in owl)=(120000 * 40 * 10**9 * 240 *1000) =1.2 OWL

alfetopito:
> All good comments, but I'm more interested in opinions on
> 
> > what would be the best way to get MIN_ECONOMICAL_VIABLE_FEE_IN_OWL:
> > as a fixed value
> > or calculate like in #1227
> > 2.1. get ETH price In OWL from market endpoint
> > 2.2. => min economical viable fee = (120k gas)(40GWEI)(Etherprice in owl)=(120000 * 40 * 10**9 * 240 *1000) =1.2 OWL

2 (calculate).

src/utils/minFee.ts
```diff
@@ -0,0 +1,19 @@
+import { BigNumber } from 'bignumber.js'
+
+export const DEFAULT_GAS_PRICE = 40e9 // 40 Gwei
+export const ETH_PRICE_IN_OWL = 240 * 1000
```
alfetopito, line 4: That's not really accurate, is it?

Velenir: super not accurate, especially the last few days 📈 
as a placeholder for now

src/api/wallet/WalletApi.ts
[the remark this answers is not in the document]
anxolin: Aha! that logic is only if the gas API fails. Then all good.

I'm not a fan of using this kind of fallbacks, at least by default, cause it can bring bugs that are very difficult to reproduce. Gas price estimator is an important service for the app, and I would be OK assuming it work (or fail, or we handle the case)

Anyways, in this case, it's not a super strong opinion

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -1,21 +1,94 @@
...
+  return useMemo(() => {
+    if (
+      isPriceLoading ||
+      isWETHPriceLoading ||
+      priceEstimation === null ||
+      wethPriceInOwl === null ||
+      gasPrice === null
+    )
+      return { isLoading: true }
+    console.log('priceEstimation of', sellToken.symbol, 'in OWL', priceEstimation.toString(10))
```
alfetopito, line 71: please wrap in braces, very easy to miss if has a single statement

```suggestion
    if (
      isPriceLoading ||
      isWETHPriceLoading ||
      priceEstimation === null ||
      wethPriceInOwl === null ||
      gasPrice === null
    ) {
      return { isLoading: true }
    }

    console.log('priceEstimation of', sellToken.symbol, 'in OWL', priceEstimation.toString(10))
```

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -49,6 +124,15 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken })
         Your order starts {formatTimeInHours(validFrom || 0, 'now')}{' '}
         {validUntil && `and will expire ${formatTimeInHours(validUntil, 'Never')}`}
       </p>
+      {!isLoading && isLowVolume && (
+        <Warning>
+          <span>
+            This is a small order. Please keep in mind that solvers might not include your order if it does not generate
```
alfetopito, line 130: Not sold on the wording. Maybe:

```suggestion
            This is a low volume order. Please keep in mind that solvers might not include your order if it does not generate
```

src/hooks/usePriceEstimation.ts
```diff
@@ -112,3 +112,43 @@ export function usePriceEstimationWithSlippage(params: SlippageParams): Result {
 
   return { priceEstimation, isPriceLoading }
 }
+
```
alfetopito, line 115: I created a new hook for exactly this https://github.com/gnosis/dex-react/blob/a9dfb5a33854f9c797f5e2ff31ae77c580b99dd0/src/hooks/useOwlAmountInBaseTokenUnits.ts

Maybe they could be merged somehow?

src/utils/minFee.ts
```diff
@@ -0,0 +1,31 @@
...
+  gasPrice,
+  ethPriceInOwl,
+}: CalcMinTradableAmountInOwlParams): BigNumber => {
+  const MIN_ECONOMICAL_VIABLE_FEE_IN_OWL = ethPriceInOwl
+    .multipliedBy(TRADE_TX_GASLIMIT * gasPrice)
+    .dividedBy(OWL_DECIMAL_UNITS)
+  console.log('MIN_ECONOMICAL_VIABLE_FEE_IN_OWL', MIN_ECONOMICAL_VIABLE_FEE_IN_OWL.toString(10))
+
+  const MIN_FEE = MIN_ECONOMICAL_VIABLE_FEE_IN_OWL.multipliedBy(BUFFER_MULTIPLIER).dividedBy(SUBSIDIZE_FACTOR)
+  return MIN_FEE.multipliedBy(FEE_FACTOR * SETTLEMENT_FACTOR)
```
alfetopito, line 30: minor nitpick: no need for caps, not constants
I mean, they are `const`, but you got the point.

src/hooks/usePriceEstimation.ts
Velenir: mine is more generic :robot:

PR #1302, opened 2020-08-04
add getLastCreditBatchId to depositApi & useTokenBalances claimable check

### LINKED ISSUE #1136: Withdraw edge case
image

W3stside:
Closes #1136 

add getLastCreditBatchId to depositApi & useTokenBalances claimable check

Not sure how best to check, any ideas?

### DIFF
--- a/src/hooks/useTokenBalances.ts
+++ b/src/hooks/useTokenBalances.ts
@@ -39,7 +39,6 @@ async function fetchBalancesForToken(
     pendingDeposit,
     pendingWithdraw,
     currentBatchId,
-    lastCreditBatchId,
     walletBalance,
     allowance,
   ] = await Promise.all([
@@ -47,7 +46,6 @@ async function fetchBalancesForToken(
     depositApi.getPendingDeposit({ userAddress, tokenAddress, networkId }),
     depositApi.getPendingWithdraw({ userAddress, tokenAddress, networkId }),
     depositApi.getCurrentBatchId(networkId),
-    depositApi.getLastCreditBatchId({ userAddress, tokenAddress, networkId }),
     erc20Api.balanceOf({ userAddress, tokenAddress, networkId }),
     erc20Api.allowance({ userAddress, tokenAddress, networkId, spenderAddress: contractAddress }),
   ])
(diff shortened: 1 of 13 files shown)


src/hooks/useTokenBalances.ts
```diff
@@ -57,7 +59,9 @@ async function fetchBalancesForToken(
     totalExchangeBalance: calculateTotalBalance(exchangeBalance, currentBatchId, pendingDeposit),
     pendingDeposit,
     pendingWithdraw,
-    claimable: pendingWithdraw.amount.isZero() ? false : pendingWithdraw.batchId < currentBatchId,
+    claimable: pendingWithdraw.amount.isZero()
+      ? false
+      : lastCreditBatchId !== currentBatchId && pendingWithdraw.batchId < currentBatchId,
```
anxolin, line 64: I think it's good to expose also `creditedOnLastBatch` and `withdrawHasMatured` so we can handle it better in the UI

gnosis-info:
Travis automatic deployment:
 https://pr1302--dexreact.review.gnosisdev.com

alfetopito:
> I like that it was added in the balance, and the logic is correct, however I have a small concerns about performance
> 
> This is adding 1 call per token on every block. Is there a way to not do this so often? normally only if we trade can change

What about this extra call is only performed in case there's a pending withdraw?
Might make those calls slower, but won't be done at every block for every token.

W3stside:
Screenshot from 2020-08-12 10-54-27

anxolin:
If we have already the "lastCreditedBatchId", why do we fail only when the user tries to claim?
Wouldn't this be better in the request withdraw modal before the user does the request?

W3stside:
> ~If we have already the "lastCreditedBatchId", why do we fail only when the user tries to claim? Wouldn't this be better in the request withdraw modal before the user does the request?~
> 
> edit: Scratch the above. The error is when we claim, not when we do the revert.
> If we have the state already, why not representing this edge case differently, so the user don't need to click in the button.
> 
> For example, instead of "0.1 [claim]" we show the clock and the number. not sure if we can do something else to make the user more aware that this case is different than the "maturing the trade" (although what he has to do is the same, wait :P )

Unless I'm not thinking about it carefully, this approach puts us back to where we were intially - having the `claim` status checked against `lastCreditedBatchId` inside `useTokenBalances` which we voted against as it was intensive on node calls.

We agreed to put it inside the `onClaim` function which is what we have here - I don't see  a way of doing what you suggested without reverting to the original changes since the `isClaimable` variable is derived/calculated inside `useTokenBalances`

PR #1303, opened 2020-08-04
1040/update orderbook automatically

### LINKED ISSUE #1040: Update orderbook automatically

alfetopito:
Closes #1040 

Depends on #1196 

Currently set to refresh every 30s


gnosis-info:
Travis automatic deployment:
 https://pr1303--dexreact.review.gnosisdev.com

src/components/OrderBookChart/index.tsx
```diff
@@ -131,6 +130,19 @@ export const Chart: React.FC<ChartProps> = props => {
       networkId,
     })
 
+    let firstLoad = true
```
W3stside, line 133: can't this be achieved the same essentially with `onComponentMount` via

`useEffect(() => { ... }, [])`?

this is all this is doing no?

src/components/OrderBookChart/index.tsx
```diff
@@ -147,11 +159,10 @@ export const Chart: React.FC<ChartProps> = props => {
 
         const initialZoom = calcInitialZoom(bids, asks)
 
+        // Zoom in, only if this is the first load of this token pair
+        adjustZoomOnFirstLoad(initialZoom)
```
W3stside, line 163: ah by token pair, could also be done with an effect based on token choice i guess. anyways this works too

src/components/OrderBookChart/index.tsx
alfetopito: Unless there's a way I'm not seeing, I don't think so.

The reasoning is:

I only want to change the zoom, when the data is loaded for the first time, to center the graph.
The zoom coordinates are calculated every time the data is loaded, in case the user wants to recenter it after playing around.
If I update the zoom in another `useEffect`, I need to depend on the zoom coordinates.
If I depend on the coordinates, the `useEffect` will trigger every time it changes, which is every time the data is loaded, which is not what I want.
Thus, the `firstLoad` flag tracks when it has been loaded already and prevents resetting the zoom.

@W3stside do you see another way of doing that?

anxolin:
I think this option solves the issue, but then. The problem is the Zoom.
image

My understanding is that refresh with new data messes up with the zoom. If that's the case, identifying the start price and end price you are displaying and maintaining that fixed between refreshes is an option?
in other words, instead of mantaining the zoom, would it make sense to calculate how to zoom in order to have a given value as min price, and another value as max price

alfetopito:
> I think this option solves the issue, but then. The problem is the Zoom.
> image
 
I did try it out before but gave it up pretty quickly.
The data was not updated properly. It was weird...

I did again a quick change. It prevents flickering on the graph, but I can see flickering on the minimap on top.
Might need more effort into it to make it work as expected.

> My understanding is that refresh with new data messes up with the zoom. If that's the case, identifying the start price and end price you are displaying and maintaining that fixed between refreshes is an option?
> in other words, instead of mantaining the zoom, would it make sense to calculate how to zoom in order to have a given value as min price, and another value as max price

That might be an option. Right now, I'm keeping the same zoom level simply by not changing anything in the view when new data is loaded.

Anyway, seems to be working the way it is.
Take a look now.

Also, reduced update frequency to 5s.

PR #1306, opened 2020-08-06
Fix EnvironmentPlugin warning

Velenir:
screenshot_2020-08-04_12-58-43

Setting env variables by default to `null` as per [EnvPlugin docs](https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values:~:text=To%20specify%20an%20unset%20default%20value%2C%20use%20null%20instead%20of%20undefined)
Though it doesn't make them _unset_ in the end, semantically fits

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -179,11 +179,11 @@ module.exports = ({ stats = false } = {}) => ({
       // AUTOCONNECT: only applies for mock implementation
       AUTOCONNECT: 'true',
       PRICE_ESTIMATOR_URL: process.env.PRICE_ESTIMATOR_URL || 'develop',
-      APP_ID: process.env.APP_ID,
-      INFURA_ID: process.env.INFURA_ID,
-      WALLET_CONNECT_BRIDGE: process.env.WALLET_CONNECT_BRIDGE,
-      ETH_NODE_URL: process.env.ETH_NODE_URL,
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1306--dexreact.review.gnosisdev.com

webpack.config.babel.js
```diff
@@ -179,11 +179,11 @@ module.exports = ({ stats = false } = {}) => ({
       // AUTOCONNECT: only applies for mock implementation
       AUTOCONNECT: 'true',
       PRICE_ESTIMATOR_URL: process.env.PRICE_ESTIMATOR_URL || 'develop',
-      APP_ID: process.env.APP_ID,
-      INFURA_ID: process.env.INFURA_ID,
-      WALLET_CONNECT_BRIDGE: process.env.WALLET_CONNECT_BRIDGE,
-      ETH_NODE_URL: process.env.ETH_NODE_URL,
-      LIQUIDITY_TOKEN_LIST: process.env.LIQUIDITY_TOKEN_LIST,
```
alfetopito, line 186: Question: env values are still read in case they are passed in, right?
Only when not set they receive null, correct?

Velenir: yes, basically `process.env.LIQUIDITY_TOKEN_LIST = process.env.LIQUIDITY_TOKEN_LIST || null`

Issue #1307, opened 2020-08-06
Orderbook not showing worst bid order.


nlordell:
It looks like the orderbook is not showing the worst bid order. If I setup a mock price estimation service that returns the following for the WETH-USDC orderbook:
```json
{
  "asks":[
    {
      "price":4.004004004004005e-10,
      "volume":"1500000000000000000"
    },
    {
      "price":4.1041041041041036e-10,
      "volume":"500000000000000000"
    }
  ],
  "bids":[
    {
      "price":3.94605e-10,
      "volume":"999999999999999872"
    },
    {
      "price":3.8960999999999996e-10,
      "volume":"2000000000000000000"
    }
  ]
}
```

For the following orders:
```
order buy 1   ETH  for 395 USDC
order buy 2   ETH  for 780 USDC

order buy 600 USDC for 1.5 ETH
order buy 205 USDC for 0.5 ETH
```

The following graph shows up:

Notice that you only really see 1 bid order (there is a hidden second 1 with a volume of 0, although the tooltip is correct).

![image](https://user-images.githubusercontent.com/4210206/89531729-faea3f80-d7f0-11ea-9811-14bdd5e4ab0c.png)

However, I would expect somthing more like this:

![image](https://user-images.githubusercontent.com/4210206/89531860-2cfba180-d7f1-11ea-82a0-06da56c85fdc.png)

anxolin:
I don't understand. Why you say bid volume 0, and not something like:
![image](https://user-images.githubusercontent.com/2352112/89631547-35abb080-d8a1-11ea-8300-50dd30998f3c.png)

nlordell:
@anxolin Sorry! that was an artifact from the screenshot. I was refering to the shape of the curve. Notice how in the first image, it looks like there is only one bid order despite there being two bids, however, the curve should look like the second example where you clearly see the two bids and the total volume is 3.

For reference, here are the two bid orders:
```
order buy 1   ETH  for 395 USDC
order buy 2   ETH  for 780 USDC
```

alfetopito:
Closed in favor of https://github.com/gnosis/gp-ui/issues/11

PR #1308, opened 2020-08-06
Add an 'empty' price point so that all bid prices are included in graph

nlordell:
This draft PR is just to illustrate a bandaid solution to the issue described in #1307 by adding an empty bid price point to make sure the last pricepoint (i.e. worst bid order) get correctly displayed. I don't know if this has a real impact for the front-end since the worst bids are generally very bad and unimportant.

Also, I imagine that when using a linear orderbook such as in #1196 that this will no longer become an issue.


gnosis-info:
Travis automatic deployment:
 https://pr1308--dexreact.review.gnosisdev.com

PR #1316, opened 2020-08-13
Clickable datapoints

Velenir:
Adds clickable datapoints to Orderbook chart
To set price from there

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -9,14 +9,14 @@ import Modali from 'modali'
 import BN from 'bn.js'
 import { isAddress } from 'web3-utils'
 
-import { encodeTokenSymbol, decodeSymbol } from '@gnosis.pm/dex-js'
+import { encodeTokenSymbol, decodeSymbol, invertPrice } from '@gnosis.pm/dex-js'
 
 // assets
 import { SwitcherSVG } from 'assets/img/SVG'
 import arrow from 'assets/img/arrow.svg'
 
 // const, types
-import { ZERO } from 'const'
+import { ZERO, PRICE_ESTIMATION_PRECISION } from 'const'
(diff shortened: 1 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1316--dexreact.review.gnosisdev.com

src/components/TradeWidget/index.tsx
```diff
@@ -488,6 +489,20 @@ function buildUrl(params: {
...
+}
+
+const OrderbookContext = React.createContext<OrderbookContextI>({
+  setSellToken: () => void 0,
+  setReceiveToken: () => void 0,
+  setPrice: () => void 0,
+})
+
+export const useOrderbookContext = (): OrderbookContextI => React.useContext(OrderbookContext)
+
```
alfetopito, line 505: nit: move to a hook file?

Velenir:
> Shouldn't the volume be filled in as well?
Do we want that? A user may want to sell specific amount and went to orderbook to look up a good price

> whatever is currently displayed in the tooltip be set. Can that be done?

I wonder. Tooltip isn't clickable by itself, it seems. And just clicking anywhere gets the whole chart as an `env.target`. Therefore dataPoints is the only specific point that is directly clickable. StepLineSeries uses those same dataPoints to draw steps, and to sticky tooltips to. Just as we can't put a datapoint in the middle of a step, can make it a  clickable target.

And if we just detect what is in the open tooltip when we click anywhere it will lead to misclicks as tooltips snap to steps, and cursor can be rather far off

Velenir:
We should discuss all this separately.
Therefore this is a draft to see what is possible

> if the user clicks in a point in the graph, one idea to iterate is, he sets a horizontal and vertical line

And then the user would like to move those lines, and that's a whole lot of extra code

> I don't think you are handling the case that the user changes the tokens in the current modal

Handling already, by changing URl and tokens

PR #1318, opened 2020-08-13
1297/balance-page-is-so-wide-css-makes-it-less-wide

W3stside:
Brings Balance page back to respecatable widths (only really applied on mainnet when token names got stupid long)

I *know* it sounds ugly, but give it a go - `overflow-x: auto` on token names

It comes with fancy scrollbars tho so it's great

image

### DIFF
--- a/src/components/DepositWidget/index.tsx
+++ b/src/components/DepositWidget/index.tsx
@@ -80,6 +80,44 @@ export const BalancesWidget = styled(CardWidgetWrapper)`
             letter-spacing: 0;
             line-height: 1.2;
             flex-flow: row nowrap;
+
+            > div {
+              overflow-y: auto;
+              padding: 0.3rem 0;
+              white-space: nowrap;
+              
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1318--dexreact.review.gnosisdev.com

W3stside:
> Yep, I think we have enough space to let the first colum grow a bit. and if it's not enough, maybe is better to let the name jump to the next line, or add a ellipsis if it's really really long
> 
> In this case, there's more than enough space
> image

yes but i guarantee that the amounts will do the same thing in the future - at least this approach it is 100% that the spillover is contained somewhere inside the scrolling - agree that it isn't the most pleasant UI. ok will try widening row instead with ellipsis

alfetopito:
Agree with Dima and Anxo. Would prefer an ellipsis and show the full name on title hover.

W3stside:
closing in favour of #1325

PR #1324, opened 2020-08-17
Recommend min sell amount

### LINKED ISSUE #1322: Show min amount for lowVolume warning
When displaying a lowVolume warning :warning: (#1268) also show the minimum amount that would satisfy the solver

Round that amount up to the next 500$, or 250$

Velenir:
Closes #1322 

Message is up for discussion, as well as further styling, of course.
rec

### DIFF
--- a/src/components/TradeWidget/TxMessage.tsx
+++ b/src/components/TradeWidget/TxMessage.tsx
@@ -196,8 +196,12 @@ const useLowVolumeAmount = ({ sellToken, sellTokenAmount, networkId }: LowVolume
 }
 
 export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken, networkId }) => {
+  // simple watch('sellToken') doesn't work
+  // as not every setValue causes rerender
+  const [, forceUpdate] = useSafeState({})
+
   const [orderHelpVisible, showOrderHelp] = useSafeState(false)
-  const { getValues } = useFormContext<TradeFormData>()
+  const { getValues, setValue } = useFormContext<TradeFormData>()
   const {
     price,
     priceInverse,
@@ -214,6 +218,17 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken, n
     networkId,
     sellTokenAmount,
   })
+
+  const adjustedRecommendedAmount = useMemo(
+    () => (recommendedAmount ? adjustPrecision(recommendedAmount?.toString(), 2) : ''),
+    [recommendedAmount],
+  )
+
+  const setRecommendedSellAmount = (): void => {
+    if (!adjustedRecommendedAmount) return
+    setValue('sellToken', adjustedRecommendedAmount, true)
+    forceUpdate({})
+  }
 
   return (
     <TxMessageWrapper>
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1324--dexreact.review.gnosisdev.com

Velenir:
> I would add the rounding, to next 500$ block

Velenir:
> Also, please review the price, it is styled as a LINK

I don't understand. Price is just a `<span>` no specific styling

Velenir:
Added display in $
But I maintain that we need to show the actual `sellAmount` we allow to click on, because $1500 doesn't automatically translate to token amount. User sets token amount, so it should be displayed directly and clickable, not some abstraction

localhost_8080_trade_DAI-USDC_sell=1 price=1 00885 from= expires=

anxolin:
Aha, I see your point. You are right, but then I would show the units in the message:

`We recommend selling at least 1488.05 DAI (approximately $1500).`

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -194,14 +220,34 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken, n
...
+  })
+
+  const adjustedRecommendedAmount = useMemo(
+    () => (recommendedAmount ? adjustPrecision(recommendedAmount?.toString(), 2) : ''),
+    [recommendedAmount],
+  )
+
+  const setRecommendedSellAmount = (): void => {
+    if (!adjustedRecommendedAmount) return
+    setValue('sellToken', adjustedRecommendedAmount, true)
```
W3stside, line 241: i think this may be fixed in v6 with `setValue('sellToken', adjustedRecommendedAmount, { shouldDirty: true })` but im not sure

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -271,20 +317,30 @@ export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken, n
...
+              <a onClick={setRecommendedSellAmount}>
+                <strong>
+                  {adjustedRecommendedAmount} {symbolOrAddress(sellToken)}
+                </strong>
+              </a>{' '}
+              (approximately <strong>${roundedAmountInUSD?.toString(10)}</strong>) of the token.
+            </p>
+            <p>
+              Please keep in mind that solvers may not include your order if it does not generate enough fees to pay
+              their running consts. Learn more{' '}
```
nlordell, line 333:
```suggestion
              their running costs. Learn more{' '}
```

src/utils/minFee.ts
```diff
@@ -30,3 +30,17 @@ export const calcMinTradableAmountInOwl = ({
   const minFee = minEconomicalViableFeeInOwl.multipliedBy(BUFFER_MULTIPLIER).dividedBy(SUBSIDIZE_FACTOR)
```
nlordell, line 30: The `SUBSIDIZE_FACTOR` was recently changed to `5`. Maybe this computation is something that our services can expose as part of a REST API?

anxolin: Yep, in reality we were told we should delete it, so change it to "1".
@Velenir  will soon share one PR (demo) on the what he calls "Traders go away, don't trade, PR" and removes any subsidy

anxolin: > The SUBSIDIZE_FACTOR was recently changed to 5

In production too? this might be an issue for us, since we set the expectation with "10" in the current web
@Velenir we need to make sure the subsidize reduction hit 1.3.0 too

PR #1329, opened 2020-08-19
Subsidize factor change

Velenir:
Setting subsidize factor to 1
pr1329--dexreact review gnosisdev com_trade_DAI-USDC_sell=10 price=1 01256 from= expires=

### DIFF
--- a/src/components/TradeWidget/TxMessage.tsx
+++ b/src/components/TradeWidget/TxMessage.tsx
@@ -203,12 +203,8 @@ const useLowVolumeAmount = ({ sellToken, sellTokenAmount, networkId }: LowVolume
 }
 
 export const TxMessage: React.FC<TxMessageProps> = ({ sellToken, receiveToken, networkId }) => {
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1329--dexreact.review.gnosisdev.com

anxolin (approved):
Thanks!

This will scare any trader, but it is the way it is...

anxolin:
Minor comments

Make sure we format using the thousands separator. Also, the functionality of clicking in the DAI amount to change the sell amount, I think is not a good idea:
- functionally, this modal is just a confirmation, only options is to accept or not
- we shouldn't change sth in the back of the screen, sure we show it in the modal too, but we cannot change it to any other amount
- the price could be not valid any more for the given amount. Note the price depend on the volume, therefore if u increase the volume u decrease the price. But you are maintaining it

W3stside (approved):

PR #1341, opened 2020-08-24
easier logout process

W3stside:
Light mobile:
Screenshot from 2020-08-24 11-36-32

Light web:
Screenshot from 2020-08-24 11-36-40

Dark mobile:
Screenshot from 2020-08-24 11-38-21

### DIFF
--- a/src/components/Layout/Header/Header.styled.ts
+++ b/src/components/Layout/Header/Header.styled.ts
@@ -24,6 +24,7 @@ export const HeaderWrapper = styled.header`
   ${UserWalletWrapper} {
     order: 1;
     margin: 3.2rem 3rem 3.2rem 0;
+    width: auto;
 
     @media ${MEDIA.mobile} {
       margin: 2rem 1rem 2rem 0;
--- a/src/components/Layout/Header/index.tsx
+++ b/src/components/Layout/Header/index.tsx
@@ -44,6 +44,7 @@ const CountDownStyled = styled.div`
   display: flex;
   flex-flow: column;
   order 2;
+  min-width: 15rem;
 
  > div {
   display: flex;
@@ -88,10 +89,10 @@ const BatchCountDown: React.FC = () => {
   return (
     <CountDownStyled>
       <div>
-        Next batch in: <strong>{formatSeconds(timeRemainingInBatch)}</strong>
+        Next batch: <strong>{formatSeconds(timeRemainingInBatch)}</strong>
       </div>
       <div>
-        Current batch: <strong>{dateToBatchId()}</strong>
+        Batch: <strong>{dateToBatchId()}</strong>
         &nbsp;
         <HelpTooltip tooltip={DevdocTooltip} />
       </div>
--- a/src/components/UserWallet/UserWallet.styled.ts
+++ b/src/components/UserWallet/UserWallet.styled.ts
@@ -12,7 +12,7 @@ export const UserWalletWrapper = styled.div`
   position: relative;
 
   .QRCode {
-    margin: 0 auto;
+    margin: 4rem auto;
   }
 `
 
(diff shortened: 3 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1341--dexreact.review.gnosisdev.com

Velenir (approved):
Veeery beautiful :sparkles: 

But dark-on-dark QR can be a problem. For one, my phone can't read it. Consider adding white `padding|border|outline|box-shadow`

anxolin (commented):
Thanks for this! Makes thing better.

What do you think about leaving only one button (make the etherscan thing just a link):

image

Also, I wouldn't use the magnifying glass icon, since this is associated with search. Could u use the same Metamask uses? 
image

Please, take in consideration Dima's comment in darkmode, we had issues before.
Probably works if you add the qr in a small white box with just a super small padding.

anxolin:
Also, consider moving the icon up, and make it a bit wider and with slightly more padding
the reasoning is:
- if it's wider and with more padding, is more prominent and easy to click, but most importantly it doesn't create one strange effect when it resizes while it shows "logging out" message
- Moving up is because the toast messages cover the button in responsive mode. See image:

IMG_6582

IMG_6583

W3stside:
@Velenir @anxolin I prefer the link as a button, but I'll leave it as a vote

button: :stop_button: 
link: :link:

W3stside:
which u cant do directly on the comment cause GH is dumb

Vote: :stop_button:

alfetopito:
link

W3stside:
> link

Excuuuuse me did I ask you? :tipping_hand_woman:

alfetopito:
> > link
> 
> Excuuuuse me did I ask you? tipping_hand_woman

Too late now, public page :shrug:

anxolin:
link, with same icon as MM :P

anxolin (changes requested):
still not address issues, like move the to the top, and make the etherscan link a link. It doesn't need so much protagonism

W3stside:
> still not address issues, like move the to the top, and make the etherscan link a link. It doesn't need so much protagonism

didn't understand the `move the icon to the top` part, top of the toast notification? which I did, or move the whole wallet card to the top of the screen? Guessing the latter is what u meant now. For the link i was checking as I wanted to see what others wanted, can't always make what you suggest without asking! I prefer the button

anxolin:
> move the icon to the top part, top of the toast notification

I was just flagging that when you have any toast, the button is not clickable, and maybe moving it to a upper position would solve the issue.
image

> For the link i was checking as I wanted to see what others wanted, can't always make what you suggest without asking! I prefer the button

I didn't mean you blindly do what I suggest. I try to encourage just the opposite. If you feel strongly about anything, you should try to make your points in public channels. I think you asking was a good thing. Why I said that "still not addressed issues" was because I was asked to review again, but this issue was not resolved (you asked, and for now people prefer link). I would expect that we reach a decision, and we send the request again, otherwise I'm reviewing the changes I disagreed in the first place. 

Having said that, I don't care that much about button or link. My argument is, the main action there is to sign out, the link is just a "satellite" functionality that doesn't need so much attention. I don't care too much what we decide in this matter. Ask maybe Michel and he can decide :)

Velenir:
My preference
pr1341--dexreact review gnosisdev com_trade_DAI-USDC_sell=0 price=0 from= expires=

W3stside:
@anxolin moved the entire wallet up, maybe that's better, check it now pls

anxolin (approved):
> @anxolin moved the entire wallet up, maybe that's better, check it now pls

Actually, I just meant the button, and not the whole wallet pop up, but I'm glad you understand it like this, I like your suggestion more.


Some small detail in responsive, the copy icon shows in the next line. Also, I think I should be able to click in the address and icon, not just in the icon. Remember the fat finger (like me) people are people too 
image

biocom:
Just to add my comment: I like that last screenshot shared by Anxo. In my view 'Logout' might be more of a primary action than 'View on Etherscan'. Making 'Logout' the single button therefore I think makes the most sense. Also for comparison this is what Uniswap uses:

Screenshot 2020-08-26 at 13 25 17

anxolin:
> Just to add my comment: I like that last screenshot shared by Anxo. In my view 'Logout' might be more of a primary action than 'View on Etherscan'. Making 'Logout' the single button therefore I think makes the most sense. Also for comparison this is what Uniswap uses:

I like that they call disconnect, it would make it more coherent. Note we call it connect VVVVV
image



Just regarding the Etherscan link, I think should be a link link, I see it was kept as a "secondary button" that looks like a link until you hover on it. It looks better, but I don't see a big reason to make it stand out so much. hehe, I leave it up to u guys 😅

PR #1345, opened 2020-08-26
[HiddenOrderbook] Allow to fetch historic books by batchId

fleupold:
Convenience feature to see the orderbook at arbitrary points in the past. This only applies to the hidden orderbook page under http://localhost:8080/book

I'd really appreciate some styling help how to align this text box a bit nicer.

### Test Plan

Screen Shot 2020-08-26 at 11 55 32

Also making sure that we can still load the normal orderbook at the most recent batch.

### DIFF
--- a/src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
+++ b/src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
@@ -21,6 +21,7 @@ interface OrderBookParams {
   baseTokenId: number
   quoteTokenId: number
   hops?: number
+  batchId?: number
 }
 
 /**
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1345--dexreact.review.gnosisdev.com

anxolin (approved):
Cool!
That's really nice we can now input the batch id.

How is this batch id used in the backend, you convert it to the epoch, and check in that time?
wouldn't make sense to add a datepicker instead?



Given the low risk of this PR, if u want to see it in PRO asap, u can cherry pick the commit and merge in `release/1.3` branch. Otherwise, we can add it in 1.4 if it goes to develop.

fleupold:
> How is this batch id used in the backend,

It computes the order book that the solver saw for that batch. We also support passing in arbitrary block numbers and in the future timestamps (to also see the state mid-order-collection) but for now I think it's particularly interesting to see that book for some crucial auctions (which we usually refer to by batchId).

I agree that for convenience in the future we should add a date picker that then translates into a timestamp and renders the orderbook at that second.

anxolin:
Okay, we can add in the future. Cool.

by the way, I've made a PR on top of this one, and by mistake one of these commit I've pushed it to this branch. 
This is why hard pushed later undoing that commit.

Now this branch should only have your commits

fleupold:
It looks like coverage is keeping it from being merged. Do I need to add a unit test somewhere?

anxolin:
> by the way, I've made a PR on top of this one, and by mistake one of these commit I've pushed it to this branch.
>This is why hard pushed later undoing that commit.
>
>Now this branch should only have your commits

I got it wrong, I thought I accidentally pushed my commits into this branch too, but it was you that merged it.
So I delete that merge, sorry 🙈

I've re-merged it:
https://github.com/gnosis/dex-react/pull/1348


> It looks like coverage is keeping it from being merged. Do I need to add a unit test somewhere?
It should be fine. U just need one more approval and we can merge

Velenir (approved):

anxolin:
@fleupold I'll merge this PR, it has already the approvals. 
I had to rebase with develop and solve some conflicts. We are doing some changes in the orderbook, this is why we need it in develop.
cc/ @Velenir

fleupold:
Thanks. Sorry I forgot to merge (I think I had to update the branch and then forgot to hit the merge button after CI passed). Merge when green would be a nice feature you could add to this repo to automate this.

PR #1350, opened 2020-08-27
Automatic Quote selection: Use priorities for price/market selection

### LINKED ISSUE #1273: Smart selection of quote token
Smart selection of quote token and base token

Rafa's suggestion for simple and good enough solution is:
- We need to know when a token is a stable coin
- If it's stable coin vs non stable coin, we choose stable coin as a quote
- In any other case, we continue as now: sell is the quote

With the above, we can eliminate the need of showing the 2 prices.
We can, just in case allow to invert the price anyways.

_Estimation: 3.5_

W3stside:
Closes #1273 

Open and supersedes #1344 

Tests: `yarn test test/utils/adjustMarketPrice.spec.ts test/hooks/quoteTokenPriorities.hooks.test.tsx`
- Runs both the util function that adjusts token priorities and the hook that returns tokenList with priorities

There was difficulty getting the correct prices displayed considering how many places price is manipulated, e.g:
1. `PriceEstimations` calculates price based on base/quote token, but this led to price errors when the priority algorithm and hook would change the order, resulting in inverse price
2. `Price` the inputs are conditionally shown based on `inversePrice` and show price accordingly. This is tricky if the market is switched due to high priority market pair being traded, thus re-inversing the price.
3. SwapPrice and the inversing of markets with adjusted pairs was similar to point 3


W3stside:
@anxolin I believe I addressed all your comments from the previous, closed PR, #1344

gnosis-info:
Travis automatic deployment:
 https://pr1350--dexreact.review.gnosisdev.com

W3stside:
Still a WIP as price is not shown properly and there is test code lingering...

src/components/TradeWidget/index.tsx
```diff
@@ -543,8 +558,7 @@ const TradeWidget: React.FC = () => {
   const defaultValidFrom = trade.validFrom || validFromParam
   const defaultValidUntil = trade.validUntil || validUntilParam
 
-  const [priceShown, setPriceShown] = useState<'INVERSE' | 'DIRECT'>('INVERSE')
-
+  const [priceShown, setPriceShown] = useState<'INVERSE' | 'DIRECT'>('DIRECT')
```
anxolin, line 561: ubernit: and also not from this PR, but can we just give the type a name (create the type)
i.e.
```ts
type PriceRepresentation =  'INVERSE' | 'DIRECT'
```

src/hooks/usePrioritiseTokensForPrice.ts
```diff
@@ -0,0 +1,70 @@
...
+export function checkMarketAndSmartAdjust({ sellToken, receiveToken }: HookProps): SmartTokenStruct {
+  let baseToken = receiveToken
+  let quoteToken = sellToken
+  let wasPriorityAdjusted = false
+
+  // Prio: 1    - STABLE USD
+  // Prio: 2    - STABLE EUR
+  // Prio: 3    - WETH
+  // Prio: 99   - VOLATILE
+  const basePrio = baseToken.priority || 99
```
anxolin, line 32: 99 is arbitrary, I would use `Number.MAX_SAFE_INTEGER`

anxolin: Also, if you do a small function getPriority, you can refactor it
```js
const basePrio = getPriority(baseToken)
const quotePrio = getPriority(quotePrio)
```

Also, I wouldn't shorten `priority` word

test/utils/adjustMarketPrice.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+// QUOTE gives price
+//     BASE/QUOTE
+// e.g WETH/TUSD
+
+describe('Check market and check base/quote token priorities', () => {
+  const [WETH, TUSD, USDT, , , , , FAKE] = tokenList
+
+  test('STABLE/STABLE => return with no change', () => {
+    // QUOTE
+    const sellToken = TUSD
```
anxolin, line 13: Why don't you define the constants for the token `USDT`, `TUSD` , so they have already the priority (instead of change in it in every test)

I think, if you do this, the tests become a bit shorter, easier to read, but more importantly less error prone, since you cannot mess up with priorities. 
We can use prio 1 for TUSD and USDT, 2 for EURO, 3 for WETH, ... (and do not change it)

Also, you can refactor the test to be onliners if you build a function:

```ts
 assertMarket({ TUSD, USDT  },  {  baseToken: TUSD, quote: USDT }) 
```

src/components/TradeWidget/Price.tsx
```diff
@@ -225,10 +227,10 @@ const Price: React.FC<Props> = ({
   return (
     <Wrapper>
       <strong>
-        Limit Price <OrderBookBtn baseToken={receiveToken} quoteToken={sellToken} />
+        Limit Price <OrderBookBtn baseToken={baseToken} quoteToken={quoteToken} />
       </strong>
       {/* using display: none to hide to avoid hook-form reregister */}
-      <PriceInputBox hidden={priceShown !== 'DIRECT'}>
+      <PriceInputBox hidden={(isPriceInverted && wasPriorityAdjusted) || (!isPriceInverted && !wasPriorityAdjusted)}>
```
anxolin, line 233: I think the logic is a bit complex, instead of complicating when rendering the view, can we just use "aux" variables for the quote/base returned by the function, and make the final selection of quote/base based on that plus `isPriceInverted`. 

This way, you move the logic to the JS part

src/components/TradeWidget/Price.tsx
```diff
@@ -156,13 +156,14 @@ export const PriceInputBox = styled.div<{ hidden?: boolean }>`
...
-  receiveToken: TokenDetails
+  quoteToken: TokenDetails
+  baseToken: TokenDetails
   priceInputId: string
   priceInverseInputId: string
   tabIndex?: number
   swapPrices: () => void
-  priceShown: 'INVERSE' | 'DIRECT'
+  isPriceInverted: boolean
+  wasPriorityAdjusted: boolean
```
anxolin, line 166: Why do we need the `wasPriorityAdjusted`  why is it relevant?
I mean, the important thing is what's the new quote, not if it's the one we thought originally, or not. Probably I'm missing sth

test/utils/adjustMarketPrice.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+import tokenList from '../data/tokenList'
+
+// QUOTE gives price
+//     BASE/QUOTE
+// e.g WETH/TUSD
+
+describe('Check market and check base/quote token priorities', () => {
+  const [WETH, TUSD, USDT, , , , , FAKE] = tokenList
+
+  test('STABLE/STABLE => return with no change', () => {
```
anxolin, line 11: Ubernit: you can create inner "describe" and group by "STABLE/STABLE", then you can do the inner tests for this two.

USDT/USDC  --> USDT/USDC
USDC/USDT  --> USDC/USDT

not ubernit: I think you haven't tested the second case above ^^^.

anxolin (commented):
That's much better!
And I really like the tests

src/components/TradeWidget/Price.tsx
W3stside: BASE_TOKEN/QUOTE_TOKEN: `DAI/WETH`

`DAI` = BASE TOKEN
`WETH` = QUOTE TOKEN

`DAI` PRIORITY = 1
`WETH` PRIORITY = 3

SHOW MARKET AS: `WETH/DAI` i.e
`DAI` = QUOTE TOKEN
`WETH = BASE TOKEN

In this scenario, `PriceSuggestions` receive the `baseToken` and `quoteToken` props as:
(they were switched above, according to the priority scheme)
```
<PriceSuggestions
  ...
  baseToken={WETH}
  quoteToken={DAI}
  isPriceInverted={priceInvertedState}
/>

where, inside `PriceSuggestions` - the `usePriceWithSlippage` hook is run using the `baseToken` `quoteToken` specified props above

So the price using the adjusted tokens will return the `inversePrice` essentially as the token pair has been switched, meaning we need some way of knowing when it has been switched to adjust the pricing shown

Open to any easier approaches. One thing to keep in mind while testing is the user's choice to `inversePrice` and how to accurately portray that in the multiple places we show the inverse price button and the `Receive at least` + `Limit Price` inputs

File: test/utils/adjustMarketPrice.spec.ts (hunk 3 above)
Comment (W3stside) [reply test/utils/adjustMarketPrice.spec.ts:13]: i like doing extra work, obviously

File: src/components/TradeWidget/index.tsx [hunk 7]
```diff
@@ -394,6 +395,20 @@ export interface TradeFormData {
 
 const validationResolver = resolverFactory<TradeFormData>(validationSchema)
 
+// TESTING
```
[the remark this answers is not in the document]
W3stside: nope

anxolin:
At least in Rinkeby, the selection is not correct, look at this picture:
image

src/components/TradeWidget/Price.tsx
anxolin: I understand why `isPriceInverted` is needed, but not `wasPriorityAdjusted`.

If you hand the final base/quote and the isPriceInverted, any component should be able to represent it.
Also, most components wouldn't even need the "isPriceInverted" probably. I can see why for price suggestions does.

> Open to any easier approaches

One idea is modeling as having JUST `sellToken`, `buyToken`,  `baseToken` and `quoteToken`.
the two first is what you have selected in the form
the two last define the market. Just thought this might model more what we have in there.

anxolin:
By the way, the lint issues shouldn't be fixed in this PR. It's not feasible to review a PR with code changes and 144 changes

src/data/quoteTokenPriorities.ts
```diff
@@ -0,0 +1,85 @@
...
+      ],
+    },
+  },
+  {
+    // OWL
+    priority: 2,
+    addresses: {
+      [Network.Mainnet]: ['0x1a5f9352af8af974bfc03399e3767df6370d82e4'], // OWL
+      // Rinkeby
+      [Network.Rinkeby]: [],
```
anxolin, line 51: You didn't add OWL in Rinkeby for any particular reason?

src/hooks/usePrioritiseTokensForPrice.ts
anxolin: You resolved my comment but I still see the same issue...

anxolin (commented):
You still have some unaddressed comments.

I have still those same questions as last review. Please, make sure you address or reply before re-requesting review. If if you think is easier to discuss in a zoom, that's an option too.

W3stside:
@anxolin ah yes I changed the `Number.MAX_INTEGER` into the base priority setting area - `useTokenList`

```
const map: Map<string, number> = new Map()
  quoteTokenPrioritiesList.forEach(({ priority, addresses }) => {
    if (addresses[networkId] && addresses[networkId].length) {
      addresses[networkId].forEach((address: string) =>
        map.set(address.toLowerCase(), priority || Number.MAX_SAFE_INTEGER),
      )
    }
  })
```

src/data/quoteTokenPriorities.ts
W3stside: nope, guess it slipped my mind :+1:

src/hooks/usePrioritiseTokensForPrice.ts
W3stside: it's in a different section, `useTokenList` but i will remove this as it's no longer needed

src/components/TradeWidget/index.tsx
anxolin: given that's a easy change, could you consider doing this?

anxolin (commented):
It looks broken

image

image

src/hooks/useTokenList.ts
```diff
@@ -3,21 +3,66 @@ import { TokenDetails } from 'types'
...
+  quoteTokenPrioritiesList?: typeof quoteTokenPriorities
+}): TokenDetails[] {
+  if (!networkId) return tokenList
+  const stablecoinMap = getNetworkCorrectStableCoinList(networkId, quoteTokenPrioritiesList)
+  console.debug('stablecoinMap', stablecoinMap)
+  // no networkId or map is empty as network id returns no tokens
+  if (!stablecoinMap?.size) return tokenList
+
+  return tokenList.map<TokenDetails>((token) =>
+    Object.assign({}, token, { priority: stablecoinMap.get(token.address.toLowerCase()) || Number.MAX_SAFE_INTEGER }),
```
Velenir, line 46:
```suggestion
    ( {...token, priority: stablecoinMap.get(token.address.toLowerCase()) || Number.MAX_SAFE_INTEGER }),
```

test/hooks/quoteTokenPriorities.hooks.test.tsx
```diff
@@ -0,0 +1,147 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { act } from 'react-dom/test-utils'
```
Velenir, line 3: Why suddenly do native react renders, when we use `enzyme` in all other component tests?

src/components/TradeWidget/index.tsx
```diff
@@ -73,17 +76,73 @@ import { updateTradeState } from 'reducers-actions/trade'
...
+        padding: '1rem',
+        height: open ? 'auto' : '3.45rem',
+        minWidth: '29rem',
+        overflow: 'hidden',
+      }}
+    >
+      <a
+        href="#"
+        onClick={handleClose}
+        style={{
```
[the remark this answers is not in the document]
W3stside: Gonna remove the whole thing, was mainly for testing

src/components/TradeWidget/index.tsx
```diff
@@ -73,17 +76,73 @@ import { updateTradeState } from 'reducers-actions/trade'
...
+        borderRadius: '0 0 1.5rem 0',
+        color: '#fff',
+        padding: '1rem',
+        height: open ? 'auto' : '3.45rem',
+        minWidth: '29rem',
+        overflow: 'hidden',
+      }}
+    >
+      <a
+        href="#"
```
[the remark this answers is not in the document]
W3stside: Thought this was best practice when using non standard anchor tags in react. Nvm

test/hooks/quoteTokenPriorities.hooks.test.tsx
W3stside: For hook testing enzyme works? I honestly just wrote it from memory using act as I'm used to it. Can change if important

src/components/TradeWidget/index.tsx
```diff
@@ -621,20 +626,28 @@ const TradeWidget: React.FC = () => {
...
-            baseToken={receiveToken}
-            quoteToken={sellToken}
-            tabIndex={1}
+            baseToken={baseToken}
+            quoteToken={quoteToken}
+            receiveToken={receiveToken}
+            sellToken={sellToken}
             onSwapPrices={swapPrices}
-            priceShown={priceShown}
+            isPriceInverted={priceShown === 'INVERSE'}
```
anxolin, line 634: Refactor isPriceInverted as a const, u have it also in line 649

src/components/trade/Price.tsx
```diff
@@ -225,15 +229,17 @@ export const Price: React.FC<Props> = ({
     inputId: priceInverseInputId,
     precision: DEFAULT_PRECISION,
   })
-  const isDirect = priceShown === 'DIRECT'
+
+  const hideInverse = (!isPriceInverted && wasPriorityAdjusted) || (isPriceInverted && !wasPriorityAdjusted)
```
anxolin, line 233: This part is a bit strange, very complex to understand, minimum we need a comment to clarify.
Also, is not true that only one price will be visible. Therefore, only one const should be enought

src/components/trade/Price.tsx
```diff
@@ -225,15 +229,17 @@ export const Price: React.FC<Props> = ({
...
+  const hideDirect = (isPriceInverted && wasPriorityAdjusted) || (!isPriceInverted && !wasPriorityAdjusted)
 
   return (
     <Wrapper>
       <strong>
         Limit Price <OrderBookBtn baseToken={baseToken} quoteToken={quoteToken} />
       </strong>
       {/* using display: none to hide to avoid hook-form reregister */}
-      <PriceInputBox hidden={!isDirect}>
+      <PriceInputBox hidden={hideInverse}>
```
anxolin, line 242: This is the direct price, not the inverse, isn't it? why does it say "hide inverse"?

src/components/trade/Price.tsx
```diff
@@ -272,9 +278,9 @@ export const Price: React.FC<Props> = ({
           <SwapPrice
             baseToken={baseToken}
             quoteToken={quoteToken}
-            isPriceInverted
-            showBaseToken
+            isPriceInverted={!wasPriorityAdjusted}
```
anxolin, line 281: This make it very difficult to understand, why `isPriceInverted = !wasPriorityAdjusted` 🤯

src/components/trade/PriceSuggestions/PriceSuggestionWidget.tsx
```diff
@@ -19,28 +27,41 @@ export const PriceSuggestionWidget: React.FC<Props> = (props) => {
...
+    sellToken,
     isPriceInverted,
     priceInputId,
     priceInverseInputId,
+    wasPriorityAdjusted,
     onSwapPrices,
   } = props
-  const { id: baseTokenId, decimals: baseTokenDecimals } = baseToken
-  const { id: quoteTokenId, decimals: quoteTokenDecimals } = quoteToken
+  const { id: baseTokenId, decimals: baseTokenDecimals } = receiveToken
```
anxolin, line 38: Why this change? is not better to express in terms of base/quote the price suggestions?
Also, I don't think this component should have "wasPriorityAdjusted". This component should know only the base/quote and wether the price has been inverted or not.

src/components/trade/PriceSuggestions/PriceSuggestions.tsx
```diff
@@ -118,7 +117,9 @@ export interface Props {
   // Market
   baseToken: TokenDetails
   quoteToken: TokenDetails
+  sellToken: TokenDetails
   isPriceInverted: boolean
+  wasPriorityAdjusted: boolean
```
anxolin, line 122: Please, let's not propagate `wasPriorityAdjusted` this prop down all the components. you just let the component know what's your selection of base/quote and if it is inverted or not.

anxolin:
Functionally I believe it works as expected.
There's still some complexities added that can be simplified imo.

src/components/trade/Price.tsx
W3stside: fair enough on 1 constant required

W3stside: I am going to rework thru the other comments to try and cut complexity but i still think the 2nd flag is required to balance the inverse/direct price with the act of changing the token order when getting pricing. Will add comment reflecting this

src/components/trade/Price.tsx
W3stside: yup

W3stside:
closing

PR #1349, opened 2020-08-26
Storybook integration

Velenir:
Run `yarn storybook` and go to http://localhost:6006/?path=/story/mesa-etherscanlink--primary

`webpack.config` was changed from ES6 imports to commonjs to be able to utilize in `.storybook/main.js`

~Not everything works yet. For example, prod build doesn't...~
Prod build works.
Custom folder override works (added About story to test)

Access at https://pr1349--dexreact.review.gnosisdev.com/storybook/

PR #1351, opened 2020-08-28
Cleanup examples add some stories

anxolin:
This PR makes some suggestions for #1349

- Delete the examples, I think is best to leave only our code
- Moved the stories next to it's component. This way, it's easy to make new stories, find the story of a component, and know if a component has it or know, just by looking at the file tree
- Added some scripts for generating documentation
- Add new story for the spinner
- Refactor spinner to allow to set another icon, I just wanted a spinning skull story!
- Added some representative stories to etherscan link, I know you can play with the controls/params, but I think is nice to have some "typical cases"
- Removed the decorator in the etherscan link stories that adds alerts, it's cool how you intercept it, but I think is nicer to just try the link in this case and open etherscan

Smaller other changes:
- `.storybook/main.js` was using JSON format for 2 props, just used JS style (without quotes in the prop name)
- Moved the page decorator to it's own file, since it will be likely used in more places, probably it can be moved to a more central place, like in `.storybook/preview` (I understand we need to make a TS version of it to add these global decorators)


gnosis-info:
Travis automatic deployment:
 https://pr1351--dexreact.review.gnosisdev.com

package.json
```diff
@@ -19,7 +19,9 @@
     "verify": "npm run lint && npm run test",
     "format": "prettier --write \"src/**/*.ts\" \"src/**/*.js\"",
     "storybook": "BROWSER=NONE start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "docs": "start-storybook --docs",
```
Velenir, line 23:
```suggestion
    "docs": "npm run storybook -- --docs",
```
unless you want to override the flags
also this adds composability

Velenir (approved):

src/components/EtherscanLink.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const networkIds = Object.values(Network).filter(Number.isInteger)
+
+export default {
+  title: 'Components/EtherscanLink',
+  component: EtherscanLink,
+  argTypes: {
+    label: { control: 'text' },
+    networkId: { control: { type: 'inline-radio', options: networkIds } },
+  },
```
alfetopito, line 16: What if you don't have this?
Storybook should be smart enough to get the arg types from component props.
At least most of the time.

src/pages/About.stories.tsx
```diff
@@ -1,21 +1,13 @@
...
   component: About,
-  decorators: [
-    (Story): JSX.Element => (
-      <>
-        <GlobalStyles />
-        <Story />
-      </>
-    ),
-  ],
+  decorators: [PageDecorator],
```
alfetopito, line 10: I think you'll want to style all your components with the global style, no?

Then, just add it to global decorators like https://github.com/gnosis/safe-cmm-app-react/blob/development/.storybook/preview.js#L9

anxolin: Yep, Dima did that. I talked before with it. He did a PR moving it to the `preview.js`

src/components/EtherscanLink.stories.tsx
anxolin: I understand u don't get this nice network id suggestions :)

anxolin: image
Maybe label can be removed! I will do that

anxolin: Amending my last comment, if you don't say the "label" is a test, we need to quote it to test it. I guess is because of this:

image

package.json
anxolin: I see your point, ok, I'll do it as u suggest also for the build-docs (although it has no args)

src/components/EtherscanLink.stories.tsx
alfetopito: Understood

PR #1352, opened 2020-08-28
Order book reposition

biocom:
# Done
- Re-organised the layout to have the orders panel below. 
- Reserved a container on the top right to showcase the order book.

Screenshot 2020-08-28 at 11 59 23
Screenshot 2020-08-28 at 12 02 13

# Need help integrating:
- Adding the actual order book component. Basically the current order book as shown in the modal, but without the pair selector (given that's already present in the order form on the left). Right now I simply added a placeholder container component.
- For mobile I would suggest for this iteration to keep the 'View order book' button and hide the upfront container we show on desktop. This means that on desktop we won't show the 'View order book' button and show the order book container instead.
- For a next iteration we can add footer tabs for this like this example (@anxolin):
image (1)

### DIFF
--- a/src/components/DepositWidget/index.tsx
+++ b/src/components/DepositWidget/index.tsx
@@ -42,7 +42,7 @@ export const BalancesWidget = styled(CardWidgetWrapper)`
   height: 100%;
 
   > .filterToolsBar {
-    height: 10%;
+    min-height: 5.2rem;
   }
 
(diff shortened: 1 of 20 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1352--dexreact.review.gnosisdev.com

PR #1353, opened 2020-08-28
Order book reposition (supersede #1352)

anxolin:
Supersede #1352, 
* It didn't start from develop (was having some 1.3 changes that shouldn't be in this PR --> it would squash them later causing conflicts)
* Bad formatting: breaking some linting rules. Applied autofix from eslint

# Done
- Re-organised the layout to have the orders panel below. 
- Reserved a container on the top right to showcase the order book.

Screenshot 2020-08-28 at 11 59 23
Screenshot 2020-08-28 at 12 02 13

# Need help integrating:
- Adding the actual order book component. Basically the current order book as shown in the modal, but without the pair selector (given that's already present in the order form on the left). Right now I simply added a placeholder container component.
- For mobile I would suggest for this iteration to keep the 'View order book' button and hide the upfront container we show on desktop. This means that on desktop we won't show the 'View order book' button and show the order book container instead.
- For a next iteration we can add footer tabs for this like this example (@anxolin):
image (1)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -72,10 +72,10 @@ const FAQ = React.lazy(() =>
     'pages/FAQ'
   ),
 )
-const OrderBook = React.lazy(() =>
+const OrderBookPage = React.lazy(() =>
   import(
     /* webpackChunkName: "OrderBook_chunk"*/
-    'pages/OrderBook'
+    'pages/OrderBookPage'
   ),
 )
 const Settings = React.lazy(() =>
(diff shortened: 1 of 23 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1353--dexreact.review.gnosisdev.com

anxolin:
I really like how much more efficient is now with the space! It looks amazing already
I like also the header for desktop and where we have the connect wallet.

Some first comments, they can be all addressed in separated PRs to `release/2.0` branch too, too keep PRs simpler:
- The balance pages and liquidity page look a bit strange now (can be other PRs) 
- We need to create some issues, to filter the orders by the selected token pair
- In other PR too, it would be great to add in the header the current market. It can be pure static HTML, we would review how to make it work and be a global state
- We can make another iteration in the overall design. The design of a box floating I think is not suitable for the trading UI. Maybe is good this part to wireframe it first?
- The orders take now too much space, we can look how to better show them. dydx approach is nice, we can see also other alternatives. This can be part of the redesign mentioned in my previous point.
- If a project wants to put a Logo somewhere, it would be nice that we allow that

Velenir (commented):
Like a lot how it looks
But there are bound to be a thousand conflicts with the OrderBook refactor

anxolin:
> Like a lot how it looks
But there are bound to be a thousand conflicts with the OrderBook refactor

Yes, this PR will be recreated brand new

PR #1354, opened 2020-08-28
Stories mod

Velenir:
- Extend `@storybook` types to have `Template.bind({}).args` already typed
- Global decorators with `<GlobalStyles/>`
- markdown rules + `FAQ.story`


gnosis-info:
Travis automatic deployment:
 https://pr1354--dexreact.review.gnosisdev.com

.storybook/preview.js
```diff
@@ -1,4 +1,16 @@
...
+import React from 'react'
+import GlobalStyles from 'styles/global'
+
 export const parameters = {
   actions: { argTypesRegex: "^on[A-Z].*" },
-}
\ No newline at end of file
+}
+
+export const decorators = [
```
anxolin, line 9: nice! 

So this file still cannot easily be a TS? (I mean without nast workarounds)

src/types/storybook.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Story } from '@storybook/react/types-6-0'
+declare module '@storybook/addons' {
+  export interface BaseStory<Args, StoryFnReturnType> {
+    bind(thisArg: Parameters<Function['bind']>[0]): Story<Args>
```
anxolin, line 4: I'm lost in this change. What is this for? 
Maybe, add it also as a comment to

anxolin (approved):
Nice one!

src/types/storybook.d.ts
Velenir: so you don't have to do `.args = {} as Props`

.storybook/preview.js
Velenir: I just tried and turns out you can a23efe9

src/types/storybook.d.ts
alfetopito: :open_mouth:

anxolin: 对我来说那是中国人，迪玛

PR #1357, opened 2020-09-01
Allow to specify an initial list of tokens in the config

anxolin:
The app depends on an initial list of tokens in dex-js

This list is used for an initial warm up of the app, when there's no persisted local storage, and the TCR hasn't been loaded.
The list of tokens are the initial thing forks change.

Judging from other projects, they end up being super creative on how to hack this list into the app.
This PR tries to make it easier for them to add a custom list. So:
- adds `initialTokenList` config params, with this list
- Makes `getTokensByNetwork` in tokenList api to not depend on dex-js anymore. It just receives the list by parameter
- Modifies tokenList API to receive the list in the constructor
- Now the initialization of the APIs, load the list from the config, and inject it to the API

### DIFF
--- a/config-default.yaml
+++ b/config-default.yaml
@@ -133,3 +133,137 @@ disabledTokens:
     - address: '0x6Fb0855c404E09c47C3fBCA25f08d4E41f9F062f'
     - address: '0xFC4B8ED459e00e5400be803A9BB3954234FD50e3'
   4: [] # Rinkeby
+
+# Initial Token List
+#   List of token loaded by default
+#   This list is used while the app loads the TCR (if the TCR setting is used)
+initialTokenList:
+  - name: Wrapped Ether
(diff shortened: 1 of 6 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1357--dexreact.review.gnosisdev.com

Velenir (approved):
Should work the same as the different is only in where `tokenList` is imported from, `dex-js VS CONFIG`?

W3stside (approved):
works for me, nice

anxolin:
> Should work the same as the different is only in where tokenList is imported from, dex-js VS CONFIG?

Yes. One fork had to fork dex-js to change this list. Other fork, used a modified version of `TokenListMock`. So I want to make it easier by moving the list to the config

PR #1358, opened 2020-09-01
Allow to change the default initial selection of tokens

anxolin:
Brother PR of #1357, same motivation:

> The app depends on an initial list of tokens in dex-js
> This list is used for an initial warm up of the app, when there's no persisted local storage, and the TCR hasn't been loaded.
> The list of tokens are the initial thing forks change.

In this PR it's been added a new `initialTokenSelection` that allows you to set the default selection in the trade page. By default, we select DAI and USDC as the initial selection for sell token, and receive token.

One fork could, thanks to #1357 change the initial list, and then with this one, could change the initial selection (choosing 2 tokens from his custom list).

### DIFF
--- a/config-default.yaml
+++ b/config-default.yaml
@@ -133,3 +133,8 @@ disabledTokens:
     - address: '0x6Fb0855c404E09c47C3fBCA25f08d4E41f9F062f'
     - address: '0xFC4B8ED459e00e5400be803A9BB3954234FD50e3'
   4: [] # Rinkeby
+
+# Initial token selection for the sell token and buy token
+initialTokenSelection:
+  sellToken: DAI
+  receiveToken: USDC
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -89,9 +89,21 @@ const Settings = React.lazy(() =>
 import { withGlobalContext } from 'hooks/useGlobalState'
 import { rootReducer, INITIAL_STATE } from 'reducers-actions'
 import PrivateRoute from './PrivateRoute'
+import { assertNonNull } from 'utils'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const Router: typeof BrowserRouter & typeof HashRouter = (window as any).IS_IPFS ? HashRouter : BrowserRouter
+
+function getInitialUrl(): string {
+  // Validate tha
+  assertNonNull(CONFIG.initialTokenSelection, 'initialTokenSelection config is required')
+  const { sellToken: initialSellToken, receiveToken: initialReceiveToken } = CONFIG.initialTokenSelection
+  assertNonNull(initialSellToken, 'sellToken is required in the initialTokenSelection config')
+  assertNonNull(initialReceiveToken, 'receiveToken is required in the initialTokenSelection config')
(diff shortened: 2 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1358--dexreact.review.gnosisdev.com

src/App.tsx
```diff
@@ -89,10 +89,22 @@ const Settings = React.lazy(() =>
 import { withGlobalContext } from 'hooks/useGlobalState'
 import { rootReducer, INITIAL_STATE } from 'reducers-actions'
 import PrivateRoute from './PrivateRoute'
+import { assertNonNull } from 'utils'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const Router: typeof BrowserRouter & typeof HashRouter = (window as any).IS_IPFS ? HashRouter : BrowserRouter
 
+function getInitialUrl(): string {
+  // Validate tha
```
Velenir, line 98: ?
```suggestion
  // Validate that
```

src/App.tsx
```diff
@@ -89,10 +89,22 @@ const Settings = React.lazy(() =>
...
+  assertNonNull(CONFIG.initialTokenSelection, 'initialTokenSelection config is required')
+  const { sellToken: initialSellToken, receiveToken: initialReceiveToken } = CONFIG.initialTokenSelection
+  assertNonNull(initialSellToken, 'sellToken is required in the initialTokenSelection config')
+  assertNonNull(initialReceiveToken, 'receiveToken is required in the initialTokenSelection config')
+  return '/trade/' + initialSellToken + '-' + initialReceiveToken + '?sell=0&price=0'
```
Velenir, line 103: I remember us having problems with some token symbols with special chars and dashes `-`
Likely need to do same here,who knows what default a fork can choose

`import { encodeTokenSymbol, decodeSymbol } from '@gnosis.pm/dex-js'` etc

Velenir (approved):

src/App.tsx
W3stside: this comment is supposed to be deleted i guess

W3stside (approved):

src/App.tsx
anxolin: hehe, ok. I'm happy if we just support normal tokens. anyways, I'll add that, thx!

anxolin: That function requires an address as mandatory param, which is not loaded at this point
I think in the super edge case of someone wanting a token with dashes in their name, they can encode it themselves.

Velenir: `import { encodeSymbol } from '@gnosis.pm/dex-js'`

config-default.yaml
```diff
@@ -134,6 +134,11 @@ disabledTokens:
     - address: '0xFC4B8ED459e00e5400be803A9BB3954234FD50e3'
   4: [] # Rinkeby
 
+# Initial token selection for the sell token and buy token
+initialTokenSelection:
+  sellToken: DAI
+  receiveToken: USDC
```
alfetopito, line 140: What about using token addresses instead?
No chance of picking the wrong token by accident, and also helps in case the token is not yet part of the UI
(Although if someone is forking it would be added anyway :thinking: )

PR #1359, opened 2020-09-01
fix error label if edge-case where tokens are the same

W3stside:
We have this weird edge case where if the user puts in an unknown token address/name or whatever the URL grabbing code defaults to market DAI/USDC - this allows users to potentially have the same market e.g DAI/DAI or USDC/USDC and we have an error code to deal

error message looked awful, fixes that

this is more of a pave over of a bigger problem but it looks better at least and users are blocked from doing anything until they change to a compatible token pair

Screenshot from 2020-09-01 17-11-08

### DIFF
--- a/src/components/Layout/PageWrapper.tsx
+++ b/src/components/Layout/PageWrapper.tsx
@@ -62,7 +62,7 @@ export const StandaloneCardWrapper = styled.div`
   }
 `
 export const PageWrapper = styled.section`
-  height: 71rem;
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1359--dexreact.review.gnosisdev.com

anxolin (approved):
Good call, yes the handling is not super good, but nice to improve a bit the style.

PR #1370, opened 2020-09-02
Fix warning when priceEsimaion = 0

Velenir:
An edge case when `priceEstimation = 0 --> minTradableAmount  = Infinity` for a token, happens when that token has not been traded yet. 

localhost_8080_trade_DAI-SUSHI_sell=3 price=2 12896 from= expires=

### DIFF
--- a/src/components/TradeWidget/TxMessage.tsx
+++ b/src/components/TradeWidget/TxMessage.tsx
@@ -158,6 +158,13 @@ const useLowVolumeAmount = ({ sellToken, sellTokenAmount, networkId }: LowVolume
   const gasPrice = useGasPrice({ defaultGasPrice: DEFAULT_GAS_PRICE, gasPriceLevel: 'fast' })
 
   return useMemo(() => {
+    if (priceEstimation !== null && priceEstimation.isZero()) {
+      // no price data for token
+      logDebug('No priceEstimation data for', sellToken.symbol, 'in OWL')
+
+      return { isLoading: false, isLowVolume: false }
+    }
+
     if (
       isPriceLoading ||
       isWETHPriceLoading ||


gnosis-info:
Travis automatic deployment:
 https://pr1370--dexreact.review.gnosisdev.com

src/components/TradeWidget/TxMessage.tsx
```diff
@@ -158,6 +158,13 @@ const useLowVolumeAmount = ({ sellToken, sellTokenAmount, networkId }: LowVolume
   const gasPrice = useGasPrice({ defaultGasPrice: DEFAULT_GAS_PRICE, gasPriceLevel: 'fast' })
 
   return useMemo(() => {
+    if (priceEstimation !== null && priceEstimation.isZero()) {
+      // no price data for token
+      logDebug('No priceEstimation data for', sellToken.symbol, 'in OWL')
+
+      return { isLoading: false, isLowVolume: false }
```
anxolin, line 165: So if there's no price estimation, we don't show the warning at all?
When does this occur? It's because an issue with the service?
I mean, gasPrice we should always have, so it's basically the price of ETH in OWL. Should we have a fallback too for this?

anxolin (commented):
Great catch dima

src/components/TradeWidget/TxMessage.tsx
Velenir: > happens when that token has not been traded yet

No issue with the service

with ETH it's not a problem, even in case it is reported as 0 (which won't happen), warning won't show as `minTradableVolume` will be calced to 0.

W3stside:
@Velenir did you lose your "T" key on your keyboard? :joy:

W3stside (approved):

Velenir:
> @Velenir did you lose your "T" key on your keyboard?

don' know wha you are alking abou

anxolin (approved):

PR #1371, opened 2020-09-02
[LimitOrder] Common refactors

anxolin:
Part of #1381

This is the first step for trying to clean and refactor the TradeWidget.

The idea is to make multiple PRs and progressively refactor the Trading widget. It looks to big to be refactored in one go. It has way to many hooks, form, and context dependencies. Just the `index` contains more than 1000 lines of code, and his subcomponents have a high level of coupling.

In this PR I create the namespace `commons` for some shared dumb components, that are used across all the app.

image

I started with a small refactor of the component FormMessage and FormInputError:
- Before they were defined in the same file. I splitted into 2
- They were inside `components/TradeWidget`, but they were used in for example the liquidity page, so their scope is more general

I also moved 2 other components we had already storybooks for: EtherscanLink and Spinner.


gnosis-info:
Travis automatic deployment:
 https://pr1371--dexreact.review.gnosisdev.com

src/components/DepositWidget/Form.tsx
```diff
@@ -9,7 +9,7 @@ import { formatAmountFull, parseAmount, abbreviateString } from 'utils'
 
 // Components
 import { CardDrawer } from 'components/Layout/Card'
-import { Spinner } from 'components/Spinner'
```
W3stside, line 12: shouldnt these then also be called `common/...`

src/components/TxNotification.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { EtherscanLink } from 'components/EtherscanLink'
```
W3stside, line 2: luckily with VS studio you can do `find and replace` and change all these at once

Velenir: You are using VS Studio? Why?

src/components/DepositWidget/Form.tsx
anxolin: Yes, I'm going little by little. The change is big, the ones I'm chaging mostly are related to the Trade widget that is the big monster in the app :) 

Within the Trade Widget, I'ms starting with the price components

src/components/TxNotification.tsx
anxolin: What Is VS Studio, I only know vim.

jokes aside, VS studio hasn't being that good helping to rename imports when I "rename" or move a file around. Sometimes was "forgetting" some files to update, or he didn't do a `git mv` so git was detecting a "delete and a add" of a file

Velenir: Have you tried `Rename Symbol F2`? It renames across files

src/components/common/FormInputError.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { FormInputError, Props } from 'components/common/FormInputError'
+
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
+import { Meta, Story } from '@storybook/react/types-6-0'
+
+export default {
+  title: 'Common/FormInputError',
```
Velenir, line 8: Why you don't provide a `Component`?

src/components/common/FormMessage.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { FormMessage, Props } from 'components/common/FormMessage'
+
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
+import { Meta, Story } from '@storybook/react/types-6-0'
+
+export default {
+  title: 'Common/FormMessage',
```
Velenir, line 8: No component? Book sad

src/components/common/FormInputError.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (errorMessage) {
+      return errorMessage
+    } else {
+      if (visibilityHidden) {
+        return 'Stop using developer tools! :D'
+      }
+
+      return null
+    }
+  }, [errorMessage, visibilityHidden])
```
Velenir, line 33: Don't use memo for a few if conditions

src/components/common/FormInputError.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { FormMessage } from 'components/common/FormMessage'
+
+const Wrapper = styled.div<{ $margin?: string; $padding?: string }>`
+  margin: ${({ $margin = '0' }): string => $margin};
+  padding: ${({ $padding = '0' }): string => $padding};
+`
+
+export interface Props {
+  errorMessage?: string
```
Velenir, line 12: You can add multiline doc comments that get shown in storybook docs tab
like in EtherscanLink component

Velenir (approved):

src/components/TxNotification.tsx
anxolin: Yep, that's how I normally do it. Also, moving the file renames the imports (when it works)

src/components/common/FormInputError.stories.tsx
anxolin: No reason, cause it worked without it. Now I read more carefully I see:

> The component field is optional (but encouraged!), and is used by addons for automatic prop table generation and display of other component metadata. title should be unique, i.e. not re-used across files.

src/components/common/FormMessage.stories.tsx
anxolin: I'll add it, so Book happy 🥳

src/components/common/FormInputError.tsx
anxolin: I would swear I delete this memo, my memo is blurry right now about this

anxolin: Actually I think I also deleted some props we don't need, this might be in some local branch of mine...

src/components/common/FormInputError.tsx
anxolin: Yep, although I think for FormInput it's ok to leave it undocumented.
We can try to document a bit more the components that are important. This one in particular, is self explanatory (it has just one field)

PR #1321, opened 2020-08-17
Release/1.3.0

Velenir:
**DON'T SQUASH**

# :notebook: Changes
- **Completely revamps the Advanced Settings** (previously order validity settings) to include custom calendar date picking options, relative starting/ending times, and the ability to select start/end times by batch ID

- **Adds a new column in the order book showing Order ID** and the ability to click **Order ID** and **Market** cells to pre-populate the order filters

- **Show provider name/logo** at the top of the app page in the user's wallet section

- **Introduce a default sell/receive token section inside the current configuration file** to pre-populate the trading page if no tokens passed in URL

- **Footer redesign - adds *Powered by Gnosis Protocol* logo and responsive design changes**

- **Travis build: auto release notes**

- **Show recommended minimum sell amount** to users for clarity on better order experience

- **GP solver subsidize factor** update to better align with current parameters

- **Storybook integration** to facilitate further app development

### :detective: Changelog
cdf11a7 Order validity upgrade (#1269)
e38d570 Allow to change the default initial selection of tokens (#1358)
0ac6ae0 Allow to specify an initial list of tokens in the config (#1357)
fa08458 preliminary Order ID column in OrdersWidget(#1261)
ea4e91e Storybook integration (#1349)
a1ce473 Easier log in/out process (#1341)
f442843 Footer changes: GP logo add, responsive format (#1326)
a72ce62 Subsidize factor change (#1329)
501424c Release notes file for travis builds (#1328)
27e98b3 Use ellipsis and title tag (#1325)
6bab734 Simplify build script
9317341 Recommend min sell amount (#1324)
5e1a671 Provider name in WalletComponent (#1270)

### :bug: Bugfixes/Fixes
16d4cf1 Fix error label for edge-case where selected tokens from URL are the same (#1359)
03b0623 Fix wallet connect/disconnect styling (#1356)
ccae05f Fix Price Suggestions wording (#1346)
995b584 Price Suggestions display and css to better fit responsive/web (#1305)
39d2c19 Fix price validation error on token change (#1340)
f0d0100 Fix HTML validity error (#1338)

PR #1375, opened 2020-09-03
TX Confirmation modal sometimes not popping up

W3stside:
`SubmitButton type="submit" ... >` now => `SubmitButton type="button" ... >` - should eliminate any race-conditions from form submitting vs confirmation modal popping up

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "start": "NODE_ENV=development webpack-dev-server --mode development --hot",
+    "start:browserstack": "yarn start --host 0.0.0.0 --disable-host-check",
     "mock": "MOCK=true yarn start",
     "build": "rimraf ./dist && NODE_ENV=production webpack --mode production",
     "lint": "eslint src --ext .ts,.tsx,.js",
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -582,7 +582,7 @@ const TradeWidget: React.FC = () => {
       <TokensAdder tokenAddresses={tokenAddressesToAdd} networkId={networkIdOrDefault} onTokensAdded={onTokensAdded} />
       {/* Toggle Class 'expanded' on WrappedWidget on click of the <OrdersPanel> <button> */}
       <FormProvider {...methods}>
-        <WrappedForm onSubmit={onConfirm} autoComplete="off" noValidate>
+        <WrappedForm onSubmit={(e): void => e.preventDefault()} autoComplete="off" noValidate>
           {sameToken && (
             <>
               <WarningLabel className="warning">Tokens cannot be the same! </WarningLabel>
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -650,7 +650,7 @@ const TradeWidget: React.FC = () => {
           >
             <SubmitButton
               data-text="This order might be partially filled."
-              type="submit"
+              type="button"
               disabled={isSubmitting || sameToken}
               tabIndex={1}
               onClick={(e): void => {
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -4,6 +4,7 @@ import { useForm, useWatch, FormProvider, SubmitHandler } from 'react-hook-form'
 import { useParams } from 'react-router'
 import { toast } from 'toastify'
 import BN from 'bn.js'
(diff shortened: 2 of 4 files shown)


gnosis-info:
Travis automatic deployment:
 https://pr1375--dexreact.review.gnosisdev.com

anxolin (commented):
This doesn't solve the issue I think

`Issue 2` still happens: https://github.com/gnosis/dex-react/pull/1321#issuecomment-686397150

I can see that in responsive we left the GEAR ⚙️ icon. Was this because there was no space in mobile?
We received feedback that this was not clear enough, I thought the change was also in mobile...

I can see that the gear is a button, and If I remove it, I can see how issue 2 disappear. So I think this 2 things can be solved at the same time.

anxolin:
This pr introduces back the issue it had the PR in the past.

`issue 5`: 
* Go to https://pr1375--dexreact.review.gnosisdev.com
* Enter amount and price
* Click on "Submit limit order button"
* No click again!

Velenir (approved):
Works great for me :+1:

anxolin (commented):
Almost, 4 out of 5.

✅ issue 1
✅ issue 2
✅Issue 3 
❌ Issue 4
✅issue 5

Also, I will create an issue for the connecting problem button problem I mentioned in Slack. It can go in v1.4 since it's an error we already had: https://github.com/gnosis/dex-react/issues/1378

package.json
```diff
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "start": "NODE_ENV=development webpack-dev-server --mode development --hot",
+    "start:browserstack": "yarn start --host 0.0.0.0 --disable-host-check",
```
anxolin, line 12: Why not adding this flags in the start?

https://webpack.js.org/configuration/dev-server/#devserverdisablehostcheck

I see the `THIS IS NOT RECOMMENDED` warning, but I understand this is if you serve your app for production using the `webpack-dev-server`, which looks like a weird thing to do in the first place (normally you build the static files and use a regular HTTP server)

Is there anything im missing?

src/components/TradeWidget/useNumberInput.tsx
```diff
@@ -50,8 +50,16 @@ export function useNumberInput(params: Params): Result {
   )
 
   const onKeyPress = useCallback(
-    (event: React.KeyboardEvent<HTMLInputElement>): void =>
-      event.key === 'Enter' ? (event.preventDefault(), removeExcessZeros(event)) : preventInvalidChars(event),
+    (event: React.KeyboardEvent<HTMLInputElement>): void => {
```
anxolin, line 53: good call 👌

anxolin (approved):
All 🥬🥦🥒

✅ issue 1
✅ issue 2
✅Issue 3
✅ Issue 4
✅issue 5

package.json
Velenir: this looks like a security concern, please don't turn it on by default
does it make you exposed when in develop?

src/components/TradeWidget/index.tsx
```diff
@@ -582,7 +582,7 @@ const TradeWidget: React.FC = () => {
       <TokensAdder tokenAddresses={tokenAddressesToAdd} networkId={networkIdOrDefault} onTokensAdded={onTokensAdded} />
       {/* Toggle Class 'expanded' on WrappedWidget on click of the <OrdersPanel> <button> */}
       <FormProvider {...methods}>
-        <WrappedForm onSubmit={onConfirm} autoComplete="off" noValidate>
+        <WrappedForm onSubmit={(e): void => e.preventDefault()} autoComplete="off" noValidate>
```
Velenir, line 585: So this event was the problem?

Velenir (approved):

src/components/TradeWidget/index.tsx
W3stside: yeah on iOS only it was submitting form

package.json
W3stside: https://www.browserstack.com/question/759

iOS10+ restrictions force browserstack to push to a new URL (`bs-local`) so testing localhosted content wasn't working for iOS only.

didn't add it to `start` as i guess we don't technically need it there, only for browserstack usage - felt this was a bit safer

anxolin: But why is this a security concern? In case you are in a public network and someone access dex-react in your IP?
If I understand what it does, I don't see it very worrying, and it would help development when we need to test locally from our phones

PR #1377, opened 2020-09-04
Fix CONFIG.initialTokenList

Velenir:
- Fixed `CONFIG.initialTokenList`
- Added tests
- 
- Added a separate `tsconfig` for `./.storybook` because putting `includes: [**]` into main `tsconfig` would require explicit mention of all folders to traverse. For example `tests`. Without `includes` all required folders were already included, but not `.storybook`. As I understand by default all folders in the rootdir are included bit not those that start with `.` Better to have one-for-all `tsconfig` at root, and one for the specific case of `.storybook`

### DIFF
--- a/.storybook/tsconfig.json
+++ b/.storybook/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "../tsconfig.json"
+}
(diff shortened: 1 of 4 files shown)


anxolin (approved):

gnosis-info:
Travis automatic deployment:
 https://pr1377--dexreact.review.gnosisdev.com

W3stside (approved):

PR #1379, opened 2020-09-04
Orderbook refactor + stories

Velenir:
- Refactor OrderBook components to have container/dumb component separation
- Stories for OrderBook


src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -22,6 +23,25 @@ interface OrderBookParams {
   hops?: number
 }
 
+/**
+ * Price point as defined in the API
+ * Both price and volume are numbers (floats)
+ *
+ * The price and volume are expressed in atoms
+ */
+export interface RawPricePoint {
```
anxolin, line 32: Maybe we should rename to `Quote` or `PriceQuote`?

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+    name: 'Quote Token',
+    symbol: 'QUOTE',
+    address: '0x2',
+    decimals: 18,
+  },
+  networkId: defaultNetworkId,
+}
+
+// Token symbols to use in control selector
+const configTokenSymbols = [defaultParams.baseToken, defaultParams.quoteToken, ...tokenList].map(
```
anxolin, line 49: ubernit: I think calling this  `tokenSymbols` is just enough

gnosis-info:
Travis automatic deployment:
 https://pr1379--dexreact.review.gnosisdev.com

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+} as Meta
+
+//-------------------REAL-DATA------------------
+
+// Template for real data
+// fetched from price-estimation service
+const RealTemplate: Story<OrderBookChartProps> = (args) => <OrderBookChart {...args} />
+
+const DAI = tokenList.find((token) => token.symbol === 'DAI')
+const USDC = tokenList.find((token) => token.symbol === 'USDC')
```
anxolin, line 78: I think u can do a refactor of lines 77, 79, 162 and 169

src/components/OrderBookWidget.tsx
```diff
@@ -212,115 +131,84 @@ const processData = (
...
-      const pricePoints = bids.concat(asks)
+interface ProcessRawDataParams {
+  data: RawApiData
+  baseToken: Pick<TokenDetails, 'decimals' | 'symbol'>
+  quoteToken: Pick<TokenDetails, 'decimals' | 'symbol'>
+}
 
-      // Sort points by price
-      pricePoints.sort((lhs, rhs) => lhs.price.comparedTo(rhs.price))
+export const processRawApiData = ({ data, baseToken, quoteToken }: ProcessRawDataParams): PricePointDetails[] => {
```
anxolin, line 152: Maybe we should promote methods like this one to a util, like `src/utils/orderbook.ts`
With that, amazing if we add some very basic test too.

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+    data: { control: null },
+  },
+  decorators: [(Story): JSX.Element => <div style={{ height: '97vh' }}>{Story()}</div>],
+} as Meta
+
+//-------------------REAL-DATA------------------
+
+// Template for real data
+// fetched from price-estimation service
+const RealTemplate: Story<OrderBookChartProps> = (args) => <OrderBookChart {...args} />
```
anxolin, line 75: Why "RealTemplate", either Template or OrderBookTemplate

Also, I'm confused why `realData` is call liked that. It's just some mock data, right?

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+  orderbookProps: Omit<OrderBookChartProps, 'data'>
+}
+// normalizing sample data
+// because real data is slightly more complex
+const produceOrderBookProps = ({ sampleData, orderbookProps }: ProduceOrderBookArgsParams): OrderBookChartProps => {
+  const { baseToken, quoteToken } = orderbookProps
+
+  const { asks = [], bids = [] } = sampleData
+  const normalizedRawData: RawApiData = {
+    asks: asks.map(({ amount, price }) => ({
```
anxolin, line 105: Amount should be afaik an string, not a number. They were changing that in the backend, not sure if they did it already. We should try to operate with Bignumbers anyways, so we don't make float calculating errors, and so that change from the backend becomes transparent to us.

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+    })),
+  }
+
+  const processedData = processRawApiData({ data: normalizedRawData, baseToken, quoteToken })
+
+  return { ...orderbookProps, data: processedData }
+}
+
+// data samples
+const [
```
anxolin, line 123: why you stract manually instead of iterating using a loop?
If we add new examples, they should show automatically

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+    ...defaultSampleParams,
+    ...data,
+  }
+
+  Story.storyName = data.name
+
+  return Story
+}
+
+export const LiquidMarket = buildSampleStory(liquidMarket)
```
anxolin, line 217: Okay, I see why you go manual. Probably if you just export a default object, with the Key being the cammelCase of the example, it would work too

src/components/OrderBookChart.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  .amcharts-AxisLabel,
+  .amcharts-CategoryAxis .amcharts-Label-group > .amcharts-Label,
+  .amcharts-ValueAxis-group .amcharts-Label-group > .amcharts-Label {
+    fill: var(--color-text-primary);
+  }
+`
+
+export enum Offer {
```
anxolin, line 57: nit: maybe `QuoteType` or  `PriceQuoteType`  (I know the names are not from this PR)

anxolin (approved):
This PR is beyond awesome! 💥

Just nitpicks comments!

src/components/OrderBookChart.stories.tsx
Velenir: No, real data from https://dex-price-estimator.gnosis.io/api/v1/markets/7-4?atoms=true&hops=2

src/components/OrderBookChart.stories.tsx
Velenir: No how exports work, can't do `export {...AllExports}`, has to be statical named exports
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

anxolin:
I've made some tests on how to dynamically load the examples stories. 

I see two options. 

* We add a single story that loads the cases as a select or `<ul>`
* We use `storiesOf` API. I'm inclined to this second one. 

I've made this working example:
```ts
import React from 'react'
import Spinner, { SpinnerProps } from './Spinner'
import { faSkull } from '@fortawesome/free-solid-svg-icons'
import { storiesOf } from '@storybook/react';


const Template = (args: SpinnerProps) => () => <Spinner {...args} />

const cases = {
  Default: Template({}),
  Skull: Template({
    icon: faSkull,
  })
}


const stories = storiesOf('Components/Spinner', module)
for (let caseName in cases) {
  stories.add(caseName, cases[caseName])
}
```

Velenir:
> We add a single story that loads the cases as a select or <ul>

Will be a long list, and slow to render (docs were crazy slow until I disabled a bunch of stories). Also less controllable

> We use storiesOf API. I'm inclined to this second one.

I wonder if they'll remove that api at some point

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,256 @@
...
+// normalizing sample data
+// because real data is slightly more complex
+const produceOrderBookProps = ({ sampleData, orderbookProps }: ProduceOrderBookArgsParams): OrderBookChartProps => {
+  const { baseToken, quoteToken } = orderbookProps
+
+  const { asks = [], bids = [] } = sampleData
+  const normalizedRawData: RawApiData = {
+    asks: asks.map(({ amount, price }) => ({
+      // turn small numbers into wei or decimal-aware units
+      // otherwise get filtered as too small by src/components/OrderBookWidget.tsx#L77
```
alfetopito, line 105: You know this comment will be outdated in the next change to OrderBookWidget file, right?

src/components/OrderBookChart.stories.tsx
```diff
@@ -0,0 +1,256 @@
...
+  description,
+  asks,
+  bids,
+  networkId,
+  baseToken: baseTokenSymbol,
+  quoteToken: quoteTokenSymbol,
+}) => {
+  const baseToken = useMemo(() => {
+    return (
+      tokenList.find((token) => token.symbol === baseTokenSymbol || token.address === baseTokenSymbol) ||
```
alfetopito, line 160: Can't you use the token find method from utils?

alfetopito (approved):
Awesome work!
Really like the many orderbook use cases.

That being said, I wonder if the stories file could be simplified.
Feels to me there's way too many things happening there to get stories rendered.

As for the use cases, I think it would be worth to add a 'orderbook breaking order' situation, where one order's price is so big that nothing is displayed, like what we already had IRL.

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
Velenir: it could be confused with price for Quote Token then

src/components/OrderBookChart.stories.tsx
Velenir: https://dex-price-estimator.gnosis.io/api/v1/markets/7-4?atoms=true&hops=2 returns numbers only

src/components/OrderBookChart.tsx
Velenir: Still think it could be mistaken for `Price for Quote Token`

src/components/OrderBookChart.stories.tsx
Velenir: They are not equivalent, `getToken` from utils doesn't have a fallback like `symbol || address`,not without starting another `.find`

W3stside (approved):
holy shit, nice work. while i agree that the stories file is large etc, i don't see a way around it.

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
anxolin: Okok, can we at least drop the Raw?
raw for me made sense in the chart context (it was private in that module). But if you make it a general exported data type, all data is raw

src/components/OrderBookChart.stories.tsx
anxolin: My point is, it should be string. If they send it in number, we should convert it to string before operating, right? 
On top of preventing calculation issues, it will make the change in the backend trivial for us.

Maybe is not for this PR if it's not a trivial change..

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -109,6 +129,23 @@ export class DexPriceEstimatorApiImpl implements DexPriceEstimatorApi {
     return `${baseUrl}markets/${baseTokenId}-${quoteTokenId}?atoms=true&hops=${hops}`
   }
 
+  public async getOrderBookData(params: OrderBookParams): Promise<RawApiData> {
+    try {
+      const url = await this.getOrderBookUrl(params)
+
+      const res = await fetch(url)
+      return await res.json()
```
W3stside, line 137: why return `await`?

Velenir: so that it triggers `catch` here
otherwise we would return a potentially rejected promise

src/components/OrderBookChart.stories.tsx
Velenir: they do get turn into BigNumber inside `processData`

Issue #1380, opened 2020-09-04
Get the minimum trade amount in USD from the backend


anxolin:
![image](https://user-images.githubusercontent.com/2352112/92246086-8ecd2b00-eec5-11ea-8ae2-61cd46ddcc22.png)

The app is showing the minimal trade in USD as a dynamic value that computes taking some parameters into account (gas price, ether price in OWL, subsidy factor, etc..)

The solvers will be changed into using a fixed amount in USD as the minimal trade. This value will be exposed from a backend endpoint so solvers and front end are more aligned.

It's likely that in the future we use this calculation again, but for now it should to change into a fixed value coming from the backend.

More details in. this slack thread
https://gnosisinc.slack.com/archives/CC3C5SMB5/p1599224119069800?thread_ts=1599223469.068400&cid=CC3C5SMB5

@e00E @fleupold  can you let us know the signature of the endpoint in advance, so we can mock it while is not implemented?

e00E:
I interpreted the slack conversation so that we would only use the fixed amount for that one specific time period but in general we'll stay with the factor. And I'm not sure if we will have that endpoint by Wednesday (but we will have the hard coded number in the solver).
Not exactly what the format would be but I imagine something like `GET http://solver-constants.../`
```json
{
    "economic_viability_factor": 5.0,
}
```

but then if we have to differentiate fixed and factors maybe

```json
{
    economic_viability_factor: {
        "fixed": 1000
}
```

and 
```json
{
    economic_viability_factor: {
        "factor": 5.0
}
```

anxolin:
I've missed this message, yep sounds good to me.

What about a slight change in the data structure, like `<BASE_URL>/economic-viability-factor`:
```json
{
    "type": "fixed-owl",
     "amount": 2500
}
```

```json
{
    "type": "subsidy-factor",
     "amount": 5.0
}
```

anxolin:
We are having this conversations now in 2 places. Let's continue here https://github.com/gnosis/dex-react/pull/1383

PR #1383, opened 2020-09-07
Fetch min trade amount in OWL

Velenir:
~Mock fetch `subsidizeFactor`, defaulting to `1`~

Fetch min trade amount in OWL


anxolin (commented):
Buut at the end, we get the subsidy factor or the actual minimum trade in OWL?
https://gnosisinc.slack.com/archives/CC3C5SMB5/p1599224119069800?thread_ts=1599223469.068400&cid=CC3C5SMB5

gnosis-info:
Travis automatic deployment:
 https://pr1383--dexreact.review.gnosisdev.com

Velenir:
> Buut at the end, we get the subsidy factor or the actual minimum trade in OWL?

Here -- subsidy factor
Did we decide that backend would provide `minTradeAmount`?

W3stside (approved):

anxolin:
My understanding is that they will hardcode it to `2500 OWL` the solver, but we just have to align with them. @nlordell and @e00E probably can confirm

nlordell:
I think the 2500 OWL minimum trade amount will be set for the next ICO, but not something permanent (@josojo is this correct?). Ultimately, we need to provide an endpoint to expose the value to the front end, so that it we don't need to make code changes everytime the configuration changes.

anxolin:
> I think the 2500 OWL minimum trade amount will be set for the next ICO, but not something permanent (@josojo is this correct?)

Yes, my understanding too

> Ultimately, we need to provide an endpoint to expose the value to the front end, so that it we don't need to make code changes every time the configuration changes.

Cool! For us is fine if this is for now, just a dummy endpoint that return the hardcoded 2500. In the future, we reiterate on it to be more in sync with the solver? This way, we can deploy 1.4 version

Velenir:
So we will be getting a `min tradable amount in OWL`, by calling `something/min-tradable/<token-id>`?

e00E:
As a first version we will have a route that is not specific to the token and returns the result always in owl.
Something like `GET /api/v1/minimum-order-size-in-owl` -> `2500` (2500 in owl base units ~= 2500 usd), if you wanted to convert to atoms it would be `2500 * 10**18`).
Or we could make it return in atoms immediately? What do you prefer?

Then later we think it makes sense to have a version per token that returns the amount in that token instead of owl.

anxolin:
I've just answered a message on the same conversation we are having in here:
https://github.com/gnosis/dex-react/issues/1380

Let's continue in this thread. 
I copy here my proposal (based on Valentin's one in that issue).

What about a slight change in the data structure, like `<BASE_URL>/economic-viability-factor`:
```json
{
    "type": "fixed-owl",
     "amount": 2500
}
```

```json
{
    "type": "subsidy-factor",
     "amount": 5.0
}
```

e00E:
The more complicated structure that you're quoting was my idea when I thought we would have this as a separate service whose only job is to expose the configuration of the driver.
Now that we have decided to implement this in the price estimator we can reduce the api to the single number from my previous post. This works because the price estimator can fetch eth price and gas price and do the calculation itself.

anxolin:
> The more complicated structure that you're quoting was my idea when I thought we would have this as a separate service whose only job is to expose the configuration of the driver.
Now that we have decided to implement this in the price estimator we can reduce the api to the single number from my previous post. This works because the price estimator can fetch eth price and gas price and do the calculation itself.

Aha, u mean that now you can offer the estimation in OWL always, so when we go back to using a subsidy factor, u would still return it in OWL? If that's it, sounds great to me

e00E:
Yes exactly. The frontend won't have to know about the subsidy factor. It will get a final absolute value always.

e00E:
The route is now available in staging:

`curl -X GET "https://price-estimate-mainnet.dev.gnosisdev.com/api/v1/minimum-order-size-owl" -H  "accept: application/json"`
`870876662379071000000`

It returns "the minimum size of an order in owl atoms for it to be considered by the solver according to current economic viability constraints". I still need to update the staging config with the economic viablity constraints but it turns out we default to a subsidy factor of 10 which happens to be correct at the moment. So this value above which is roughly 870 owl is correct.

anxolin:
Thanks @e00E ! That's great. 
But this 870 is what will change soon to 2.500? right?

e00E:
2500 is what we thought we would set it to. But now the gas price is low so that with our subsidy factor we get something lower than 2500 so there is no reason to artificially increase it.

anxolin:
> 2500 is what we thought we would set it to. But now the gas price is low so that with our subsidy factor we get something lower than 2500 so there is no reason to artificially increase it.

Yes, makes sense, but is it capped to 2500 if it raises again?

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -151,6 +156,26 @@ export class DexPriceEstimatorApiImpl implements DexPriceEstimatorApi {
...
+    const baseUrl = this._getBaseUrl(networkId)
+    return `${baseUrl}minimum-order-size-owl`
+  }
+
+  public async getMinOrderAmounInOWL(networkId: number): Promise<BigNumber> {
+    try {
+      const url = this.getMinOrderAmounInOWLURL(networkId)
+      const res = await fetch(url)
+      // not res.json() because backend returns "8738236863863283268688" big number of OWL in atoms
+      const minAmount = await res.text()
```
W3stside, line 169: :+1:

src/hooks/useMinTradableAmountInOWL.ts
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react'
+import useSafeState from './useSafeState'
+import { dexPriceEstimatorApi } from 'api'
+import BigNumber from 'bignumber.js'
+
+export function useMinTradableAmountInOWL(networkId: number): null | BigNumber {
```
W3stside, line 6: out of curiousity - why no `useMemo` here instead of the effect? is it for the initial `null` render?

W3stside (approved):

src/hooks/useMinTradableAmountInOWL.ts
Velenir: How would I use `useMemo`? `dexPriceEstimatorApi.getMinOrderAmounInOWL` is async

W3stside: ah yes, true. didnt even notice that

src/utils/minFee.ts
```diff
@@ -2,7 +2,6 @@ import { BigNumber } from 'bignumber.js'
 import { logDebug } from './miscellaneous'
 
 export const DEFAULT_GAS_PRICE = 40e9 // 40 Gwei
-export const SUBSIDIZE_FACTOR = 1 // no subsidy
 
 // account for a 20% change in the time it takes to mine the tx, and start the batch
 export const BUFFER_MULTIPLIER = 1.2
```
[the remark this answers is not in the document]
anxolin: well, I guess after this PR is not too relevant since you don't use it anymore. Agree with this PR that is good to keep the calculation around just in case come!

anxolin (approved):

src/api/dexPriceEstimator/DexPriceEstimatorApi.ts
```diff
@@ -151,6 +156,26 @@ export class DexPriceEstimatorApiImpl implements DexPriceEstimatorApi {
...
 
+  public getMinOrderAmounInOWLURL(networkId: number): string {
+    const baseUrl = this._getBaseUrl(networkId)
+    return `${baseUrl}minimum-order-size-owl`
+  }
+
+  public async getMinOrderAmounInOWL(networkId: number): Promise<BigNumber> {
+    try {
+      const url = this.getMinOrderAmounInOWLURL(networkId)
+      const res = await fetch(url)
```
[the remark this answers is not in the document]
Velenir: We don't block the user.

If we want to timeout at some point, should timeout on all fetch requests, for orderbook, for price-estimates.
Either in a general way `DexPriceEstimatorApi` or have a wrapper over `fetch`
In a separate PR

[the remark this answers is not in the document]
Velenir: Timeout everywhere it makes sense is a good idea

PR #1388, opened 2020-09-08
Swap price fix

anxolin:
* Addresses one TODO in the swap tokens storybook
* Improves the stories
* Defines a reusable decorator for when we want to "frame and center" the components. For small fragments of UI like this one I find it useful.
* There was a small issue with the CSS that made the icon to be displayed below the token symbol (instead of besides)


gnosis-info:
Travis automatic deployment:
 https://pr1388--dexreact.review.gnosisdev.com

src/components/common/SwapPrice.stories.tsx
```diff
@@ -1,23 +1,33 @@
...
-Basic.args = {
-  baseToken: GNO,
-  quoteToken: DAI,
-  isPriceInverted: true,
+  return (
+    <SwapPrice
+      baseToken={GNO}
+      quoteToken={DAI}
+      isPriceInverted={isPriceInverted}
+      onSwapPrices={(): void => {
```
Velenir, line 23: is anything supposed to happen onClick? I see actions logged in the Actions tab. But no console.logs, not even inside iframe context

anxolin: Yes, that's an issue. It happens because https://github.com/gnosis/dex-react/blob/develop/.storybook/preview.tsx#L7 intercepts the call overriding me!

I'm not sure I fully understand actions here, so probably missing the point why this was added in the first place, but I would say actions can be defined locally to the specific story if they are needed

I'll remove it in this PR, but let me know your thoughts

anxolin: I think is better to keep actions local instead of having the global one. 
This way, we are more declarative with the properties, and we don't override the `on<Something>` automatically.

In my case I believe it was overridden, cause I allow to provide as parameters a Partial of the Props, so I can define the defaults, and let anyone use the template to override any prop. In this case, I'm being injected the `onSwapPrices` automatically because of the naming convention

src/storybook/decorators.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
+import { Frame } from 'components/common/Frame'
+import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/types'
+
+export const CenteredAndFramed = (DecoratedStory: () => StoryFnReactReturnType): JSX.Element => (
+  <div style={{ textAlign: 'center' }}>
+    <Frame style={{ display: 'inline-block' }}>{DecoratedStory()}</Frame>
```
Velenir, line 8: While this is ok, it only centers horizontally. Why not go for a `flex` like we do everywhere else

src/components/trade/Price.stories.tsx
```diff
@@ -9,7 +9,6 @@ import { GNO, DAI } from 'storybook/tokens'
 
 export default {
   title: 'Trade/Price',
-  component: Price,
```
Velenir, line 12: ?

Velenir (approved):

src/storybook/decorators.tsx
W3stside: or `inline-flex`

W3stside (approved):

src/storybook/decorators.tsx
anxolin: Cause I didn't mean to center it vertically :) 

this is just when we want this frame:
image


On top that I would prefer to see it in the top, If we center vertically probably the actions can easily be hiding part of the component.

I'll merge it, let me know if u are still not convinced, I can easily make a PR to address it after.

PR #1389, opened 2020-09-08
Allow to connect from non-localhost

anxolin:
Simplifies the `package.json` so we don't have a script just for browserstack.
Also this way we can access the dev server from localhost or from our local network. This is convenient for access from our mobile using WIFI

Additionally, makes it coherent with storybook, which we already allowed:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "node": "12"
   },
   "scripts": {
-    "start": "NODE_ENV=development webpack-dev-server --mode development --hot",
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1389--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1389--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):

W3stside (approved):

PR #1404, opened 2020-09-11
Trade layout

anxolin:
* Adds a new layout for the v2.0 so we can start shaping it
* Add a new route for `/v2` that displays a trading page
* Maintains all the old routes (so we can maintain v2 and v1 in the same branch)
* Creates the story books for the new trading page
* Creates the story books for the new and old layout

**NOTE 1**: Not found page was duplicated, I know I can refactor, but probably we want to let them be different. We can see what we do after we have some designs.

**NOTE 2**: It doesn't use the GlobalStyles on purpose. I think v2 can define styles in the components and avoid using this global styles.

**TEST**:
* https://pr1404--dexreact.review.gnosisdev.com/v2
* https://pr1404--dexreact.review.gnosisdev.com/v2/one-weird-page

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -127,7 +127,7 @@ const App: React.FC = () => (
           <TradingLayout>
             <React.Suspense fallback={null}>
               <Switch>
-                <Route path="" exact component={Trading} />
+                <Route path="/v2" exact component={Trading} />
                 <Route component={NotFound2} />
               </Switch>
             </React.Suspense>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -121,7 +121,6 @@ const initialUrl = getInitialUrl()
 // App
 const App: React.FC = () => (
   <>
-    <GlobalStyles />
     <Router basename={process.env.BASE_URL}>
       <Switch>
         <Route path="/v2">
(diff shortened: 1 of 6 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1404--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1404--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/App.tsx
```diff
@@ -110,24 +123,38 @@ const App: React.FC = () => (
...
-            <Route component={NotFound} />
-          </Switch>
-        </React.Suspense>
-      </SwapLayout>
+      <Switch>
+        <Route path="/v2">
+          <TradingLayout>
+            <React.Suspense fallback={null}>
+              <Switch>
+                <Route path="" exact component={Trading} />
```
Velenir, line 130:
```suggestion
                <Route path="/v2" exact component={Trading} />
```
there are no relative routes in react-router v5, this one becomes a catch-all with `path=""`
You can look forward to [v6](https://reacttraining.com/blog/react-router-v6-pre/#relative-route-path-and-link-to) though

Velenir (changes requested):

Velenir (approved):

PR #1406, opened 2020-09-14
Apollo + Last trades

anxolin:
## Included

This PR is just some small test with the following objectives:
* Integrate Apollo, so we can easily query Gnosis Protocol subgraph
* Implement a small widget, that query the graph and at the same time separates the logic and the presentation. I wanted to see how it would look with the query logic closer to the widget, instead of using APIs
* For the ^^^ I implemented a basic LastTrades component and it's widget
* I added some stories for both

This PR adds also a new decorator that can be used in any stories that requires using thegraph API.

image

## Not included
Since I did this on Friday (not part of the sprint), I didn't get to do some of the things that should
Some examples of what future PRs could include
* Add better support for typescript
* Add better parametrization of the widget
* Pagination?
* Update automatically (change https for WS)
* Highlight prices when they change
* Do a design and style the component
* Better formatting

## Test
Check out the storybook


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1406--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1406--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/trade/LastTrades/LastTradesWidget.tsx
```diff
@@ -0,0 +1,51 @@
...
+    size: new BigNumber(sellVolume).div(new BigNumber(10).pow(18 - sellTokenDecimals)),
+    time: new Date(Number(tradeEpoch) * 1000),
+  }
+}
+
+export const LastTradesWidget: React.FC<Props> = (props) => {
+  const { loading, error, data } = useQuery(GET_TRADES)
+
+  const { quoteToken } = props
+  const lastTrades: LastTradesItem[] = data ? data.trades.map(toLastTradeItem) : []
```
Velenir, line 48: memoize?

src/components/trade/LastTrades/LastTrades.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const LastTrades: React.FC<Props> = (props) => {
+  const { loading, error, trades } = props
+  const lastTrades = trades.map((trade) => {
+    const { id, size, price, time } = trade
+    return (
+      <li key={id}>
+        {formatPrice(size)} - {formatPrice(price)} - {formatDateLocaleShortTime(time.getTime())}
+      </li>
+    )
+  })
```
Velenir, line 39: memoize?

src/components/trade/LastTrades/LastTradesWidget.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import { TokenDetails } from 'types'
+import { LastTrades, LastTradesItem } from './LastTrades'
+
+import { useQuery, gql } from '@apollo/client'
+import BigNumber from 'bignumber.js'
+import { calculatePrice } from '@gnosis.pm/dex-js'
+
+const GET_TRADES = gql`
+  query GetTrades {
```
Velenir, line 10: I wonder if `subscription` will work. With WebSockets

src/storybook/decorators.tsx
```diff
@@ -2,9 +2,22 @@ import React from 'react'
...
 
 export const CenteredAndFramed = (DecoratedStory: () => StoryFnReactReturnType): JSX.Element => (
   <div style={{ textAlign: 'center' }}>
     <Frame style={{ display: 'inline-block' }}>{DecoratedStory()}</Frame>
   </div>
 )
+
+const apolloClient = new ApolloClient({
+  uri: 'https://api.thegraph.com/subgraphs/name/gnosis/protocol',
+  cache: new InMemoryCache(),
```
Velenir, line 16: Depending on a lot of stuff, if we use Apollo outside of storybook in the actual app, we may need to setup cache invalidation, or subscriptions, or something

I suspect this will just keep resolving queries from cache after first fetch.

Velenir (commented):
Decentralization... :wave:

anxolin:
> Decentralization... 👋

Thegraph is decentralized (in a way), and are in their way of further decentralize the API service. If we are building a classical UI we'll need to consume the graph, or you think we can do it onchain?

src/components/trade/LastTrades/LastTradesWidget.tsx
anxolin: I think so, I haven't got to try it, but I understand it's transparent and that's the magic of using their `useQuery`. Anyways, I just kept things simple in this PR as explained in the description

src/components/trade/LastTrades/LastTradesWidget.tsx
anxolin: Sure, I can memoize, but is it worth it at this point for pure data and given how simple tries to be this component?

image

src/components/trade/LastTrades/LastTrades.tsx
anxolin: Happy to do it, but for this early state I would consider it also premature optimization

src/storybook/decorators.tsx
anxolin: Good point. Should we create an issue with all the things I mentioned in my description, plus this, so we keep track of them as we learn more about apollo?

This PR was not part of our sprint, so I didn't go too fine grained into details. This `InMemoryCache` works great for now :P

Velenir (approved):
For use in Storybook this is good

PR #1408, opened 2020-09-14
Fix unwrapped promise cache result

Velenir:
Using `CacheMixin` cached promise result of a method, but didn't change return type of that method.
So when you would expect `method()` to return a Promise and would call `method().then()` it would fail if returned from cache

This PR removes caching of Promise result. Still leaves in place cache invalidation for rejected promises.

### DIFF
--- a/src/api/proxy/CacheMixin.ts
+++ b/src/api/proxy/CacheMixin.ts
@@ -62,10 +62,6 @@ export class CacheMixin {
       // store possibly a promise
       this.store(hash, value, ttl)
       Promise.resolve(value)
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1408--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1408--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin (approved):

PR #1413, opened 2020-09-14
Change skull for a warning sign

anxolin:
Some users are scared :) 

BEFORE
image

AFTER:
image

I personally like it less 😢
but we don't want to scare people


Velenir (approved):
☣️☠

PR #1416, opened 2020-09-16
Add debug in case of estimation error, and when sending a tx

anxolin:
Adds some logs for when the estimation failed. 

It helped me now to debug an issue, I think is good to add. Also, now we also show the transaction details in the log even if we succeed on the estimation

Also, it adds logs at API level, for methods that send transactions

### DIFF
--- a/src/api/deposit/DepositApi.ts
+++ b/src/api/deposit/DepositApi.ts
@@ -149,6 +149,8 @@ export class DepositApiImpl implements DepositApi {
     txOptionalParams,
   }: DepositParams): Promise<Receipt> {
     const contract = await this._getContract(networkId)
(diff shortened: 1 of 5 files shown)


Velenir (approved):

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1416--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1416--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin:
Sorry dima I added a few commits after your review, I've requested re-review

Velenir (approved):

PR #1422, opened 2020-09-17
Fix wrong quote

### LINKED ISSUE #1421: Price Quote token is bad, again
image

In the picture, is showing the price in WETH `0.00401`, but then it says the units are `DAI` 

The logic is reversed:
image

anxolin:
Fixes #1421 

The inverse logics are a bit chaotic 😅

I tried to get to the root cause of why the token were wrong, and at the end there was many part together having the inverted logic, some errors were making other errors right! What a nightmare of debugging! This parts need to be refactored asap.

I tried to do every change as an individual commit, so maybe is interesting to review the commits in order. Each one, tries to make the code more readable regarding prices, or fixes a price logic error at one point of the app. 
Note that when something was fixed, it was showing another error in other part of the app. So the process was continue correcting every component until it was making sense again.

Please, review this PR in detail in case there's new issues.

### DIFF
--- a/src/components/trade/Price.stories.tsx
+++ b/src/components/trade/Price.stories.tsx
@@ -9,6 +9,7 @@ import { GNO, DAI } from 'storybook/tokens'
 
 export default {
   title: 'Trade/Price',
+  component: Price,
 } as Meta
 
 const Template: Story<Props> = (props) => {
--- a/src/components/trade/Price.stories.tsx
+++ b/src/components/trade/Price.stories.tsx
@@ -24,8 +24,8 @@ const Template: Story<Props> = (props) => {
 
 export const Basic = Template.bind({})
 Basic.args = {
-  sellToken: GNO,
-  receiveToken: DAI,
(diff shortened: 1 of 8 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1422--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1422--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/trade/Price.stories.tsx
```diff
@@ -24,8 +24,8 @@ const Template: Story<Props> = (props) => {
 
 export const Basic = Template.bind({})
 Basic.args = {
-  sellToken: GNO,
-  receiveToken: DAI,
+  quoteToken: GNO,
```
Velenir, line 27: While you are at it, add `component` to `export default`

Velenir (approved):
Good work :+1:

PR #1424, opened 2020-09-17
Refactor swap price

anxolin:
Refactor of the 3 places where we show the denomination of the price:

1. In the confirm modal: It was using it's own component. Now uses SwapPrice component:
image

2. In the limit price input: For some reason, we stopped showing the "per BASE-TOKEN" part. I've added this in this PR, plus used the refactorized SwapPrice component:
image

3. In the price suggestions: It's the same component, just with a `showBaseToken = false` 
image

PR #1425, opened 2020-09-18
Refactor story data

anxolin:
Refactors the data from the stories in storybook, and organize a new structure for utils and types so we can keep it more organized.

Also, I've added a new list of mock tokens for building stories.

Lastly, I addressed the comments in https://github.com/gnosis/dex-react/pull/1424 and added the controls for selecting the quote/base token.

Technical note: I didn't use `useMemo` on purpose for the selection of base/quote, I think we don't need to optimize this and is easier to read. Also, I don't think it makes a big difference in this case ;)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1425--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1425--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/storybook/utils/index.ts
```diff
@@ -0,0 +1,18 @@
+import { TokenDetails } from 'types'
+
+// searches for Token by symbol or address
+// accepts default in case none found
+export function findFromListHoC(tokens: TokenDetails[]): typeof _findToken {
+  function _findToken(symbolOrAddress: string): TokenDetails | undefined
+  function _findToken(symbolOrAddress: string, defaultToken: TokenDetails): TokenDetails
+  function _findToken(symbolOrAddress: string | undefined, defaultToken: TokenDetails): TokenDetails
```
Velenir, line 8: you call `findToken(undefined, defaultT)` somewhere?

Velenir (approved):

src/storybook/utils/index.ts
anxolin: Yes, https://github.com/gnosis/dex-react/pull/1425/files#diff-d7d520ccdf63c7771001d0c9e57d5b23R39

since I provide the default, it was strange to me not to allow undefined too there. It was convenient

PR #1445, opened 2020-09-25
Extra gas per order for order cancellation

Velenir:
> Multiply 5K by the number of IDs being deleted, and add it to the normal estimation

Closes ##1444

### DIFF
--- a/src/api/earmark.ts
+++ b/src/api/earmark.ts
@@ -1,4 +1,5 @@
 import { logDebug } from 'utils'
+import { hexToNumber, numberToHex } from 'web3-utils'
 
 // 0.1 Gwei, reasonable gasPrice that would still allow flags replacement
 export const MIN_GAS_PRICE = (1e8).toString(10)
@@ -22,3 +23,17 @@ export const earmarkGasPrice = (gasPrice: string, userPrint: string): string => 
 
 // simple concatenation, with '0x' for empty data to have `0x<userPrint>` at the least
 export const earmarkTxData = (data = '0x', userPrint: string): string => data + userPrint
+
+const GAS_PER_DATA_CHAR = 8 // wei
+
+// increases pre-estimated gas by the amount needed to save extra earmark data to chain
+// 16 wei per 2 characters
+export const calcEarmarkedGas = (gas: string | number, userPrint: string): string => {
+  const gasNumber = hexToNumber(gas)
+
+  const earmarkGas = GAS_PER_DATA_CHAR * userPrint.length
+
+  const newGas = gasNumber + earmarkGas
+
+  return numberToHex(newGas)
+}
--- a/src/api/exchange/ExchangeApi.ts
+++ b/src/api/exchange/ExchangeApi.ts
@@ -190,6 +190,8 @@ export interface ExchangeApiParams extends DepositApiDependencies {
 
 type TradeEvent = BatchExchangeEvents['Trade']
 
+const EXTRA_GAS_PER_ORDER_CANCELLATION = 5000
+
 /**
  * Basic implementation of Stable Coin Converter API
  */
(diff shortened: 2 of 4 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1445--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1445--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin (commented):
Just what we saw about the earmark gas estimator accounting some extra gas in case the estimation comes from before

src/api/earmark.ts
```diff
@@ -22,3 +23,17 @@ export const earmarkGasPrice = (gasPrice: string, userPrint: string): string =>
 
 // simple concatenation, with '0x' for empty data to have `0x<userPrint>` at the least
 export const earmarkTxData = (data = '0x', userPrint: string): string => data + userPrint
+
+const GAS_PER_DATA_CHAR = 8 // wei
+
+// increases pre-estimated gas by the amount needed to save extra earmark data to chain
+// 16 wei per 2 characters
+export const calcEarmarkedGas = (gas: string | number, userPrint: string): string => {
```
anxolin, line 31: nit: I would have make this function responsibility only to get the gas for earmark, also I understand that can be static. The earmark size is not changing

Then the SUM of the two gas would happen in the middleware

anxolin (approved):

src/api/earmark.ts
Velenir: Earmark size doesn't change in our particular use case, because we have a fixed number of flags. But for a generic use case it is nice to make this a function of both gas and userPrint, always

I think the responsibility of both gas and data lies with whatever changes any one of them, walletApi in this case
That way middleware doesn't have to bother with hex->num->hex conversion and is more opaque that way.

src/api/earmark.ts
```diff
@@ -22,3 +23,17 @@ export const earmarkGasPrice = (gasPrice: string, userPrint: string): string =>
 
 // simple concatenation, with '0x' for empty data to have `0x<userPrint>` at the least
 export const earmarkTxData = (data = '0x', userPrint: string): string => data + userPrint
+
+const GAS_PER_DATA_CHAR = 8 // wei
```
alfetopito, line 27: Interesting.

I know about the issue with the cancellation lacking gas, but was this much extra data also causing problems?

alfetopito (approved):

src/api/earmark.ts
anxolin: It was a nit, so it's not a big deal but I'm referring to two things here:
- https://github.com/gnosis/dex-react/pull/1445/files#diff-515503171f912f8f40997a12b3f94afcR371 This line is modifying the gas price, adding some "extra". I think this "getting extra" should be independent from the calculation on the result amount. For me that function doesn't need the "gas" param. What you do with that extra gas is the middleware responsability imo (I'm if we leave this untouched, this is a ubernit)
- Once you have a function that returns the extra gas for the 8 hex digits, you don't need to calculate it over and over for every call, since the earmark data doesn't change. It could be statically calculated or cached. `userPrint` don't change. 

Neither of the above are really important

src/api/earmark.ts
anxolin: We estimated before the earmark. Cause the gas estimation override happens before the transaction is sent.
The earmark has some costs. It's the responsibility of the middleware that adds some extra data, to adjust the gas to cover the cost of the new data

anxolin: The gas difference in our tests was:
`72185 - 72073 = 112`

`112 / 14 digits = 8 gas per digit`

This is how we arrived at the magic number

src/api/earmark.ts
Velenir: > Once you have a function that returns the extra gas for the 8 hex digits, you don't need to calculate it over and over for every call, since the earmark data doesn't change. It could be statically calculated or cached. userPrint don't change.

You convinced me. Will change, on monday. `walletApi.userPrint = {dataSuffix: string, extraGas: number}` or something

src/api/earmark.ts
alfetopito: Not questioning the number, just asking whether 112 gas might cause the tx to fail

anxolin: if I understand the question, yes. This missing 112 would make the tx fail. Cause it really needs this gas, so the tx cannot complete

PR #1451, opened 2020-09-28
Require(*).default storybook workaround

Velenir:
When using `require.context` with our webpack config a required module takes shape of
```ts
{
    default: "/5a86806e7c0681c126888fe301054197.png",
    Symbol(Symbol.toStringTag): "Module",
    __esModule: true
 }
```

When using with storybook module becomes `'static/media/0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d.5a86806e.png'`

Issue storybookjs/storybook#11610

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -34,6 +34,20 @@ module.exports = {
       customRules: custom.module.rules,
     })
 
+    const rulesWithMarkdownAndImages = replaceRulesForFile({
+      filename: 'token_address.png',
+      baseRules: rulesWithMarkdown,
+      customRules: [{
+        // what storybook uses
+        test: /\.(svg|ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|cur|ani|pdf)(\?.*)?$/,
+        loader: 'file-loader',
+        // only change from storybook -- esModule: false -> true
+        // this way require() and require.context need module.default
+        // same as in webpack config
+        options: { name: 'static/media/[name].[hash:8].[ext]', esModule: true }
+      }]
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1451--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1451--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin (commented):
I was hopping the solution would be sth like a change in our webpack, babel or TS config.

Anyways, I'm ok with the solution. Only thing, please move it to a util, and use it also in the other require.context use (wallets)

anxolin:
I was thinking it had to do with:
https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

Velenir:
> I was thinking it had to do with:
> https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs

Also though it was something there, still think so. But `import()` specifically works. Babel config is something to look into of course

anxolin (commented):
Great!!
Do you think you can remove then the "temporal fix" logic?

Velenir:
> Do you think you can remove then the "temporal fix" logic?

But I did :confused:  Did I leave something in?

anxolin:
> But I did 😕 Did I leave something in?

Weird, now it shows the other change, I was only seeing the webpack config, not sure what happened.... 
All good!

anxolin (approved):

W3stside (approved):

PR #1452, opened 2020-09-29
Fix xDAI token break

Velenir:
When switching to xDAI chain and refreshing the page it would break with `symbol of undefined` error (@anxolin knows what I'm talking about). First experienced at https://pr1449--dexreact.review.gnosisdev.com/

That would happen because we cast TS types for `getToken` to `TokenDetails` when it could be `undefined`
When there are no tokens for the network (in `config.yaml` or elsewhere) `tokens` array becomes empty
So after adding xDAI tokens to config, the error would fix itself, but still not a good idea

This PR allows for `tokens = []` and as a side-effect allows to switch to an unsupported chain without the whole app coming down.
Try Ropsten, for example.

### DIFF
--- a/src/api/tokenList/TokenListApi.ts
+++ b/src/api/tokenList/TokenListApi.ts
@@ -3,7 +3,7 @@ import { getTokensByNetwork } from './tokenList'
 import { logDebug } from 'utils'
 import GenericSubscriptions, { SubscriptionsInterface } from './Subscriptions'
 import { TokenDetailsConfig } from '@gnosis.pm/dex-js'
-import { DISABLED_TOKEN_MAPS } from 'const'
+import { DISABLED_TOKEN_MAPS, EMPTY_ARRAY } from 'const'
 
 const addOverrideToDisabledTokens = (networkId: number) => (token: TokenDetails): TokenDetails => {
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1452--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1452--dexreact.review.gnosisdev.com/storybook/)**: Component stories

alfetopito (approved):
The app doesn't horribly break anymore when on an unsupported network \o/

anxolin (approved):
This is great!

I really like the side effect.
We'll probably need to handle the "unknown network logic" higher in the app, and show a nicer message with the list of supported networks. But can be done in a separate PR.

nit: it creates a flickery effect, showing the widget with a couple of tokens, and then disappears. I thought it was on first load only (like, load default network, and then select the actual one), but ir does it every time I change tabs. 
Not a big deal if we address what I mentioned above

Velenir:
> flickery effect

fixed

anxolin (approved):
>> flickery effect
> fixed

Still happens. Also, is not a big deal

Velenir:
> Still happens. Also, is not a big deal

Build broke, try again

anxolin (approved):
It works! great fix

Issue #1453, opened 2020-09-30
Remove id in `initialTokenList` config


anxolin:
The ID is currently required for the config of the token list
However, it shouldn't be, the ID is loaded from the contract, and can change by network.

During some refactor of the types in `dex-js` (https://github.com/gnosis/dex-js/pull/187) and it's coresponding dex-react PR, it was added some type to signal the intent to remove the ID. There's a TODO now in the code pointing to this issue
`type TokenDetailsConfigLegacy = TokenDetailsConfig & WithId`

https://github.com/gnosis/dex-js/pull/187/files#diff-6c58ae6ccad72d41e6c1b57e29ac7621R35

Issue #1442, opened 2020-09-24
safeTokenName util from dex-js doesn't need some of the properties it requires


anxolin:
It requires a `TokenDetails`, and this includes an `id` that is not required.

In general, I think we have a lot of duplications in `TokenDetails` like types. 
I think we can define lesser types, and give them proper names. 

Also, #1440 removes the `image` from the de token details, probably we can define something in this direction:
* TokenErc20:  what we call now `MinimalTokenDetails`?
* Token: It's a `TokenErc20` with an optional id and addressMainnet
* TokenDex: Token with id as mandatory
* TokenDetails: Token used for bootstrapping some token config. Probably we can instead of adding the id directly, we add it per network.

PR #1454, opened 2020-09-30
Refactor and simplify token types

anxolin:
Addresses #1442 

This is a big refactor of the types. 
There was too many Token definitions. Now most of the token definitions are in `dex-js`

Also, addresses one annoying thing, that we declared `TokenDetails` in dex-js and dex-react, making it super confusing.
Now TokenDetail is only in dex-react. 

dex-js has TokenDex (former TokenDetail) and these other definitions:
```
export type TokenErc20 = WithSymbolAndName & WithAddress & WithDecimals
export type Token = TokenErc20 & WithAddressMainnetOpt & Partial<WithId>
export type TokenDex = TokenErc20 & WithAddressMainnetOpt & WithId
```

Depends on dex-js `v0.7.2` that shoud be tagged as soon as we merge https://github.com/gnosis/dex-js/pull/187

Also, I leave this for other PR because is 100% independent of this one: https://github.com/gnosis/dex-react/issues/1453
Here I just gave a name signaling the intent of changing the config so we don't use the `id` anymore.

### DIFF
--- a/src/components/TokenOptionItem.tsx
+++ b/src/components/TokenOptionItem.tsx
@@ -153,19 +153,16 @@ const generateMessage = ({
   switch (reason) {
     // not registered --> advise to register
     case TokenFromExchange.NOT_REGISTERED_ON_CONTRACT:
-      if (!token)
+      if (!token) {
         return (
           <a href="https://docs.gnosis.io/protocol/docs/addtoken1/" rel="noopener noreferrer" target="_blank">
             Register token on Exchange first
           </a>
         )
+      }
+
       return (
-        <OptionItem
-          name={token.name}
-          symbol={token.symbol}
-          address={token.address}
-          addressMainnet={token.addressMainnet}
-        >
+        <OptionItem name={token.name} symbol={token.symbol} address={token.address}>
           <ExtraOptionsMessage
             href="https://docs.gnosis.io/protocol/docs/addtoken1/"
             rel="noopener noreferrer"
--- a/src/services/factories/getTokenFromErc20.ts
+++ b/src/services/factories/getTokenFromErc20.ts
@@ -1,4 +1,3 @@
-import { TokenDetails } from 'types'
 import { getErc20Info } from 'services/helpers'
 import { Erc20Api } from 'api/erc20/Erc20Api'
 import Web3 from 'web3'
--- a/src/services/factories/getTokenFromErc20.ts
+++ b/src/services/factories/getTokenFromErc20.ts
@@ -3,6 +3,8 @@ import { getErc20Info } from 'services/helpers'
 import { Erc20Api } from 'api/erc20/Erc20Api'
 import Web3 from 'web3'
 import { logDebug } from 'utils'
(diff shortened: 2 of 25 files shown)


src/services/factories/getTokenFromErc20.ts
```diff
@@ -14,10 +15,8 @@ export interface TokenFromErc20Params {
   networkId: number
 }
 
-export type TokenFromErc20 = Omit<TokenDetails, 'id'> | null
-
 export function getTokenFromErc20Factory(closureParams: Params) {
-  return async (params: TokenFromErc20Params): Promise<TokenFromErc20> => {
+  return async (params: TokenFromErc20Params): Promise<TokenErc20 | null> => {
```
Velenir, line 19: Why wan't this `| null` before?

src/services/index.ts
```diff
@@ -159,30 +159,35 @@ export const fetchTokenData = async ({
...
+    if (!tokenInExchange) {
+      // Token is not registered in the contract
       return {
         token: erc20Token,
         reason: erc20Token ? TokenFromExchange.NOT_REGISTERED_ON_CONTRACT : TokenFromExchange.NOT_ERC20,
         tokenAddress: params.tokenAddress,
       }
-
-    if (!erc20Token)
+    } else if (!erc20Token) {
```
Velenir, line 169: Why add `else` if we have `return` above anyway?

Velenir (approved):

src/services/index.ts
anxolin: I just thought in this case it would be nice to show there's 3 cases, and they are all covered

not a big change. Consider a case where u accidentally remove the word "return" in one of the branches, it would still compile, but it would be an error

src/services/factories/getTokenFromErc20.ts
anxolin: It was https://github.com/gnosis/dex-react/pull/1454/files#diff-dfb2c72865d565e8a1e493c4f13fbce1L17

it was just a bit more obscure declaration

src/components/TokenOptionItem.tsx
```diff
@@ -153,19 +153,16 @@ const generateMessage = ({
   switch (reason) {
     // not registered --> advise to register
     case TokenFromExchange.NOT_REGISTERED_ON_CONTRACT:
-      if (!token)
+      if (!token) {
```
W3stside, line 156: 😤 :stuck_out_tongue_closed_eyes:

W3stside (approved):

src/components/TokenOptionItem.tsx
anxolin: what?? :) 
Do you really think is better to skip the brackets?

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1454--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1454--dexreact.review.gnosisdev.com/storybook/)**: Component stories

PR #1455, opened 2020-10-01
StyledModal - a re-export of Modali with custom Modal

W3stside:
Re-exports `modali` inside `components/common/StyledModal` with our custom `Modal` component that wraps `GlobalStyles` and removes unnecessary global CSS override

Re-export facilitates changing rest of app to new Modal code as it is now just the export path that needs to change, the rest remains the same


src/components/common/StyledModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+const Modal: React.FC<Modali.ModalProps> = (props) => (
+  <>
+    <ModaliGlobalStyle />
+    <Modali.Modal {...props} />
+  </>
+)
+
+// To import default Modali and not change much code elsewhere
+const StyledModali = { ...Modali, Modal }
```
W3stside, line 170: is this a performance concern @Velenir @anxolin

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1455--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1455--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/StyledModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+  body.modali-open .modali-wrapper-centered {
+    top: 0!important;
+  }
+  
+  body.modali-open .modali-body-style {
+    padding: 0;
+  }
+`
+
+const Modal: React.FC<Modali.ModalProps> = (props) => (
```
anxolin, line 162: An export is missing here, right?

src/components/common/StyledModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+  <>
+    <ModaliGlobalStyle />
+    <Modali.Modal {...props} />
+  </>
+)
+
+// To import default Modali and not change much code elsewhere
+const StyledModali = { ...Modali, Modal }
+
+export { StyledModali as default, ModalHook, toggleModaliComponent, useModali }
```
anxolin, line 172: I would declare the export of the modaly stuff on the top of the file, right after the exports.
I we could just do:
`export { ModalHook, toggleModaliComponent as toggleModalComponent, useModali as useModal } from 'modali'`

anxolin (commented):
It's an approved, great PR!
but, can we just rename the file from `StyledModal` to `Modal`. We pretend from the other parts of the app "modali" don't exist.

src/components/common/StyledModal.tsx
Velenir: Not at all :+1:

src/components/common/StyledModal.tsx
Velenir: Can't.
You need to import to get variables into the module scope. This is an error:
```js
export {useModali} from 'modali'
console.log(useModali)
```
 and this is just duplicating code:
```js
export { ModalHook, toggleModaliComponent as toggleModalComponent, useModali as useModal } from 'modali'
import { ModalHook, toggleModaliComponent as toggleModalComponent, useModali as useModal } from 'modali'
```

Velenir: I'm all for renaming vars to _useModal~i~_ at `export` or `import` time

Velenir (approved):

src/components/common/StyledModal.tsx
W3stside: no its being used below in extension of `modali`

src/components/common/StyledModal.tsx
W3stside: From what I understand my code here is correct.. right?

src/components/common/StyledModal.tsx
Velenir: Why do you want to export Modal when we have it inside `defaultexport.Modal`
That is if we want to follow modali compound components pattern and match modali namespace:
```js
import Modali from 'modali'
Modali.Button
Modali.Modal
```

so
```js
import Modal, { useModal, ModalHook } from 'components/common/StyledModal'
Modal.Button
Modal.Modal
```

Velenir (approved):

src/components/common/StyledModal.tsx
Velenir: Right.
Unless... @anxolin ?

src/components/common/StyledModal.tsx
```diff
@@ -0,0 +1,172 @@
+import React from 'react'
+import { createGlobalStyle } from 'styled-components'
+import Modali, { ModalHook, toggleModaliComponent, useModali } from 'modali'
+
+import variables from 'styles/variables'
+import { MEDIA } from 'const'
+
+const ModaliGlobalStyle = createGlobalStyle`
+// global root variables
+  ${variables}
```
anxolin, line 10: Do we need this?

src/components/common/StyledModal.tsx
anxolin: Is my suggestion that bad? I just find it less verbose

```js
import { Modal, useModal } from 'components/common/Modal';

const App = () => {
  const [modal, toggleModal] = useModal();

  return (
    <div>
      <button onClick={toggleModal}>Open modal</button>
      <Modal {...modal}>🔥Modal🔥</Modal>
    </div>
  );
};
```

src/components/common/StyledModal.tsx
anxolin: Well, I don't think mixing import and export is duplicating code.

In my opinion is more clear what we are doing, reexported some things renamed, and exporting a restyled Modal object.

```js
import { Modal } from 'modali'
export { ModalHook, toggleModaliComponent as toggleModalComponent, useModali as useModal } from 'modali'
```

src/components/common/StyledModal.tsx
W3stside: well to keep it as similar to the original import syntax it was kept this way. otherwise it involves changing 10+ files to use a slightly shorter syntax

src/components/common/StyledModal.tsx
W3stside: i'll keep it inside the `global.scss` file as one of the few styles we keep there and remove it here :+1:

src/components/common/StyledModal.tsx
anxolin: Well, the refactor is a find and replace, and the use seems to me natural. I feel that if you are abstracting the use of Modali and making a component is a fair suggestion

Not a big deal either way. So if you feel strong leave it like this. The rename I would do it (I don't like the "styled" word in our component)

PR #1459, opened 2020-10-01
Fix style scoping

Velenir:
Enough is enough. This error has been there for 3 months, at the least. Who knows how many more there are.
WE NEED LINTIING for **styled-components**

Vote on this with :+1: or :-1:

### DIFF
--- a/src/components/Settings/WalletConnect.tsx
+++ b/src/components/Settings/WalletConnect.tsx
@@ -105,7 +105,6 @@ const OrSeparator = styled.div`
   @media ${MEDIA.mobile} {
     padding: 1rem 1.5em;
   }
-}
 `
 
 const ErrorWrapper = styled.p`


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1459--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1459--dexreact.review.gnosisdev.com/storybook/)**: Component stories

alfetopito (approved):
:scream: jeez, and it still works? :shrug: 

is there linting for styled components?

anxolin (commented):
Yep, pls, let's add linting? Maybe we could even add it in this PR, so we also solve any other broken style

How did you see this? by chance of was this breaking something?

W3stside (approved):
lol

Velenir:
> How did you see this?

When that particular file was open I saw that problem reported in VS Code problems by some ts-styled-plugin which is used internally by some VSCode extension

Before you ask, that plugin doesn't help if added to tsconfig, only works as part of VScode

Velenir:
Superseded by #1471

PR #1472, opened 2020-10-02
Smart Quote PR Comment edits (to make simpler)

W3stside:
Merges into #1273 

Edits addressed in PR (simplification of some logic, comments to describe some logic)

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -158,9 +158,6 @@ const TradeWidget: React.FC = () => {
   const defaultValidFrom = calculateValidityTimes(trade.validFrom || validFromParam)
   const defaultValidUntil = calculateValidityTimes(trade.validUntil || validUntilParam)
 
-  const [priceShown, setPriceShown] = useState<PriceType>('DIRECT')
-  const swapPrices = (): void => setPriceShown((oldPrice) => (oldPrice === 'DIRECT' ? 'INVERSE' : 'DIRECT'))
-
   const defaultFormValues: TradeFormData = {
     [sellInputId]: defaultSellAmount,
     [receiveInputId]: '',
@@ -186,6 +183,10 @@ const TradeWidget: React.FC = () => {
       defaultTokenSymbol: initialReceiveToken,
     }),
   )
+
+  const [priceShown, setPriceShown] = useState<PriceType>('DIRECT')
+  const priceIsInverse = priceShown === 'INVERSE'
+  const swapPrices = (): void => setPriceShown((oldPrice) => (oldPrice === 'DIRECT' ? 'INVERSE' : 'DIRECT'))
 
   const { baseToken, quoteToken, wasPriorityAdjusted } = usePrioritiseTokensForPrice({ receiveToken, sellToken })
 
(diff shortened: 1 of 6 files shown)


src/components/trade/Price.tsx
```diff
@@ -230,16 +230,21 @@ export const Price: React.FC<Props> = ({
...
   })
 
-  const hideInverse = (!isPriceInverted && wasPriorityAdjusted) || (isPriceInverted && !wasPriorityAdjusted)
-  const hideDirect = (isPriceInverted && wasPriorityAdjusted) || (!isPriceInverted && !wasPriorityAdjusted)
+  // Show Inverse Price ... IF
+  // DIRECT + ADJUSTED || INVERSE + UNADJUSTED
+  // aka
+  // 1. price = DIRECT && market = ADJUSTED
+  // 2. price = INVERSE && market = UNADJUSTED
+  const marketAdjustedisPriceInverted =
```
anxolin, line 238: I understand now, and is more clear. Thanks for reducing the duplication and comments
`marketAdjustedisPriceInverted`

My suggestion, to keep doing it more readable is to get rid everywhere of `wasPriorityAdjusted` with the only exception of where this data is actually calculated: In the trade widget index.ts
From the subcomponents point of view, they don't need this information and it really makes it less difficult to reason about.

The index.ts would need to do probably something else. Probably is just adjusting `priceIsInverse` to have a similar logic as the one you did in this line. 

Probably, something like:
```js
const priceIsInverted = priceShown === 'INVERSE'
const priorityWasAdjusted = baseToken == receiveToken // Note you didn't need it as part of the function

// price is inverted, either because it was inverted by the user, or the token priority made the base/quote selection to change
const priceIsInverse = priceIsInverted && !priorityWasAdjusted || !priceIsInverted && priorityWasAdjusted`
```

See how now the children don't need to know "the reason" why is inverted. Only that the price is the other way around.

anxolin (commented):
Sorry david, still feels overcomplicated

src/components/TradeWidget/utils.ts
```diff
@@ -8,7 +8,7 @@ import { BATCH_START_THRESHOLD } from './validationSchema'
 import { BATCH_TIME_IN_MS } from 'const'
 import { TokenDetails } from 'types'
 
-export function calculateReceiveAmount(priceValue: string, sellValue: string): string {
+export function calculateReceiveAmount(priceValue?: string, sellValue?: string): string {
```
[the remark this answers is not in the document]
Velenir: can it be empty string?

src/components/TradeWidget/index.tsx
```diff
@@ -233,13 +234,7 @@ const TradeWidget: React.FC = () => {
...
   const { control, handleSubmit, reset, setValue, trigger } = methods
 
-  const {
-    sellToken: sellValue,
-    validFrom: validFromValue,
-    validUntil: validUntilValue,
-    price: priceValue,
-    priceInverse: priceInverseValue,
-  } = useWatch({
+  const { sellToken: sellValue, validFrom: validFromValue, validUntil: validUntilValue, price: priceValue } = useWatch({
```
Velenir, line 237: Should we switch to `useWatch` everywhere we use `watch`?

Velenir (commented):
[Build broke](https://travis-ci.com/github/gnosis/dex-react/builds/187839550#L733)

W3stside:
closing

Issue #1476, opened 2020-10-02
Smart quote selection


anxolin:
Solving this issue will also close #1273

The issue is to break a PR that got complicated. 
We need to break #1350 into smaller ones.
Note there's also #1344 (which addresses some review comments), superseded bt #1472

This will be the first of several PR. 
Additionally, there was some other tasks related to Smart Quote Selection, that would be required. This issue includes them in the following lists.

## Part of breaking PR #1350: Smart quote in Trading Widget
PRs will be merge to the WIP "smart-quote"
* [x] Add Priorities data file - #1474
* [x] Add utiltity / Test - #1474
* [x] useCannonicalMarket hook, probably we don't need it, the util should suffice --> not needed
* [x] Review why the original PR changed this type, they are not equivalent! https://github.com/gnosis/dex-react/pull/1350/files#diff-7a5f1dde90cfb4e531bbeb45baa98158R1
* [x] Consider deleting `wasPriorityAdjusted` from the utility `getMarket`
* [x] Load priorities for tokens (in API) #1478
* [x] Make the smart selection of the market in Trade Widget #1479
* [ ] Make the smart selection of the market in Orders

##  Part of breaking PR #1350: But not part of Smart quote selection 
PR will be merged against branch "develop"
* [x] Add get best ask endpoint - #1477
* [x] Create useBestAsk hook, use it in Price suggestions
~Update hooks library~ :point_left: won't do

## Improvements of Orders
* [x] Use the smart quote selection for the orders #1491
* [x] Allows to invert the quote, per row #1491
* [x] Additionally, adds a title for further clarification of the price :) #1491
* [x] Create `SmartPrice` for showing a price and the invert logic
* [ ] Storybook for `SmartPrice`. Good to have, not required
- [x] Allow to invert the market #1494
- [x] Allow to invert the market (just limit price)
- [x] We could consider not displaying 2 prices. Maybe we find a better
- [x] Everything is more clear if we break Filled/Total into 2 columns
- [x] Add horizontal scrolling to table 
- [x] Improve layout to match Trades new layout

## Improvements in Trades
These are not part of the initial PR that it's being reworked, but are part of the Smart Market Selection tasks.

- [x] Shows the market as a canonical in Trades #1494
- [x] Implement the concept of SIDE for Trades #1494
- [x] Add units to the price #1494
- [x] Rearrange Trades/Orders Row into more readable layout
- [x] Allow to invert the market #1494
- [x] Allow to invert the market (just limit price)
[...]

PR #1479, opened 2020-10-02
[Smart Quote]  User automatic market selection

anxolin:
Part of #1476

Make use of the smart market selection. Use it for the trade widget

This PR:
* Simplifies a bit how we model that the user inverted the market price. Now is a flag, before was DIRECT | INVERSE, which makes it more difficult to reason about. 
* Now `isPriceInverted` means, the user didn't like the "default market selection", so he changed how we quote the prices (in our base token)
* `PriceSuggestions` now receives info on the market and if this was inveted. He doesn't care what are we selling or buying
*  Same as above for the `Price` component
* `IMPORTANT`: Lastly, and key for this PR to work. Now we have to make sure we calculate the receive tokens properly. Before, we were using always the "directPrice" to calculate the receive tokens. Now we just check if the sellToken is the quote, to decide if we use the direct or inverse price


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1479--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1479--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):

src/components/TradeWidget/index.tsx
```diff
@@ -313,10 +314,17 @@ const TradeWidget: React.FC<TradeWidgetProps> = ({
...
+  // Get canonical market
+  const { baseToken, quoteToken } = useMemo(() => getMarket({ receiveToken, sellToken }), [receiveToken, sellToken])
+
   // Update receive amount
   useEffect(() => {
-    priceValue && sellValue && setValue(receiveInputId, calculateReceiveAmount(priceValue, sellValue))
-  }, [priceValue, priceInverseValue, setValue, sellValue])
+    if (priceValue && sellValue) {
+      // If price is quoted in sell tokens, we use it, otherwise we use the inverse
+      const priceUsedForReciveAmount = sellToken === quoteToken ? priceValue : priceInverseValue
```
W3stside, line 324: should be `priceUsedForReceiveAmount` (typo)

W3stside:
Confirmation modal is incorrect:
Screenshot from 2020-10-06 13-40-23

src/components/TradeWidget/index.tsx
W3stside, line 324: why not just use `wasPriorityAdjusted` here? isn't that the same thing

anxolin:
> Confirmation modal is incorrect:

Indeed. It was not using the canonical market. Check this commit 47def0a05e392fb89c5032ea4a532f65bd2353ed

W3stside (approved):
nice, thanks

src/components/TradeWidget/index.tsx
anxolin: Cause this is easier to reason about. In the old PR `wasPriorityAdjusted ` was being drilled down to components, and is not super clear what it means. 

wasPriorityAdjusted + isPriceInverse + sell/buy tokens, quote/base tokens ---> 🤯

Here it's easier to read:
* if sell token is the quote, I use the price
* otherwise, I use the inverse 

I don't care if user flipped the market, or the smart logic did it, I care about the units of the price

Issue #1398, opened 2020-09-10
XDAI integration


anxolin:
https://www.xdaichain.com/

## Tasks Web v1.5
* [x] Add xDAI to dex-js and update contracts: https://github.com/gnosis/dex-js/pull/184
* [x] Add xDAI test tokens, config, network utils, etc:  #1429
* [X] Handle TxLink (former EtherscanLink) generation  #1430 
* [x] Issue with token selection: #1435
* [x] Handle gas price estimation
* [x] Handle Wrap/Unwrap
* [x] Add "DAI logo" for wxDAI and remove DAI from the default list  https://github.com/gnosis/dex-react/pull/1487
* [X] TCR for xDAI. Ours? Will dxDAO will have it's own list? --> not needed, implemented a filter using config
* [X] Filter using config for networks with no TCR https://github.com/gnosis/dex-react/pull/1489
* [X] Add price estimation endpoint for production (pending) -->Not needed, we use web3
* [x] Make sure we use production price estimation, now that it's deployed
* [x] Wallet connect?
* [ ] Websocket connection
* [x] Add xDAI RPC field in settings page 
* [X] We changed our sorting of the token list because how we load them now. 
Should we add alphabetical order? Maybe with the exception of WETH and wxDAI, that should be shown as the first option? https://github.com/gnosis/dex-react/pull/1513
* [x] Test: Make test in IOS, Android, Desktop. Both Wallet Connect and in-app browser.
   * [ ] The Graph
   * [x] Token Load
   * [x] Workflow approve, deposit
   * [x] Workflow withdraw
   * [x] Workflow trading
   * [x] Workflow wrap/unwrap
   * [x] Workflow liquidity page 

**Not blocking web 1.5, but part of xDAI**
* [x] Review documentation, add xDAI into our docs and WIKI #1505
* [x] Telegram bot https://github.com/gnosis/dex-telegram/pull/212
* [x] Telegram bot: Allow to specify a list of addresses by network instead of a TCR. https://github.com/gnosis/dex-telegram/pull/212 made the TCR optional, but now it shows all addresses as "Maybe". It would be good to add a hardcoded list of trusted tokens by network
* [x] Further refactors, nice to have:
   * [x] Given all the changes required in the PRs, can we make it easier to add networks. Theoretically it should be as easy as add the new network in a couple of places, but there was some repetition: See https://github.com/gnosis/dex-react/pull/1429
  * [x] Improve boot of the app, handle crash

## Contract and tokens
[...]

PR #1489, opened 2020-10-05
Refactor and allow to filter in networks with no TCR

anxolin:
Part of https://github.com/gnosis/dex-react/issues/1398

This PR was created to avoid having to deploy a TCR for xDAI. 

The current behaviour of the app, is that **if there's no TCR or it's list is empty, then it doesn't apply any filters, so it will display ALL THE TOKENS**

This behaviour is a bit undesirable. For xDAI we have a default list of some basic tokens. Anyone deploying the app can deploy the TCR for xDAI but if it's not done, it will use the default list.

Also, moves the part of getting the ids and filtering them to its own function. The logic was already pretty complex, and was using too many loops. 

**NOT PART OF THIS PR**
The token loading is unnecessarily complex. This PR doesn't try to tackle that, although it does a small refactor on some parts it touches, the goal is not to do a full rewrite of this. It would be a good idea to invest in cleaning up this.

See: https://github.com/gnosis/dex-react/pull/1489#discussion_r500142151


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1489--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1489--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/services/factories/tokenList.ts
```diff
@@ -130,42 +130,57 @@ export function getTokensFactory(factoryParams: {
...
+      logDebug(`[tokenListFactory][${networkId}] Not using a TCR. Filtering by tokens in the config file`)
+      const tokensConfig = tokenListApi.getTokens(networkId)
+      const filteredAddressAndIds = tokensConfig
+        .map((token): [string, number] | null => {
+          const tokenId = listedAddressesAndIds[token.address]
+          return [token.address, tokenId] || null
+        })
+        .filter(notEmpty)
+
+      return new Map<string, number>(filteredAddressAndIds)
```
W3stside, line 162: you don't want to move this outside the `if/else` with a `let filteredAddressAndIds` outside the scope? seeing as you're returning `new Map(filteredAddressAndIds` in both cases

src/services/factories/tokenList.ts
```diff
@@ -191,7 +206,7 @@ export function getTokensFactory(factoryParams: {
           `[tokenListFactory][${networkId}] Got details for address ${partialToken.address}: symbol '${partialToken.symbol}' name '${partialToken.name}'`,
         )
         // Get id from address/id mapping
-        const id = addressesAndIds.get(partialToken.address) as number
+        const id = filteredAddressesAndIds.get(partialToken.address) as number
```
W3stside, line 209: can't this be `undefined`?

W3stside (approved):

src/services/factories/tokenList.ts
anxolin: I considered it, but just went for this option. I'm ok change it too. In this case I don't feel makes a huge difference

anxolin: 5b8e19c3e83925d232bdfb2ac7dcbab98bb0b40f

src/services/factories/tokenList.ts
anxolin: I don't think so. The logic for loading the tokens is a bit unnecessarily complex. This is why these things are required.
I did one small refactor in this PR, but they could have more refactors and probably this `as number` would disapear. But this is not part of this PR refactor all.

It cannot happen to be undefined, cause we made sure in the code above that only tokens listed in the exchange (with id) are in this `partialTokens` array

src/services/factories/tokenList.ts
```diff
@@ -130,42 +130,57 @@ export function getTokensFactory(factoryParams: {
...
     )
 
     const promises: Promise<TokenErc20 | string>[] = []
 
     // Go over all value (id) and key (tokenAddress) pairs registered on the contract
-    addressesAndIds.forEach((id, tokenAddress) => {
+    filteredAddressesAndIds.forEach((id, tokenAddress) => {
       if (!localAddressesMap.has(tokenAddress)) {
         // New token not in our local list, fetch erc20 details for it
         logDebug(`[tokenListFactory][${networkId}] Address ${tokenAddress} with id ${id} not in local list, fetching`)
```
Velenir, line 186: This looks complicated

Is the idea to:

1. get intersection of tokens from `tokenListApi.getTokens(networkId)` and `filteredAddressesAndIds`
2. fetch tokens that are in `filteredAddressesAndIds` but not in `tokenListApi.getTokens(networkId)` as ERC20
3. fill in the blanks of tokens from 2 from chain as ERC20
4. assign ids to fetched ERC20

If I'm correct, then I feel you have too many loops and a pretty complicated logic in general. Would the following do the same?

```ts
async function updateTokenDetails(networkId: number, numTokens: number): Promise<void> {
  // Get filtered ids and addresses
  const filteredAddressesAndIds = await _getFilteredIdsMap(networkId, numTokens)
  const localTokens: TokenDetails[] = []

  const allTokensFromApi = new Map(tokenListApi.getTokens(networkId).map((t) => [t.address, t]))
  const TokenDetailsPromises: Promise<TokenDetails | undefined>[] = []

  filteredAddressesAndIds.forEach((id, tokenAddress) => {
    const token = allTokensFromApi.get(tokenAddress)
    if (token) {
      localTokens.push(token)
    } else {
      const fullTokenPromise = getErc20DetailsOrAddress(networkId, tokenAddress).then((partialToken) => {
        if (typeof partialToken === 'string') {
          // We replaced potential null responses with original tokenAddress string for logging purposes
          logDebug(`[tokenListFactory][${networkId}] Address ${partialToken} is not a valid ERC20 token`)
          return
        }
        // If we got a valid response
        logDebug(
          `[tokenListFactory][${networkId}] Got details for address ${partialToken.address}: symbol '${partialToken.symbol}' name '${partialToken.name}'`,
        )
        // build token object
        const token: TokenDetails = { ...partialToken, id }

        return token
      })
      TokenDetailsPromises.push(fullTokenPromise)
    }
  })

  const fetchedAndFilledTokenDetails = (await Promise.all(TokenDetailsPromises)).filter(notEmpty)

  const tokenList = localTokens.concat(fetchedAndFilledTokenDetails)

  // Persist it \o/
  tokenListApi.persistTokens({ networkId, tokenList })
}
```

Velenir (changes requested):
This won't work, sorry

src/services/factories/tokenList.ts
```diff
@@ -130,42 +130,57 @@ export function getTokensFactory(factoryParams: {
...
 
-    logDebug(`[tokenListFactory][${networkId}] TCR contains ${tcrAddressesSet.size} addresses`)
+    if (tcrAddressesSet.size > 0) {
+      // If filtering by TCR
+      logDebug(`[tokenListFactory][${networkId}] TCR contains ${tcrAddressesSet.size} addresses`)
+      logDebug(`[tokenListFactory][${networkId}] Token id and address mapping:`)
+      const filteredAddressAndIds = Array.from(listedAddressesAndIds.entries()).filter(([address, id]) => {
+        // Remove addresses that are not on the TCR, if any
+        const isInTcr = tcrAddressesSet.has(address)
+        logDebug(`[tokenListFactory][${networkId}] ${id} : ${address}. On TCR? ${isInTcr}`)
```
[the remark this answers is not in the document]
anxolin: Great catch! I think I rushed this PR too much!

src/services/factories/tokenList.ts
```diff
@@ -130,42 +130,57 @@ export function getTokensFactory(factoryParams: {
...
+      })
+      return new Map<string, number>(filteredAddressAndIds)
+    } else {
+      // If not using a TCR, filter by default list
+      logDebug(`[tokenListFactory][${networkId}] Not using a TCR. Filtering by tokens in the config file`)
+      const tokensConfig = tokenListApi.getTokens(networkId)
+      const filteredAddressAndIds = tokensConfig
+        .map((token): [string, number] | null => {
+          const tokenId = listedAddressesAndIds[token.address]
+          return [token.address, tokenId] || null
```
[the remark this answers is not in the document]
anxolin: Yep, this is derived from the previous error on how I was accessing the Map in a Object fashion

src/services/factories/tokenList.ts
Velenir: @anxolin did you consider this refactor?

W3stside: :point_up: @Velenir's logic looks correct and is easier to read imo

anxolin: Proposal that reiterate on Dima's ideas. I think it simplifies it a bit further and fix some weird signature we had in the API
https://github.com/gnosis/dex-react/pull/1498/files

Velenir (approved):

PR #1491, opened 2020-10-06
[Smart Quote] Simplify prices in order row

anxolin:
WIP:
1. Use the smart quote selection for the orders

image

2. Allows to invert the quote, per row

3. Additionally, adds a title for further clarification of the price :) 
image

4. Moves the logic of displaying the "price + smart quote" into it's own component --> `SmartPrice`


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1491--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1491--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/OrdersWidget/OrderRow.tsx
```diff
@@ -94,24 +99,14 @@ interface OrderDetailsProps extends Pick<Props, 'order' | 'pending'> {
...
       <div className="order-details">
-        <strong>
-          {priceInverse} {displayTokenSymbolOrLink(sellToken)}
-        </strong>
-        <br />
-        {price} {displayTokenSymbolOrLink(buyToken)}
+        <SmartPrice
+          sellToken={sellToken}
+          buyToken={buyToken}
+          price={{ numerator: order.priceNumerator, denominator: order.priceDenominator }}
```
Velenir, line 108: This is recreated on each render, so inside SmartPrice `useMemo` deps change every time

src/components/common/SmartPrice.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface Props {
+  buyToken: TokenDetails
+  sellToken: TokenDetails
+  price: Fraction
+}
+
+export const SmartPrice: React.FC<Props> = ({ buyToken, sellToken, price: priceFraction }) => {
+  const [isPriceInverted, setIsPriceInverted] = useState(false)
+  const { baseToken, quoteToken } = getMarket({ sellToken, receiveToken: buyToken })
```
Velenir, line 22: Does `getMarket` return const refs for tokens? Maybe also memoize to have const deps in `useMemo`

Velenir (approved):

src/components/common/SmartPrice.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const [price, priceInverse] = useMemo((): string[] => {
+    const buyOrderPrice = calculatePrice({
+      numerator: { amount: priceFraction.numerator, decimals: buyToken.decimals },
+      denominator: { amount: priceFraction.denominator, decimals: sellToken.decimals },
+    })
+    const buyOrderPriceInverse = invertPrice(buyOrderPrice)
+    const sellTokenIsQuote = buyToken === quoteToken
```
W3stside, line 30: ? `sellTokenIsQuote` if `buyToken` = `quoteToken` doesn't make sense, should it be `const buyTokenIsQuote`

src/components/common/SmartPrice.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <span title={`${priceDisplayed} ${safeTokenName(quoteDisplayed)} per ${safeTokenName(baseDisplayed)}`}>
+      {priceDisplayed}
+      &nbsp;
+      <SwapPrice
+        baseToken={baseToken}
+        quoteToken={quoteToken}
+        isPriceInverted={isPriceInverted}
+        onSwapPrices={(): void => setIsPriceInverted(!isPriceInverted)}
```
W3stside, line 61:
```suggestion
        onSwapPrices={(): void => setIsPriceInverted((isPriceInverted) => !isPriceInverted)}
```

src/components/common/SmartPrice.tsx
W3stside: this is why I prefered `getMarket` as a hook, but can be easily combined now to make one using the `util` function and a hook

src/components/OrdersWidget/OrderRow.tsx
anxolin: d49d8beccde88e691bd2c8ae7ed7c92342f5ed50

src/components/common/SmartPrice.tsx
anxolin: hehe, yep, good catch. I guess I reversed the logic at some point and didn't update the var name

src/components/common/SmartPrice.tsx
anxolin: Why is this better? I mean, is it possible that `isPriceInverted` doesn't have the right value when `onSwapPrices` is called?

I guess, not in a normal case. Only way I can think off is when accumulate several calls to the function. In that case, is your option better? I mean, should we allow 2 executions that happen before a re-rendender?

src/components/common/SmartPrice.tsx
anxolin: Not sure I understand 100%

It’s not always clear to me what we expect to memoize and what not.
In this case, I thought about memoizing, but then I thought it would be a fast computation, so not sure if it would be worth it.
the function just return the original objects (tells u what is the base/quote) so, instances don’t change.
Where do you see the line in what to memoize and what not? Why a hook here is helpful? Is it to allow to reuse the util+memo logic in other parts of the app?

src/components/common/SmartPrice.tsx
W3stside: i just think it's safer. if you feel strongly opposed to the change then it probably doesn't matter

src/components/common/SmartPrice.tsx
W3stside: Yes you're right that in this case it's probably fine. "Expensive" calculation is a subjective term in some regards and i think this function is relatively simple enough.

src/components/common/SmartPrice.tsx
anxolin: nup, I'm ok with the change actually. I'm more curious than anything

PR #1494, opened 2020-10-06
[Smart Quote] Show buy sell fills

anxolin:
Part of https://github.com/gnosis/dex-react/issues/1476

So now that have a smarter way to decide what market is the user viewing, we can use it also for the trades.
Actually, now we could bring the concept of "Side" of the trade.

The side of the trade would be if we are buying or selling.

This PR:
- [x] Shows the market as a canonical 
- [x] Decide if we are selling or buying: decide the side
- [x] Add units to the price
- [x] Allow to invert the market
- [x] Memoize, the whole thing :) 

image

If we invert the market, the user effectively selects a different QUOTE, change the price value and it's units. Also, the TRADE is being seen as a BUY if it was a SELL before, and the other way around:
image

Future PRs could:
- Move "Side" to its own column. We could get inspiration in https://trade.dydx.exchange/ or our 2.0 designs. @W3stside could you help me with this one?
- Allow to invert the market (the whole row)
- We could consider not displaying 2 prices. Maybe we find a better
- Everything is more clear if we break Sold/Bought into 2 columns. And we called `Amount` and `Filled` for example. Check in other exchanges and our designs for v2.0

All pending tasks are defined in https://github.com/gnosis/dex-react/issues/1476

### DIFF
--- a/src/components/TradesWidget/TradeRow.tsx
+++ b/src/components/TradesWidget/TradeRow.tsx
@@ -78,6 +78,61 @@ const TypePill = styled.span<{
   text-transform: uppercase;
 `
 
+interface MarketAndPricesParams extends Pick<Trade, 'buyToken' | 'sellToken' | 'limitPrice' | 'fillPrice'> {
+  isPriceInverted: boolean
+}
+
+interface MarketAndPrices {
+  side: string
+  quoteTokenLabel: React.ReactNode
+  sellTokenLabel: React.ReactNode
+  buyTokenLabel: React.ReactNode
+  market: string
+  limitPrice: BigNumber | undefined
+  fillPrice: BigNumber
+}
+
+function getMarketAndPrices(params: MarketAndPricesParams): MarketAndPrices {
+  const { sellToken, buyToken, isPriceInverted, fillPrice, limitPrice } = params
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1494--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1494--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/TradesWidget/TradeRow.tsx
```diff
@@ -121,40 +176,57 @@ export const TradeRow: React.FC<TradeRowProps> = (params) => {
...
-      >
-        {market}
+      <td data-label="Market" className="showResponsive">
+        <SwapIcon swap={(): void => setIsPriceInverted(!isPriceInverted)} />{' '}
+        <span
+          onClick={(): void =>
+            onCellClick({
+              target: { value: market },
+            })
+          }
```
Velenir, line 214: This looks very artificial. I understand that the func expects a proper MouseEvent, but maybe it should be changed to accept a value.

Velenir (approved):

W3stside (approved):

PR #1498, opened 2020-10-07
Improve fetch tokens

anxolin:
As we commented in PR https://github.com/gnosis/dex-react/pull/1489, the logic for getting the list of tokens is too complex.

Dima, made a proposal in this comment:
https://github.com/gnosis/dex-react/pull/1489#discussion_r500157569

This PR is a reiteration in his proposal

* Simplifies logic, less loops
* Break the logic in more scoped and defined functions
* Removes some weird API signature, where we were returning a String instead of TokenDetails when there was not a valid ERC20

### DIFF
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -130,7 +130,11 @@ export function getTokensFactory(factoryParams: {
     return new Set(tcrApi && (await tcrApi.getTokens(networkId)))
   }
 
-  async function _getFilteredIdsMap(networkId: number, numTokens: number): Promise<Map<string, number>> {
+  async function _getFilteredIdsMap(
+    networkId: number,
+    numTokens: number,
+    tokensConfig: TokenDetails[],
+  ): Promise<Map<string, number>> {
     const [tcrAddressesSet, listedAddressesAndIds] = await Promise.all([
       // Fetch addresses from TCR
       retry(() => fetchTcrAddresses(networkId)),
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1498--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1498--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/services/factories/tokenList.ts
```diff
@@ -164,62 +166,60 @@ export function getTokensFactory(factoryParams: {
...
+    const tokensConfigMap = new Map(tokensConfig.map((t) => [t.address, t]))
+    const tokenDetailsPromises: (Promise<TokenDetails | undefined> | TokenDetails)[] = []
+    addressToIdMap.forEach((id, tokenAddress) => {
+      // Resolve the details using the config, otherwise fetch the token
+      const token: TokenDetails | undefined | Promise<TokenDetails | undefined> = tokensConfigMap.has(tokenAddress)
+        ? tokensConfigMap.get(tokenAddress)
+        : _fetchToken(networkId, id, tokenAddress)
+
+      if (token) {
+        tokenDetailsPromises.push(token)
```
Velenir, line 183: This works, but we may get slightly unexpected order of tokens
Before it was `localTokens + fetchedTokens`, now they will be interleaved. Before new tokens that get added to TCR would appear at the end of the list, now -- who knows.

Not a big problem though.

Velenir (approved):

src/services/factories/tokenList.ts
anxolin: Should we sort alphabetically?

W3stside (approved):

PR #1508, opened 2020-10-07
Modal (prev. modali) Storybook stories

W3stside:
Adds stories for Modal component


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1508--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1508--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/Modal.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+  ),
+}
+
+export const LoremOneButton = Template.bind({})
+LoremOneButton.args = {
+  options: {
+    ...LoremDefaultModalOptions,
+    buttons: [
+      <>&nbsp;</>,
+      <Modal.Button label="Close" key="yes" isStyleDefault onClick={(): void => console.log('Closing modal')} />,
```
Velenir, line 39: This is a good chance to use [actions](https://storybook.js.org/docs/react/essentials/actions) instead of console.logs
Though you may have to use them [like this](https://github.com/storybookjs/storybook/blob/master/addons/actions/ADVANCED.md#manually-specified-actions)

Velenir (approved):

src/components/common/Modal.stories.tsx
W3stside: nice idea

anxolin (commented):
Shouldn't there be an overlay when the modal is being shown?

W3stside:
> Shouldn't there be an overlay when the modal is being shown?

It comes from the `global` styles. All the variables are there - also the dark/light mode style switching is there so I'm not so sure about adding it to the Modal css

anxolin:
> It comes from the global styles. All the variables are there - also the dark/light mode style switching is there so I'm not so sure about adding it to the Modal css

My understanding was that we try to make the components self contained. Not dependant of globals. At least for some nice defaults.

anxolin (commented):
> It comes from the global styles. All the variables are there - also the dark/light mode style switching is there so I'm not so sure about adding it to the Modal css

My understanding was that we try to make the components self contained. Not dependant of globals. At least for some nice defaults.

W3stside:
> > It comes from the global styles. All the variables are there - also the dark/light mode style switching is there so I'm not so sure about adding it to the Modal css
> 
> My understanding was that we try to make the components self contained. Not dependant of globals. At least for some nice defaults.

it was to make the components as self contained *as possible* - I had the variables inside the modal in a previous commit but you asked me to move them out. There needs to be *some* global variables.. In this case, the overlay is there, it's just that the colour is defaulting to `white` as it's trying to pull a variable colour from global, which doesn't exist. 

Best solution, and as we already discussed, is to just add the bare minimum global styles (h1, anchor tag, native stuff + colour variables) and add that into our stories global

anxolin (approved):

PR #1509, opened 2020-10-08
[Price Impact]: Base

W3stside:
## Price Impact

**Base PR for all waterfalling price impact related PRs**

- [x] Confirmation Modal Warnings #1550 
- [x] Price impact percentage on trade page
- [x] PriceSuggestions price impact warnings #1544 
- [x] insert useBestAsk hook where appropriate and TODOs specifiy
- [x] Check errors/styles are correct


src/components/trade/PriceImpact.tsx
```diff
@@ -0,0 +1,100 @@
...
+  ])
+
+  if (!priceImpact) return null
+
+  return (
+    <PriceSuggestionsWrapper>
+      <div className="container">
+        <span>
+          <span>Price impact </span>
+          <BoldColourTag className={getImpactColourClass(priceImpact.toFixed(0))}>
```
Velenir, line 91: There's no need to cast `priceImpact` to number, `getImpactColourClass` can deal with BN

src/components/trade/PriceImpact.tsx
```diff
@@ -0,0 +1,100 @@
...
+    baseTokenId,
+    baseTokenDecimals,
+    quoteTokenId,
+    quoteTokenDecimals,
+  })
+
+  const priceImpact = React.useMemo(() => calculatePriceImpact({ bestAskPrice, limitPrice }), [
+    bestAskPrice,
+    limitPrice,
+  ])
```
Velenir, line 82: You can put `getImpactColourClass()` into this same memo

Velenir (approved):

Velenir:
[Build broke](https://travis-ci.com/github/gnosis/dex-react/builds/188707869#L718)
This has happened before and will happen again, unless...
Since linter doesn't catch type errors in the `pre-commit` hook, should we maybe add `pre-prush` hook with `build` 

Actually, @W3stside can you try something:

1. Run `yarn build`
2. Note the errors
3. Run `npx tsc --noEmit`

If errors are the same, I propose adding `pre-push` hook with `tsc --noEmit`

/cc @anxolin @alfetopito

src/components/trade/PriceImpact.tsx
W3stside: :+1:

src/components/trade/PriceImpact.tsx
W3stside: :+1:

W3stside:
> [Build broke](https://travis-ci.com/github/gnosis/dex-react/builds/188707869#L718)
> This has happened before and will happen again, unless...
> Since linter doesn't catch type errors in the `pre-commit` hook, should we maybe add `pre-push` hook with `build`
> 
> Actually, @W3stside can you try something:
> 
> 1. Run `yarn build`
> 2. Note the errors
> 3. Run `npx tsc --noEmit`
> 
> If errors are the same, I propose adding `pre-push` hook with `tsc --noEmit`
> 
> /cc @anxolin @alfetopito

how does `yarn lint` aka `eslint` not catch these?

W3stside:
@Velenir `npx tsc --noEmit` is WAY faster and catches the same errors, please make this PR so we can add asap!

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1509--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1509--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir:
> how does `yarn lint` aka `eslint` not catch these?

because they are types
```ts
const n: number = 'string'
```
won't trigger a linter error, they just do different things. You can explore what triggers linter [here](https://astexplorer.net/#/gist/688644b570396ff42446ea2acfe14c7c/16c2858d486593bde4ea227963a5ac5af5f71af1)

anxolin:
It show infinite:

image

What do you mean with `BELOW` why is it GREEN and negative?
Price impact, as I see it, should be always positive, and between 0 and 100%

anxolin (commented):
The tooltip content is right. We have the same as Uniswap, but in their case they compare the market price with the current price. 

We are comparing limit price with current price (honestly, I'm not 100% sure if it makes sense). Price impact officially would be to compare our best ask and market price. I would be concern if users get confused by price impact, I guess, we can try it out and ask for feedback once the demo is working

@W3stside @Rafanator I don't want to


Also: Tooltip is broken:
image

Rafanator:
I think we should leave the text as it is. A purist definition of `market price` would be a sort of averaged price amongst all exchanges. Given that we won't do that, I believe it is also fair to call `market price` the price of the `best ask`, as that would be the `market price` in our exchange. 

I would just do some small arrangements: 

`The difference between the market price and the estimated fill price due to order size` 

@anxolin Please let me know what you think.

W3stside:
> What do you mean with `BELOW` why is it GREEN and negative?
> Price impact, as I see it, should be always positive, and between 0 and 100%
> 
> Alsop, It show infinite:
> 
> image

fair enough, changes made

anxolin:
1. Why if the user doesn't have a price (therefor we cannot show the price impact), we still show a 0%?

image

2. Can we use also a nicer green. Like in the "price surplus" PR. I believe that green was used for light mode, but in dark mode is not good. Same for the red color in the price impact, it looks a bit purple-brownish. 

3. 
Regarding rafa's reply
> Given that we won't do that, I believe it is also fair to call market price the price of the best ask, as that would be the market price in our exchange.

@Rafanator  (and david), with my comment I don't mean this. My point is Uniswap calls Price Impact to the price the user gets, that is the same the user selects as his limit price. We use the limit price, not the "fill price". Not saying is wrong, but ther's a tiny part I don't like. My point is explained here https://gnosisinc.slack.com/archives/CPZA1AGKY/p1603127944112400

4. @W3stside , I think it would be good if you mimic exactly Uniswap behaviour, ranges for colors, colors and labels. For example, this is how they show tiny amounts.
image

image

src/components/trade/PriceImpact.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // PRICE_IMPACT = 100 - (BEST_ASK / LIMIT_PRICE * 100)
+  const impact = ONE_HUNDRED_BIG_NUMBER.minus(bestAskPrice.div(limitPriceBigNumber).times(ONE_HUNDRED_BIG_NUMBER))
+  return impact.lt(ZERO_BIG_NUMBER) ? ZERO_BIG_NUMBER : impact
+}
+
+function PriceImpact({ limitPrice, baseToken, quoteToken, networkId }: PriceImpactProps): React.ReactElement | null {
+  const { id: baseTokenId, decimals: baseTokenDecimals } = baseToken
+  const { id: quoteTokenId, decimals: quoteTokenDecimals } = quoteToken
+
+  // TODO: useBestAskPrice hook here
```
anxolin, line 80: Hehe, good, maybe make sure is in the check list

anxolin:
* I still have the same concern as the one expressed in my former comment (3). 
I think @Rafanator is looking into it

* I still have the same concern as the one expressed in my former comment (4). The label is fine now, but the colors...

* 5) Are we sure we want to hide the price impact if we have a price selected? I understand in this case, we can show it. Even if it's tiny or 0, it should show <0.01% imo
image

anxolin (commented):
* I still have the same concern as the one expressed in my former comment (3). 
I think @Rafanator is looking into it

* I still have the same concern as the one expressed in my former comment (4). The label is fine now, but the colors...

* 5) Are we sure we want to hide the price impact if we have a price selected? I understand in this case, we can show it. Even if it's tiny or 0, it should show <0.01% imo
image

src/components/trade/PriceImpact.tsx
Rafanator: @anxolin Regarding point `3`, I did a suggestion on [PR1544](https://github.com/gnosis/dex-react/pull/1544). Let me know what you think.

anxolin (approved):

PR #1512, opened 2020-10-08
delay first eth_getTransactionReceipt

Velenir:


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1512--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1512--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/api/wallet/composeProvider.ts
```diff
@@ -101,6 +101,27 @@ export const composeProvider = <T extends Provider>(
...
+  let lastTransaction = ''
+
+  engine.push(
+    createConditionalMiddleware<string[]>(
+      (req) => req.method === 'eth_getTransactionReceipt',
+      async (req) => {
+        const tx = req.params?.[0]
+
+        console.log('eth_getTransactionReceipt for tx:', tx, req)
+        if (tx && tx !== lastTransaction) {
```
anxolin, line 113: I don't understand, why `lastTransaction` is a kind off global var?

what happens with 
eth_getTransactionReceipt(tx1)
eth_getTransactionReceipt(tx2)
eth_getTransactionReceipt(tx1)

Velenir: Outer closure scope var.

So that we delay only the first call for a tx. When is a fresh, just sent tx. Calls after the first when `tx === lastTx` we don't delay.

anxolin: Dima, sorry, I still don’t understand how that works. Why that vat is out of the middleware

This is what I would understand by reading, I understand that sending 2 concurrent tx would do:

```js
eth_getTransactionReceipt(tx1)  -->  { lastTx = “”, tx=tx1   }    --->  delay
eth_getTransactionReceipt(tx1)  -->  { lastTx = “tx1", tx=tx1   }    --->  do not delay
eth_getTransactionReceipt(tx2)  -->  { lastTx = “tx1”, tx=tx2   }    --->  delay
eth_getTransactionReceipt(tx1)  -->  { lastTx = “tx2", tx=tx1   }    --->  delay
```

Maybe I'm missing what was the problem in the first place? 🤷‍♂️

Velenir: So I allowed myself a few assumptions:

1. call can break when made too fast on a newly sent tx
2. there won't be too many instances of concurrent txs
3. we can safely delay some calls
4. after successfully getting a receipt there won't be further calls for that one tx

so I expect sequential calls for one tx for the most part. And later for the next tx


`lastTx` var has to be in outer scope, because we are keeping state
Would you rather delay every call indiscriminately? (:unamused: )
Or keep a set of all txs so we know which are new? (:disappointed: )
Or implement a more involved logic where we track `eth_sendTransaction`return of tx hash and add it to `newTxs` set? (:thinking: )

I'm ok with doing the last one, it sounds interesting implementation-wise

PR #1513, opened 2020-10-08
Deterministic sorting

anxolin:
Solves the issue that the sorting is not deterministic.

Changing the implementation on how we get the tokens exposed that our order is not deterministic.
When we changed how we iterate and compose the final list of tokens, it moved our token orders.

* This PR sorts alphabetically the tokens
* It also handles a special case. The Native token / and Wrapper comes always first: WETH for mainnet and rinkeby, and wxDAI for xDAI
* WETH comes second in xDAI network
* To facilitate the sorting, it was created an artificial derived property (label), with the `safeTokenName`

image

NOTE:
* The property `label` was added, not only for sorting. I think that it's just nice to have already a not nullable displayable name for debugging and for the UI. We use this function over and over in many parts of the app. This PR doesn't do the refactor to take advantage of this new property. There will be a separate issue for that
* See https://github.com/gnosis/dex-react/issues/1514


W3stside:
[build is broken](https://travis-ci.com/github/gnosis/dex-react/builds/188850229#L1143)

src/services/factories/tokenList.ts
```diff
@@ -175,18 +176,56 @@ export function getTokensFactory(factoryParams: {
     const tokenDetailsPromises: (Promise<TokenDetails | undefined> | TokenDetails)[] = []
     addressToIdMap.forEach((id, tokenAddress) => {
       // Resolve the details using the config, otherwise fetch the token
-      const token: TokenDetails | undefined | Promise<TokenDetails | undefined> = tokensConfigMap.has(tokenAddress)
-        ? tokensConfigMap.get(tokenAddress)
+      const token: undefined | Promise<TokenDetails | undefined> = tokensConfigMap.has(tokenAddress)
```
W3stside, line 179: nit: can we in the future call unresolved vars `promised<varName>`? I find it easier to read.

W3stside (approved):

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1513--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1513--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/services/factories/tokenList.ts
```diff
@@ -175,18 +176,56 @@ export function getTokensFactory(factoryParams: {
...
         : _fetchToken(networkId, id, tokenAddress)
 
       if (token) {
+        // Add a label for convenience
+        token.then((token) => {
+          if (token) {
+            token.label = safeTokenName(token)
+          }
+          return token
+        })
```
Velenir, line 190: As we don't await this specific promise, it could be a race condition
It won't be a race condition in our case because this microtask is scheduled before `await Promise.all`, but still a bad approach.

src/services/factories/tokenList.ts
```diff
@@ -218,8 +258,20 @@ export function getTokensFactory(factoryParams: {
...
+    tokenListApi.persistTokens({ networkId, tokenList })
+  }
+
+  function _tokenComparer(a: TokenDetails, b: TokenDetails): number {
+    if (a.label < b.label) {
+      return -1
+    } else if (a.label > b.label) {
+      return 1
+    }
+    return 0
```
Velenir, line 274: considering what a lable may be (unicode string)and for simpicity, I recommend `String.localeCompare` (or `Intl.collator.compare` if you fancy that)

src/services/factories/tokenList.ts
```diff
@@ -175,18 +176,56 @@ export function getTokensFactory(factoryParams: {
...
+    const tokenIndex = tokenList.findIndex((t) => t.address === tokenAddress)
+    if (tokenIndex !== -1) {
+      const token = tokenList[tokenIndex]
+      tokenList.splice(tokenIndex, 1)
+
+      return [token, ...tokenList]
+    }
+
+    return tokenList
+  }
```
Velenir, line 209: Please, no :cry: 
This is like 3 loops more than necessary

src/services/factories/tokenList.ts
```diff
@@ -218,8 +258,20 @@ export function getTokensFactory(factoryParams: {
     // Get token details for each filtered token
     const tokenDetails = await _fetchTokenDetails(networkId, filteredAddressesAndIds, tokensConfig)
 
+    // Sort tokens
+    const tokenList = _sortTokens(networkId, tokenDetails)
```
Velenir, line 262: This is misleading, Sort is mutative, so we don't get a new `tokenList` out of that. One might be inclined to treat `tokenDetails` as unchanged after this line, though we don't do it.

Just writing
```js
_sortTokens(networkId, tokenDetails)
```
explicitly shows  that `tokenDetails` was mutated

src/services/factories/tokenList.ts
```diff
@@ -175,18 +176,56 @@ export function getTokensFactory(factoryParams: {
...
+
+    // Make sure wxDAI and WETH are the first tokens
+    switch (networkId) {
+      case Network.Mainnet:
+        return _moveTokenToHeadOfArray(WETH_ADDRESS_MAINNET, tokensSorted)
+      case Network.Rinkeby:
+        return _moveTokenToHeadOfArray(WETH_ADDRESS_RINKEBY, tokensSorted)
+      case Network.xDAI:
+        tokensSorted = _moveTokenToHeadOfArray(WETH_ADDRESS_XDAI, tokensSorted)
+        return _moveTokenToHeadOfArray(WXDAI_ADDRESS_XDAI, tokensSorted)
```
Velenir, line 223: Please, no no :sob:

src/services/factories/tokenList.ts
```diff
@@ -175,18 +176,56 @@ export function getTokensFactory(factoryParams: {
...
+      const token = tokenList[tokenIndex]
+      tokenList.splice(tokenIndex, 1)
+
+      return [token, ...tokenList]
+    }
+
+    return tokenList
+  }
+
+  function _sortTokens(networkId: number, tokens: TokenDetails[]): TokenDetails[] {
```
Velenir, line 211: Can do the same in one pass.

Velenir (commented):
Consider #1517

anxolin:
@Velenir we are closing a RC1 with this. I'm ok with the proposal of improvement for the sorting, however the PR https://github.com/gnosis/dex-react/pull/1517 don't have the right order

On monday, when we close 1.5 u can merge this one. I'll use this one for the RC1

PR #1515, opened 2020-10-09
pre-push types check

Velenir:
Check for type correctness before `git push`
Aims to avoid broken builds in travis

image


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1515--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1515--dexreact.review.gnosisdev.com/storybook/)**: Component stories

W3stside (approved):
NOICE

package.json
```diff
@@ -24,11 +24,13 @@
     "storybook": "BROWSER=NONE start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "docs": "start-storybook --docs",
-    "build-docs": "build-storybook --docs"
+    "build-docs": "build-storybook --docs",
+    "check-types": "tsc --noEmit"
```
anxolin, line 28: maybe we can call it  `build:check`?  for coherence with the other tsc tasks?

anxolin (approved):

package.json
Velenir: `build:check` implies a type of build -- `namespace::field`
it's not building anything, it can't build anything because only webpack can with our config.
It's similar to a linter in what it does,but nota linter

alfetopito (approved):
Like it! Will do this on safe CMM app as well :+1:

PR #1516, opened 2020-10-09
Account for pending orders being undefined

anxolin:
Model the pending transactions state differently, so we don't assume, we let the thing be undefined

We received a bug report of that part, maybe a race condition. Still revieinging, but for now, let's not make the assumption

### DIFF
--- a/src/reducers-actions/pendingOrders.ts
+++ b/src/reducers-actions/pendingOrders.ts
@@ -50,10 +50,12 @@ export const removePendingOrdersAction = (payload: {
   payload,
 })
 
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1516--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1516--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):

W3stside (approved):

PR #1519, opened 2020-10-09
Deprecate DAI

anxolin:
Explicitly forbid DAI trading

Not sure if this is good or bad thing. 
Theoretically, it's good to let the user use it if he wants, but this PR allows to show the DAI token, and inform the user that he better use wxDAI

Is likely an user doesn't know what's wxDAI, so maybe is a good thing
image

I leave it as a draft


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1519--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1519--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin:
We won't follow this approach. People should be able to trade DAI if they want

PR #1525, opened 2020-10-12
1486.5/fix wrong tokens orders logic update

W3stside:
Updates logic to avoid the flash of bad data

Still re-mounts data but at least doesn't show wrong numbers. Though I think the flag is probably a bad idea, it looks better. Maybe in the future we move everything to global state and use a flag `isDefault` list or something to let `useOrders` know not to render anything first time.

Thoughts @Velenir ?

### DIFF
--- a/src/api/tokenList/TokenListApiMock.ts
+++ b/src/api/tokenList/TokenListApiMock.ts
@@ -5,6 +5,8 @@ import GenericSubscriptions from './Subscriptions'
 export class TokenListApiMock extends GenericSubscriptions<TokenDetails[]> implements TokenList {
   private _tokenList: TokenDetails[]
 
+  public isListReady = false
+
   public constructor(tokenList: TokenDetails[]) {
     super()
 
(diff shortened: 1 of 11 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1525--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1525--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/api/tokenList/TokenListApiMock.ts
```diff
@@ -34,6 +36,10 @@ export class TokenListApiMock extends GenericSubscriptions<TokenDetails[]> imple
     this._tokenList = tokenList
     this.triggerSubscriptions(tokenList)
   }
+
+  public setListReady(state: boolean): void {
+    this.isListReady = state
+  }
```
Velenir, line 42: Should this not also `triggerSubscriptions`? I'm not sure `isListReady` will be picked up by `useOrders` hook immediately without.

src/hooks/useOrders.ts
```diff
@@ -121,7 +122,7 @@ export function useOrders(): Result {
       cancelled = true
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [offset, isLoading])
+  }, [offset, isLoading, isListReady])
```
Velenir, line 125: This is the main change, right?

Velenir (approved):

src/api/tokenList/TokenListApiMock.ts
W3stside: It is picked up in the beginning but maybe your right

PR #1529, opened 2020-10-13
Investigate orders load speed

W3stside:
Addresses several issues:
1. data load speed on xDAI (ordersWidget)
2. isListReady, necessary for not showing stale data, mixed with useEffect logic was blocking updates
3. types

### DIFF
--- a/src/services/factories/tokenList.ts
+++ b/src/services/factories/tokenList.ts
@@ -288,10 +288,6 @@ export function getTokensFactory(factoryParams: {
     areTokensUpdated.add(networkId)
 
     try {
-      // Set token list readiness to false
-      // and prevent stale data being presented in app
-      tokenListApi.setListReady(false)
-
       const numTokens = await retry(() => exchangeApi.getNumTokens(networkId))
       const tokens = tokenListApi.getTokens(networkId)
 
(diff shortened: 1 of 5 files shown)


src/services/factories/tokenList.ts
```diff
@@ -312,16 +308,20 @@ export function getTokensFactory(factoryParams: {
...
     }
   }
 
   return function (networkId: number): TokenDetails[] {
     if (!areTokensUpdated.has(networkId)) {
       logDebug(`[tokenListFactory][${networkId}] Will update tokens for network`)
+
+      // Set token list readiness to false
+      // and prevent stale data being presented in app
+      tokenListApi.setListReady(false)
```
Velenir, line 320: Should this not be awaited?

Velenir (approved):

src/services/factories/tokenList.ts
W3stside: why? its not async

Velenir: Wrong line, updateTokens function

W3stside: ah shit

PR #1539, opened 2020-10-19
Encapsulate CSS into Toast and remove Toast styles from global CSS file

W3stside:
Moves out global Toast styles into component using the only available method: extending via `Styled Components` and using `ToastContainer` rendered inside `App.tsx`

Not seeing any re-renderings of DOM tree due to toast renders so I think we're safe there (previous implementation rendered to `body` scoped `div` BUT this is impossible to extend with `Styled` components or style anywhere except or global styles)

See the [toastify code here](https://github.com/fkhadra/react-toastify/blob/439c60adc6ce8110d2840b3980cc66d1090d9adc/src/core/toast.tsx#L78)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1539--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1539--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin (approved):

PR #1543, opened 2020-10-20
App Error Handling: Incorrect Network

### LINKED ISSUE #283: The app breaks when you are not in the right network
We should inform the user that network is not supported

W3stside:
Closes #283 

Deals with incorrect network using CONFIG list's default accepted networks and uses `toast.error` pointing users to a our repo's wiki about networks.

**THE URL IN THE TOAST ERROR NEEDS TO POINT ELSEWHERE!! Right now it is pointing to THIS repo.**

Thoughts on toast approach?

### DIFF
--- a/src/hooks/useNetworkCheck.tsx
+++ b/src/hooks/useNetworkCheck.tsx
@@ -1,0 +1,82 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { ToastId } from 'react-toastify'
+import { toast } from 'toastify'
+
+import { useWalletConnection } from 'hooks/useWalletConnection'
+import useSafeState from 'hooks/useSafeState'
+
+import { Network } from 'types'
+
+const availableNetworks = CONFIG.exchangeContractConfig.config.map(({ networkId }) => networkId)
+
+const NetworkErrorMessage: React.FC = () => (
+  <div>
+    <div>Please switch to a supported network!</div>
+    <strong>{`${availableNetworks.map(
+      (id, idx, arr) => (idx === arr.length - 1 ? ' or ' : ' ') + Network[id],
+    )}`}</strong>
+    <br />
+    <br />
+    <div>
+      Click{' '}
+      <a href="https://github.com/gnosis/dex-react/wiki#-compatible-networks" target="_blank" rel="noreferrer">
+        here
+      </a>{' '}
+      for more information
+    </div>
+  </div>
+)
+
+function runCheckOnValidNetworks(networkId?: number): boolean {
+  if (!networkId) return true
+
+  return availableNetworks.includes(networkId)
+}
+
+const useNetworkCheck = (): void => {
+  const [currentNetworkId, setCurrentNetworkId] = useSafeState<Network | undefined>(undefined)
+  const [currentNetworkErrorId, setCurrentNetworkErrorId] = useSafeState<ToastId | undefined>(undefined)
+
+  const { networkId } = useWalletConnection()
+
+  const isValidOrNonNetwork = React.useMemo(() => runCheckOnValidNetworks(networkId), [networkId])
+
+  // User switched network, dismiss current error
+  React.useEffect(() => {
+    const isNewNetworkError = !isValidOrNonNetwork && currentNetworkId !== networkId
+
+    async function dealWithNewNetworkError(): Promise<void> {
+      if (isValidOrNonNetwork) {
+        if (currentNetworkErrorId) {
+          await toast.dismiss(currentNetworkErrorId)
+          setCurrentNetworkErrorId(undefined)
+        }
+      }
+
+      // Connected to a network but not one of the accepted ones
+      else if (isNewNetworkError) {
+        currentNetworkErrorId && (await toast.dismiss(currentNetworkErrorId))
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1543--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1543--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/hooks/useNetworkCheck.tsx
```diff
@@ -0,0 +1,82 @@
...
+      </a>{' '}
+      for more information
+    </div>
+  </div>
+)
+
+function runCheckOnValidNetworks(networkId?: number): boolean {
+  if (!networkId) return true
+
+  return availableNetworks.includes(networkId)
```
Velenir, line 34: better make a Set for `availableNetworks`, then won't even need `useMemo` as lookup is O(1)
Not that you need useMemo for `<short_array>.includes`

src/hooks/useNetworkCheck.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Network } from 'types'
+
+const availableNetworks = CONFIG.exchangeContractConfig.config.map(({ networkId }) => networkId)
+
+const NetworkErrorMessage: React.FC = () => (
+  <div>
+    <div>Please switch to a supported network!</div>
+    <strong>{`${availableNetworks.map(
+      (id, idx, arr) => (idx === arr.length - 1 ? ' or ' : ' ') + Network[id],
+    )}`}</strong>
```
Velenir, line 18: Here you are using the fact that `array.toString()` basically does `array.join(',')`
Shouldn't do it like that
Better use:
```ts
availableNetworks.map((id, idx, arr) => (idx === arr.length -1 ? 'or ' : '') + Network[id]).join(', ')
```

and move it outside the component, no need to recalc on render.

src/hooks/useNetworkCheck.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <br />
+    <br />
+    <div>
+      Click{' '}
+      <a href="https://github.com/gnosis/dex-react/wiki#-compatible-networks" target="_blank" rel="noreferrer">
```
Velenir, line 23:
```suggestion
      <a href="https://github.com/gnosis/dex-react/wiki#-compatible-networks" target="_blank" rel="noopener noreferrer">
```

src/hooks/useNetworkCheck.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    dealWithNewNetworkError()
+  }, [
+    currentNetworkErrorId,
+    currentNetworkId,
+    isValidOrNonNetwork,
+    networkId,
+    setCurrentNetworkErrorId,
+    setCurrentNetworkId,
+  ])
```
Velenir, line 79: This should behave the same, roughly. Test first,of course
```ts

React.useEffect(() => {
    const isValidOrNonNetwork = runCheckOnValidNetworks(networkId)

    if (isValidOrNonNetwork) return

    const promisedToastID = toast.error(<NetworkErrorMessage />, {
      autoClose: false,
    })

    return (): void => {
      promisedToastID.then((toastID) => toast.dismiss(toastID))
    }
  }, [networkId])
```

src/hooks/useNetworkCheck.tsx
W3stside: sure, can i ask why that's better though? genuinely curious

src/hooks/useNetworkCheck.tsx
W3stside: :thinking: looks waay cleaner your approach, will try it thx

src/hooks/useNetworkCheck.tsx
W3stside: true true, not sure why i didnt think of that! :+1:

src/hooks/useNetworkCheck.tsx
W3stside: works fucking beautifully, i could kiss you

src/hooks/useNetworkCheck.tsx
W3stside: (not the moving out, the .join)

Velenir: Specific `toString` behaviors are implementation details, that are generally not a good thing to depend on. They are there to make some data structures human readable in case of logging out using in string concatenation.
You wouldn't really on an object producing a nice string, you'll get `[object Object]` or specific class/constructor name. Functions will produce the whole function internals as a string. So I would not really on arrays too.
Nope there's nothing inherently wrong with stringifying arrays, just the conversation is very implicit, and can be overridden at times.

Velenir (approved):

anxolin (commented):
I see this "minimalistic web" 😅 if I open the App:
image

Velenir:
> I see this "minimalistic web"  if I open the App:

wow, wasn't the case yesterday
It's not just blank page,it looks like react didn't render anything, `div#root` is empty. As if JS is disabled

W3stside:
weird PR behaviour concerning PRaul so reopening in another PR...

PR #1544, opened 2020-10-21
[Price Impact] Warnings in PriceSuggestions

W3stside:
Part of #1509 

Adds warnings as outlined in [this doc](https://docs.google.com/document/d/13U6B9rXiqyxJ_ELExxmfrJDhlMCvA-BEZFLLKhzJKkc/edit#) when limit price is set to certain amount above/below Fill and/or Best Ask prices

Example:
Screenshot from 2020-10-21 17-28-49


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1544--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1544--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    // CASE 5: Limit price LESS THAN Fill price AND Best Ask price
+    if (isLimitPriceLessThanFillAndBestAsk) {
+      warning = 'Your order might only be filled when the market price reaches your limit price'
+    } else {
+      // CASE 3 & 4: Limit price LESS THAN Fill price but EQUAL TO or GREATER than Best Ask price
+      warning = 'Your order might only be partially filled'
+    }
+  } else {
+    // CASE 6: Limit price is GREATER THAN Fill price (and Best Ask)
+    warning = 'Deep in the market: Your order might be fully executed at this price.'
```
anxolin, line 57: image


I wouldn't show this message, right away, right @Rafanator 
u can set a bit more of price threshold.

What % makes sense here?

Rafanator:
I have a suggestion on how we can make the `Deep in the market` warning less shocking. 

**Uniswap does the following** 

1. If `Price Impact` < 1%, show `Price Impact` with Green (pictured below:
Screenshot 2020-10-22 at 11 59 34

2. If  3% > `Price Impact` ≥ 1, show `Pice Impact` with normal white color: 
Screenshot 2020-10-22 at 11 59 56

3. If `Price Impact` ≥ 5%, show `Price Impact` in red:
Screenshot 2020-10-22 at 12 05 22

**What we should do**

I would say that we do the same thing when it comes to colours:

- `Price impact < 1%`: Use green, include text saying `Your order might be fully executed at this price.`

- `Price impact ≥ 1%`: Use white (or standard font color), include same text `Your order might be fully executed at this price.`

- `Price Impact ≥ 3%`: Use orange, include text saying `Your order might be fully executed at this price.`

- `Price Impact ≥ 5%`: Use red, include text saying `⚠️Deep in the market: Your order might be fully executed at this price.`

@anxolin I think this should reduce the excessive warning that you had suggested before.

W3stside:
@Rafanator but the price impact can still be huge even when selecting the fill price. In this case should we still show the price impact, even though the user selected the suggested price AND show the Deep in the Market label if the impact is >= 5? Seems strange to me

Screenshot from 2020-10-22 21-55-33

anxolin:
@Rafanator I agree with your colors comment, although that's for this other PR https://github.com/gnosis/dex-react/pull/1509
in fact, I did exactly the same comment (less detailed), it's my point (4)

Also, please, note that the price impact has no relationship with the warning "Your order might be fully executed at this price."
* Price impact: Relationship between the user's "Limit Price" and the market's "Best ask" --> in our app, not sure how standard is this
* The warnings: Compares the user's "Limit Price" to the "Fill Price". If you price is too cheap, then we warn "Deep in the market: Your order might be fully executed at this price." if the price is to expensive, we warn "Your order might only be filled when the market price reaches your limit price". 

My issue is with "Deep in the market: Your order might be fully executed at this price.", it's show to quick imo. For me, makes sense the user being able to add a bit of slippage, in case of the price moving while the trade is not happening.


@W3stside , regarding

>  but the price impact can still be huge even when selecting the fill price.
Yes, that's the whole point. The price impact will tell you how does your price compare with the best ask (that is the closer thing we have to the spot price). If your order drops the price 10% in order to get fill, in other words, if to fulfill the order we need to be 10% cheaper than the best offer in the orderbook, then the price impact for that price is in fact 10%.

src/components/trade/PriceImpact/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  limitPrice,
+  fillPrice,
+  baseToken,
+  quoteToken,
+  networkId,
+}: PriceImpactProps): React.ReactElement | null {
+  const { id: baseTokenId } = baseToken
+  const { id: quoteTokenId } = quoteToken
+
+  // TODO: useBestAskPrice hook here
```
anxolin, line 49: No you do, right? do we need this TODO?

src/components/trade/PriceImpact/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  })
+
+  const { priceImpactSmart, className, priceWarning } = React.useMemo(() => {
+    const priceImpact = calculatePriceImpact({ bestAskPrice, limitPrice })
+    // Smart format, if possible
+    const priceImpactSmart =
+      (priceImpact &&
+        !priceImpact.isZero() &&
+        formatSmart({ amount: parseAmount(priceImpact.toString(), 4) as BN, precision: 4, smallLimit: '0.01' })) ||
+      '<0.01'
```
anxolin, line 63: I personally don't love this long complex lines. I feel that, `AA && BB || CC` pattern are meant for simpler expressions. Also, good to add line breaks when you make calls and u have IF/ELSE logics

One suggestion is:
```suggestion
    let priceImpactSmart
    if (priceImpact && !priceImpact.isZero()) {
      priceImpactSmart = formatSmart({
        amount: parseAmount(priceImpact.toString(), 4) as BN,
        precision: 4,
        smallLimit: '0.01',
      })
    } else {
      priceImpactSmart = '<0.01'
    }
```

Or even, it's possible to take my suggestion and move it (with the calculatePriceImpact too) to it's own function, so you also simplify the hook, maybe even the whole hook content

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
+import BigNumber from 'bignumber.js'
+
+import { parseBigNumber } from 'utils'
+import { ONE_HUNDRED_BIG_NUMBER, ONE_BIG_NUMBER, ZERO_BIG_NUMBER } from 'const'
+
+import { PriceImpactArgs, PriceImpactArgsBase } from '../PriceImpact'
+
+// Limit as to where price colour changes
+const HIGH_PRICE_IMPACT_THRESHOLD = new BigNumber(5)
+const MID_PRICE_IMPACT_THRESHOLD = new BigNumber(3)
```
anxolin, line 10: uber: A comment to the right helps a bit to know is a percentage, and no units for example:
`const HIGH_PRICE_IMPACT_THRESHOLD = new BigNumber(5) // 5%`

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+const LOW_PRICE_IMPACT_THRESHOLD = ONE_BIG_NUMBER
+// TODO: remove for slippage toggle
+// 0.5% placeholder slippage
+const PLACEHOLDER_SLIPPAGE = new BigNumber(0.005)
+
+const determineImpactLevel = (impact: BigNumber | null): number => {
```
anxolin, line 16: for impact level, u defined some magic numbers. I think the function should return a enum

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  return impactLevel
+}
+
+// sets colour of strong tag presenting price impact
+const getImpactColourClass = (impact: BigNumber | null): string => {
+  const impactLevel = determineImpactLevel(impact)
+
+  let className = ''
+  if (impactLevel) {
```
anxolin, line 36: why not a switch?

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { PriceImpactArgs, PriceImpactArgsBase } from '../PriceImpact'
+
+// Limit as to where price colour changes
+const HIGH_PRICE_IMPACT_THRESHOLD = new BigNumber(5)
+const MID_PRICE_IMPACT_THRESHOLD = new BigNumber(3)
+const LOW_PRICE_IMPACT_THRESHOLD = ONE_BIG_NUMBER
+// TODO: remove for slippage toggle
+// 0.5% placeholder slippage
+const PLACEHOLDER_SLIPPAGE = new BigNumber(0.005)
```
anxolin, line 14: The deep in the market percentage should be bigger imo. 
Maybe ask rafa, but at least 5%?

0.5% or 1% could be for the lighter version of the warning

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  return className
+}
+
+function determinePriceWarning(params: PriceImpactArgs, impact: BigNumber | null): string | null {
+  const { limitPrice, fillPrice, bestAskPrice } = params
+
+  const limitPriceBigNumber = limitPrice && parseBigNumber(limitPrice)
+  const fillPriceRounded = fillPrice?.toFixed(5)
+
+  // CASE 1 & 2: no warning
```
anxolin, line 55: I don't think we should refer to the CASES in the document. It would be good if the comment is self explanatory without another document. I feel for some complex cases can be justified, but here is not the case.

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  // round fillPrice to 5 decimals points and string compare against limit
+  if (
```
anxolin, line 57: if you fund yourself doing too many conditions in a IF, some with negations, etc.
I think is better to define some previous vars 

`const noLimitPrice = !limitPriceBigNumber || limitPriceBigNumber.isZero()`

this way, it's easier to understand every statement

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+    !limitPriceBigNumber ||
+    limitPriceBigNumber.isZero() ||
+    !fillPrice ||
+    !bestAskPrice ||
+    fillPriceRounded === limitPrice
+  )
+    return null
+
+  // Is user's limit price less than the suggested fill price?
+  const isLimitPriceLessThanFillPrice = limitPriceBigNumber.lt(fillPrice)
```
anxolin, line 67: Some declarative way to name this `isLimitPriceLessThanFillPrice` is `orderWontBeFullyExecuted`

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Is user's limit price less than suggested fill price AND best ask?
+  // One could assume that if the limit price is lower than best ask then it's lower than fill
+  // but just to be safe
+  const isLimitPriceLessThanFillAndBestAsk = isLimitPriceLessThanFillPrice && limitPriceBigNumber.lt(bestAskPrice)
```
anxolin, line 71: Some declarative way to name this isLimitPriceLessThanFillAndBestAsk is orderWontBeExecuted

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  let warning: string
+  if (isLimitPriceLessThanFillPrice) {
+    // CASE 5: Limit price LESS THAN Fill price AND Best Ask price
+    if (isLimitPriceLessThanFillAndBestAsk) {
+      warning = '⚠️ Your order might only be filled when the market price reaches your limit price'
```
anxolin, line 77: You don't need 2 nested IFs `isLimitPriceLessThanFillAndBestAsk` is already a subcase of `isLimitPriceLessThanFillPrice`

src/utils/miscellaneous.ts
```diff
@@ -208,3 +208,5 @@ export const RequireContextMock: __WebpackModuleApi.RequireContext = Object.assi
 export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
   return value !== null && value !== undefined
 }
+
+export const isStrNumTruthy = (value?: string | number): boolean => !!value && !!+value
```
anxolin, line 212: I had to scratch my head to understand it by name

what about sth like`isNonZeroNumber`?

anxolin (commented):
Mostly nitpicks.

Other than the comments included, what we discussed about adding a buffer so we don't show alerts in case the user overpays just a little (have sth like 0.5% tolerance)

src/components/trade/PriceImpact/index.tsx
Velenir: is `useBestAskPrice` and `useBestAsk` the same?

src/components/trade/PriceImpact/index.tsx
Velenir: I kinda want to adopt [`do` ](https://babeljs.io/docs/en/babel-plugin-proposal-do-expressions) expressions for cases like this
But only kinda

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,107 @@
...
+  let impactLevel = 0
+  if (impact) {
+    if (impact.gte(HIGH_PRICE_IMPACT_THRESHOLD)) {
+      impactLevel = 3
+    } else if (impact.gte(MID_PRICE_IMPACT_THRESHOLD)) {
+      impactLevel = 2
+    } else if (impact.lt(LOW_PRICE_IMPACT_THRESHOLD)) {
+      impactLevel = 1
+    }
+  }
```
Velenir, line 26: To echo @anxolin here: Why not a `switch(true)`

src/components/trade/PriceImpact/index.tsx
W3stside: yeah theyre the same

W3stside: and yeah it's already in, will remove comment

src/components/trade/PriceImpact/index.tsx
W3stside: made it an entire hook in another PR, good advice thanks

src/components/trade/PriceImpact/utils.ts
W3stside: Sorry I don't really get what you mean here

src/components/trade/PriceImpact/utils.ts
W3stside: Normally I would do that, but for some reason TS was giving me hell here:

Screenshot from 2020-10-26 14-35-52

src/components/trade/PriceImpact/utils.ts
W3stside: I will change this but imo this is totally subjective - I understand more clearly my variable name

src/components/trade/PriceImpact/utils.ts
W3stside: :+1:

src/components/trade/PriceImpact/utils.ts
W3stside: I think i handled this in the new logic so will resolve, take a look at fixes please

src/components/trade/PriceImpact/utils.ts
```diff
@@ -64,35 +71,38 @@ function determinePriceWarning(params: PriceImpactArgs, impact: BigNumber | null
...
+
+  const isPriceMidImpactAndExecutable =
+    impactLevel >= ImpactLevel.LOW &&
+    !orderWontBeFullyExecuted &&
+    limitPriceBigNumber.gte(fillPrice.times(MID_PLACEHOLDER_SLIPPAGE))
+
+  switch (true) {
+    // CASE 6: Limit price is GREATER THAN Fill price AND upper threshold 5%
+    case isPriceAboveDeepMarketThreshold:
+      return '⚠️ Deep in the market: Your order might be fully executed at this price.'
```
Velenir, line 93: do these `⚠️`  get displayed in one of our included fonts or in a system font?
If it's a system font, not every user would necessarily have them
Probably ok, though

W3stside: ah i thought it would always be displayed but just not necessarily in our font, will it be broken on some systems?

Velenir: Font needs to support it. Or will display an empty block
But it may already be alright, if this text is displayed with the fonts we use
Just need to check

Velenir: I even have it displayed differently in Chrome and Firefox, ugh
Check diff browsers

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+// Limit as to where price colour changes
+const HIGH_PRICE_IMPACT_THRESHOLD = new BigNumber(5) // 5%
+const MID_PRICE_IMPACT_THRESHOLD = new BigNumber(3) // 3%
+const LOW_PRICE_IMPACT_THRESHOLD = ONE_BIG_NUMBER // 1%
+// TODO: remove for slippage toggle
+// 0.5% placeholder slippage
+const HIGH_PLACEHOLDER_SLIPPAGE = new BigNumber(1.05) // 5%
+const MID_PLACEHOLDER_SLIPPAGE = new BigNumber(1.03) // 3%
+
+enum ImpactLevel {
```
anxolin, line 17: nice! Although, a nitipick is that I think u don't need to specify the numbers, since we work only with the enum, right. They are self assigned

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+  HIGH = 3,
+  MID = 2,
+  LOW = 1,
+  NONE = 0,
+}
+
+const determineImpactLevel = (impact: BigNumber | null): ImpactLevel => {
+  const impactLevel = ImpactLevel.NONE
+
+  switch (true) {
```
anxolin, line 27: nice @W3stside and @Velenir actually I totally forgot about this switch(true) thing, my proposal was not good. Glad u shaped it into this

src/components/trade/PriceImpact/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+  const impactLevel = determineImpactLevel(impact)
+
+  switch (impactLevel) {
+    case ImpactLevel.HIGH:
+      return 'highImpact'
+    case ImpactLevel.MID:
+      return 'midImpact'
+    case ImpactLevel.LOW:
+      return 'lowImpact'
+    default:
```
anxolin, line 50: I prefer in this case no DEFAULT, and be exhaustive with all enum values

typescript is smart enough to know u forgot about one if you don't add the DEFAULT, however in your code, you won't be notified if you forgot about one enum value.

anxolin (approved):
I think now you nailed it

As a nitpick, for cases where the impact is negative (our price is super good for us, but it won't execute cause there's no offers at that price) --> then saying the price impact is `<0.01` is misleading. We don't even have a price impact

src/components/trade/PriceImpact/utils.ts
Velenir: Firefox:
Screen Shot 2020-10-27 at 11 46 59

Chrome:
pr1544--dexreact review gnosisdev com_trade_USDC-DAI__sell=12 price=0 93 from= expires=

Velenir: The reason for the difference is very, very interesting:

If you copy that warning symbol and inspect it you'll have:
```js
'⚠️'.length
2
'⚠️'.charAt(1)
"️"
'⚠️'.charAt(0)
"⚠"
'⚠️'.charCodeAt(0)
9888
'⚠️'.charCodeAt(1)
65039
```
First char is the [warning symbol](https://unicode-table.com/en/26A0/). Second -- [ Variation Selector-16](https://unicode-table.com/en/FE0F/) that says _display the symbol before me as emoji_

But Chrome ignores it because of [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1033453)
That's why Firefox uses an Emoji font but Chrome does, Chrome doesn't recognize that there's an emoji

As you can't force emoji in Chrome, to have the same cross-browser experience you'll need to enforce text rendering of non emoji: `⚠` (believe you me this particular one is `length = 1`) or better yet more explicitly `&#x26A0;&#xFE0E;` (html entities), where `&#xFE0E;` is [Variation Selector-15](&#xFE0E;) to enforce non-emoji display.


Further reading [Unicode Variation Selector](https://blog.tomayac.com/2019/12/12/same-same-but-different-unicode-variation-selector-16/) , [How to prevent Unicode characters from rendering as emoji in HTML from JavaScript?](https://code-examples.net/en/q/1f6401d)

Velenir: Or just use font-awesome icons

src/components/trade/PriceImpact/utils.ts
W3stside: done in #1550

src/components/trade/PriceImpact/utils.ts
W3stside: done in #1550

PR #1554, opened 2020-10-28
[Smart Quote] Fix broken **limitPrice** amount

W3stside:
`usePriceEstimationWithSlippage` returns price in `quote` token (as it should) but the app wasn't inverting the price when `smart-quote` logic was running thus showing incorrect prices in FE

This fixes that.

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -674,14 +674,23 @@ const TradeWidget: React.FC<TradeWidgetProps> = ({
             isPriceInverted={isPriceInverted}
           />
           <PriceSuggestions
+            // Network
             networkId={networkIdOrDefault}
+            // Tokens
+            // Canonical market
             baseToken={baseToken}
             quoteToken={quoteToken}
+            // Keep original selling pair
+            receiveToken={receiveToken}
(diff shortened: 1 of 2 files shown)


Velenir (approved):

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1554--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1554--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/TradeWidget/index.tsx
```diff
@@ -674,14 +674,23 @@ const TradeWidget: React.FC<TradeWidgetProps> = ({
             isPriceInverted={isPriceInverted}
           />
           <PriceSuggestions
+            // Network
```
anxolin, line 677: I like the comments. When they have to many components, if there's not another solution like break the component or simplify it signature, it's nice to at least add these comments 👌

anxolin (commented):
Why there's no price impact here??
image

anxolin (approved):
Approving since as @W3stside commented, this branch doesn't have the price impact yet

PR #1561, opened 2020-10-30
1511 - Show WETH/WXDAI >> Balances when filtering zero balance

### LINKED ISSUE #1511: Issue with filtering non-zero balances
WETH should show if we have balance. In fact, and maybe easier to solve, WETH and wxDAI should show always

image

image

W3stside:
Closes #1511 

Always show WETH/WXDAI when filtering non-zero balances

Doesn't apply to text search however


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1561--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1561--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):

anxolin (commented):
I think the idea was to not hide Ether if it has some balance

Also, it feels a bit complex solution.
That widget has too much logic around the filters, and some hooks that although they are related to each other, from the component perspective, we only need to know `displayedBalances`

I wonder if it's better to move all the complex hooks into a single hook, with a simpler interface

W3stside:
> I think the idea was to not hide Ether if it has some balance
> 
> Also, it feels a bit complex solution.
> That widget has too much logic around the filters, and some hooks that although they are related to each other, from the component perspective, we only need to know `displayedBalances`
> 
> I wonder if it's better to move all the complex hooks into a single hook, with a simpler interface

I don't disagree, but honestly, I think it's fine as is. Since we want to move on from this and have tons of work on OBA wdyt? @Velenir

anxolin:
I agree is not a prio this issue, although also I think we shouldn't merge a rushed code. It makes more complex a component that is already too complex. The solution is not too complication and leaves the code in a better state. If we feel other things are more important and this takes time, I even prefer to handle it at a later point thant increment our tech debt.

Reason is, we'll likely need to come back to this code base. Plus, we'll likely have to keep working slightly in this project too

anxolin:
💥nuke this PR

PR #1580, opened 2020-11-03
[Price Impact] - Fix price calculation and price warnings

W3stside:
Fixes price calculation and fixes warnings from small calc errors

### DIFF
--- a/src/components/TradeWidget/index.tsx
+++ b/src/components/TradeWidget/index.tsx
@@ -301,10 +301,10 @@ const TradeWidget: React.FC<TradeWidgetProps> = ({
   useEffect(() => {
     if (priceValue && sellValue) {
       // If price is quoted in sell tokens, we use it, otherwise we use the inverse
-      const priceUsedForReceiveAmount = sellToken === quoteToken ? priceValue : priceInverseValue
+      const priceUsedForReceiveAmount = sellToken.address === quoteToken.address ? priceValue : priceInverseValue
       setValue(receiveInputId, calculateReceiveAmount(priceUsedForReceiveAmount, sellValue))
     }
-  }, [isPriceInverted, quoteToken, sellToken, priceValue, priceInverseValue, setValue, sellValue])
+  }, [isPriceInverted, quoteToken.address, sellToken.address, priceValue, priceInverseValue, setValue, sellValue])
 
   const url = buildUrl({
     sell: sellValue,
--- a/src/components/TradeWidget/utils.ts
+++ b/src/components/TradeWidget/utils.ts
@@ -17,9 +17,7 @@ export function calculateReceiveAmount(priceValue: string, sellValue: string): s
 
     if (sellAmount && price) {
       const receiveBigNumber = sellAmount.dividedBy(price)
-      // Receive at least input value
-      // uses amount floor < 0.0001
-      // but Limit Price set shows full amount
+      // Format the "Receive at least" input amount same as PriceSuggestions price
       receiveAmount =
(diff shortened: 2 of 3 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1580--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1580--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/TradeWidget/index.tsx
```diff
@@ -301,10 +301,10 @@ const TradeWidget: React.FC<TradeWidgetProps> = ({
   useEffect(() => {
     if (priceValue && sellValue) {
       // If price is quoted in sell tokens, we use it, otherwise we use the inverse
-      const priceUsedForReceiveAmount = sellToken === quoteToken ? priceValue : priceInverseValue
+      const priceUsedForReceiveAmount = sellToken.address === quoteToken.address ? priceValue : priceInverseValue
```
anxolin, line 304: is this related to the issue? was the problem related to "identity"?

src/components/TradeWidget/utils.ts
```diff
@@ -16,7 +17,11 @@ export function calculateReceiveAmount(priceValue: string, sellValue: string): s
 
     if (sellAmount && price) {
       const receiveBigNumber = sellAmount.dividedBy(price)
-      receiveAmount = receiveBigNumber.isNaN() || !receiveBigNumber.isFinite() ? '0' : receiveBigNumber.toString(10)
+      // Receive at least input value
+      // uses amount floor < 0.0001
+      // but Limit Price set shows full amount
```
anxolin, line 22: I get the change below, but I didn't understand the comment

src/components/trade/PriceSuggestions/PriceSuggestionItem.tsx
```diff
@@ -21,54 +21,67 @@ export interface Props {
...
   priceFormatted: string
-  invertedPriceFormatted: string
+  inversePriceFormatted: string
+  inversePriceRaw: string
+  priceRaw: string
 }
 
-function formatPriceToPrecision(price: BigNumber): string {
-  return price.toFixed(PRICE_ESTIMATION_PRECISION)
+const LOW_PRICE_FLOOR = new BigNumber(0.0001)
```
anxolin, line 29: I think it's better to use string in the constructor, so there's never float point issues

src/components/trade/PriceSuggestions/PriceSuggestionItem.tsx
```diff
@@ -21,54 +21,67 @@ export interface Props {
 
 interface FormattedPrices {
   priceFormatted: string
-  invertedPriceFormatted: string
+  inversePriceFormatted: string
+  inversePriceRaw: string
+  priceRaw: string
```
anxolin, line 26: good call

anxolin: good call

I would expect raw being a BigNumber (judging by the name)
Since all these are "formatted" (all are strings), I would suggest:
- priceLabel
- inversePriceLabel
- priceValue
- inversePriceValue

I think, then the `<button>` is more clear

anxolin (approved):

src/components/TradeWidget/index.tsx
W3stside: I made this change as I was seeing a lot of re-renders in this hook, not sure if the references between the objects `sellToken` and `quoteToken` is stable. anyways can remove

src/components/TradeWidget/utils.ts
W3stside: changed it

PR #1585, opened 2020-11-05
Test integration test

anxolin:
Adds travis integration tests, only in `develop`, `master` and `tagged versions`


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1585--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1585--dexreact.review.gnosisdev.com/storybook/)**: Component stories

.travis.yml
```diff
@@ -4,10 +4,17 @@ node_js:
   # TODO: stable (v13.0.1 at the time) breaks sha3 package compilation
   # Sticky v12
   - '12'
-cache: 'yarn'
+cache:
+  yarn: true
+  directories:
+    - ~/.cache
```
Velenir, line 10: You sure we need this,`.cache` in home folder?
I've seen [`$HOME/.cache`](https://docs.travis-ci.com/user/caching/#yarn-cache) somewhere before
Also this cache setup looks like a blanket cache of everything. Did you see it recommended somewhere?

anxolin: https://docs.cypress.io/guides/getting-started/installing-cypress.html#Binary-cache

I was considering consolidating all the caches in this folder if this doesn't add any troubles (like not existing intermediate dir `.cache/`)

anxolin: Also https://docs.cypress.io/guides/guides/continuous-integration.html#Example-travis-yml-config-file

Velenir: On, you convinced me

Velenir (approved):

.travis.yml
```diff
@@ -10,9 +10,14 @@ cache:
...
     - .eslint-cache
     - .jest-cache
+    - ~/.cache
 script:
   - yarn test-coverage
   - yarn build
+  # Integration test, only in tags, master and develop
+  - if [[ $TRAVIS_TAG != "" || $TRAVIS_BRANCH = master || $TRAVIS_BRANCH = develop ]]; then
+      yarn test:integration;
+    fi
```
alfetopito, line 20: Why not on PRs?

I would expect to have the integration tests to run on every PR change, as they are supposed to assure no changes are breaking the code.

It's true that this stops a failed change to be deployed into develop/master/tags, but doesn't prevent a breaking change from being merging into those branches.

PR #1589, opened 2020-11-05
Remove push hook

anxolin:
Now commiting is again super fast!
The problem is pushing is SLOW. I personally prefer to let travis fail, we already work in branches, and it saves so much time.

The linter already caches most of the issues, all the other should be catched in the "build" of travis.

I wouldn't add `yarn check-types` at all, I did it here in this PR, just for experimenting. We can leave it if our concern is we want to FAIL asap. But, I think it's better to leave that out

EDIT: Following @Velenir tip, I'll try with https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ storybook-static
 .jest-cache
 .eslint-cache
 .stylelint-cache
+tsconfig.tsbuildinfo
 
 cypress/videos
 cypress/screenshots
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1589--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1589--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):

PR #1590, opened 2020-11-05
Cache coverage tests

anxolin:
Adds coverage cache

I was unsure if it was safe to cache the coverage, but it looks like it works!
Same report, and the time goes does from 38s to 8s in my machine. Travis will be happy about this.

Try to run:

```bash
yarn test:coverage

# now again
yarn test:coverage

# now change any test

# now test again
yarn test:coverage
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
     - .eslint-cache
     - .jest-cache
 script:
-  - yarn test-coverage
+  - yarn test:coverage
   - yarn build
 before_install:
   # Needed to deploy pull request and releases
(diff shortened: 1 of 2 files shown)


Velenir:
I have a concern: you using same folder for caching tests with and without coverage. I'm not so certain the caches end up the same.
Can you compare check assume of folders. Mauve better to use different folders

Velenir:
Though it's probably ok. I imagine coverage would at most as to cache, not override it

Velenir (approved):

anxolin:
> I have a concern: you using same folder for caching tests with and without coverage. I'm not so certain the caches end up the same.
> Can you compare check assume of folders. Mauve better to use different folders

It's a fair concern, but I've tried it, and it just works like a charm

PR #1593, opened 2020-11-06
Add about test

anxolin:
Just a small test on the about page

For some reason I'm getting sometimes a timeout, like if the "about" link is not visible for some reason (although I see it)

I suspect is a re-rendering or sth..

EDIT: Super annoying issue, at the end it was a scroll issue. Cypress was complaining because the user was not seeing the link (it was not because it was not visible in the DOM, they were referring to the viewport)
For that you need to scroll to the bottom. The problem is that cypress don't work that well with `scroll: smooth` as explained here https://github.com/cypress-io/cypress/issues/3200#issuecomment-672788861 

Now the test should succeed and be faster too.

This PR:
* Solves the issue witht the scroll in about page
* For that, it introduces another way of navigating: `navigateApp` instead of `navigate` (command, in cypress terminology)
* Adds some convenient scripts in the package.json 

## Note for dima
Regarding your comment in #1594 regarding why eslint was added, removed and added back.
* I thought I needed, because commands are in JS, and they need a global "Cypress" object that is magically injected. 
* Then I thought I didn't need it, since I was already using typescript, so I said, I'll do the commands in Typescript so I can get rid of the globals too
* It turned out, they don't like that, it didn't work so I had to go back to JS + definitions file, so I need globals again

## test

```bash
yarn start
yarn cypress
```

then click on the test


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1593--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1593--dexreact.review.gnosisdev.com/storybook/)**: Component stories

anxolin:
Merging and deciding later what to do, if we just kill cypress and live playwrite, or not. Either way, the test passes and applies the recomended solution to the scrollgy thingy :)

PR #1600, opened 2020-11-10
Rename old layout to Legacy

anxolin:
This PR, although has a lot of changes, **it only renames the `SwapLayout` into `LegacyLayout`**

It makes it obvious also, by looking at the modified files that the layout is not generic enough. For example, the card component shouldn't be having to do anything to the layout, however, we can see that it was declared inside it.

In future PRs, we could take individual components we need for v2 and make them more independent


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1600--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1600--dexreact.review.gnosisdev.com/storybook/)**: Component stories

biocom (approved):

Velenir (approved):

W3stside (approved):

PR #1601, opened 2020-11-10
Generic header

anxolin:
This PR creates a new generic Header

This header has a basic parametrization, mainly, it allows to inject a menu and a navTool.
The idea is, each app could use its own menu and tools

Future PRs will include an implementation of the menu and the navTool


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1601--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1601--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/layout/GenericLayout/Header/index.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import styled from 'styled-components'
+
+// Assets
+import LogoImage from 'assets/img/logo-v2.svg'
+
+const HeaderStyled = styled.header`
```
W3stside, line 7: styles in own `ComponentName.styled.tsx` please

src/components/layout/GenericLayout/Header/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+`
+
+interface Props {
+  menu?: React.ReactNode
+  tools?: React.ReactNode
+}
+
+export const Header: React.FC<Props> = ({ menu, tools }) => (
```
Velenir, line 58: I'm not a fan how we use different approaches for subcomponents.
NavTools uses children and enabler-props.
This uses prop-as-component (aka slots).

I think with components as Header, which are implied to be Layout-centric,it's best to stick to children only.

Big IMHO

Velenir (approved):

PR #1602, opened 2020-11-10
Add generic menu

anxolin:
This PR adds a generic menu component that can contain any links

Also, adds the Links for the TradeApp


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1602--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1602--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/app/TradeApp/index.tsx
```diff
@@ -20,25 +21,37 @@ const Trading = React.lazy(
...
-      <Switch>
-        <Route path="/v2" exact component={Trading} />
-        <Route component={NotFound} />
-      </Switch>
-    </React.Suspense>
-  </GenericLayout>
-)
+export const TradeApp: React.FC = () => {
+  const menu = (
+    <Menu>
```
W3stside, line 32: we already have code for this no? NavLinks or sth

W3stside: Well i guess this is a _generic menu_ but here is the code I am [referring to](https://github.com/gnosis/dex-react/blob/v2-add-menu/src/components/layout/LegacyTradeLayout/Header/Navigation.tsx)

src/components/layout/GenericLayout/Menu/index.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { PropsWithChildren } from 'react'
+import styled from 'styled-components'
+
+import { MEDIA } from 'const'
+
+export const NavStyled = styled.ol`
```
W3stside, line 6: move files into own folder please

W3stside (approved):
Approved with minor comments

src/components/layout/GenericLayout/Menu/index.tsx
Velenir: Why `ol`? If you want a list, use`ul`at least. Plus I don't see any reason fort his comp to be a list. You override list display and other innate rendering anyway. And list enforces `<li>` as children
Just use div.

src/app/TradeApp/index.tsx
```diff
@@ -20,25 +21,37 @@ const Trading = React.lazy(
...
+      <li>
+        <a href="#">Trade</a>
+      </li>
+      <li>
+        <a href="#">Swap</a>
+      </li>
+      <li>
+        <a href="#">Liquidity</a>
```
Velenir, line 40: Can we already use `Link`components, just to not change them later.

biocom: I think it's an ordered list, where arguably order matters? OL is a valid semantic list. A <div> with <a>'s work, but arguably an ordered (or unordered) list is more correct in terms of accessibility?

src/components/layout/GenericLayout/Menu/index.tsx
biocom: I think it's an ordered list, where arguably order matters? OL is a valid semantic list. A

with 's work, but arguably an ordered (or unordered) list is more correct in terms of accessibility?

biocom (approved):

src/app/TradeApp/index.tsx
anxolin: Good point. I'm refactoring the new design. The code u linked seems a bit overcomplicated, maybe is good to add it, or maybe we don't need it so complex for v2, I'll add a task for the TODO to review this

src/components/layout/GenericLayout/Menu/index.tsx
anxolin: @W3stside this is similar to the comment u did in the header PR, I have the same opinion in this case (small components can have small css in the same file), maybe we should discuss it in the other thread. Also I'll add it to the TODO list agree on the procedure on when to make your one file for the styles.

src/app/TradeApp/index.tsx
anxolin: but in reality we don't even have pages... but as u please

anxolin: done

src/app/TradeApp/index.tsx
W3stside: I'll look it over rn

PR #1603, opened 2020-11-10
Add generic nav tools

anxolin:
Adds a parametrized nav tools

Empty by default, allows to add optionally the wallet, settings, notifications and arbitrary links

Starting to fit all together 🚀
image


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1603--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1603--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/app/TradeApp/index.tsx
```diff
@@ -21,11 +28,26 @@ const Trading = React.lazy(
     ),
 )
 
-const NavTools: React.FC = () => (
-  <Frame>
-    <footer>Tools</footer>
-  </Frame>
-)
+const PortfolioLink = styled.li`
```
W3stside, line 31: can you please move all styled component stuff into a `ComponentName.styled.ts` file?

src/components/layout/GenericLayout/NavTools/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { MEDIA } from 'const'
+import { NavStyled } from '../Menu'
+import { Frame } from 'components/common/Frame'
+
+import SettingsImage from 'assets/img/settings.svg'
+import SettingsImageWhite from 'assets/img/settings-white.svg'
+import NotificationsImage from 'assets/img/bell.svg'
+import NotificationsImageWhite from 'assets/img/bell-white.svg'
+
+const SettingsLink = styled.li`
```
W3stside, line 13: move styled stuff into own file please

src/components/layout/GenericLayout/NavTools/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+      display: block;
+      width: 0.1rem;
+      height: 2.4rem;
+      background: var(--color-border);
+      margin: 0 1.6rem;
+      position: absolute;
+      right: -2.8rem;
+    }
+
+    @media ${MEDIA.mediumDown} {
```
W3stside, line 82: you could move this rule into the already present @media above. For me it's easier to read when it's all in the same media query BUT i do agree it's easier to code it the way it is now. Up to you

W3stside (approved):

src/components/layout/GenericLayout/NavTools/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    &:not(:last-of-type)::after {
+      content: '';
+      display: block;
+      width: 0.1rem;
+      height: 2.4rem;
+      background: var(--color-border);
+      margin: 0 1.6rem;
+      position: absolute;
+      right: -2.8rem;
+    }
```
Velenir, line 80: Is this for pipe | separator? This is quite a lot of artificial code. Maybe better have a `<Separator/>` component used inside NavTools
Plus it probably shouldn't be css-only semantically-wise.

src/components/layout/GenericLayout/NavTools/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+export const NavTools: React.FC<Props> = ({ children, hasWallet, hasSettings, hasNotifications }) => (
+  <Wrapper>
+    {children}
+    {hasWallet && <UserWallet />}
+    {hasSettings && (
+      <SettingsLink>
+        <a href="#" title="View Settings Menu"></a>
+      </SettingsLink>
+    )}
+    {hasNotifications && (
```
Velenir, line 111: What separates children passed it from children enabled by `has*`
NavTools either should be a layout only component, then `<Wrapper>{children}</Wrapper>`
Or the whole shebang,then no children only enabling props.
IMHO

Velenir (commented):
Here are my nitpicks

src/components/layout/GenericLayout/NavTools/index.tsx
biocom: Yeah, in terms of 'grouping' all rules for a specific selector, I always opt to keep the media queries closely aligned (nested) for the element it applies to. Keeps the code dryer in terms of not having to replicate writing the selector. Thoughts on this?

src/components/layout/GenericLayout/NavTools/index.tsx
biocom: I see your point. I think it's a preference of keeping markup dry vs. using CSS. As for the semantic question, I wonder if the pipe should be treated as a semantic item or rather a visual 'separator'. Purely semantically it's a valid list with list items. I usually opt for keeping the markup (HTML) dry and try to solve it with CSS. But up to your evaluation.

Velenir (approved):

src/components/layout/GenericLayout/NavTools/index.tsx
Velenir: Don't have a strong opinion either way. Separator is more versatile and can be added where needed without ` &:not(:last-of-type)::after` specification, which is brittle and is either everywhere the selector mathces or you'll have to add classes to specific sections you want to separate.

And css frameworks usually go for separators, e.g. MaterialUI has `<Divider/>

src/components/layout/GenericLayout/NavTools/index.tsx
W3stside: Agreed @biocom but it can be easy to miss things when changing media query code later imo

src/components/layout/GenericLayout/NavTools/index.tsx
biocom: Is the alternative to create a <Separator /> element we then conditionally (with a function?) add on the markup, except for the last item in a list?

src/components/layout/GenericLayout/NavTools/index.tsx
biocom: Can you name a scenario where you'd change a media query somewhere, and that impacting a specific media query nested in particular element?

src/components/layout/GenericLayout/NavTools/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    display: block;
+    margin: 0;
+    width: 3.6rem;
+    height: 1.4rem;
+    content: '';
+    background: url(${NotificationsImage}) no-repeat center/contain;
+  }
+
+  &:hover > a {
+    background: url(${NotificationsImageWhite}) no-repeat center/contain;
```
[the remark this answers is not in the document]
biocom: Agree makes the code dryer. Have to watch for making sure the overrides are inherited properly on hover (no-repeat center/contain etc.) cross browser.

src/components/layout/GenericLayout/NavTools/index.tsx
Velenir: yes, map over children and add `<Separator/> for example
or add `<Separator/>` ourselves together with the children for better customizability

biocom: Fine with me. From one side I find it overkill to create a specific component for it. On the other hand I fully understand your argument and specifically about a potential re-use of the component (to display the | separator in a styled way).

src/components/layout/GenericLayout/NavTools/index.tsx
W3stside: sure, the size of the media query

src/components/layout/GenericLayout/NavTools/index.tsx
W3stside: i agree with @biocom here, the CSS is fine and is reusable. Adding a separator component to me is overkill

src/components/layout/GenericLayout/NavTools/index.tsx
biocom: The actual value of the media query is handled in the global variables file. Globally changing the variable name in all files can be an easy search 'n replace. Or you mean something else?

src/app/TradeApp/index.tsx
anxolin: hehe, I'm happy to do it, but let's have a chat about it on the reasons why do it or why don't
I added a task also in the TODO

src/components/layout/GenericLayout/NavTools/index.tsx
anxolin: If I get this right, I would lean to Michel approach, where the CSS of one component is close to the functional component it applies. I don't have a strong opinion, in this case, it wouldn't hurt my eyes to have the `ol` and `li` media defined close to each other (since the css is small, and is easy to maintain either way)

anxolin: I'll merge this PR as it is, but let's decide and agree what is best

src/components/layout/GenericLayout/NavTools/index.tsx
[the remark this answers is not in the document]
anxolin: Ok, applying, I really need to finish consolidating this PRs!
It took more time than anticipated. If there's issue we review as a new PR, I'll apply dimas commit

src/components/layout/GenericLayout/NavTools/index.tsx
anxolin: Agree with David and Michel, this is what CSS is for :) 
let's try to be semantic with the mark up

src/components/layout/GenericLayout/NavTools/index.tsx
anxolin: Agreed, let do as u sugges

PR #1605, opened 2020-11-10
Use already made formatPercentage function as used elsewhere

W3stside:
Use already created `utils/format.ts` > `formatPercentage` function

Properly formats to 2 decimals places as seen in other parts of app


W3stside:
need to re-open, travis issues. Initially had branch set as `develop` which is breaking all the things. Can't seem to change the base branch in travis back to 1.6-fixes

Issue #1604, opened 2020-11-10
Refactor for v2


anxolin:
@biocom did a great PR with the new design in #1591 

The issue, with that PR is that it modifies too many files, also creates a coupled architecture and introduces duplication of code.

For this reason, this PR is going to be broken in different parts:
* [ ] Global styles: https://github.com/gnosis/dex-react/pull/1596  
* [ ] Rename TradeLayout to generic layout: https://github.com/gnosis/dex-react/pull/1597
* [ ] TradeApp: https://github.com/gnosis/dex-react/pull/1598
* [ ] Refactor legacy app:  https://github.com/gnosis/dex-react/pull/1599 
* [ ] Rename old layout to Legacy: https://github.com/gnosis/dex-react/pull/1600 
* [ ] Generic header:  https://github.com/gnosis/dex-react/pull/1601 
* [ ] Add generic menu:  https://github.com/gnosis/dex-react/pull/1602 
* [ ] Refactor UserWallet  
* [ ] Refactor WalletComponent
* [ ] Improve global styles: Maybe apply a similar patter as uniswap: Globals, and GlobalsTheme
* [ ] Review some styles I removed from the globals because I thought it didn't belong there (see below)

## Additional notes
Note, review some styles I removed from the globals because I thought it didn't belong necessary there, or maybe we don't even need them 🤷‍♂️

```ts
import { createGlobalStyle } from 'styled-components'
import checkWhite from 'assets/img/check-white.svg'

const GlobalStyles = createGlobalStyle`

#root {
  height: 100%;
  font-size: 1.3rem;
}

button {
  background-color: var(--color-background-CTA);
  color: var(--color-text-CTA);
  font-family: inherit;
  cursor: pointer;
  transition: all 0.2s ease-in-out;
  transition-property: color, background-color, border-color, opacity;
  border: 0;
  font-weight: var(--font-weight-bold);
  outline: 0;

  &:hover {
    background-color: var(--color-background-button-hover);
    color: var(--color-text-button-hover);
  }

  &.cancel {
    background: transparent;
    color: var(--color-text-active);

    &:hover {
      background-color: var(--color-background-button-hover);
      color: var(--color-text-button-hover);
    } 
  }

  &:disabled,
  &[disabled]{
    &:hover {
      background-color: var(--color-background-button-disabled-hover);
    }
    opacity: .35;
    pointer-events: none;
  }
  &.success {
    border-color: var(--color-button-success);
    color: var(--color-button-success);
[...]

PR #1609, opened 2020-11-12
Add order form /w components

biocom:
- Starting to work on adding a OrderFrom component.
- Cleanup global styles and variables

### DIFF
--- a/src/components/TokenPairSelector/index.tsx
+++ b/src/components/TokenPairSelector/index.tsx
@@ -5,7 +5,7 @@ import { TokenPairSelectorList } from 'components/TokenPairSelectorList'
 
 import ArrowWhite from 'assets/img/arrow-white.svg'
 
-const TokenPairSelectorStyled = styled.div`
+const Wrapper = styled.div`
   display: flex;
   width: 100%;
 
@@ -58,13 +58,13 @@ const TokenPairSelectorStyled = styled.div`
 `
 
 export const TokenPairSelector: React.FC = () => (
-  <TokenPairSelectorStyled>
+  <Wrapper>
     <button>
       <b>ETH/USDC</b>
       <i>Select Pair</i>
       <TokenPairSelectorList />
     </button>
-  </TokenPairSelectorStyled>
+  </Wrapper>
 )
 
 export default TokenPairSelector
--- a/src/components/TokenPairSelector/index.tsx
+++ b/src/components/TokenPairSelector/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 
-import { TokenPairSelectorList } from 'components/TokenPairSelectorList'
+import TokenPairSelectorList from 'components/TokenPairSelectorList'
 
 import ArrowWhite from 'assets/img/arrow-white.svg'
 
@@ -57,7 +57,7 @@ const Wrapper = styled.div`
   }
 `
 
-export const TokenPairSelector: React.FC = () => (
+const TokenPairSelector: React.FC = () => (
   <Wrapper>
     <button>
       <b>ETH/USDC</b>
--- a/src/components/TokenPairSelector/index.tsx
+++ b/src/components/TokenPairSelector/index.tsx
@@ -1,0 +1,70 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { TokenPairSelectorList } from 'components/TokenPairSelectorList'
+
+import ArrowWhite from 'assets/img/arrow-white.svg'
+
+const TokenPairSelectorStyled = styled.div`
+  display: flex;
+  width: 100%;
+
+  > button {
+    width: 100%;
+    height: var(--height-bar-default);
+    background: var(--color-primary);
+    color: var(--color-text-secondary1);
+    appearance: none;
+    border: 0;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: 0 1.6rem;
+    box-sizing: border-box;
+    border-bottom: 0.1rem solid var(--color-border);
+    outline: 0;
+    cursor: pointer;
+    transition: color 0.3s ease-in-out;
+
+    &:hover {
+      color: var(--color-text-primary);
+    }
+  }
+
+  > button > b {
+    font-size: 1.5rem;
+    letter-spacing: 0.03rem;
+  }
+
+  > button > i {
+    font-size: 1.2rem;
+    font-style: normal;
+    display: flex;
+    flex-flow: row nowrap;
+    justify-content: center;
+    align-items: center;
+    font-weight: var(--font-weight-medium);
+
+    &::after {
+      display: block;
+      height: 1.2rem;
+      width: 1.2rem;
+      content: '';
+      background: url(${ArrowWhite}) no-repeat center/contain;
+      transform: rotate(90deg);
+      margin: 0 0 0 0.7rem;
+    }
+  }
+`
+
+export const TokenPairSelector: React.FC = () => (
+  <TokenPairSelectorStyled>
+    <button>
+      <b>ETH/USDC</b>
+      <i>Select Pair</i>
+      <TokenPairSelectorList />
+    </button>
+  </TokenPairSelectorStyled>
+)
+
+export default TokenPairSelector
--- a/src/components/TokenPairSelectorList/index.tsx
+++ b/src/components/TokenPairSelectorList/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 
-const TokenPairListStyled = styled.div`
+const TokenPairSelectorListStyled = styled.div`
   width: 100%;
   display: none;
   position: absolute;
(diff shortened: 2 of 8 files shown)


src/components/TokenPairSelector/index.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import styled from 'styled-components'
+
+import { TokenPairSelectorList } from 'components/TokenPairSelectorList'
+
+import ArrowWhite from 'assets/img/arrow-white.svg'
+
+const TokenPairSelectorStyled = styled.div`
```
biocom, line 8: Do we specifically want the styled component in a separate file? And what is the threshold for when we opt to do this vs. just keeping it in the FC file?

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1609--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1609--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/TokenPairSelector/index.tsx
anxolin: My opinion is that we shouldn't while the file is small
it's nice to have a small markup, and small css in a small file
Probably @W3stside  would disagree with me

src/components/TokenPairSelector/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  display: flex;
+  width: 100%;
+
+  > button {
```
anxolin, line 12: I'm wondering. Should we have our own styled buttons instead of doing it here?
same approach as Uniswap.

src/components/layout/GenericLayout/variablesCss.ts
```diff
@@ -1,9 +1,12 @@
 import { css } from 'styled-components'
 
 const AllColors = `
+  /* HEIGHTS */
+  --height-bar-default: 6.2rem;
+  /* ------------------------------ */
+
   /* FONTS */
   --font-default: "Inter", "Helvetica Neue", Helvetica, sans-serif;
-  --font-mono: "Roboto Mono", monospace;
```
anxolin, line 6: good call to do all this cleanup. We can leave it in this PR. But this one is better of in it's own independent PR
The smaller the PR, the better

src/components/TokenPairSelector/index.tsx
anxolin, line 8: ubernit: We usually call the outermost styled element `wrapper` (instead of `TokenPairSelectorStyled ` in this case). It's just a convention we've used. Open for discussion if it's not good. In my eyes, I like it. Wdyt?

src/components/TokenPairSelector/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+`
+
+export const TokenPairSelector: React.FC = () => (
+  <TokenPairSelectorStyled>
+    <button>
+      <b>ETH/USDC</b>
+      <i>Select Pair</i>
+      <TokenPairSelectorList />
```
anxolin, line 65: I think this is a select component. Can we do a generic component for any select?

src/components/TokenPairSelector/index.tsx
biocom: In this case, while using a `<button>` it will not be a button like we will use else where. Just decided that using a `<button>` element might be most appropriate here. So I regard this a specific use case. 

That being said, I think it would be good to have a reusable styled button where applicable.

src/components/TokenPairSelector/index.tsx
biocom: Yeah sounds good. So in this case `TokenPairSelectorStyled` is only used in this index.tsx file, so then just calling it `wrapper` would be a consistent approach of naming it, correct?

src/components/TokenPairSelector/index.tsx
biocom: It depends what we define under a `select` component. In my opinion this will be a full blown select component where you can do multiple things (tabs, star token pairs, etc.). In my mind it only makes sense to have a generic `select` component for something we can re-use (like we did with the V1 token selector). Curious on your thoughts.

src/components/TokenPairSelector/index.tsx
W3stside: when scaffolding or starting and it is small, keep them in it's fine. But once it grows, move them. From there start to think about whether or not a component can be extrapolated into a more generic form and inherited around the app.

src/components/TokenPairSelector/index.tsx
W3stside: see [my comment here](https://github.com/gnosis/dex-react/issues/1604#issuecomment-725997685) regarding this. It should be a styled button we inherit from for more specific cases. But fine to flesh out as pure CSS now and rework later

src/components/TokenPairSelector/index.tsx
biocom: Updated.

src/components/TokenPairSelector/index.tsx
W3stside: should this not be `react-select` or are we scrapping that?

biocom: Would our purposed token pair select menu, fit the use of `react-select`? (never used it extensively)

src/components/TokenPairSelector/index.tsx
biocom: I can try a start with that. I guess we try to fit all types of button logic into a generic `Button` component? Even if they look/behave entirely different, correct?

src/components/TokenPairSelector/index.tsx
anxolin: But in this specific case, I don't think we should move it to it's own styled file.
I do think we should create a reusable button, and a reusable select. But being so small, it's totally fine for me to be in the same file

src/components/TokenPairSelector/index.tsx
anxolin: Good point :) 

Yes and no.

I think in this case it would be more a button that looks like a select? 

Select pair here:
image

Opens this:
image


So it might be that is not even a select

biocom: It rather is:
- On click of the `<button>` element (Select pair) it opens a new component.
- The new component consists of:
  - Tabs
  - Each content of a selected tab contains a scrollable list + inline search function

src/components/TokenPairSelector/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    justify-content: center;
+    align-items: center;
+    font-weight: var(--font-weight-medium);
+
+    &::after {
+      display: block;
+      height: 1.2rem;
+      width: 1.2rem;
+      content: '';
+      background: url(${ArrowWhite}) no-repeat center/contain;
```
anxolin, line 53: This will likely be an optional image passed as a prop to the button
We could parametrize the use of an image to the right or to the left of the button

Doesn't need to be in this PR though. Since it can easily get complicated...
we can mege soon this to develop, and then work in refactoring this into finer pieces

anxolin (approved):

src/components/TokenPairSelector/index.tsx
anxolin: cool, it works. So, in reply to @W3stside we don't need the react-select, cause is a button

src/components/TokenPairSelector/index.tsx
anxolin: > I can try a start with that. I guess we try to fit all types of button logic into a generic Button component? Even if they look/behave entirely different, correct?

correct!

src/components/TokenPairSelector/index.tsx
W3stside: my point is mainly that now, at this early re-factor stage, we should be doing this actively, instead of retro-actively. Largely because it is easiest now.

Keeping styled in one folder is fine, yes. Though that is what we started off doing and then files got huge and the rest is history. So while it's not the end of the world in anyway to have too many styled comps in a main component file, it's just my 2 cents

src/components/TokenPairSelector/index.tsx
Velenir: let's not use `react-select` I remember the pain...

src/components/TokenPairSelectorList/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+  display: none;
+  position: absolute;
+  top: 0;
+  height: 100vh;
+  background: yellow;
+  left: 0;
+`
+export const TokenPairSelectorList: React.FC = () => <Wrapper>-Token Pair List Element -</Wrapper>
+
+export default TokenPairSelectorList
```
Velenir, line 15: two exports?

Velenir (approved):

src/components/TokenPairSelectorList/index.tsx
biocom: Ah yeah, fixed now :)

src/components/TokenPairSelector/index.tsx
anxolin: Well, we are open for the change. But we need to consider we have a working select with functionality to add tokens and so on. If we can leverage that is cool. If we don't need it, we can can make another proposal. Anyways, for now we don't even have a select I think

PR #1610, opened 2020-11-12
Rate limit calls

Velenir:
Sample implementation of rate limiting provider calls

### DIFF
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -330,7 +330,8 @@ export const composeProvider = <T extends Provider>(
 
   if (CHAIN_CALLS_RATE_LIMIT > 0) {
     const passThroughMware: JsonRpcMiddleware = (_req, _res, next) => next()
-    const rateLimitedPassThrough = rateLimit(passThroughMware, CHAIN_CALLS_RATE_LIMIT) // a more visible delay of 100ms
+    // consecutive calls with CHAIN_CALLS_RATE_LIMIT in between
+    const rateLimitedPassThrough = rateLimit(passThroughMware, CHAIN_CALLS_RATE_LIMIT)
 
     engine.push((req, res, next, error) => {
       // tx signing is wallet-dependent
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -299,6 +299,61 @@ export const composeProvider = <T extends Provider>(
     ),
   )
 
+  // Simple delay implementation
+  // engine.push(
+  //   createConditionalMiddleware(
+  //     (req) => req.method !== 'eth_sendTransaction',
+  //     async (req) => {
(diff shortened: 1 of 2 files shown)


anxolin:
This PR is not super required right now, if we want to reuse the work, we could just disable by default the rate limit (by default is 0ms, that wouldn't add the middleware)

I understand this would allow us to have equivalent code, and be able to enable the rate limit for testing with just an env var

anxolin (commented):
This PR is not super required right now, if we want to reuse the work, we could just disable by default the rate limit (by default is 0ms, that wouldn't add the middleware)

I understand this would allow us to have equivalent code, and be able to enable the rate limit for testing with just an env var

Velenir:
Though `0` would remove the delay, it would still can end up >0 because of event queue
I vote for closing this PR, and reopening when/if needed

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1610--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1610--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/api/wallet/composeProvider.ts
```diff
@@ -276,6 +276,61 @@ export const composeProvider = <T extends Provider>(
...
+      queue.push(args)
+      if (!timer) {
+        processQueue() // start immediately on the first invocation
+        timer = setInterval(processQueue, delay)
+      }
+    }
+  }
+
+  const passThroughMware: JsonRpcMiddleware = (_req, _res, next) => next()
+  const rateLimitedPassThrough = rateLimit(passThroughMware, 100) // a more visible delay of 100ms
```
anxolin, line 322: should we have an env for this?, and when the env is 0 we just don't add the middleware?

anxolin: 0 should be the default of that env imo

src/api/wallet/composeProvider.ts
```diff
@@ -276,6 +276,61 @@ export const composeProvider = <T extends Provider>(
     ),
   )
 
+  // Simple delay implementation
```
anxolin, line 279: remove dead code

src/api/wallet/composeProvider.ts
Velenir: ok, that makes sense
let me do it quickly

anxolin (approved):

W3stside (approved):
the code looks fine... but I'm scared this will have some unintended side effects. Are we sure this won't unexpectedly drop transactions?

Velenir:
> the code looks fine... but I'm scared this will have some unintended side effects. Are we sure this won't unexpectedly drop transactions?

It wouldn't, and it's disabled for now because of 0 delay

anxolin:
@W3stside merging this one here!

PR #1614, opened 2020-11-12
Sanitize error.data

Velenir:
In case when there's ` res.error.data = {originalError}` it breaks `res.error.data.substring()` inside web3
This PR reassigns `data` form `originalError` right after receiving response from provider

Works around `execution reverted: Only the proxy can call` error for deprecated SNX and sUSD

PR #1615, opened 2020-11-12
Sanitize error data

anxolin:
Cherry-picked solution from #1614 

Intercepts the error handling, making sure the `data` field is present.

This solves an issue with web3 failing in their error handling, leaving the promises unfulfilled.

### DIFF
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -12,14 +12,6 @@ import { isWalletConnectProvider, Provider } from './providerUtils'
 import { logDebug } from 'utils'
 import { web3 } from 'api'
 
-import {
-  addTxPendingApproval,
-  areTxsPendingApproval,
-  openWaitForTxApprovalModal,
-  removeAllTxsPendingApproval,
-  removeTxPendingApproval,
-} from 'components/OuterModal'
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const sanitizeErrorData = (jsonRpcError?: JsonRpcError<any>): void => {
   if (!jsonRpcError) return
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -12,6 +12,23 @@ import { isWalletConnectProvider, Provider } from './providerUtils'
 import { logDebug } from 'utils'
 import { web3 } from 'api'
 
+import {
+  addTxPendingApproval,
+  areTxsPendingApproval,
+  openWaitForTxApprovalModal,
+  removeAllTxsPendingApproval,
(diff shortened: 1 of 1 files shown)


Velenir (approved):

src/api/wallet/composeProvider.ts
```diff
@@ -12,6 +12,15 @@ import { isWalletConnectProvider, Provider } from './providerUtils'
 import { logDebug } from 'utils'
 import { web3 } from 'api'
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const sanitizeErrorData = (jsonRpcError?: JsonRpcError<any>): void => {
+  if (!jsonRpcError) return
```
W3stside, line 17: can jsonRpcError be an empty object?

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1615--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1615--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/api/wallet/composeProvider.ts
```diff
@@ -54,7 +70,11 @@ function providerAsMiddleware(provider: Provider): JsonRpcMiddleware {
 
     provider[sendFName](req, (err: JsonRpcError<unknown>, providerRes: JsonRpcResponse<unknown>) => {
       // forward any error
-      if (err) return end(err)
+      if (err) {
+        sanitizeErrorData(err)
+        return end(err)
+      }
+      sanitizeErrorData(providerRes.error)
```
W3stside, line 77: Bit confused here.

What do we expect `providerRes.error` to be when `err === true` and when `err === false`? My confusion stems from line 77, that `sanitizeErrorData` will never run because of line 75

But I'm probably understanding this wrong is all ;)

W3stside (approved):

src/api/wallet/composeProvider.ts
Velenir: It shouldn't, as long as types are respected, and just in general.
But even if it is, it won't break

src/api/wallet/composeProvider.ts
Velenir: Errors can be handled in two ways
We get Object or we get error prop on response
Need to handle them both, because they both can be indirectly formatted

W3stside: so the return doesn't break line 77? or it's one or the other. either way if you see what I mean and it's fine then you've handled it in which case i say CASE CLOSED

Velenir: We call sanitizer whether before return or in else clause (after)

PR #1616, opened 2020-11-12
Do not load balance from snx

anxolin:
WIP, try not to load some tokens that we know they break when trying to load their balance

For this, one new field `display` is added to the deprecated list of tokens. So we can indicated that a deprecated token shouldn't even be displayed (because for example, we cannot load the balances)

Applies filter early in the token loading strategy:
image

But for some reason, it is loading them:
image

### DIFF
--- a/config-default.yaml
+++ b/config-default.yaml
@@ -36,7 +36,7 @@ tcr:
     lists:
       - networkId: 1
         listId: 1
-        contractAddress: '0x1854dae560abb0f399d8badca456663ca5c309d0'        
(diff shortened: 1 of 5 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1616--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1616--dexreact.review.gnosisdev.com/storybook/)**: Component stories

PR #1618, opened 2020-11-13
Fix issue approving tokens

anxolin:
This solves Anna's issue with approving tokens using Wallet Connect.

> TLDR: The issue, is that the estimation is failing if we provide a `gasPrice`

yes, really. I'm betting this has to do with yesterday hardfork in some way. https://www.coindesk.com/ethereums-hard-fork-disruption

So this is how we estimate. We just send the same params we would use in the tx, to the `estimateGas` function:
```js
web3.eth.estimateGas({data: "0x095ea7b30000000000000000000000006f400810b62df8e13fded51be75ff5393eaa841fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdec0de02120132", from: "0xd5cbc5e664813c7f1ce0a136aa154ce525e97e21", to: "0x6810e776880c02933d47db1b9fc05908e5386b96", gasPrice: "0xcce416bb4"}, console.log)
```

> Throws error: `react_devtools_backend.js:2450 MetaMask - RPC Error: gas required exceeds allowance (160) {code: -32000, message: "gas required exceeds allowance (160)"}`

The original params has a `priceEstimation`, so it goes with the data of the estimation too.
You would think that if the estimation don't need that info, they would ignore it. It seems is not the case! at least now, I bet it used to be and not anymore!

https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#methods-mymethod-estimategas

```js
web3.eth.estimateGas({data: "0x095ea7b30000000000000000000000006f400810b62df8e13fded51be75ff5393eaa841fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdec0de02120132", from: "0xd5cbc5e664813c7f1ce0a136aa154ce525e97e21", to: "0x6810e776880c02933d47db1b9fc05908e5386b96"}, console.log)
```

> Returns: `44374`

Only difference, I removed the price estimation

### DIFF
--- a/src/api/wallet/composeProvider.ts
+++ b/src/api/wallet/composeProvider.ts
@@ -201,10 +201,16 @@ export const composeProvider = <T extends Provider>(
         if (!txConfig) return false
 
         if (!txConfig.gas) {
-          const gasLimit = await web3.eth.estimateGas(txConfig).catch((error) => {
-            console.error('[composeProvider] Error estimating gas, probably failing transaction', txConfig)
-            throw error
-          })
+          const gasLimit = await web3.eth
+            .estimateGas({
+              from: txConfig.from,
+              gas: txConfig.gas,
+              value: txConfig.value,
+            })
+            .catch((error) => {
+              console.error('[composeProvider] Error estimating gas, probably failing transaction', txConfig)
+              throw error
+            })
           logDebug('[composeProvider] No gas Limit. Using estimation ' + gasLimit)
           txConfig.gas = numberToHex(gasLimit)
         } else {


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1618--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1618--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):
Great job debugging it :+1: 
Though I would have expected tsConfig to be sanitized inside web3 or provider

anxolin:
> Though I would have expected tsConfig to be sanitized inside web3 or provider

But `gasPrice` is not even part of the web3 params, so I don't get why they use it.
More over, what has silently changed. Or changed not silently and I'm unaware.

This is another interesting test, if I send any prop they don't know about, like `gasPrice2` in the following example, it works. So it ignores the extra properties as it should. Is just the `gasPrice` is now being used for some reason! (and is not part of the params)
```js
web3.eth.estimateGas({data: "0x095ea7b30000000000000000000000006f400810b62df8e13fded51be75ff5393eaa841fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdec0de02120132", from: "0xd5cbc5e664813c7f1ce0a136aa154ce525e97e21", to: "0x6810e776880c02933d47db1b9fc05908e5386b96", gasPrice2: "0xcce416bb4"}, console.log)
```

Velenir:
> > Though I would have expected tsConfig to be sanitized inside web3 or provider
> 
> But `gasPrice` is not even part of the web3 params, so I don't get why they use it.
> More over, what has silently changed. Or changed not silently and I'm unaware.
> 
> This is another interesting test, if I send any prop they don't know about, like `gasPrice2` in the following example, it works. So it ignores the extra properties as it should. Is just the `gasPrice` is now being used for some reason! (and is not part of the params)
> 
> ```js
> web3.eth.estimateGas({data: "0x095ea7b30000000000000000000000006f400810b62df8e13fded51be75ff5393eaa841fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdec0de02120132", from: "0xd5cbc5e664813c7f1ce0a136aa154ce525e97e21", to: "0x6810e776880c02933d47db1b9fc05908e5386b96", gasPrice2: "0xcce416bb4"}, console.log)
> ```

types allow it, though
[estimateGas](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth/types/index.d.ts#L346-L347) > [TransactionConfig](https://github.com/ethereum/web3.js/blob/0df0fea0bc/packages/web3-core/types/index.d.ts#L131)

PR #1621, opened 2020-11-16
Use same TCR as mesa for testing

anxolin:
Use same TCR as mesa for testing

Any project can for dex-react and use their own TCR. For testing the TCR functionality, we could use same as Mesa, since I assume they would provide a rationale on their listing and deprecate old tokens (and check for compatibilities before listing) 

I left as a comment the old TCR we were using, since it can be handy to use back in the future for some tests.

### DIFF
--- a/config-default.yaml
+++ b/config-default.yaml
@@ -36,6 +36,7 @@ tcr:
     lists:
       - networkId: 1
         listId: 1
+        # contractAddress: '0x1854dae560abb0f399d8badca456663ca5c309d0' # Testing TCR with rare tokens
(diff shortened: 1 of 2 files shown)


Velenir (approved):

Velenir:
Just fix tests, and good to go

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1621--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1621--dexreact.review.gnosisdev.com/storybook/)**: Component stories

W3stside (approved):

PR #1622, opened 2020-11-16
Allow to not specify hops

anxolin:
Remove hops param from URL for the orderbook

Also, allow in the `/book` to leave hops empty, and make that the default.

@fleupold is there a maximum number of hops we should set (for the validation only)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1622--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1622--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/pages/OrderBook.tsx
```diff
@@ -84,8 +84,19 @@ const OrderBook: React.FC = () => {
   const { tokens: tokenList } = useTokenList({ networkId: networkIdOrDefault })
   const [baseToken, setBaseToken] = useSafeState<TokenDetails | null>(null)
   const [quoteToken, setQuoteToken] = useSafeState<TokenDetails | null>(null)
-  const [hops, setHops] = useSafeState(ORDER_BOOK_HOPS_DEFAULT)
   const [batchId, setBatchId] = useSafeState<number | undefined>(undefined)
+  const [hopsValue, setHopsValue] = useSafeState('')
```
W3stside, line 88: why not default this to `undefined`

src/pages/OrderBook.tsx
```diff
@@ -132,16 +143,11 @@ const OrderBook: React.FC = () => {
...
               type="number"
               min="0"
               max={ORDER_BOOK_HOPS_MAX.toString()}
-              onChange={(e: ChangeEvent<HTMLInputElement>): void => {
-                const hopsValue = e.target.value
-                if (hopsValue && !Number.isNaN(Number(hopsValue))) {
-                  setHops(Number(hopsValue))
-                }
-              }}
+              onChange={(e: ChangeEvent<HTMLInputElement>): void => setHopsValue(e.target.value)}
```
W3stside, line 150: could/should create a handler that accepts `e` as default

```const handleChange = (e: EventType): void => setHopsValue(e.target.value)```

then just pass it as 
```onChange={handleChange}```

File: src/components/OrderBookWidget.stories.tsx [hunk 3]
```diff
@@ -17,7 +17,7 @@ const defaultParams: OrderBookProps = {
   baseToken: baseTokenDefault,
   quoteToken: quoteTokenDefault,
   networkId: defaultNetworkId,
-  hops: ORDER_BOOK_HOPS_DEFAULT,
+  hops: undefined,
```
Velenir, line 20: No that there are 30 hops max, I wonder if it will be better to change `control.type`
pr1622--dexreact review gnosisdev com_storybook__path=_story_orderbook-sample-data--liquid-market

Velenir (approved):

W3stside (approved):

src/pages/OrderBook.tsx
anxolin: I usually like to do as you suggest, however in this case, the code is so small, I decide against separating the handler from it's basic implementation. I believe it's still easy to read (it does one thing)

Let me know if u feel strong about this one

src/components/OrderBookWidget.stories.tsx
anxolin: Good one!

PR #1623, opened 2020-11-16
Adding a generic Tabs component.

biocom:
- Adding a generic Tab component.
- Adding some supporting global variables
- Start making use of the Tab component in the OrderBuySell component


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1623--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1623--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/OrderBuySell/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import Tabs from 'components/common/Tabs/Tabs'
+
+export interface TabItemType {
+  readonly id: number
+  title: string
+  content: string
+  readonly activeColor?: string
+}
+
+// pass string props to indicate what content/component to show
```
W3stside, line 12: remove

src/components/OrderBuySell/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+    id: 1,
+    title: 'BUY',
+    content: '- buy component -',
+    activeColor: '--color-buy',
+  },
+  {
+    id: 2,
+    title: 'SELL',
+    content: '- sell component -',
+    activeColor: '--color-sell',
```
W3stside, line 25: not sure if we're keeping `var(--color-style-thing)` variables in CSS, @Velenir @anxolin any input? My understanding was that we move to JS style CSS and theming

biocom: Yeah probably better, as I saw in your Button component. So in this case we could use a JS theme variable like `sellColor` or something. Perhaps for a future PR?

src/components/OrderBuySell/index.tsx
biocom: Remove the comment, or?

src/components/common/Tabs/TabContent.tsx
```diff
@@ -0,0 +1,20 @@
...
+  tabItems: TabItemType[]
+  activeTab: number
+}
+
+const TabContent: React.FC<Props> = (props) => {
+  const { tabItems, activeTab } = props
+  return (
+    <div>
+      {tabItems.map((tab) => {
+        return tab.id === activeTab ? tab.content : ''
```
Velenir, line 14: This would render `''` for non-active tabs.
Better do
```suggestion
        return tab.id === activeTab ? tab.content : null
```

to not render anything

src/components/common/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react'
+import styled from 'styled-components'
+
+interface TabProps {
+  title: string
+  readonly id: number
+  readonly isActive: boolean
+  readonly activeColor?: string
```
Velenir, line 8: Faaancy :+1: 
@W3stside , @anxolin why aren't we using `readonly` more, we totally should

src/components/common/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+    border-bottom-left-radius: var(--border-radius-default);
+  }
+
+  &:last-of-type {
+    border-top-right-radius: var(--border-radius-default);
+    border-bottom-right-radius: var(--border-radius-default);
+  }
+
+  &.isActive {
+    background: ${(props): string => (props.activeColor ? `var(${props.activeColor})` : 'var(--color-primary)')};
```
Velenir, line 41: If you want to pass in `activeColor` then I would expect `activeColor: cyan` to be acceptable
So better
```suggestion
    background: ${(props): string => (props.activeColor || 'var(--color-primary)')};
 ```
and the whole `activeColor: 'var(--***)'` whenever you create it
```

Velenir (approved):

src/components/common/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+    color: var(--color-text-primary);
+  }
+`
+
+const TabItem: React.FC<TabProps> = (props) => {
+  const { onTabClick, id, title, isActive, activeColor } = props
+
+  return (
+    <Wrapper activeColor={activeColor} className={isActive ? 'isActive' : ''} onClick={(): void => onTabClick(id)}>
```
W3stside, line 50: @Velenir @anxolin i think we should finalise on this concept:

1. use classes 
2. use styled props

to affect change in styled comps. If `2` is preferred, this could be rewritten as:
```ts
const Wrapper = styled.li.attrs(({ $activeColor, $isActive }) => ({
  $backgroundColor: `var(${$activeColor ? $activeColor : '--color-primary'})`,
  $color: `var(--color-text${$isActive ? '--color-primary' : '--text-secondary2'})`
}))`
  background: ${({ $backgroundColor }): string => $backgroundColor};
  color: ${({ $color }): string => $color};
  height: var(--height-button-default);
  display: flex;
  align-items: center;
  flex: 1 1 0;
  font-weight: var(--font-weight-bold);
  font-size: var(--font-size-large);
  letter-spacing: 0.1rem;
  text-align: center;
  justify-content: center;
  cursor: pointer;

  &:first-of-type {
    border-top-left-radius: var(--border-radius-default);
    border-bottom-left-radius: var(--border-radius-default);
  }
  &:last-of-type {
    border-top-right-radius: var(--border-radius-default);
    border-bottom-right-radius: var(--border-radius-default);
  }
`
```

src/components/common/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,54 @@
...
+import TabContent from 'components/common/Tabs/TabContent'
+
+interface Props {
+  tabItems: TabItemType[]
+}
+
+const Tabs: React.FC<Props> = (props) => {
+  const [activeTab, setActiveTab] = useState<number>(1)
+
+  const onTabClick = (id: number): void => {
```
W3stside, line 16: how is this fn different from the original state setting fn?

src/components/common/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return setActiveTab(id)
+  }
+
+  const Wrapper = styled.div`
```
W3stside, line 20: never write styled components inside other components! they are remade on every render.

src/components/OrderBuySell/index.tsx
W3stside: yes

src/components/common/Tabs/TabItem.tsx
W3stside: /shrug

W3stside (dismissed):
mainly the styled comp written inside the Tabs component

src/components/common/Tabs/TabItem.tsx
biocom: So this way we enforce to pass a var(--**) type of color instead any type of string/named color, correct?

src/components/common/Tabs/TabItem.tsx
biocom: I think this makes sense, given the class was only applied to target it with CSS. So keeping it in the styled component sounds like we can keep styling logic contained.

src/components/common/Tabs/Tabs.tsx
biocom: Oooh good call, will engrave that in my brain from now on.

src/components/common/Tabs/Tabs.tsx
Velenir: Good eye :+1:

src/components/common/Tabs/Tabs.tsx
Velenir: Good eye, I didn't even notice

src/components/common/Tabs/TabItem.tsx
Velenir: the opposite
`activeColor` can be any CSS color value, both css vars and just colors

src/components/common/Tabs/Tabs.tsx
W3stside: Thanks, eye read carefully :eye:

src/components/common/Tabs/Tabs.tsx
biocom: Not sure I follow. You mean why use the useState hook vs. something else, or?

W3stside: what does `onTabClick` do any different than `setActiveTab` ? Look at the way you're using it on line `43`

biocom: Ah I see now. I could simply pass the setActiveTab directly without having to put it in a separate function :)

src/components/common/Tabs/TabItem.tsx
anxolin: Yes, I vote for `2.` too

src/components/common/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+  readonly activeColor?: string
+}
+
+interface WrapperProps {
+  readonly activeColor?: string
+  isActive: boolean
+}
+
+const Wrapper = styled.li<WrapperProps>`
+  background: ${(props): string => `var(${props.isActive ? props.activeColor : '--color-primary'})`};
```
biocom, line 18: @W3stside Is this approach fine too? Wasn't sure if the `.attr` is required, unless I'm changing actual attributes of the element? E.g. data-%, type, className, etc.

But my hunch from your example, is that your store these variables to reuse in the CSS below. Let me know as I'm a bit blank on that specific syntax.

W3stside: it's not necessary no, this is fine :+1:

src/components/OrderBuySell/index.tsx
W3stside: yeah later is fine totally

PR #1624, opened 2020-11-16
[V2] common button component + stories

W3stside:
Start of theming with styled-components and common button component aimed at reuse

Would love feedback on data structure of theme object

### DIFF
--- a/src/components/common/Button/index.stories.tsx
+++ b/src/components/common/Button/index.stories.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 // also exported from '@storybook/react' if you can deal with breaking changes in 6.1
 import { Story, Meta } from '@storybook/react/types-6-0'
-import { ButtonBase } from '.'
+import { ButtonBase, ButtonBaseProps } from '.'
 import { DarkModeThemeToggler } from 'storybook/decorators'
 
 export default {
@@ -22,7 +22,9 @@ export default {
   },
 } as Meta
 
-const Template: Story = (args) => <ButtonBase {...args}>{args.label}</ButtonBase>
+const Template: Story<ButtonBaseProps & { label: string | React.ReactNode }> = (args) => (
+  <ButtonBase {...args}>{args.label}</ButtonBase>
+)
 
 export const PrimaryButton = Template.bind({})
 PrimaryButton.args = {
--- a/src/components/common/Button/index.stories.tsx
+++ b/src/components/common/Button/index.stories.tsx
@@ -1,0 +1,37 @@
+import React from 'react'
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
+import { Story, Meta } from '@storybook/react/types-6-0'
+import { ButtonBase } from '.'
+import { DarkModeThemeToggler } from 'storybook/decorators'
+
+export default {
+  title: 'Common/Button',
+  component: ButtonBase,
+  decorators: [DarkModeThemeToggler],
+  argTypes: {
+    label: { control: 'text' },
+    theme: {
+      control: null,
+    },
+    as: {
+      control: null,
+    },
+    forwardedAs: {
+      control: null,
+    },
+  },
+} as Meta
+
+const Template: Story = (args) => <ButtonBase {...args}>{args.label}</ButtonBase>
+
+export const PrimaryButton = Template.bind({})
+PrimaryButton.args = {
+  label: 'Main Button',
+}
+
+export const SecondaryButton = Template.bind({})
+SecondaryButton.args = {
+  label: 'SecondaryButton',
+  $alt: true,
+  $border: true,
+}
--- a/src/components/common/Button/index.tsx
+++ b/src/components/common/Button/index.tsx
@@ -1,0 +1,95 @@
+import styled, { ThemedStyledProps } from 'styled-components'
+import { ColourTheme, ThemeType } from 'styles/theme'
+
+const { MAIN, SECONDARY } = ThemeType
+
+export const ButtonBase = styled.button.attrs(({ $alt, $border, theme }: { // $alt button theme toggle
+  $alt?: boolean; $border?: boolean; theme: ThemedStyledProps<ColourTheme, ColourTheme> }) => ({
+  // Handle button theming
+  button: {
+    bg1: !$alt ? theme.button[MAIN].bg1 : theme.button[SECONDARY].bg1,
+    bg1Hover: !$alt ? theme.button[MAIN].bg1Hover : theme.button[SECONDARY].bg1Hover,
+    text1: !$alt ? theme.button[MAIN].text1 : theme.button[SECONDARY].text1,
+    text1Hover: !$alt ? theme.button[MAIN].text1Hover : theme.button[SECONDARY].text1Hover,
+  },
+  // Show/Hide border toggle
+  $border,
+}))`
+  border: 0.1rem solid;
+  border-color: ${({ $border, button }): string => ($border ? button.text1 : 'transparent')};
+  background-color: ${({ button }): string => button.bg1};
+  color: ${({ button }): string => button.text1};
+
+  &:hover {
+    border-color: ${({ button }): string => button.bg1Hover};
+    background-color: ${({ button }): string => button.bg1Hover};
+    color: ${({ button }): string => button.text1Hover};
+  }
+
+  border-radius: 2rem;
+  cursor: pointer;
+  font-weight: bold;
+  outline: 0;
+  padding: 0.5rem 1rem;
+
+  transition: all 0.2s ease-in-out;
+  transition-property: color, background-color, border-color, opacity;
+
+  &.cancel {
+    background: transparent;
+    color: var(--color-text-active);
+
+    &:hover {
+      background-color: var(--color-background-button-hover);
+      color: var(--color-text-button-hover);
+    }
+  }
+
+  &:disabled,
+  &[disabled] {
+    &:hover {
+      background-color: var(--color-background-button-disabled-hover);
+    }
+    opacity: 0.35;
+    pointer-events: none;
+  }
+
+  &.success {
+    border-color: var(--color-button-success);
+    color: var(--color-button-success);
+    :hover {
+      background-color: var(--color-button-success);
+      border-color: var(--color-button-success);
+      color: var(--color-background-pageWrapper);
+    }
+  }
+
+  &.danger {
+    border-color: var(--color-button-danger);
+    color: var(--color-button-danger);
+    :hover {
+      background-color: var(--color-button-danger);
+      border-color: var(--color-button-danger);
+      color: var(--color-background-pageWrapper);
+    }
+  }
+
+  &.secondary {
+    border-color: var(--color-button-secondary);
+    color: var(--color-button-secondary);
+    :hover {
+      border-color: black;
+      color: black;
+    }
+  }
+
+  &.big {
+    font-size: 1.2rem;
+    padding: 0.65rem 1rem;
+  }
+
+  &.small {
+    font-size: 0.6rem;
+    padding: 0.3rem 0.5rem;
+  }
+`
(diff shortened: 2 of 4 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1624--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1624--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/Button/index.tsx
```diff
@@ -0,0 +1,95 @@
+import styled, { ThemedStyledProps } from 'styled-components'
+import { ColourTheme, ThemeType } from 'styles/theme'
+
+const { MAIN, SECONDARY } = ThemeType
+
+export const ButtonBase = styled.button.attrs(({ $alt, $border, theme }: { // $alt button theme toggle
+  $alt?: boolean; $border?: boolean; theme: ThemedStyledProps<ColourTheme, ColourTheme> }) => ({
```
Velenir, line 7: That's a lot of types, make an interface out of them outside the component

src/components/common/Button/index.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+    label: { control: 'text' },
+    theme: {
+      control: null,
+    },
+    as: {
+      control: null,
+    },
+    forwardedAs: {
+      control: null,
+    },
```
Velenir, line 21: These are styled-component intrinsic props? And you disable controls because we don't need them or because they break on edit (`theme` would breaks as it is an object I imagine)

src/styles/theme.ts
```diff
@@ -0,0 +1,201 @@
...
+
+interface ColourThemeBase {
+  [t: string]:
+    | {
+        [key in ThemeType]?:
+          | {
+              [prop: string]: string
+            }
+          | string
+          | undefined
```
Velenir, line 22: This is a bit hard to read.
Mind extracting into its own type?

src/styles/theme.ts
```diff
@@ -0,0 +1,201 @@
...
+  SECONDARY = 'secondary',
+}
+
+interface ColourThemeBase {
+  [t: string]:
+    | {
+        [key in ThemeType]?:
+          | {
+              [prop: string]: string
+            }
```
Velenir, line 20: This can be `Record<string, string>`

Velenir: All these types can be Records, but maybe they would become less readable.
Up to you

src/styles/theme.ts
```diff
@@ -0,0 +1,201 @@
+const white = '#FFF'
+const blue = '#218DFF'
+const blueDark = '#0b66c6'
+const bgLight = '#edf2f7'
+const bgDark = '#2e2e2e'
+const blackLight = '#181a1b'
+// const black = '#000'
```
Velenir, line 7: ?

src/components/common/Button/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // Handle button theming
+  button: {
+    bg1: !$alt ? theme.button[MAIN].bg1 : theme.button[SECONDARY].bg1,
+    bg1Hover: !$alt ? theme.button[MAIN].bg1Hover : theme.button[SECONDARY].bg1Hover,
+    text1: !$alt ? theme.button[MAIN].text1 : theme.button[SECONDARY].text1,
+    text1Hover: !$alt ? theme.button[MAIN].text1Hover : theme.button[SECONDARY].text1Hover,
+  },
```
Velenir, line 14: Maybe do
```ts
const {bg1, bg1Hover, text1, text1Hover} = theme.button[$alt ? SECONDARY : MAIN]
return {button: {bg1, bg1Hover, text1, text1Hover}, $border}
// if you don't need to pick only specific fields from theme.button, then just
return {button: theme.button[$alt ? SECONDARY : MAIN], $border}
```

or even
```ts
({ $alt, $border, theme: {[$alt ? SECONDARY : MAIN]: button} }) => ({button, $border})
```

src/components/common/Button/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  font-weight: bold;
+  outline: 0;
+  padding: 0.5rem 1rem;
+
+  transition: all 0.2s ease-in-out;
+  transition-property: color, background-color, border-color, opacity;
+
+  &.cancel {
+    background: transparent;
+    color: var(--color-text-active);
```
Velenir, line 40: I'm really not a fan of how we mix styled-comp theme and css vars theme, but whatever

src/storybook/decorators.tsx
```diff
@@ -8,6 +8,31 @@ import { ApolloProvider } from '@apollo/client'
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faMoon, faSun } from '@fortawesome/free-regular-svg-icons'
+
+import { ThemeProvider } from 'styled-components'
+
+export const DarkModeThemeToggler = (DecoratedStory: () => JSX.Element): JSX.Element => {
+  const [darkMode, setDarkMode] = React.useState(false)
+  const theme = colors(darkMode)
+
+  const handleDarkMode = React.useCallback(() => setDarkMode((darkMode) => !darkMode), [])
```
Velenir, line 22: No need to memoize here, as you pass it not to a custom Component but to `<button>` which rerenders anyway all the time

Velenir, line 22: Wanna extract this into a hook?

src/styles/theme.ts
```diff
@@ -0,0 +1,201 @@
...
+          | {
+              [prop: string]: string
+            }
+          | string
+          | undefined
+      }
+    | string
+}
+
+export type ColourTheme = ColourThemeBase & {
```
Velenir, line 27: Any reason you chose `type &` instead of `interface extends`? Not that it matters

W3stside:
@Velenir thanks, will review. Thoughts on theme data structure? Feels like it could be optimised

src/components/common/Button/index.tsx
W3stside: Will do

src/components/common/Button/index.stories.tsx
W3stside: Yeah intrinsic. Not necessary to be able to edit so removed. Theme works as an object but isn't useful enough. In a perfect world I would love a way to edit the theme obj prop in a cleaner way like the `{ control: 'color' }` would work with a flat `backgroundColor` prop.

src/styles/theme.ts
W3stside: Yes 😊

src/styles/theme.ts
W3stside: Yeah I opted for prop string as a more readable choice

src/styles/theme.ts
W3stside: Nyeh

src/components/common/Button/index.tsx
W3stside: Hmm ok looks cleaner thx

src/components/common/Button/index.tsx
W3stside: Me neither, will rework for prop based logic 🤙

src/styles/theme.ts
W3stside: Not really

src/storybook/decorators.tsx
W3stside: Yes sir 🤛

W3stside: Ah ok, cool thx

W3stside: Ah ok, cool thx

src/components/common/Button/index.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  },
+} as Meta
+
+const Template: Story = (args) => <ButtonBase {...args}>{args.label}</ButtonBase>
```
Velenir, line 25: I would expect `Story<ButtonProps>` here, like in other stories

Velenir: For the fun of it if you are having trouble extracting props from StyledComponent, try inference
```ts
type PropsFromStyled<T extends StyledComponent<any,any>> = T extends StyledComponent<React.FC<infer U>, any> ? U : T

type ButtonProps = PropsFromStyled<typeof ButtonBase>
//or
type PropsFromFC<T extends React.FC> = T extends React.FC<infer U> ? U : T

Comment (W3stside) [reply src/components/common/Button/index.stories.tsx:25]: when i can't understand the types, i dont do inference :D

File: src/styles/theme.ts [hunk 11]
```diff
@@ -0,0 +1,201 @@
...
+  button: {
+    main: {
+      text1: blue,
+      text1Hover: white,
+      get bg1(): string {
+        return darkTheme.bg1
+      },
+      bg1Hover: blueDark,
+    },
+    secondary: {
```
Velenir, line 43: `main and secondary` end up the same, don't they?

W3stside: wdym?

Velenir: ignore me, I was looking at something different

src/storybook/decorators.tsx
```diff
@@ -8,6 +8,31 @@ import { ApolloProvider } from '@apollo/client'
...
+export const DarkModeThemeToggler = (DecoratedStory: () => JSX.Element): JSX.Element => {
+  const [darkMode, setDarkMode] = React.useState(false)
+  const theme = colors(darkMode)
+
+  const handleDarkMode = (): void => setDarkMode((darkMode) => !darkMode)
+
+  return (
+    <>
+      <ThemeProvider theme={theme}>
+        <Frame style={{ backgroundColor: 'lightgray' }}>{DecoratedStory()}</Frame>
```
anxolin, line 27: I wouldn't add this gray frame. 
I think, we can add a second decorator for the frame and leave the default color.

src/storybook/decorators.tsx
```diff
@@ -8,6 +8,31 @@ import { ApolloProvider } from '@apollo/client'
...
+import { ThemeProvider } from 'styled-components'
+
+export const DarkModeThemeToggler = (DecoratedStory: () => JSX.Element): JSX.Element => {
+  const [darkMode, setDarkMode] = React.useState(false)
+  const theme = colors(darkMode)
+
+  const handleDarkMode = (): void => setDarkMode((darkMode) => !darkMode)
+
+  return (
+    <>
```
anxolin, line 25: I think the background of the whole story should be the background of the app

src/styles/theme.ts
```diff
@@ -0,0 +1,198 @@
...
+  },
+}
+
+export const colors = (darkMode: boolean): ColourTheme => {
+  const theme = darkMode ? darkTheme : lightTheme
+
+  return theme
+}
+
+// FOR REFERENCE
```
anxolin, line 74: ok, but this is from v1?

not sure if we will need it even as a reference. We have the colors in the original design. 
what do you think @biocom

biocom: For the trading interface, I currently have these colors/variables defined here https://github.com/gnosis/dex-react/blob/0f9feb35788877e80aebbfe3027a8b56f3afbe83/src/components/layout/GenericLayout/variablesCss.ts

biocom: Not sure if we want to consolidate theming between Simple Swap <-> Trading Interface. But I'd say in terms of colors they likely 90% overlap.

W3stside: cool @biocom this is what i needed. will transition my latest pr #1626

src/storybook/decorators.tsx
W3stside: good idea. i'll make these changes in whichever PR is preferred #1624 or #1626

W3stside:
closing in favour for the BIGGER, the BETTER, the SEXIER, #1626 and co.

PR #1625, opened 2020-11-17
Playwright wallet connection test

Velenir:
- Adds Wallet injection into playwright browser (`window.ethereum`)
- Compilation for browser injects step when needed
- Integration test for connecting user wallet

To see it perform and to test/debug please run
```sh
#serve the app
serve -s -l 8080 dist
#in a different terminal
yarn playwright:debug --testNamePattern=wallet
```

Actually please just run that so we can check it works on different machines

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "cypress": "cypress open",
     "cypress:run": "cypress run",
     "playwright": "jest --roots playwright --no-cache",
+    "playwright:rebuild": "ts-node ./playwright/utils/build_injects.ts",
     "playwright:debug": "PWDEBUG=1 jest --roots playwright",
     "playwright:record": "start-server-and-test start:dist http://localhost:8081 'playwright-cli codegen http://localhost:8081 -o output.js'",
     "test:playwright": "BASE_URL=http://localhost:8081 start-server-and-test start:dist http://localhost:8081 playwright"
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "@storybook/addon-knobs": "^6.0.21",
     "@storybook/addon-links": "^6.0.21",
     "@storybook/react": "^6.0.21",
+    "@truffle/hdwallet-provider": "^1.2.0",
     "@types/bn.js": "^4.11.6",
     "@types/combine-reducers": "^1.0.0",
     "@types/enzyme": "^3.10.4",
--- a/package.json
+++ b/package.json
@@ -166,6 +166,7 @@
     "stylelint-prettier": "^1.1.2",
     "ts-jest": "^26.4.1",
     "ts-loader": "^8.0.3",
+    "ts-node": "^9.0.0",
     "typescript": "^4.0.3",
     "webpack": "^4.41.6",
     "webpack-bundle-analyzer": "^3.6.0",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "check-types": "tsc --noEmit --incremental",
     "cypress": "cypress open",
     "cypress:run": "cypress run",
-    "playwright": "jest --roots playwright --no-cache",
+    "playwright": "jest --roots playwright",
     "playwright:rebuild": "ts-node ./playwright/utils/build_injects.ts",
     "playwright:debug": "PWDEBUG=1 jest --roots playwright --runInBand",
     "playwright:record": "start-server-and-test start:dist http://localhost:8081 'playwright-cli codegen http://localhost:8081 -o output.js'",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "cypress:run": "cypress run",
     "playwright": "jest --roots playwright --no-cache",
     "playwright:rebuild": "ts-node ./playwright/utils/build_injects.ts",
-    "playwright:debug": "PWDEBUG=1 jest --roots playwright",
+    "playwright:debug": "PWDEBUG=1 jest --roots playwright --runInBand",
     "playwright:record": "start-server-and-test start:dist http://localhost:8081 'playwright-cli codegen http://localhost:8081 -o output.js'",
     "test:playwright": "BASE_URL=http://localhost:8081 start-server-and-test start:dist http://localhost:8081 playwright"
   },
--- a/playwright/integration/connect_wallet.test.ts
+++ b/playwright/integration/connect_wallet.test.ts
@@ -1,0 +1,24 @@
+import { page, baseURL } from '../utils/test_setup'
+import { account } from '../utils/provider_setup'
+
+describe('Connect Wallet', () => {
+  it('Can connect a Wallet', async () => {
+    // GIVEN: At home page
+    await page.goto(baseURL)
+
+    // WHEN: connected a Wallet
+    // Click text=/.*Connect Wallet.*/
+    await page.click('text=/.*Connect Wallet.*/')
+
+    await page.click('text=Web3')
+
+    // WHEN: Wallet popup is open
+    await page.click('header >> text=Web3')
+
+    // THEN: wallet connected & address available
+    //                          XPATH:    closest div ancestor of element with text "Copy address to clipboard"
+    const textWithAddress = await page.textContent('//*[text()="Copy address to clipboard"]//ancestor::div[1]')
+
+    expect(textWithAddress).toContain(account.address)
+  })
+})
--- a/playwright/integration/connect_wallet.test.ts
+++ b/playwright/integration/connect_wallet.test.ts
@@ -1,5 +1,12 @@
-import { page, baseURL } from '../utils/test_setup'
-import { account } from '../utils/provider_setup'
+import { baseURL } from '../utils/test_setup'
+import { account, setupProvider } from '../utils/provider_setup'
+import { Page } from 'playwright'
+
+let page: Page
+
+setupProvider((variables) => {
+  ;({ page } = variables)
+})
 
 describe('Connect Wallet', () => {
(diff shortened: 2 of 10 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1625--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1625--dexreact.review.gnosisdev.com/storybook/)**: Component stories

playwright/integration/connect_wallet.test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { account } from '../utils/provider_setup'
+
+describe('Connect Wallet', () => {
+  it('Can connect a Wallet', async () => {
+    // GIVEN: At home page
+    await page.goto(baseURL)
+
+    // WHEN: connected a Wallet
+    // Click text=/.*Connect Wallet.*/
+    await page.click('text=/.*Connect Wallet.*/')
```
biocom, line 11: This doesn't look like a specific (HTML) selector but rather (regex?) looking for a specific text string. Unless I misunderstand this piece, is it better to scope down to the selectors so that their presence is validated too?

Velenir: The idea is to do testing from user perspective
`page.click('text=/.*Connect Wallet.*/')` represents
1. User comes to a page
2. User sees text, not a selector, not even necessarily link or button
3. User doesn't care, text looks clickable
4. User clicks

That's what [playwright docs recommend](https://playwright.dev/#version=v1.6.1&path=docs%2Fselectors.md&q=best-practices) and what some other libraries (like react-testing-lib) advocate for

Velenir: Selector presence is for unit (not even integration) testing with libs that like comparing snapshots, like Enzyme
Look up Enzyme vs react-testing-library approaches

biocom: Does that open up possible collusions/complexity around identical naming conventions for text/labels? Or do you see that as an advantage in terms of enforcing a better user experience for the user? I like the thinking here otherwise, but curious on your thoughts.

playwright/integration/connect_wallet.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    await page.click('text=Web3')
+
+    // WHEN: Wallet popup is open
+    await page.click('header >> text=Web3')
+
+    // THEN: wallet connected & address available
+    // XPATH: closest ancestor with 0x* text of element with text "Copy address to clipboard"
+    const textWithAddress = await page.textContent(
+      '//*[text()="Copy address to clipboard"]//ancestor::*[starts-with(., "0x")][1]',
+    )
```
Velenir, line 22: This is a good example of playwright approach.
As there aren't better ways to distinguish where the address is, instead of `<wallet popup class which is pseudo-random with hash 'cause generated by styled-comps> > text that looks like address`
I opted for `Copy address to clipboard button > closest parent with address`
dex dev gnosisdev com_

playwright/integration/connect_wallet.test.ts
biocom: Although after reading the docs, I wonder if using an actual user-facing attribute like 'aria-label' or 'title' (etc.) would be better. It would thick both the 'selector' and 'user facing attribute/text' boxes.

playwright/integration/connect_wallet.test.ts
biocom: I think this is quite thorough, which is good. As opposed to testing for selectors.

What if there a multiple of these components doing the same thing for some reason? page.Textcontent() only returns the first found item? E.g. would be good to test the validity of any other similar/identical item on page.

playwright/integration/connect_wallet.test.ts
Velenir: Yes, it does, you are completely right. Trade offs are trade offs
That's why I use `header >> text=Web3` in another click, for specificity

You can even argue, as you are saying, for better user xp. For example we have `Connect Wallet` both in header and in TradeOrders component (when disconnected). But they do the same, so we shouldn't care which user clicks

dex dev gnosisdev com_ (1)
dex dev gnosisdev com_ (2)

playwright/integration/connect_wallet.test.ts
Velenir: Very well may be
Like [here](https://github.com/gnosis/dex-react/pull/1625#discussion_r524995607)

playwright/integration/connect_wallet.test.ts
biocom: But the test will only click the first found item? E.g. I guess we want to be sure to test all possible instance for working properly as expected?

Velenir: Yep
We want to be sure of the whole user flow
So to properly test we would need to do txs
And not rely on rinkeby random address
So setup ganache and deploy contracts locally and who know what else

So yes,and no :wink:

biocom: Makes sense.

anxolin (commented):
Do the test take so much time to you?
image


Also, do they fail, I get this error?

image

package.json
```diff
@@ -31,7 +31,8 @@
     "cypress": "cypress open",
     "cypress:run": "cypress run",
     "playwright": "jest --roots playwright --no-cache",
-    "playwright:debug": "PWDEBUG=1 jest --roots playwright",
+    "playwright:rebuild": "ts-node ./playwright/utils/build_injects.ts",
+    "playwright:debug": "PWDEBUG=1 jest --roots playwright --runInBand",
```
anxolin, line 35: Good call creating this script

Velenir:
> Is it possible is a bit flaky? Why do they’ll take so much time?

This is strange. How did you run? what command?
@biocom , @W3stside can you also try
First `yarn test:playwright` for the whole test
After
```
#serve the app
serve -s -l 8080 dist
#in a different terminal
yarn playwright:debug --testNamePattern=wallet
```

playwright/utils/test_setup.ts
```diff
@@ -0,0 +1,32 @@
...
+  slowMo = 2000
+} else {
+  // close after only when not debugging
+  afterAll(() => browser.close())
+  afterEach(() => page.close())
+}
+
+beforeAll(async () => {
+  // launch chrome,headless by default
+  browser = await chromium.launch({ slowMo })
```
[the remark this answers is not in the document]
Velenir: we have to hardcode something
can of course do `BROWSER=chromium|firefox|webkit` and later pick it

anxolin:
With `test:playwrite` is taking less time 8s and succeeding 

image

The debug worked too now

playwright/utils/test_setup.ts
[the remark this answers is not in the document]
biocom: Also can think about an optional `executablePath` argument, to use Chromium but use your local installed Chrome browser. Should in theory be the same I suppose (Chromium <-> Chrome) but there can be nuances in practice when testing.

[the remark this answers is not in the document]
Velenir: Nah, can't count on Chrome extensions not screwing everything up
Because if you just use local Chrome, and have MMask installed, it will conflict with `window.ethereum` injected by playwright
And if you use local Cjrome with `--sandbox` or whatever no-extension flag, then it's almost the same as Chromium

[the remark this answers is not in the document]
biocom: I guess my thinking is:
- Even having 'expected' extensions like MM, might simulate/catch some side effects/bugs as opposed to run it in a clean env. like Chromium.
- Same for running Chromium vs Chrome, like you say they are almost the same but I'd regard it as a more realistic test if you know what I mean.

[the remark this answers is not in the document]
Velenir: Now this is interesting indeed. Can do
```sh
EXECUTABLE_PATH=$(which google-chrome) yarn test:playwright
```

biocom (approved):

Velenir:
Added `BROWSER and EXECUTION_PATH` env vars.
Would be cool if you tried
```sh
serve -s -l 8080 dist

BROWSER=chromium EXECUTABLE_PATH=$(which google-chrome) yarn test:playwright #or brave browser
BROWSER=firefox yarn test:playwright
BROWSER=webkit yarn test:playwright
```
`webkit` doesn't work for me, though

W3stside (approved):

playwright/utils/test_setup.ts
```diff
@@ -0,0 +1,60 @@
...
+// PWDEBUG is playwright specific flag
+// sets up window.playwright and other stuff
+const isDebug = !!process.env.PWDEBUG
+let slowMo: number | undefined
+
+// timeout to have time to account for async nature of browser actions
+jest.setTimeout(30000)
+if (isDebug) {
+  // slow down playright actions
+  slowMo = 2000
```
anxolin, line 45: Maybe add a const on top, and explain that artificially delay the test to visually see what is going on

anxolin (approved):
Very good job.
The injection was a bit hacky, I mean. programmatically using webpack to compile the wallet, ... 

I haven't gone over all the doc, so not sure if there's simpler way. I would think so, but just glad u found a solution.

playwright/integration/connect_wallet.test.ts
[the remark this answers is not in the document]
Velenir: This pattern was made by playwright itself during `playwright:record`, I just didn't think twice to change it

PR #1628, opened 2020-11-18
Fix issue gas estimation

anxolin:
Correcting what was obvious an issue, since in PR hotfix #1618 the `data` was ignored to solve an issue with the hardfork last week.  Only the fields from this method were added https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#methods-mymethod-estimategas, we should have used https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html?highlight=estimateGas#estimategas

Actually, the fix removes the `gasPrice` that breaks the estimation for approving tokens for Anna after last hardfork, although my understanding it should be fine to add them according to JSON-rpc spec

https://eth.wiki/json-rpc/API

image

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
     'react/prop-types': [1, { skipUndeclared: true }],
     'react-hooks/rules-of-hooks': 'error',
     'react-hooks/exhaustive-deps': 'warn',
+    '@typescript-eslint/no-unused-vars': ['error', { ignoreRestSiblings: true }],
   },
   overrides: [
     {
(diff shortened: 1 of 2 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1628--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1628--dexreact.review.gnosisdev.com/storybook/)**: Component stories

Velenir (approved):
Works for me :+1:

W3stside (approved):
looks good.

PR #1629, opened 2020-11-18
V2 Tabs comments

W3stside:
@biocom It's not major changes but changes that should have been included. 

Nice work on this stuff btw, just make sure to get 2 approvals eh? :)

### DIFF
--- a/cypress/support/index.d.ts
+++ b/cypress/support/index.d.ts
@@ -1,7 +1,7 @@
 /// <reference types="cypress" />
 
 declare namespace Cypress {
-  interface Chainable {
+  export interface Chainable {
     /**
      * Substitute for "visit" command. The original visit command should not be used, when the tests depend on the
      * scroll, because cypress has an issue with "scroll-behaviour: smooth
--- a/src/components/common/Tabs/TabContent.tsx
+++ b/src/components/common/Tabs/TabContent.tsx
@@ -1,17 +1,9 @@
 import React from 'react'
-import { TabItemType } from 'components/OrderBuySell'
 
 type Props = {
-  tabItems: TabItemType[]
-  activeTab: number
+  content?: string | React.ReactNode
 }
 
-const TabContent: React.FC<Props> = ({ tabItems, activeTab }) => (
-  <div>
-    {tabItems.map((tab) => {
-      return tab.id === activeTab ? tab.content : null
-    })}
-  </div>
-)
+const TabContent: React.FC<Props> = ({ content }) => <div>{content}</div>
 
 export default TabContent
--- a/src/components/common/Tabs/TabContent.tsx
+++ b/src/components/common/Tabs/TabContent.tsx
@@ -6,15 +6,12 @@ type Props = {
   activeTab: number
 }
 
-const TabContent: React.FC<Props> = (props) => {
-  const { tabItems, activeTab } = props
-  return (
-    <div>
-      {tabItems.map((tab) => {
-        return tab.id === activeTab ? tab.content : null
-      })}
-    </div>
-  )
-}
+const TabContent: React.FC<Props> = ({ tabItems, activeTab }) => (
+  <div>
+    {tabItems.map((tab) => {
+      return tab.id === activeTab ? tab.content : null
+    })}
+  </div>
+)
 
 export default TabContent
--- a/src/components/common/Tabs/TabItem.tsx
+++ b/src/components/common/Tabs/TabItem.tsx
@@ -33,7 +33,7 @@ const Wrapper = styled.li<Pick<TabProps, 'activeColor'>>`
 `
 
 const TabItem: React.FC<TabProps> = ({ onTabClick, title, activeColor }) => (
-  <Wrapper activeColor={activeColor} onClick={onTabClick}>
+  <Wrapper role="tab" aria-selected={!!activeColor} activeColor={activeColor} onClick={onTabClick}>
     {title}
   </Wrapper>
 )
--- a/src/components/common/Tabs/TabItem.tsx
+++ b/src/components/common/Tabs/TabItem.tsx
@@ -7,18 +7,24 @@ interface TabProps {
   readonly activeColor: string | false
 }
 
-const Wrapper = styled.li<Pick<TabProps, 'activeColor'>>`
+const Wrapper = styled.button<Pick<TabProps, 'activeColor'>>`
   background: ${({ activeColor }): string => `var(${activeColor || '--color-primary'})`};
   color: ${({ activeColor }): string => `var(--color-text-${activeColor ? 'primary' : 'secondary2'})`};
   height: var(--height-button-default);
   display: flex;
   align-items: center;
+  justify-content: center;
   flex: 1 1 0;
+
   font-weight: var(--font-weight-bold);
   font-size: var(--font-size-large);
   letter-spacing: 0.1rem;
   text-align: center;
-  justify-content: center;
+
+  appearance: none;
+  border: 0;
+  outline: 0;
+
   cursor: pointer;
 
   &:first-of-type {
--- a/src/components/common/Tabs/TabItem.tsx
+++ b/src/components/common/Tabs/TabItem.tsx
@@ -3,20 +3,13 @@ import styled from 'styled-components'
 
 interface TabProps {
   title: string
-  readonly id: number
-  onTabClick: (arg: number) => void
-  isActive: boolean
-  readonly activeColor?: string
+  onTabClick: () => void
+  readonly activeColor: string | false
 }
 
-interface WrapperProps {
-  readonly activeColor?: string
-  isActive: boolean
-}
-
-const Wrapper = styled.li<WrapperProps>`
-  background: ${({ isActive, activeColor }): string => `var(${isActive ? activeColor : '--color-primary'})`};
-  color: ${({ isActive }): string => `var(--color-text-${isActive ? 'primary' : 'secondary2'})`};
+const Wrapper = styled.li<Pick<TabProps, 'activeColor'>>`
+  background: ${({ activeColor }): string => `var(${activeColor || '--color-primary'})`};
+  color: ${({ activeColor }): string => `var(--color-text-${activeColor ? 'primary' : 'secondary2'})`};
   height: var(--height-button-default);
   display: flex;
   align-items: center;
@@ -39,14 +32,10 @@ const Wrapper = styled.li<WrapperProps>`
   }
 `
 
-const TabItem: React.FC<TabProps> = (props) => {
-  const { onTabClick, id, title, isActive, activeColor } = props
-
-  return (
-    <Wrapper activeColor={activeColor} isActive={isActive} onClick={(): void => onTabClick(id)}>
-      {title}
-    </Wrapper>
-  )
-}
+const TabItem: React.FC<TabProps> = ({ onTabClick, title, activeColor }) => (
+  <Wrapper activeColor={activeColor} onClick={onTabClick}>
+    {title}
+  </Wrapper>
+)
 
 export default TabItem
--- a/src/components/common/Tabs/Tabs.tsx
+++ b/src/components/common/Tabs/Tabs.tsx
@@ -31,7 +31,7 @@ const Tabs: React.FC<Props> = ({ tabItems }) => {
 
   return (
     <Wrapper>
-      <ul>
+      <div role="tablist">
         {tabItems.map(({ title, activeColor }, index) => (
           <TabItem
             key={index}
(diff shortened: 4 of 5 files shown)


cypress/support/index.d.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference types="cypress" />
 
 declare namespace Cypress {
-  interface Chainable {
+  export interface Chainable {
```
W3stside, line 4: fixes TS problem with Chainable

src/components/common/Tabs/TabItem.tsx
```diff
@@ -39,14 +32,10 @@ const Wrapper = styled.li<WrapperProps>`
...
-  const { onTabClick, id, title, isActive, activeColor } = props
-
-  return (
-    <Wrapper activeColor={activeColor} isActive={isActive} onClick={(): void => onTabClick(id)}>
-      {title}
-    </Wrapper>
-  )
-}
+const TabItem: React.FC<TabProps> = ({ onTabClick, title, activeColor }) => (
+  <Wrapper activeColor={activeColor} onClick={onTabClick}>
```
W3stside, line 36: this is the most important one. Look at how you made the fn in your changes and see what's different with mine and what that means for props etc

src/components/common/Tabs/TabContent.tsx
```diff
@@ -6,15 +6,12 @@ type Props = {
...
-  return (
-    <div>
-      {tabItems.map((tab) => {
-        return tab.id === activeTab ? tab.content : null
-      })}
-    </div>
-  )
-}
+const TabContent: React.FC<Props> = ({ tabItems, activeTab }) => (
+  <div>
```
biocom, line 10: @W3stside  I see:
- Simple destructuring of props already in the function arguments
- Implicit return value (no need for `return`) 

Thank you.

src/components/common/Tabs/TabItem.tsx
```diff
@@ -3,20 +3,13 @@ import styled from 'styled-components'
 
 interface TabProps {
   title: string
-  readonly id: number
-  onTabClick: (arg: number) => void
-  isActive: boolean
-  readonly activeColor?: string
+  onTabClick: () => void
+  readonly activeColor: string | false
```
biocom, line 7: I see that writing 
`readonly activeColor: string | false`
is more explicit than doing:
`readonly activeColor?: string`

Because in the latter, if you don't pass anything, we wouldn't know the type?

Now if no `activeColor` value is passed, it falls back to being `false`.

src/components/common/Tabs/TabItem.tsx
```diff
@@ -3,20 +3,13 @@ import styled from 'styled-components'
...
 
-interface WrapperProps {
-  readonly activeColor?: string
-  isActive: boolean
-}
-
-const Wrapper = styled.li<WrapperProps>`
-  background: ${({ isActive, activeColor }): string => `var(${isActive ? activeColor : '--color-primary'})`};
-  color: ${({ isActive }): string => `var(--color-text-${isActive ? 'primary' : 'secondary2'})`};
+const Wrapper = styled.li<Pick<TabProps, 'activeColor'>>`
```
biocom, line 10: Here we use "Pick" to only grab certain types from TabProps?
`Pick<TabProps, 'activeColor'>`

gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1629--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1629--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/Tabs/TabItem.tsx
biocom: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

W3stside: if i changed only `TabProps` i could forget to change the Wrapper one too so this is safer

src/components/common/Tabs/TabItem.tsx
```diff
@@ -3,20 +3,13 @@ import styled from 'styled-components'
...
+  background: ${({ activeColor }): string => `var(${activeColor || '--color-primary'})`};
```
biocom, line 11: Here we destructure `activeColor` from `props` correct?

Do we need the $ in front because of it being standard interpolated template literals?

src/components/common/Tabs/TabContent.tsx
W3stside: sure. we don't always do it this way and it is actually easier for me to when testing to not use lamda return, but near the end of the pr cycle if you see it doesnt need the curly brace return pattern as it's not calcualting anything in the fn body then it's always nice to clean it up this way. Anyways this part was nothing important more of a reminder that desrtructuring is also common in the fn signature

src/components/common/Tabs/Tabs.tsx
```diff
@@ -25,24 +25,22 @@ const Wrapper = styled.div`
   }
 `
 
-const Tabs: React.FC<Props> = (props) => {
-  const [activeTab, setActiveTab] = useState<number>(1)
+const Tabs: React.FC<Props> = ({ tabItems }) => {
```
biocom, line 28: Destructure `tabItems` from props right in the argument passing.

src/components/common/Tabs/TabItem.tsx
W3stside: i didn't see a scenario where we would leave out the `activeColor` - and the pattern `someTruthyValue && valueIfTruthy` is common enough. To make it agree with TS we need to type it this way. You could also always do `someTruthyValue ? valueIfTruthy : ' '` if that's more clear. Then you could leave the type as `string | undefined` - but the base type where activeColor came from (i think Tabs?) had `activeColor` set as `string | undefined` and i saw no such scenario where `activeColor` would ever be undefined

src/components/common/Tabs/TabItem.tsx
W3stside: > Do we need the $ in front because of it being standard interpolated template literals?
do you mean in the variable name? like `$activeColor`? Normally yes we would name any prop that is passed into a `styled-component` this way but with the way we were passing it I didnt change it. Maybe not necessary for this project.

src/components/common/Tabs/Tabs.tsx
```diff
@@ -25,24 +25,22 @@ const Wrapper = styled.div`
...
+  const [activeTab, setActiveTab] = useState(1)
```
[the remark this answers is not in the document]
W3stside: but if it's easier for you, feel free to leave it. Just showing you it's a possibility

src/components/common/Tabs/TabItem.tsx
W3stside: > Here we destructure `activeColor` from `props` correct?
> 
> Do we need the $ in front because of it being standard interpolated template literals?

also dropped `isActive` entirely

src/components/common/Tabs/TabItem.tsx
biocom: I see:
- Destructuring props right from the argument passing:
- `isActive` not required to pass, given we can use `activeTab`. 
- onTabClick is already passed down as a prop, so no need to use an arrow fn with the `id` as an argument (already done in Tabs.tsx)

src/components/common/Tabs/Tabs.tsx
```diff
@@ -25,24 +25,22 @@ const Wrapper = styled.div`
...
-        {props.tabItems.map(({ id, title, activeColor }) => (
+        {tabItems.map(({ id, title, activeColor }) => (
           <TabItem
             key={id}
-            id={id}
             title={title}
             onTabClick={(): void => setActiveTab(id)}
-            isActive={activeTab === id}
-            activeColor={activeColor}
+            activeColor={activeTab === id && activeColor}
```
biocom, line 39: I wonder:
- We conditionally set the activeColor here? If activeColor has no value/false then this won't pass?
Right now if no `activeColor` is passed for a tabitem, it won't have a color at all. So likely need to add logic to fallback to a 'default'.

Btw that was the reason I wanted to get an 'isActive' to conditionally apply styles. But I understand we can still do this, but without having to pass a prop for it.

biocom: Ah it already falls back anyhow in the styles after double checking. All good.

src/components/common/Tabs/TabItem.tsx
W3stside: yes last point is the important one

src/components/common/Tabs/TabContent.tsx
biocom: Really appreciate it. Yes the way I understand it, we are defining a component here and we can assume it's implied that this will be the return value when _importing_ it elsewhere.

src/components/common/Tabs/TabItem.tsx
biocom: Thinking about it as a 'generic' Tabs component, the only scenario I can think of for not wanting to pass the activeColor, is if you want to use the default colors. Which...seems to me to be a quite common scenario?

E.g. in this case we use Tabs to have the Green/Red (buy/sell) appearance. But say, if you have a 'normal' Tab layout, you'd want to just use the default colors (not Green/Red).

src/components/common/Tabs/TabItem.tsx
biocom: I see, so that makes `Pick` superior here, given I don't have to have a 'duplicate' type interface.

src/components/common/Tabs/TabItem.tsx
W3stside: I think the best approach depends on how you want to re-use the components. One way would be to use theme colours with alternatives. Then use an `isActive` prop that toggles the highlight colour internally OR just pass an `activeColour` prop that sets it automically, overwriting the original theme colour. But we didn't need both in this case

W3stside:
I didnt do comprehensive testing but it seemed to work as it was supposed to @biocom - def check if I didnt fuck anything up haha

src/components/common/Tabs/TabContent.tsx
```diff
@@ -6,15 +6,12 @@ type Props = {
...
+    {tabItems.map((tab) => {
+      return tab.id === activeTab ? tab.content : null
+    })}
```
Velenir, line 13: `map` loops over the whole array, plus react has to render an array (even if it's mostly filled with nulls) in the end
```suggestion
    {tabItems.find((tab) => tab.id === activeTab)?.content}
```
exits early

Velenir (approved):

src/components/common/Tabs/TabItem.tsx
biocom: To @Velenir 's point, by making it optional, you always rely on the default/fallback theming for that tabs. Would that be a better approach?

I guess @W3stside you were suggesting to just use a default theme color/variable where the logic is handled elsewhere? You feel that's easier to support optional passed activeColor vars?

src/components/common/Tabs/TabItem.tsx
biocom: So in this case if we were to use $activeColor vs. activeColor (as the destructured variable from `props`) that is the same, just a specific naming convention (prepending the $ sign) to make it distinctive for a reason?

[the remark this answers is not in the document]
W3stside: we're going to change this to use colour sheet constants anyways so it doesn't really matter

src/components/common/Tabs/TabItem.tsx
```diff
@@ -3,20 +3,13 @@ import styled from 'styled-components'
...
+  color: ${({ activeColor }): string => `var(--color-text-${activeColor ? 'primary' : 'secondary2'})`};
```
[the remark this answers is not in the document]
W3stside: we're going to change this to use colour sheet constants anyways so it doesn't really matter

src/components/common/Tabs/TabContent.tsx
biocom: Can we commit this suggestion @W3stside ? Good to go otherwise.

biocom (approved):

src/components/common/Tabs/TabContent.tsx
W3stside: @Velenir @biocom I'm more concerned about why this is even mapping in the first place as you're only rendering one element. Just pass the element you need inside as a prop

biocom: Ah so, you're saying it'll be better to pass the activeTab content directly as a prop to this Tab component?

biocom: @W3stside Please review https://github.com/gnosis/dex-react/pull/1634 to address this.

W3stside:
Let's wait on the waterfall PR #1634 before merging - and squash

biocom:
> #1634

Ready to review.

PR #1631, opened 2020-11-19
Simplify text selectors

anxolin:
I just wanted to try out what I wrote in one comment of #1625 

> Why do you use a "pattern", is not needed right? (since text selector already searches for partial text matching)
> Also, I think it's more readable the short-form

My understanding is this PR has equivalent selectors


Velenir:
I don't understand what you changed. This is my PR verbatim, but directed at develop, skipping waterfall

Velenir (approved):
If this works (and it should), I'm ok with this change

PR #1634, opened 2020-11-19
V2 Tabs - Separate PR addressing comments

biocom:
V2 Tabs - Separate PR addressing comments

Addresses comments left in #1629 also waterfalls in

### DIFF
--- a/src/components/OrderBuySell/index.tsx
+++ b/src/components/OrderBuySell/index.tsx
@@ -5,8 +5,7 @@ import Tabs from 'components/common/Tabs/Tabs'
 export interface TabItemType {
   readonly id: number
   title: string
-  // content?: string | undefined
-  content?: any
+  content?: string | undefined
   readonly activeColor: string
 }
 
--- a/src/components/OrderBuySell/index.tsx
+++ b/src/components/OrderBuySell/index.tsx
@@ -5,7 +5,7 @@ import Tabs from 'components/common/Tabs/Tabs'
 export interface TabItemType {
   readonly id: number
   title: string
-  content: string
+  content?: string | undefined
   readonly activeColor: string
 }
 
--- a/src/components/OrderBuySell/index.tsx
+++ b/src/components/OrderBuySell/index.tsx
@@ -5,7 +5,7 @@ import Tabs from 'components/common/Tabs/Tabs'
 export interface TabItemType {
   readonly id: number
   title: string
-  content?: string | undefined
+  content?: string | React.ReactNode
   readonly activeColor: string
 }
 
--- a/src/components/OrderBuySell/index.tsx
+++ b/src/components/OrderBuySell/index.tsx
@@ -3,7 +3,6 @@ import styled from 'styled-components'
 import Tabs from 'components/common/Tabs/Tabs'
 
 export interface TabItemType {
-  readonly id: number
   title: string
   content?: string | React.ReactNode
   readonly activeColor: string
@@ -11,13 +10,11 @@ export interface TabItemType {
 
 const tabItems: TabItemType[] = [
   {
-    id: 1,
     title: 'BUY',
     content: '- buy component -',
     activeColor: '--color-buy',
   },
   {
-    id: 2,
     title: 'SELL',
     content: '- sell component -',
     activeColor: '--color-sell',
--- a/src/components/OrderBuySell/index.tsx
+++ b/src/components/OrderBuySell/index.tsx
@@ -5,7 +5,8 @@ import Tabs from 'components/common/Tabs/Tabs'
 export interface TabItemType {
   readonly id: number
   title: string
-  content?: string | undefined
+  // content?: string | undefined
+  content?: any
   readonly activeColor: string
 }
 
--- a/src/components/common/Tabs/TabContent.tsx
+++ b/src/components/common/Tabs/TabContent.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 
 type Props = {
-  content?: string | undefined
+  content?: string | React.ReactNode
 }
 
 const TabContent: React.FC<Props> = ({ content }) => <div>{content}</div>
--- a/src/components/common/Tabs/TabContent.tsx
+++ b/src/components/common/Tabs/TabContent.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 
 type Props = {
-  content: string
+  content?: string | undefined
 }
 
 const TabContent: React.FC<Props> = ({ content }) => <div>{content}</div>
--- a/src/components/common/Tabs/TabContent.tsx
+++ b/src/components/common/Tabs/TabContent.tsx
@@ -1,17 +1,9 @@
 import React from 'react'
-import { TabItemType } from 'components/OrderBuySell'
 
 type Props = {
-  tabItems: TabItemType[]
-  activeTab: number
+  content: string
 }
 
-const TabContent: React.FC<Props> = ({ tabItems, activeTab }) => (
-  <div>
-    {tabItems.map((tab) => {
-      return tab.id === activeTab ? tab.content : null
-    })}
-  </div>
-)
+const TabContent: React.FC<Props> = ({ content }) => <div>{content}</div>
 
 export default TabContent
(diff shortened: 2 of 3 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1634--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1634--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/OrderBuySell/index.tsx
```diff
@@ -5,7 +5,7 @@ import Tabs from 'components/common/Tabs/Tabs'
 export interface TabItemType {
   readonly id: number
   title: string
-  content: string
+  content?: string | undefined
```
W3stside, line 8: `prop:?`: stuff` is the same as `prop: stuff | undefined`

W3stside:
```suggestion
  content?: string | React.ReactNode
```

Should also accept components probably. please test this code doesn't break it

src/components/common/Tabs/TabContent.tsx
```diff
@@ -1,17 +1,9 @@
 import React from 'react'
-import { TabItemType } from 'components/OrderBuySell'
 
 type Props = {
-  tabItems: TabItemType[]
-  activeTab: number
+  content?: string | undefined
```
W3stside, line 4: same thing here

src/components/common/Tabs/Tabs.tsx
```diff
@@ -40,7 +40,7 @@ const Tabs: React.FC<Props> = ({ tabItems }) => {
           />
         ))}
       </ul>
-      <TabContent tabItems={tabItems} activeTab={activeTab} />
+      <TabContent content={tabItems.find((tab) => tab.id === activeTab)?.content} />
```
W3stside, line 43: Array positions are guaranteed, so maybe in this case it's probably simpler to save the tabcontent `index` position and not the ID in local state. Then render it here as:
```tsx
const [activeTab, setActiveTab] = useState(0)
const { content } = useMemo((): TabItemType => tabItems[activeTab], [activeTab])
...
<TabContent content={content} />
...
```

please check this first

W3stside: though tbh, i think this entire thing is over engineered - we only have 2 tabs. No need for mapping etc but it's probably fine for now as it's still small. 

Remember that all these looping things etc, while cool, take computing power and when we can avoid them, we should. Obviously some are faster than others but in this case to simple render tab content on 2 tabs it's over kill to use `Array.prototype.find(...)` and `.map` on a likely tiny array. 

Though I say this and I think that the v1 `useTabs` appraoch i want to reuse is probably similar :sweat_smile:

W3stside (changes requested):

src/components/common/Tabs/Tabs.tsx
W3stside: here is the full rework:
```tsx
const Tabs: React.FC<Props> = ({ tabItems }) => {
  // default to first tab in array position
  const [activeTab, setActiveTab] = useState(0)
  const { content } = useMemo((): TabItemType => tabItems[activeTab], [activeTab])

  return (
    <Wrapper>
      <ul>
        {tabItems.map(({ id, title, activeColor }, index) => (
          <TabItem
            key={id}
            title={title}
            // set activeTab by it's array position
            onTabClick={(): void => setActiveTab(index)}
            activeColor={activeTab === id && activeColor}
          />
        ))}
      </ul>
      <TabContent content={content} />
    </Wrapper>
  )
}
```

W3stside:
@biocom make sure to check it before commiting recommendations:
```json
ERROR in /home/travis/build/gnosis/dex-react/src/components/common/Tabs/Tabs.tsx
ERROR in /home/travis/build/gnosis/dex-react/src/components/common/Tabs/Tabs.tsx(43,19):
TS2322: Type 'ReactNode' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.
```

you can apply the change locally and then run `npx tsc --noEmit` to check errors. Looks like a type clash in this case probaly found in another file where you render the content

src/components/common/Tabs/Tabs.tsx
biocom: @W3stside Refactored it now, to use the index over the passed tabItems array of objects, instead of a specific ID of each Array object.

As for over engineering: I think the goal is to have a general purpose tabs component. In this particular instance these are 2 tabs, but in theory there could be multiple.

W3stside: ok, well then it's fine. but it is just a rehash of [what we already have](https://github.com/gnosis/dex-react/blob/michel-comments/src/hooks/useTabs.tsx) 

Like I mentioned before it's a good working exercise so all good, but let's not reinvent the wheel where we don't need to

src/components/common/Tabs/Tabs.tsx
```diff
@@ -26,21 +26,25 @@ const Wrapper = styled.div`
...
             title={title}
-            onTabClick={(): void => setActiveTab(id)}
-            activeColor={activeTab === id && activeColor}
+            onTabClick={(): void => setActiveTab(index)}
+            activeColor={activeTab === index && activeColor}
           />
         ))}
       </ul>
-      <TabContent content={tabItems.find((tab) => tab.id === activeTab)?.content} />
+      <TabContent content={content && content} />
```
W3stside, line 44: should be:
```
{content && <TabContent content={content && content} />}

src/components/common/Tabs/Tabs.tsx
biocom: I understand. Do we want to replace this with the existing logic?

W3stside: nah, let's do that another time. I think you deserve to merge all this as you busted your ass on it. But eventually either look at how we can re-emulate the logic (if we still like it) into yours. 

Nice work on all this btw :+1:

src/components/OrderBuySell/index.tsx
Velenir: > `prop:?: string` is the same as `prop: string | undefined`

almost
`{props?: string} ===roughly==== {props: string} | {}`
if it's undefined you have to pass `undefined`, [see here](https://www.typescriptlang.org/play?target=99#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChTgA7GJKTdJOANQmABMAFHMAZzgG9ScOBwQB5DgH4AXHB4woVAOakAvqQD0GuDAAWSHoxgB3CDoCeYA3BRE4AaypsUO-XJQgkAGjgAjAK7w6DD+KAA2YeZwlBAwpDCWjCzsXBC8AExwALwCohKy8oqUSqpwAD4C6qRoEJTycADCuGDJbLLIwQB0AGINADytqbwAfNlwABRg3DwAlNmjfWDD-FNpPJ15HKp9GkvVtfVN4K3p7cQwPf2D0+mjOZPTc1kLS-z4m-hwVHCrvHAAZP8ftMNuItjs9vs6vAqPIji1WGwAIJjPrw1oaYZQ+qwmDoxEAIVR+PYIjBWQARDw2BTMdiYdCSWx0iicmjmic6TVoV9GRzEekiWymZlNpTqbSsRRqLR6GhGABVShsJCYKhITjTARCMn5OQKZTlOD+ZWq9VsNTkemNZpKlVqygas5dXp9O1mx2atZ3Ca-WbzOCLZZ+0ESba7KXcnF88Duh0a1mB+Fx82YuBaOC6FAANyMZjAKB4fE21txydN8bYQqTtor5t1HCy-BN9vNqkxQA)

src/components/common/Tabs/Tabs.tsx
W3stside: will happily approve once this is changed :)

biocom: Updated @W3stside . Thank you.

Velenir (approved):

W3stside (approved):

PR #1635, opened 2020-11-19
Some Wave/WebAim recommended fixes.

biocom:


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1635--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1635--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/Tabs/TabItem.tsx
```diff
@@ -25,8 +25,13 @@ const Wrapper = styled.li<WrapperProps>`
   font-size: var(--font-size-large);
   letter-spacing: 0.1rem;
   text-align: center;
+  appearance: none;
+  border: 0;
   justify-content: center;
   cursor: pointer;
+  outline: 0;
+
+  /* TODO: Provide alternative :focus styling because of using outline: 0; */
```
anxolin, line 34: is this sth u intended to do in this PR and forgot, or sth we need to address in the future?
not sure if I understand

anxolin (approved):
it helps with the review if you add some description and/or screenshot if the change is visual

src/components/common/Tabs/TabItem.tsx
biocom: Intended to be addressed in the future. As this would apply to a lot more items.

Velenir (approved):
Where are these fixes recommended? Did you run code through some validator?
If this is because of `lat & aria` attributes, I'm pretty sure we don't have them in a thousand places
If you want them in, that would be one big PR...

biocom:
> Where are these fixes recommended? Did you run code through some validator?
> If this is because of `lat & aria` attributes, I'm pretty sure we don't have them in a thousand places
> If you want them in, that would be one big PR...

Have been using https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh which is the WAVE evaluation tool.

My approach is to just tackle things step by step, instead of doing one big PR. These are recommended changes only for improving accessibility.

src/components/common/Tabs/TabItem.tsx
anxolin: Okay. Maybe here is not worth it. For bigger things, or things we don't want to forget, we add a task as an issue or in the backlog so we don't forget. Although now we are not attending so much issues of this gpv1

W3stside:
While I think this is good, I think we should postpone this for until we have fully implemented `playwright` integratino so that it can test any potential regressions from adding these.

Not saying I think these will break anything (they shouldnt) BUT if we start changing tons of files and we mess sth up that isn't caught from linter or else, we'll be in trouble

W3stside (approved):

biocom:
> While I think this is good, I think we should postpone this for until we have fully implemented `playwright` integratino so that it can test any potential regressions from adding these.
> 
> Not saying I think these will break anything (they shouldnt) BUT if we start changing tons of files and we mess sth up that isn't caught from linter or else, we'll be in trouble

Fair point. As you pointed out, adding validation should be validated HTML by default. But in practice things can happen.

PR #1644, opened 2020-11-28
Add timeout for fetching tokens

anxolin:
We've observed in two occasions now that some Metamask fails to resolve some promise for contract calls. First with the issue of SNX/sUSD proxy deprecation, yesterday with Ledger MM (not it seems it doesn't do that anymore).

When this happens, we neither get a success or an error.

Currently, if for one token, the promise is not resolved, the overall "fetch all balances" promise don't resolve either (uses `Promise.all`)

This PR introduces a timeout for fetching individual tokens for this extreme cases. It uses 10s, maybe it could be less. But theoretically promises should resolve one way or the other!

### DIFF
--- a/src/hooks/useTokenBalances.ts
+++ b/src/hooks/useTokenBalances.ts
@@ -7,7 +7,7 @@ import { erc20Api, depositApi } from 'api'
 import useSafeState from './useSafeState'
 import { useWalletConnection } from './useWalletConnection'
 
-import { formatSmart, logDebug, isTokenEnabled, timeout } from 'utils'
+import { formatSmart, logDebug, isTokenEnabled, timeout, notEmpty } from 'utils'
 import { TokenBalanceDetails, TokenDetails } from 'types'
 import { WalletInfo } from 'api/wallet/WalletApi'
 import { PendingFlux } from 'api/deposit/DepositApi'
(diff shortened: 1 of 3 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1644--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1644--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/utils/miscellaneous.ts
```diff
@@ -210,3 +210,22 @@ export function notEmpty<TValue>(value: TValue | null | undefined): value is TVa
...
+export function timeout<T>(params: TimeoutParams<T>): Promise<T> {
+  const { time = DEFAULT_TIMEOUT, result, timeoutErrorMsg: timeoutMsg = 'Timeout' } = params
+  return new Promise((resolve, rejects) =>
+    setTimeout(() => {
+      if (result) {
+        resolve(result)
+      } else {
+        rejects(new Error(timeoutMsg))
+      }
+    }, time),
```
Velenir, line 229: We already have `delay()` util to do that

src/utils/miscellaneous.ts
```diff
@@ -210,3 +210,22 @@ export function notEmpty<TValue>(value: TValue | null | undefined): value is TVa
...
 
 export const isNonZeroNumber = (value?: string | number): boolean => !!value && !!+value
+
+export interface TimeoutParams<T> {
+  time?: number
+  result?: T
+  timeoutErrorMsg?: string
+}
+
+export function timeout<T>(params: TimeoutParams<T>): Promise<T> {
```
Velenir, line 220: Types are not exactly correct. You intentionally throw (return rejected Promise) for some `T`. Typescript can detect that if you do an overload type, for example

src/hooks/useTokenBalances.ts
```diff
@@ -83,28 +84,39 @@ async function _getBalances(walletInfo: WalletInfo, tokens: TokenDetails[]): Pro
   const contractAddress = depositApi.getContractAddress(networkId)
   assert(contractAddress, 'No valid contract address found. Stopping.')
 
-  const balancePromises: Promise<TokenBalanceDetails | null>[] = tokens.map((token) =>
-    fetchBalancesForToken(token, userAddress, contractAddress, networkId)
-      .then((balance) => {
+  const balancePromises: Promise<TokenBalanceDetails | null>[] = tokens.map((token) => {
+    const timeoutPromise = timeout<TokenBalanceDetails | null>({
```
Velenir, line 88: But this isn't `Promise<TokenBalanceDetails | null>`, this is `Promise<never>` as it will throw, always.

src/hooks/useTokenBalances.ts
```diff
@@ -83,28 +84,39 @@ async function _getBalances(walletInfo: WalletInfo, tokens: TokenDetails[]): Pro
...
+    })
+  })
 
-        return null
-      }),
-  )
   const balances = await Promise.all(balancePromises)
+
+  // TODO: Would be better to show the errored tokens in error state
   return balances.filter(Boolean) as TokenBalanceDetails[]
```
Velenir, line 120: We don't have to cast if we use a type-guard predicate `notEmpty`, which you made precisely for these cases, didn't you?

src/hooks/useTokenBalances.ts
```diff
@@ -83,28 +84,39 @@ async function _getBalances(walletInfo: WalletInfo, tokens: TokenDetails[]): Pro
...
+
+    return balancePromise.catch((e) => {
+      console.error('[useTokenBalances] Error for', token, userAddress, contractAddress, e)
 
-        const cachedValue = balanceCache[cacheKey]
-        if (cachedValue) {
-          logDebug('Using cached value for', token, userAddress, contractAddress)
-          return cachedValue
-        }
+      const cacheKey = constructCacheKey({ token, userAddress, contractAddress, networkId })
```
Velenir, line 105: No real need to have `constructCacheKey(...)` in two places

Velenir (changes requested):
I understand that this was made to be a quicker-than-quick fix. I'll make a PR with improvements

W3stside (approved):

PR #1647, opened 2020-12-01
Token image fetch optimization

Velenir:
After a suggestion by @biocom 
There are a lot of `GET https://.../<address>/logo.png 404` errors in the console for token images that aren't there. And while we can't get rid of these reports (as these are browser fetch errors), we can not retry failed images (on route change, on `TokenImage` remount)

So let's keep track of failed images and use fallback right away.

Also stop passing extra props to `Wrapper`s, that's plain bad manners :stern_look:

### DIFF
--- a/src/components/common/TokenImg.tsx
+++ b/src/components/common/TokenImg.tsx
@@ -76,7 +76,14 @@ const useFailOnceImage = ({ address, addressMainnet, symbol, name }: Omit<Props,
   }, [address, addressMainnet, symbol, name])
 }
 
+export const TokenImg: React.FC<Props> = (props) => {
+  const { faded } = props
+
   const imageProps = useFailOnceImage(props)
+
+  return <Wrapper {...imageProps} faded={faded} />
+}
+
 export const TokenImgWrapper = styled(TokenImg)`
   margin: 0 1rem 0 0;
 `
--- a/src/components/common/TokenImg.tsx
+++ b/src/components/common/TokenImg.tsx
@@ -13,9 +13,15 @@ const Wrapper = styled.img<WrapperProps>`
   padding: 2px;
   opacity: ${(props): number => (props.faded ? 0.4 : 1)};
 `
+// keep track of 404 token images
+// so we don't retry fetching them
+const failedTokenImages = new Set<string>()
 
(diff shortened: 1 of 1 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1647--dexreact.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1647--dexreact.review.gnosisdev.com/storybook/)**: Component stories

src/components/common/TokenImg.tsx
```diff
@@ -46,23 +52,36 @@ const tokensIconsFilesByAddress: Record<string, string> = tokensIconsRequire.key
...
+export const TokenImg: React.FC<Props> = (props) => {
+  const { faded } = props
 
-  // TODO: Simplify safeTokenName signature, it doesn't need the addressMainnet or id!
-  // https://github.com/gnosis/dex-react/issues/1442
-  const safeName = safeTokenName({ address, symbol, name })
+  const imageProps = useFailOnceImage(props)
 
-  return <Wrapper alt={safeName} src={iconFileUrl} onError={_loadFallbackTokenImage} {...props} />
+  return <Wrapper {...imageProps} faded={faded} />
```
biocom, line 84: Just to learn your stance here: Why is it better to pass an object with props here, than passing them as separate props? I'd be inclined to say, viewing them explicitly separated reads quicker. On the other hand, passing a single props object, arguably is easier to read too, given logic is handled separately in the `useFailOnceImage` function.

biocom (approved):

src/components/common/TokenImg.tsx
Velenir: it's not better or worse, just different
Now passing all `props` was bad, because `{address, symbol, name, ...}` ended up on the `Wrapper`, and while most of them got filtered by react, `name` gets applied, which doesn't break anything,but just not good. Also extra unnecessary props can cause extra unnecessary rerenders when they change
`imageProps` on the other hand are constructed as `Pick<ImgHTMLAttributes, **>` so they are real img attributes

anxolin:
Superseded in the new project https://github.com/gnosis/gp-ui/pull/25

PR #1655, opened 2021-01-07
Divide into apps

anxolin:
This PR restructure the codebase to support multiple apps

## What is included
Mostly is moving files:
* src/app/{OLD-APP-NAME} --> `src/app/{APP-NAME}`: Before we had 2 layouts in `src/app` I created in preparation for this. I accomodated them inside `src/apps`
* `src/apps` Since a picture is worth a thousand words. This is how the apps are structured:
image
* Note how every app now has an **INDEPENDENT** entry point and pages. Webpack only will make use of the dependencies that it needs for each app

## Help for the review
Most of the changes in this PR will be in `src/apps/swap-v1/SwapAppV1.tsx` and `src/apps/trade/TradeApp.tsx` wich in reality is a mix between the  old `src/App.tsx` (see how is deleted in this PR) and the layouts for the old app layout (see how `src/app/TradeApp/index.tsx` and ` src/app/LegacyTradeApp/index.tsx` are also deleted/moved to the new app layout)

A note in webpack. For now, in this PR we keep defaulting to v1 UI (since it's the website of this repository). This will change soon if we do the fork and deprecation of this repo.

## Future PRs
Future PRs will include a way to run only trade interface, only the old swap-v1 interface, the explorer. Also it'll generate one HTML entry point per app.


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1655--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1655--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

src/apps/trade/TradeApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+    content: '';
+    background: url(${PortfolioImage}) no-repeat center/contain;
+  }
+
+  &:hover > a::before {
+    background-image: url(${PortfolioImageWhite});
+  }
+`
+
+export const SwapAppV1: React.FC = () => {
```
W3stside, line 61: should be renamed to `TradeApp` or something

W3stside: also likely doesn't need to be exported?

src/apps/swap-v1/SwapAppV1.tsx
```diff
@@ -103,27 +109,49 @@ function getInitialUrl(): string {
...
-        <Route path="/book" exact component={OrderBook} />
-        <Route path="/connect-wallet" exact component={ConnectWallet} />
-        <Route path="/trades" exact component={Trades} />
-        <Route path="/settings" exact component={Settings} />
-        <Redirect from="/" to={initialUrl} />
-        <Route component={NotFound} />
-      </Switch>
-    </React.Suspense>
-  </LegacyTradeLayout>
+export const SwapAppV1: React.FC = () => {
```
W3stside, line 116: is already default export unless we want this elsewhere?

src/apps/trade/TradeApp.tsx
```diff
@@ -0,0 +1,112 @@
...
+        </GenericLayout>
+        {GlobalModalInstance}
+      </Router>
+      {process.env.NODE_ENV === 'development' && <Console />}
+    </>
+  )
+}
+
+export default hot(
+  withGlobalContext(
```
W3stside, line 106: this will make the new trade app share state with the legacy app as it's the same context. not saying it should be changed just a note

W3stside (approved):

src/apps/trade/TradeApp.tsx
anxolin: I kind of like the names exports, so the component decide on the name of the exported item (and also support multiple items to export). But I exported as default too, since the component it makes sense to be the default. I have no strong opinion, so I think I'll remove then just the default

src/apps/trade/TradeApp.tsx
anxolin: yep, true. There's more things like this one. I think when we address the PR of state we should revisit. I think they should have independent state

src/apps/trade/TradeApp.tsx
anxolin: I'll do it in the tip of the waterfall

anxolin: > should be renamed to TradeApp or something

I think the old web is more of a swap UI than trade ui, I think. Whatever we call it, it should have the V1 since it's not migrated for protocol 2

W3stside: Yeah i'm fine with the named export :+1:

alfetopito: I disagree with swap v1.
Even though kind of there's a swap element, there's also trade mixed in, together with liquidity.
I look at swap v1 and think it's the first version of the swap app (the swap MVP).

I'd rather call it gpV1Ui == Gnosis Protocol V1 User Interface. We could also use the long name to be very explicit

PR #1659, opened 2021-01-07
Fixes window.matchMedia incase there is no window object and/or no matchMedia function

W3stside:
As title states, was breaking inside `DepositWidget.components.test.tsx`

Conditionally runs logic involving `window` and `matchMedia` only if such exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Gnosis Protocol is a fully permissionless DEX that enables ring trades to maximi
 
 ## 📚 Wiki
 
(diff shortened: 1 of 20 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1659--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1659--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

anxolin (commented):
I don't get why this is needed

why wouldn't we have the window? would it be hiding another problem?

W3stside:
Yeah i think this can be closed, not sure why it was giving me issues :+1:

PR #1662, opened 2021-01-07
Run apps

anxolin:
This PR let us build or run ONLY one specific app if we want, make it faster for developing.

Also, we can choose now to `build:apps` (build all apps), or `build:<app>` to build an specific app


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1662--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1662--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

package.json
```diff
@@ -8,10 +8,16 @@
     "node": "12"
   },
   "scripts": {
-    "start": "NODE_ENV=development webpack-dev-server --mode development --hot --host 0.0.0.0 --disable-host-check",
+    "start": "npm run start:swap-v1",
```
W3stside, line 11: shouldn't these be using `yarn` :thinking:

webpack.config.js
```diff
@@ -24,7 +25,27 @@ const baseUrl = isProduction ? '' : '/'
 const config = overrideEnvConfig(loadConfig())
 const { name: appTitle } = config
 
-const apps = [SWAP_APP_V1, TRADE_APP]
+function getSelectedApps() {
+  const appName = process.env.APP
+  if (appName) {
```
W3stside, line 30: Could also avoid the Array.find with Map, though one could argue its maybe overcomplicated that way

```ts
const ALL_APPS = new Map([
    [SWAP_APP_V1.name, SWAP_APP_V1], 
    [TRADE_APP.name, TRADE_APP]
])

...

function getSelectedApps() {
  const appName = process.env.APP
  if (appName) {
    const app = ALL_APPS.get(appName)
    if (!app) {
      throw new Error(`Unknown App ${app}`)
    }

    return [
      {
        ...app,
        filename: 'index.html', // If we return only one app, the html web is "index.html"
      },
    ]
  } else {
    const allApps = Array.from(ALL_APPS.values())
    return allApps
  }
}

...

// Get selected apps: all apps by default
const apps = getSelectedApps()

W3stside (approved):
nice!

package.json
anxolin: all the others use npm. This way work with both

webpack.config.js
anxolin: I find the find ok, I see your option also OK. but I think I like more how it reads `const ALL_APPS = [SWAP_APP_V1, TRADE_APP]`

but I might do this in the top of the waterfall since you bother to write it and is also good option

Issue #1660, opened 2021-01-07
Reorganise and Optimise Styles + Theme [Multipart PR]


W3stside:
This series of PRs will be to address our current `theme` + `styles` organisation and logic. It is built on top of the `styled-theming` we have all already seen in PR #1626 but goes a bit further to clean up and make this more robust. It is similar to Uniswap's approach but incorporates other optimisations.

- [ ] Add `polished` CSS interpolation library for easier CSS dynamic stylings
- [ ] Global styles: slim down / separate concerns, make two global style providers: Static and Themed
- [ ] Re-organise `styled` folder into `theme` > `styled` > root files
- [ ] Add custom ThemeProvider similar to Uniswap's approach that is strictly typed and maps children with `theme` prop
- [ ] Add hooks for dealing with app theme
- [ ] Add global state to save user app theme selections
- [ ] Upgrade `common/Buttons` logic to be more legible + easier to read w/code AND `types`
- [ ] Fix `common/Buttons` stories

PR #1664, opened 2021-01-07
[V2 #1660 Style Rework] - Global Style & new Theme FS

W3stside:
Part of #1660

This creates a new `theme` fs and remove the global styles found in `GenericLayout`
Adds some subfolders and exports at `index` to prep for later waterfalled in theme code

It applies CSS as it should and can be checked inside stories (button) and by running the Trade app (new one)

fs:
```
src
-- theme (dir)
   -- styles (dir)
       - globals.ts
       - fonts.ts
       - overrides.ts
       - index.ts
```

`globals.ts` is now split apart and imports necessary chunks of css from outside. Is split between `StaticGlobalStyle` which is only, well, `static` css, and `ThemedGlobalStyle` which will be plugged-in under the (later coming) `ThemeProvider` which will allow us to dynamically compute new styles on the global style object if/when necessary with theming.

## Important
This is the _initial_ pr in the chain thus many things will not be implemented here, notably:
1. the `var(--stuff)` is still here and will go away later, once the theme is properly introduced
2. StaticGlobalStyles isn't changed much from the one we have now, it will change later once i have the correct providers in place.


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1664--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1664--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

src/theme/styles/global.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  html, body {  
+    height: 100vh;
+    width: 100vw;
+    margin: 0;
+    font-size: 62.5%;
+    line-height: 10px;
+    font-family: var(--font-default);
+    background-image: linear-gradient(0deg, var(--color-gradient-1) 0%, var(--color-gradient-2) 100%);
```
anxolin, line 71: We can address later, but I would think when you substitute `--color-gradient-1` with colors in the palette, since the palette is theme related, this settings would be theme dependent. 

The point of having a static css, is to have a nice default reset for all themes, right?

All these styles that have to do with color shouldn't be part of `ThemedGlobalStyle` instead?

For reference, look at Uniswap FixedGlobalStyle, it's nice to have some styles that are not re-render when the theme changes: 
image

src/theme/styles/global.ts
```diff
@@ -0,0 +1,112 @@
...
+  *, *:before, *:after {
+    box-sizing: inherit;
+  }
+
+  a {   
+    text-decoration: underline;
+    cursor: pointer;
+      &:link, 
+      &:visited {
+        color: var(--color-text-active);
```
anxolin, line 89: This is another example of my comment above, also the scrollbars etc..

src/theme/styles/overrides.ts
```diff
@@ -0,0 +1,115 @@
+import { css } from 'styled-components'
+
+// TODO: may not need any of this current CSS for V2
+
+export const web3ModalOverride = css`
```
anxolin, line 5: Would it make sense making`overrides` a folder, with one file per "hacked library"
Then, every app, can decide which chunk of CSS they use. I don't think they will all have `web3Modal` for example.

anxolin (approved):
Nice! 
I made some comments. We can discus it here, but if you want you can address them as separate PR or in the top of the waterfall, given that the waterfall is big.

src/theme/styles/global.ts
W3stside: yep all this changes later in the chain

src/theme/styles/global.ts
W3stside: yep all this changes later in the chain

src/theme/styles/overrides.ts
W3stside: mm yes, i agree. i will separate this in another PR

W3stside: added as a subpoint in #1660

alfetopito (approved):
:+1:

PR #1665, opened 2021-01-07
[V2 #1660 Style Rework] - Export media queries

W3stside:
Part of #1660 

Simple PR, just re-uses the media queries from V1 and exports them from `theme > style > media`


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1665--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1665--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

anxolin (approved):
Good for me, although I would think we have too many sizes:

For reference, Uniswap uses only these, that are pretty standard in the industry. More specific ones can be done ad-hoc imo, for individual components (using a helper function if we want)
```css
upToExtraSmall: ThemedCssFunction<DefaultTheme>
upToSmall: ThemedCssFunction<DefaultTheme>
upToMedium: ThemedCssFunction<DefaultTheme>
upToLarge: ThemedCssFunction<DefaultTheme>
```

anxolin:
@biocom what do you think?

W3stside:
@anxolin @biocom i think for now we keep as it's not different from before and we can trim down easily later. the uni ones are default sizes, but it's nice to have some inbetweens. Again, can just be removed later.

I also prefer this over uni's approach of adding the queries to theme which i don't really see as related in our project

alfetopito (approved):

PR #1666, opened 2021-01-07
[V2 #1660 Style Rework] Add Theme colour palette, types, and fix subsequent errors

W3stside:
Part of #1660 and :droplet: 's into #1664 

This PR adds our theme colour palette inside `theme > styles > colours` and the start of its types inside `theme > types` - relevant code all exported inside `theme > index.ts`

Colour palette is very minimal right now, can be iterated over as need be. Right now the `Color` type resembles:
```ts
interface Colors {
  // text
  primaryText1: Color
  secondaryText1: Color
  secondaryText2: Color
  textDisabled: Color

  // backgrounds / greys
  bg1: Color
  bg2: Color
  bgDisabled: Color

  // gradients
  gradient1: Color
  gradient2: Color

  // Base
  white: Color
  black: Color
  red1: Color
  red2: Color
  red3?: Color
  green1: Color
  green2: Color
  green3?: Color
  yellow1: Color
  yellow2: Color
  yellow3?: Color
  blue1: Color
  blue2: Color
  blue3?: Color
}
```

and the colour palettes themselves can be easily edited inside `theme > styles > colours` like:
```ts
const BASE_COLOURS = {
  // base
  white: '#FFF',
  black: '#000',
  red1: '#FF305B',
  red2: '#FF6871',
  red3: '#F82D3A',
  green1: '#00C46E',
  green2: '#09371d',
  green3: '#a9ffcd',
  yellow1: '#f1851d',
  yellow2: '#f1851d',
  blue1: '#2172E5',
  blue2: '#3F77FF',
}

export const LIGHT_COLOURS = {
  // text
  primaryText1: '#FFF',
  secondaryText1: '#EDEDED',
  secondaryText2: '#9797B8',
  textDisabled: '#31323E',

  // backgrounds / greys
  bg1: '#ffc1ff',
  bg2: '#F7F8FA',
  bgDisabled: '#ffffff80',

  // gradients
  gradient1: '#8958FF',
  gradient2: '#3F77FF',
}

export const DARK_COLOURS = {
  // text
  primaryText1: '#FFF',
  secondaryText1: '#EDEDED',
  secondaryText2: '#9797B8',
  textDisabled: '#31323E',

  // backgrounds / greys
  bg1: '#212429',
  bg2: '#2C2F36',
  bgDisabled: '#ffffff80',

  // gradients
  gradient1: '#8958FF',
  gradient2: '#3F77FF',

  // TODO: add to theme, not colour palette
[...]

PR #1667, opened 2021-01-07
[V2 #1660 Style Rework] - Use theme manager hook and add global user state

W3stside:
Part of #1660 and :water_buffalo: 's into #1666 

Adds global state for user,

```ts
interface UserState {
  theme: Theme
}
```
with `sideEffect` reducer to persist in localstorage

Adds hook to better manager data:
```ts
const useThemeMode = (): Theme => {
  const [state] = useGlobalState()

  return useMemo(() => state.user?.theme || Theme.AUTO, [state.user.theme])
}

export function useThemeManager(): [Theme, (newTheme: Theme) => void] {
  const [, dispatch] = useGlobalState()
  const theme = useThemeMode()

  const setNewTheme = useCallback(
    (newTheme: Theme) => {
      dispatch(updateTheme({ theme: newTheme }))
    },
    [dispatch],
  )

  return [theme, setNewTheme]
}
```

Will be applicable once proper `ThemeProvider` is created in app. Right now can not really be used


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1667--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1667--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

src/reducers-actions/user.ts
```diff
@@ -0,0 +1,41 @@
+import { Actions } from 'reducers-actions'
+import { Theme } from 'theme'
+import { setStorageItem } from 'utils'
+
+export interface UserState {
+  theme: Theme
+}
+
+type ActionTypes = 'UPDATE_USER_STATE' | 'UPDATE_THEME'
```
anxolin, line 9: What is `UPDATE_USER_STATE`??

src/reducers-actions/index.ts
```diff
@@ -39,6 +40,7 @@ export interface GlobalState {
   trade: TradeState
   trades: TradesState
   localTokens: LocalTokensState
+  user: UserState
```
anxolin, line 43: Can we make the reducer only theme related? 
I understand u called user to save any UserSetting, but I think this will be biting us when we try to reuse the state in different apps. 

The whole redux and state needs a bit of better restructuring (we can do it in a later PR), but in this PR at least I would add a separate state only for the theme/reducer

anxolin (commented):
great! I have to say, now it's making it way easier to do the review since you are addressing smaller scope PR

src/reducers-actions/user.ts
W3stside: ah, should go. it's an artifact

src/reducers-actions/index.ts
W3stside: sounds good

anxolin (approved):

alfetopito (approved):

PR #1668, opened 2021-01-08
[V2 #1660 Style Rework] - Enhance ThemeProvider

W3stside:
Part of #1660 and :water_buffalo: 's into #1667 

Enhances the default `styled-components` `ThemeProvider` wrapper and passes in our `theme` - also adds the benefit of implicitly passing it to any children

1. uses the `useThemeManager` hook and redux data from previous PR to detect selected theme and adjust accordingly.
2. Wraps ThemeToggler inside `storybook > decorators` with `withGlobalContext` inorder to access global state and use the theme provider
3. adds a `utils` file inside `theme` for some common utils. `getColourPalette` returns the `themeMode` specific `colourPalette`

### DIFF
--- a/src/components/common/Button/index.tsx
+++ b/src/components/common/Button/index.tsx
@@ -1,8 +1,9 @@
 import React from 'react'
-import styled, { css, /* DefaultTheme, */ ThemeProvider } from 'styled-components'
+import styled, { css } from 'styled-components'
 import { ThemeValue, variants } from 'styled-theming'
 
 import { COLOURS, BASE_STYLES } from 'styles'
+import ThemeProvider, { Theme } from 'theme'
 
 const {
   white,
@@ -50,7 +51,7 @@ export type ButtonSizeVariations = 'default' | 'small' | 'big'
 // <ButtonBase kind="danger" />
 export const ButtonTheme = variants('mode', 'variant', {
   default: {
-    light: css`
+    [Theme.LIGHT]: css`
       color: ${white};
       background: ${mainGradient};
 
@@ -58,7 +59,7 @@ export const ButtonTheme = variants('mode', 'variant', {
         background: ${mainGradientDarker};
       }
     `,
-    dark: css`
+    [Theme.DARK]: css`
       color: ${white};
       background: ${mainGradient};
 
@@ -71,7 +72,7 @@ export const ButtonTheme = variants('mode', 'variant', {
     return this.default
   },
   secondary: {
-    light: css`
+    [Theme.LIGHT]: css`
       color: ${blue};
       background: ${bgLight};
       border-color: ${blue};
@@ -80,7 +81,7 @@ export const ButtonTheme = variants('mode', 'variant', {
         background: ${bgDark};
       }
     `,
-    dark: css`
+    [Theme.DARK]: css`
       color: ${blue};
       background: ${bgDark};
       border-color: ${blue};
@@ -92,7 +93,7 @@ export const ButtonTheme = variants('mode', 'variant', {
     `,
   },
   success: {
-    light: css`
+    [Theme.LIGHT]: css`
       color: ${white};
       background: ${successLight};
 
@@ -101,7 +102,7 @@ export const ButtonTheme = variants('mode', 'variant', {
         border-color: ${successDark};
       }
     `,
-    dark: css`
+    [Theme.DARK]: css`
       color: ${white};
       background: ${successDark};
 
@@ -112,7 +113,7 @@ export const ButtonTheme = variants('mode', 'variant', {
     `,
   },
   danger: {
-    light: css`
+    [Theme.LIGHT]: css`
       color: ${white};
       background: ${dangerLight};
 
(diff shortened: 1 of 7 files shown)


gnosis-info:
Travis automatic deployment:
   * **🎩 [Dapp](https://pr1668--gpv1ui.review.gnosisdev.com)**: Testing web app
   * **📚 [Storybook](https://pr1668--gpv1ui.review.gnosisdev.com/storybook/)**: Component stories

src/reducers-actions/tokenRow.ts
```diff
@@ -1,6 +1,6 @@
 import { Actions } from 'reducers-actions'
 
-export const enum ActionTypes {
+export enum ActionTypes {
```
W3stside, line 3: storybook hates `const enum` though I think `enum` is safe as is here without `const`

anxolin (commented):
Great PR. Not from this PR, but I'm realising now some small things with the structure, so I better comment in the top of the waterfall

- [ ] I feel the colors logics should be closer together. There's big coupling between some different files. Ideally there would be a file called `src/theme/colors.tsx` with: `getThemePalette` (so no utils needed), all the color definitions, and the type `Color` and Colors`
- [ ] I think `StaticGlobalStyle` has too much code. Imo should be a plain reset.
- [ ] `ThemedGlobalStyle` should have some of the styles that are now in `StaticGlobalStyle`

src/theme/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+export * from './utils'
+
+import React, { useMemo } from 'react'
+import { DefaultTheme, ThemeProvider as StyledComponentsThemeProvider } from 'styled-components'
+
+import { getThemePalette } from './utils'
+import { useThemeMode } from 'hooks/useThemeManager'
+
+// Extension/override of styled-components' ThemeProvider but with our own constructed theme object
+const ThemeProvider: React.FC<{ componentKey: Partial<DefaultTheme['componentKey']> }> = ({
```
alfetopito, line 12: I don't think this is the best way to use the theme.
In my view it should be added to the top of the app hierarchy, dependent only on the global theme state.

The way it's used here every component wanting to use the theme would have to use this provider. Or am I reading this wrong?

src/components/common/Button/index.tsx
```diff
@@ -233,9 +234,7 @@ const ColouredAndSizedButtonBase = styled(ColouredButtonBase)`
 // Wrap ColouredAndSizedButtonsBase in it's own ThemeProvider which takes the toplevel app theme
 // ThemeProvider and interpolate over it's props
 const ThemeWrappedButtonBase: React.FC<React.ButtonHTMLAttributes<Element>> = ({ children, ...restProps }) => (
-  // TODO: fix when ThemePRovider is properly coded in
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  <ThemeProvider theme={({ mode }: any): any => ({ mode, componentKey: 'button' })}>
+  <ThemeProvider componentKey="button">
```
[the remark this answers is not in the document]
alfetopito: Well, my comment is pretty much the same as what Anxo is saying here.

src/theme/index.tsx
W3stside: @alfetopito it will be - right now this is simply extending the default `ThemeProvider` in case we want to use nested `ThemeProvider`s for interpolating over the entire theme, sth like: 
```js
// Define our `fg` and `bg` on the theme
const theme = {
  fg: "palevioletred",
  bg: "white"
};

// This theme swaps `fg` and `bg`
const invertTheme = ({ fg, bg }) => ({
  fg: bg,
  bg: fg
});

render(
  <ThemeProvider theme={theme}>
    <div>
      <Button>Default Theme</Button>

      <ThemeProvider theme={invertTheme}>
        <Button>Inverted Theme</Button>
      </ThemeProvider>
    </div>
  </ThemeProvider>
);
```

the global `ThemeProvider` will wrap below `StaticGlobalStyles` as `ThemedGlobalStyles` and will allow for interpolating theme logic

src/components/common/Button/index.tsx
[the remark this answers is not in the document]
W3stside: No, that's not what's going on here, tho I can see why you'd think that. Any nested ThemeProvider simply receives the theme as props and can interpolate them. What I've done here is create a sort of sub-theme - allows me to make a cleaner code style for button theme props using `styled-theming` - it does require a `key` in the top level theme to work but can then be applied via `styled-theming` to create the button's `size` prop. Or could be done for language, or font style etc without creating the typical `styled-component` template string props passing `e.g`:
```tsx
const SizableButton = styled(ButtonBase)<ButtonProps>`
  font-size: `${({}): string => size === 'BIG' ? '1.6rem' : size === 'SMALL' ? '0.6rem' : '1rem'}`;
  padding: `${({}): string => size === 'BIG' ? '0.5rem' : size === 'SMALL' ? '0.2rem' : '0.35rem'}`;
  ...
```
The logic above could surely be optimised by using functions or some such but imo the sub-theme angle can be pretty powerful for scaling

src/theme/index.tsx
alfetopito: Ook then. I trust you on this

alfetopito (approved):

PR #1672, opened 2021-01-14
Fix integration tests: Pin dependencies

anxolin:
> Originally created by @W3stside 

Versions for `web3` and `web3modal` updated in one of the PRs, causing breakage in the `playwright` tests.

`Web3Modal` likely updated and subsequently installed a higher `web3` version which took precedence.

1. `web3@1.3.1` >> `web3@1.2.9`
2. `web3modal@1.9.3` >> `web3modal@1.9.0`

Maybe we need to fix this later and not rely on pinning...

Playwright tests are passing:
Screenshot from 2021-01-14 11-53-24

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,8 +91,8 @@
     "react-toastify": "^5.5.0",
     "styled-components": "^5.0.0",
     "styled-theming": "^2.2.0",
-    "web3": "^1.2.9",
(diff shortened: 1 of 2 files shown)


anxolin:
APPROVED

alfetopito:
Self five: self five

